PR #6, opened 2016-09-08
Padhye

jitupadhye:
fixed parser to parse the following:

change 1:

>>>> no class-map

this one was easy: just added to ignore.

change 2:

class-map match-any BEST_EFFORT
  >>> match ip access-group dscp_0_non_ecn set-color yellow
  match ip access-group dscp_0

For this, I updated the rule for class map match, and added a color-setter matcher, and defined red, yellow and green as tokens,


arifogel:
After merging this, I also readded a line that you somehow removed from one of the .prefs files. Make sure to merge with master again so you don't have this conflict.

PR #7, opened 2016-09-08
Padhye

jitupadhye:
fixed parser to parse the following:

change 1:

>>>> no class-map

this one was easy: just added to ignore.

change 2:

class-map match-any BEST_EFFORT
  >>> match ip access-group dscp_0_non_ecn set-color yellow
  match ip access-group dscp_0

For this, I updated the rule for class map match, and added a color-setter matcher, and defined red, yellow and green as tokens,


jitupadhye:
management egress-interface-selection
  application http
  application snmp
  application radius
  application tacacs
  application syslog
  application ssh
  exit

PR #19, opened 2016-10-28
ignore switchport

jitupadhye:
ignore "switchport xxx xxx xxx xxx" in cisco config.

### DIFF
--- a/projects/batfish/.project
+++ b/projects/batfish/.project
@@ -7,16 +7,6 @@
 	<buildSpec>
 		<buildCommand>
 			<name>org.eclipse.ui.externaltools.ExternalToolBuilder</name>
-			<triggers>full,incremental,</triggers>
-			<arguments>
-				<dictionary>
-					<key>LaunchConfigHandle</key>
-					<value>&lt;project&gt;/.externalToolBuilders/org.eclipse.xtext.ui.shared.xtextBuilder.launch</value>
-				</dictionary>
-			</arguments>
-		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.ui.externaltools.ExternalToolBuilder</name>
 			<triggers>clean,full,incremental,</triggers>
 			<arguments>
 				<dictionary>
--- a/projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
+++ b/projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
@@ -530,6 +530,7 @@ l_null
    NO?
    (
       ABSOLUTE_TIMEOUT
+      | ACCOUNTING
       | ACTIVATION_CHARACTER
       | AUTHORIZATION
       | AUTOSELECT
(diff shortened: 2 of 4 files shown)


projects/batfish/.project
```diff
@@ -7,16 +7,6 @@
 	<buildSpec>
 		<buildCommand>
 			<name>org.eclipse.ui.externaltools.ExternalToolBuilder</name>
```
arifogel: I don't know why you had to modify this file. I don't want to merge this change. Which version of XText are you using? Perhaps if you update to the latest this change won't be necessary.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -530,6 +530,7 @@ l_null
    NO?
    (
       ABSOLUTE_TIMEOUT
```
arifogel: Both of these added keywords need to be removed from null_block_substanza or unpredictable behavior will occur.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_aaa.g4
```diff
@@ -421,7 +421,7 @@ aaa_authorization_commands
    COMMANDS level = DEC?
```
arifogel: Where are the unit tests? I don't fully understand the impact of this change. Also, please make it clear what flavor of Cisco your example lines are coming from.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_ignored.g4
```diff
@@ -410,6 +411,7 @@ null_block_substanza
          | ERSPAN_ID
          | ESCAPE_CHARACTER
          | EXCEED_ACTION
+         | EXEC
```
arifogel: I really need to see a unit test here. This might conflict with other changes I've made. In fact I'm pretty sure I removed this keyword from here recently.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_ignored.g4
```diff
@@ -249,6 +249,7 @@ null_block_stanza
       | STCAPP
       | SVCLC
       | SWITCH
+      | SWITCHPORT
```
arifogel: If indeed 'switchport' appears at the top level (a unit test would be great for confirming this), then this keyword should be removed from null_block_substanza.

projects/batfish/.project
arifogel: I committed a fix for this, I think. Please merge latest master.

arifogel: OK I figured it out. I sometimes disable the xtext builder because it causes things to slow down. Disabling that builder causes two effects:
1. It creates a file: projects/batfish/.externalToolBuilders/org.eclipse.xtext.ui.shared.xtextBuilder.launch
2. It modifies projects/batfish/.project
To avoid propagating my disabling of xtext builder to the repo, i put the file in 1. in .gitignore. However, I didn't realize that .project in 2. was also being modified. The fix was to tell git to ignore changes to .project when I feel like disabling the xtext builder. This is accomplished with:
git update-index --assume-unchanged projects/batfish/.project
before making the change.

PR #50, opened 2016-12-17
Ratul

ratulm:
Travis is stuck, and it wouldn't even let me cancel the job. Build tests pass for me.


arifogel:
Looks like tests passed. Was your issue transient?

PR #65, opened 2017-01-24
Ratul

ratulm:


projects/batfish/src/org/batfish/grammar/cisco/Cisco_snmp.g4
```diff
@@ -67,6 +67,7 @@ ss_host_null
       INFORMS
       | TRAPS
       | VERSION
+      | variable
```
arifogel: This is a breaking change, as shown in your changes to the ref file. It seems like you are trying to parse:
`snmp-server host 10.1.2.3 xyx`
But since `variable` matches all non-`NEWLINE` tokens, and `ss_host_null` includes `~NEWLINE*`, and `ss_host_null` appears above `ss_host_use_vrf`, this will make `ss_host_use_vrf` unmatchable. Furthermore, this will cause anything after `snmp-server host` to parse, and we may fail to catch unhandled variants of currently implemented matches - even if we put `ss_host_use_vrf` above `ss_host_null`. Generally speaking, you should never put `variable` in a `xx_null` rule.

test_rigs/unit-tests/configs/cisco_snmp
```diff
@@ -62,7 +62,9 @@ snmp-server enable traps vtp notifs
 snmp-server enable traps vtp vlancreate
 snmp-server enable traps vtp vlandelete
 snmp-server engineID local 000000090200000652D6C000
+snmp-server file-transfer access-group 8 protocol tftp
```
arifogel: This is a reference to an access-list, which needs to be recorded.

test_rigs/unit-tests/configs/cisco_snmp
```diff
@@ -79,6 +81,7 @@ snmp-server source-interface trap loopback0
 snmp-server source-interface trap mgmt0
 snmp-server source-interface trap Vlan1
 snmp-server tcp-session auth
+snmp-server tftp-server-list 8
```
arifogel: This is also a reference to an access-list, which needs to be recorded.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_snmp.g4
```diff
@@ -90,6 +91,7 @@ ss_null
       | CONTACT
       | CONTEXT
       | ENGINEID
+      | FILE_TRANSFER
```
arifogel: Since in your unit test you are using this to handle a line with a reference to an access-list, `FILE_TRANSFER` should not appear in `ss_null` but rather be handled. Otherwise we may have false positives for unreferenced access-lists.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_snmp.g4
```diff
@@ -103,6 +105,7 @@ ss_null
       | QUEUE_LENGTH
       | SOURCE_INTERFACE
       | TCP_SESSION
+      | TFTP_SERVER_LIST
```
arifogel: Since in your unit test you are using this to handle a line with a reference to an access-list, `TFTP_SERVER_LIST` should not appear in `ss_null` but rather be handled. Otherwise we may have false positives for unreferenced access-lists.

test_rigs/unit-tests/configs/cisco_snmp
```diff
@@ -62,7 +62,9 @@ snmp-server enable traps vtp notifs
 snmp-server enable traps vtp vlancreate
 snmp-server enable traps vtp vlandelete
 snmp-server engineID local 000000090200000652D6C000
+snmp-server file-transfer access-group 8 protocol tftp
 snmp-server group cenic-nms v2c IPv4 auth
+snmp-server host 10.1.2.3 xyx
 snmp-server host 10.1.2.3 use-vrf management
```
arifogel: The line `snmp-server host 10.1.2.3 use-vrf management` is broken by your change above (see comment).

test_rigs/parsing-tests/unit-tests.ref
```diff
@@ -11722,10 +11736,20 @@
             "      (ss_host",
             "        HOST:'host'  <== mode:DEFAULT_MODE",
             "        IP_ADDRESS:'10.1.2.3'  <== mode:DEFAULT_MODE",
-            "        (ss_host_use_vrf",
-            "          USE_VRF:'use-vrf'  <== mode:DEFAULT_MODE",
+            "        (ss_host_null",
             "          (variable",
-            "            MANAGEMENT:'management'  <== mode:DEFAULT_MODE)",
+            "            VARIABLE:'xyx'  <== mode:DEFAULT_MODE)",
```
arifogel: This is a breaking change. See comment above about `variable` in `ss_host_null`.

PR #90, opened 2017-04-03
Support additional variants of  'tacacs server' and 'logging' commands

xyzsa:


projects/batfish/src/org/batfish/grammar/cisco/CiscoLexer.g4
```diff
@@ -9435,6 +9435,11 @@ FORWARD_SLASH
    '/'
 ;
```
arifogel: This change will have a serious impact. I would be surprised if it did not break other things.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -2271,7 +2271,7 @@ s_tacacs
    TACACS
    (
       t_null
```
arifogel: Please format

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -2699,6 +2699,18 @@ t_server
    SERVER hostname = variable_hostname NEWLINE
 ;
```
arifogel: Please format

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -2699,6 +2699,18 @@ t_server
...
 
+t_address
+:
+   (
+      ADDRESS ((IPV4 IP_ADDRESS) | (IPV6 IPV6_ADDRESS)) NEWLINE
+   )
+;
+
+t_key
```
arifogel: I don't believe this is the best way to handle this. I think lexer modes would be more appropriate. As I said, having HEX_DIGIT be a token will likely have a serious impact.
See if you  can implement this by pushing into a lexer mode when you encounter KEY.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
```diff
@@ -14,13 +14,13 @@ logging_address
       | IPV6_ADDRESS
    )
    (
```
arifogel: The naming and formatting is off for these rules.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
```diff
@@ -44,7 +44,10 @@ logging_archive_null
 
 logging_buffered
 :
```
arifogel: Naming is off, and there is no need for parentheses here.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
```diff
@@ -211,4 +214,24 @@ s_logging
       logging_common
       | logging_vrf
    )
-;
```
arifogel: I don't believe these should have been converted to rules. Were you just trying to refactor?

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
```diff
@@ -106,8 +109,8 @@ logging_host
 :
    HOST iname = variable? hostname = variable
    (
```
arifogel: Must the vrf, discriminator, and port portions appear in this exact order?

projects/batfish/src/org/batfish/grammar/cisco/CiscoLexer.g4
xyzsa: It did not break any of the existing unit tests or parsing of test configs.

There is indeed a similar DIGIT: F_Digit; rule (line 9402). Checked with ANTLR4 documentation that when a terminal (like one of 0 through 9) matches two lexical rules, it is fine, with only restriction that the terminals themselves used directly (e.g. as in xyz: 1;) will be disallowed, but using corresponding LHS components (like F_Digit and F_HexDigit would be fine).

Can you please elaborate on what is going to be broken, and additional tests could be added to cover that, and in which case, F_Digit: DIGIT; case would also need a fix?

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
xyzsa: OK.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
xyzsa: OK.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
xyzsa: OK, will address in next iteration.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
xyzsa: OK..

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
xyzsa: I could find examples for VRF and Discriminator order; this can be generalized with a Kleene's closure, but then it will accept many sentences that are not legal (e.g. repeating VRF clauses, etc.) nor practical.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
xyzsa: Yes, it is about refactoring the rule fragments are used across multiple rules. For instance, there were varied uses of VRF clause, one with variable being captured, and other being ignored. This makes all uses uniform.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
xyzsa: Then the same should apply to F_Digit fragment as well; change followed the existing code base.

arifogel:
The HEX_DIGIT rule will not be matched if the DIGIT rule appears first, and
vice versa in the case of any text source '0' to '9'. Though there may be
incomplete test cases, this is likely to be a breaking change. I'd like to
look over the codebase to see whether the DIGIT rule should even exist.
I'll get back to you afterwards.

projects/batfish/src/org/batfish/grammar/cisco/Cisco_logging.g4
xyzsa: Parentheses were added to avoid a long (>80 columns) line.

projects/batfish/src/org/batfish/grammar/cisco/CiscoLexer.g4
xyzsa: Deleted as suggested.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -2701,14 +2705,21 @@ t_server
...
+      (
+         IPV6 IPV6_ADDRESS
+      )
+   ) NEWLINE
 ;
 
 t_key
 :
-    KEY DIGIT HEX_DIGIT+
+   KEY DEC VARIABLE NEWLINE
```
arifogel: I don't believe this change will work if the key happens to only include digits from 0-9 (eschewing A-F). Please add a unit test with such key. If it breaks, you will likely have to use lexer modes to solve this.

xyzsa: Will add the tests, but after checking further, the syntax allows non-hex strings (when DEC is omitted or 0). It is an encrypted string from a rotational cipher (does not have to be hexadecimal). Will update the rule accordingly.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -1773,14 +1773,18 @@ s_tacacs
    (
       t_null
       | t_server
+      (
```
arifogel: Please move this parenthetical expression to the end of t_server definition. It is the handler for t_server that would use information in this expression.

projects/batfish/src/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -1773,14 +1773,18 @@ s_tacacs
...
+      )*
    )
 ;
 
 s_tacacs_server
 :
    NO? TACACS_SERVER
    (
-      ts_host
+      ts_host t_key?
```
arifogel: Similarly please move this t_key? To the end of the definition for ts_host

Issue #127, opened 2017-05-09
Disable vlan interfaces with no active switchports using them


arifogel:
Currently, vlan interfaces are considered to be up independently of interfaces that may use them. However, any vlan interface not used by an active switchport interface via trunking or access mode should not be up.

lpapillon:
Please note that this check should also take into account the following:
If a vlan has " no autostate" in its config, then it's always up regardless of how many ports are in this vlan.
Example:
interface Vlan112
    no autostate

arifogel:
Fixed by d4c87fedbd4f03b5bacae06b646aadde67cb0644

PR #146, opened 2017-06-14
disable unused vlan interfaces

xyzsa:
Address issue 127 (https://github.com/batfish/batfish/issues/127): disable VLAN interfaces with no active switchports using them. Also take into consideration that a vlan interface can have "no autostate" specified, in which case, it should not be disabled.


projects/batfish-common-protocol/src/org/batfish/datamodel/Configuration.java
```diff
@@ -72,6 +72,10 @@
 
    private static final String TACACS_SOURCE_INTERFACE_VAR = "tacacsSourceInterface";
 
+   private static final int VLAN_NORMAL_MIN_DEFAULT = 1;
```
arifogel: Are these constants common to all VLAN implementations for all vendors? ~If not, they should go in the respective VendorConfiguration classes in org.batfish.representation.XXX.XXXConfiguration.java~ It looks like you also have some constants in the VendorConfiguration classes. Why are there constants here?

projects/batfish-common-protocol/src/org/batfish/datamodel/Interface.java
```diff
@@ -173,17 +187,22 @@ else if (name.startsWith("TenGigabitEthernet")) {
          return InterfaceType.PHYSICAL;
       }
       else if (name.startsWith("Tunnel")) {
-         return InterfaceType.VPN;
```
arifogel: This change is only OK if Tunnel interfaces are definitely not used for IPSEC VPNs. Please answer whether this is the case.

projects/batfish-common-protocol/src/org/batfish/datamodel/Interface.java
```diff
@@ -950,9 +996,8 @@ public JSONObject toJSONObject() throws JSONException {
       JSONObject iface = new JSONObject();
       iface.put("node", _owner.getName());
       iface.put("name", _key);
-      iface.put("prefix", _prefix.toString());
```
arifogel: Um what is this used by? This looks like unmaintainable hackery.

projects/batfish-common-protocol/src/org/batfish/datamodel/Configuration.java
xyzsa: Yes, these are common across all vendors, and Cisco has further refinements onto them (VLAN id is 12-bit long per 802.11Q, with 0 and 0xFFF not usable). Cisco specific rules, see for example: http://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus5000/sw/configuration/guide/cli/CLIConfigurationGuide/VLANs.html

projects/batfish-common-protocol/src/org/batfish/datamodel/Configuration.java
```diff
@@ -72,6 +72,10 @@
...
+
+   private static final int VLAN_NORMAL_MAX_DEFAULT = 4095;
```
xyzsa: This should be 4094 - will update.

projects/batfish-common-protocol/src/org/batfish/datamodel/Interface.java
xyzsa: The change here is to refactor duplicated "type" and "prefix" strings used elsewhere in the code (existing, unmaintainable hackery :-)). JSON object serialization is used in unit test result references otherwise.

projects/batfish-common-protocol/src/org/batfish/datamodel/Interface.java
xyzsa: According to http://www.cisco.com/en/US/technologies/tk583/tk372/technologies_white_paper0900aecd8029d629_ps6635_Products_White_Paper.html tunnel interfaces can be used for IPSec VPNs - will rollback this change.

xyzsa: It is actually the other way: VPN (encrypted) is more restricted than Tunnel (encapsulation, that covers VPNs along with many non-VPN cases). See IOS 12.4 documentation: http://www.cisco.com/c/en/us/td/docs/ios/12_4/interface/configuration/guide/inb_tun.html and a Tunnel interface configured for IPSec VPN at: http://www.cisco.com/en/US/technologies/tk583/tk372/technologies_white_paper0900aecd8029d629_ps6635_Products_White_Paper.html

The code does not seem to have any dependencies on InterfaceType.VPN for Tunnel interfaces for Cisco configurations; there is one reference for Juniper, that does not apply in this case.

If the current code has de

projects/batfish-common-protocol/src/org/batfish/datamodel/Configuration.java
```diff
@@ -74,7 +74,7 @@
 
    private static final int VLAN_NORMAL_MIN_DEFAULT = 1;
 
-   private static final int VLAN_NORMAL_MAX_DEFAULT = 4095;
+   private static final int VLAN_NORMAL_MAX_DEFAULT = 4094;
```
arifogel: Looks good.

PR #151, opened 2017-06-20
Changes to ease Batfish development in IntelliJ

dhalperi:
* .gitignore: add IDEA project files
* batfish: delete unused test


projects/batfish-common-protocol/test/src/org/batfish/test/TestSerialization.java
```diff
@@ -1,23 +0,0 @@
...
-
-public class TestSerialization {
-
-   @Test
-   public void test() throws JsonProcessingException, JSONException {
-      ObjectMapper mapper = new ObjectMapper();
-      BgpNeighbor neighbor = new BgpNeighbor((Prefix) null, null);
-      String val = mapper.writeValueAsString(neighbor);
-      JSONObject obj = new JSONObject(val);
-      assert obj.length() == 1; // just id
```
dhalperi: comment: the actual size is 25. The test fails in IntelliJ because it turns on java assertions. The test passes with `ant test` because the `assert` is a no-op at runtime. If I change it to use more standard junit assertThat:

```java
assertThat(obj.length(), equalTo(1));
```

then the test fails because 1 != 25.

.gitignore
```diff
@@ -3,3 +3,7 @@
 /unit-tests
 .DS_Store
 *.testout
+
+# IntelliJ project files
```
arifogel: Cool

projects/batfish-common-protocol/build.xml
```diff
@@ -101,21 +101,5 @@
      <jar destfile="${outputJar}" basedir="bin" />
 	</target>
 
-   <target name="test" depends="compileTest">
```
arifogel: Sure, just put it back after tests are added to build.sh

projects/batfish-common-protocol/test/src/org/batfish/test/TestSerialization.java
```diff
@@ -1,23 +0,0 @@
-package org.batfish.test;
```
arifogel: Yup this is a garbage test.

projects/batfish-common-protocol/test/src/org/batfish/test/TestSerialization.java
arifogel: The more you know

PR #152, opened 2017-06-22
fixed read recurse files

### LINKED ISSUE #103: add ability to recurse through configs directory
Currently only configuration files in top level of testrig `configs` folder are read. Instead, the whole tree under `configs` should be read.

kidsbear:
This closes #103.


projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -430,6 +430,7 @@ else if (!settings.printParseTree()) {
 
    private long _timerCount;
 
+
```
dhalperi: please delete extra lines added

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3554,18 +3555,17 @@ public void pushDeltaEnvironment() {
       _testrigSettings = _deltaTestrigSettings;
    }
 
+
```
dhalperi: please delete extra lines added

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3578,6 +3578,20 @@ public void pushDeltaEnvironment() {
       return configurationData;
    }
 
+   static List<Path> listAllFiles(Path configsPath) {
```
dhalperi: add a javadoc comment explaining what this function does. Something simple; you do not need `@param` etc.

```java
/**
 * Returns a list of {@link Path paths} to files under the directory indicated
 * by {@code configsPath}. Directories are recursively expanded but not
 * included in the returned list.
 *
 * <p>Temporary files (with names that start with {@code .}) are omitted from the returned list.
 */
```

is probably what I'd write.

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3578,6 +3578,20 @@ public void pushDeltaEnvironment() {
...
+      List<Path> configFilePaths;
+      try {
+         configFilePaths = Files.walk(configsPath)
+                 .filter(path -> !path.getFileName().toString().startsWith(".")
+                         && Files.isRegularFile(path))
+                 .sorted()
+                 .collect(Collectors.toList());
+      } catch (IOException e) {
+         throw new BatfishException("Failed to walk path: " + configsPath.toString(), e);
```
dhalperi: I believe the `.toString()` is redundant as `"" + configsPath` will convert `configsPath` to String automatically. Drop it?

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
+
+public class BatfishTest {
+
+    @Test(expected = BatfishException.class)
+    public void testNotExistingPath() {
+        List<Path> result = Batfish.listAllFiles(Paths.get("./test/notExist"));
```
dhalperi: It looks like the variable `result` is not used. Drop it? (like this:)

```java
Batfish.listAllFiles(Paths.get("./test/notExist"));
```

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+
+    @Test
+    public void testNoFileUnderPath() {
+        List<Path> result = Batfish.listAllFiles(Paths.get("./test/noFile"));
+        Assert.assertEquals(0, result.size());
```
dhalperi: two things:

1. Usually, functions like `assertEquals` are statically imported so that it doesn't say

    ```java
    Assert.assertEquals(x, y);
    ```

    but rather

    ```java
    assertEquals(x, y);
    ```

    (makes the code slightly easier to read).

2. A better way to do assertions is using [`assertThat`](http://junit.org/junit4/javadoc/4.12/org/junit/Assert.html#assertThat(T, org.hamcrest.Matcher)):

    ```java
    assertThat(results, empty());
    ```

    Using [`empty`](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#empty()). That way when the check fails you get a better error message. Instead of "0 != 3" you get "expected: empty list, got: [a, b, c]". Or something like that.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testReadStartWithDotFile() {
+        List<Path> result = Batfish.listAllFiles(Paths.get("./test/startWithDot"));
+        Assert.assertEquals(0, result.size());
+    }
+
+
+    @Test
+    public void testFileNameRead() {
+        List<Path> result = Batfish.listAllFiles(Paths.get("./test/unNestedDirectory"));
+        Assert.assertEquals("./test/unNestedDirectory/test1.cfg", result.get(0).toString());
```
dhalperi: here it should be something like:

```java
List<Path> expected = // build a list here of the expected paths.
assertThat(result, equalTo(expected));
```

(We use `equalTo` instead of `containsInAnyOrder` since the function is supposed to return sorted output and thus always the same answer.)

projects/batfish/src/org/batfish/main/Batfish.java
dhalperi: Should also document that the result is sorted.

projects/batfish/test/org/batfish/main/BatfishTest.java
dhalperi: Should test the contents of the exception message. I believe you can use [`ExpectedException`](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) for this.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,68 @@
...
+        List<Path> result = Batfish.listAllFiles(Paths.get("./test/nestedDirectory"));
+        Assert.assertEquals("./test/nestedDirectory/aStart/e-test.cfg", result.get(0).toString());
+        Assert.assertEquals("./test/nestedDirectory/b-test.cfg", result.get(1).toString());
+        Assert.assertEquals("./test/nestedDirectory/d-test.cfg", result.get(2).toString());
+        Assert.assertEquals("./test/nestedDirectory/innerDirectory/a-test.cfg", result.get(3).toString());
+        Assert.assertEquals("./test/nestedDirectory/innerDirectory/c-test.cfg", result.get(4).toString());
+    }
+
+
+}
```
dhalperi: generally, fewer blank lines. No need for 2 blank lines ever, nor at end of file or beginning of file.

projects/batfish/test/unNestedDirectory/test3.cfg
```diff
@@ -0,0 +1 @@
+I am Test 3!!!
```
dhalperi: Consider whether you can create the files in a temp directory inside of the test rather than check them in. I bet you can use the [`TemporaryFolder`](http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html) to handle temp files including automatically deleting them when the test finishes.

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3578,6 +3578,20 @@ public void pushDeltaEnvironment() {
       return configurationData;
    }
 
+   static List<Path> listAllFiles(Path configsPath) {
+      List<Path> configFilePaths;
+      try {
+         configFilePaths = Files.walk(configsPath)
+                 .filter(path -> !path.getFileName().toString().startsWith(".")
+                         && Files.isRegularFile(path))
```
dhalperi: q? does this handle symlinks? should it? we should decide and document

dhalperi: from [javadoc](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#isRegularFile(java.nio.file.Path,%20java.nio.file.LinkOption...)): "By default, symbolic links are followed and the file attribute of the final target of the link is read. If the option `NOFOLLOW_LINKS` is present then symbolic links are not followed."

So I believe this does what we want for symlinks.

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3578,6 +3575,28 @@ public void pushDeltaEnvironment() {
       return configurationData;
    }
 
+   /**
+    * Returns a sorted list of {@link Path paths} contains all files under the directory
+    * indicated by {@code configsPath}. Directories under {@code configsPath} are recursively
+    * expanded but not included in the returned list.
+    *
+    * <p>Temporary files(files start with {@code .} are omitted from the returned list. </p>
+    * <p>Currently skips all symlinks.</p>
```
dhalperi: newline between paragraphs.

Also, this part is wrong. It includes symlinks, does not skip them.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.collection.IsEmptyCollection.empty;
+import static org.junit.Assert.assertThat;
+
+public class BatfishTest {
```
dhalperi:
```java
/**
 * Tests of {@link Batfish}.
 */
```

dhalperi: (document new files/classes)

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+    @Rule
```
dhalperi: single newline between rules.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,97 @@
...
+        file.getParentFile().mkdir();
+        assertThat(file.createNewFile(), is(true));
+        List<Path> resultList = Batfish.listAllFiles(startWithDot.toPath());
+        assertThat(resultList, is(empty()));
+    }
+
+    @Test
+    public void testReadUnNestedPath() throws IOException {
+        File unNestedFolder= folder.newFolder("unNestedDirectory");
+        List<Path> fileList = new ArrayList<>();
```
dhalperi: rename `fileList` to `expected`:

1. It's paths, not files.
2. Do not put object types in their names. aka, no `List`.

dhalperi: and 3. no need to create it twice (once up here as `fileList`, once below as `expectedList`.

dhalperi: and same changes below.

projects/batfish/src/org/batfish/main/Batfish.java
dhalperi: (Actually, this part is not wrong if `Files.walk` excludes symlinks. Thanks!)

projects/batfish/src/org/batfish/main/Batfish.java
```diff
@@ -3578,6 +3576,29 @@ public void pushDeltaEnvironment() {
       return configurationData;
    }
 
+   /**
+    * Returns a sorted list of {@link Path paths} contains all files under the directory
```
dhalperi: my suspicion is that Ari currently has these at 80 character limit, and I think this line is over. Please double-check wrapping to 80 chars?

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.collection.IsEmptyCollection.empty;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Test for {@link Batfish#listAllFiles(Path)}.
```
dhalperi: This is too specific: the test class will eventually test more functions. "Tests (since plural) for {@link Batfish}." is right.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void testNoFileUnderPath() throws IOException {
+        File emptyFolder = folder.newFolder("emptyFolder");
+        List<Path> result = Batfish.listAllFiles(emptyFolder.toPath());
+        assertThat(result, empty());
+    }
+
+    @Test
+    public void testReadStartWithDotFile() throws IOException {
+        File startWithDot = folder.newFolder("startWithDot");
+        File file = new File(startWithDot + "/.cfg");
```
dhalperi: there should be no `/` in path names here -- that's assuming a certain OS is running the tests.

Instead, to construct `Path` objects corresponding to subchildren, use `Path#resolve`.

dhalperi: Throughout the file.

projects/batfish/test/org/batfish/main/BatfishTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void throwsExceptionWithSpecificType() {
+        thrown.expect(BatfishException.class);
+        thrown.expectMessage("Failed to walk path: ./test/notExist");
```
dhalperi: rather than making up a string, use something resolved from folder that you know does not exist. E.g., `folder.getRoot().toPath().resolve("nonexistent")`.

PR #157, opened 2017-06-23
Convert from Ant to Maven

dhalperi:
Converts Batfish from Ant to Maven. A few highlights:

* Automatic dependency management via Maven – we just list jars and versions required once. Either command-line `mvn` or IDE integration (installed by default in both Eclipse and IntelliJ) will resolve and download the jars, javadoc, and sources.
* Much smaller download as a result. And less maintenance as the ground truth for everything only appears once (in the `pom.xml` files) and all IDEs use it. Plus we don't need `lib` directories.
* Automatic dependency verification via `maven-dependency-analyze` – every module is checked to see that it actually needs all the things it depends on.
* Clean Maven integration with ANTLR – all we do is put the ANTLR files in `src/main/antlr` and the plugin does the rest.
* Separation of generated files (g4->java) in a separate `target/generated-sources` directory.
* General simplification of build scripts since Maven itself handles the cross-module dependencies.
* Smaller allinone jar, since we can easily build both a "slim" `batfish.jar` (e.g.) and a "fat" `batfish-bundle.jar` with all dependencies. Thus allinone only contains 1, not 4, copies of all transitive dependencies.
* All the tests still pass and the build works with command line `mvn`, `batfish_functions`, Eclipse, and IntelliJ on my laptop.

A few caveats:
* This build does not build Javadoc yet, but it's easy to add.
* I don't know the `batfish_functions` very well, but I tried all the commands I could find and they seemed to work.

I am sure there are regressions, but overall I think this is going to be a step in the right direction.

@arifogel , want to sit down and talk through changes + how to identify regressions tomorrow?


dhalperi:
- [ ] Determining version in tools/common.sh is a bottleneck, and some people have batfish_reload.sh-type commands in their shell scripts.

dhalperi:
Update: have switched to a simpler bash-based mechanism to extract version, so we don't run Maven to do it.

Also, the following script worked for me with Eclipse

```
RESET

1. Wipe out maven (rm -rf ~/.m2)
2. Wipe out a temp Eclipse workspace (rm -rf /tmp/workspace)
3. Wipe out all local git state (git clean -dfx)

PREP

4. Compile the entire project using maven (mvn compile in the batfish/projects/ folder)

ECLIPSE

5. Open a fresh Eclipse, using the temp workspace.
6. File > Open Projects from Filesystem
7. Choose batfish/projects/

wait.
```

It went through approximately 5 rounds of compiling, but ended up with 3 warnings no errors.

PR #159, opened 2017-06-26
Add validation of parameter value according to template variables

kidsbear:
Implemented by mimicking python code. There are still some uncertain behaviors unimplemented, added comments to indicates.


projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
 
+   /**
+    * Validate the json-encoded {@code value} matches the valid setting
+    *  specified by {@code variable}.
+    *  <p>If {@code value} does not match the valid setting,
+    *  {@link BatfishException BatfishException} will be thrown.</p>
+    */
+   static void validate(JsonNode value, Variable variable)
+         throws BatfishException {
+      if (variable.getType().getStringType() &&
```
dhalperi: Add a comment why this block is here. Add a comment before the `JSON_PATH` line explaining why that line is there and that you think this is a bug.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+         if (!value.isTextual()) {
+            throw new BatfishException("The parameter value type: " + value
+                  .getNodeType() + " doesn't match the type required for " +
+                  "variable type: " + variable.getType());
+         }
+      }
+      switch (variable.getType()) {
+         case BOOLEAN:
+            if (!value.isBoolean()) {
+               throw new BatfishException("Expected a boolean for " +
```
dhalperi: there are likely better ways to format code to construct strings so that it's readable. In general, try to avoid lots of `"" + ""`. example:

```java
throw new BatfishException(
   "Expected a boolean for variable type " + type);
```

(also, since you use it everywhere, save `variable.getType()` into a temp variable e.g., `type` in the above.)

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+                     "variable type: " + variable.getType());
+            }
+            break;
+         case COMPARATOR:
+            String textContent = value.textValue();
+            if (!(textContent.equals(">") || textContent.equals(">=") ||
```
dhalperi: Create a constant set of the valid comparators (e.g., `private static final Set<String> COMPARATORS = ...`) and then simply use `if (!COMPARATORS.contains(value.textValue()))`.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+            //TODO: need prefix range validation check
+            break;
+         case STRING:
+            break;
+         case SUBRANGE:
+            //TODO: need prefix range validation check
+            break;
+         case JSON_PATH:
+            validateJsonPath(value);
+            break;
```
dhalperi: in all of the above, please make sure the invalid value is printed as well as the name of the parameter that is invalid.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+      }
+   }
+
+   /**
+    * Validate that {@code jsonPathRegex}  contains a valid java regular
```
dhalperi: Java is written with capital J.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    * expression of a json path(Starts with '/', ends with either '/' or
+    * '/i', contains a valid java regular expression between '/').
+    *  <p>If the content of {@code jsonPathRegex} is not a valid java regular
+    *  expression of a json path,
+    *  {@link BatfishException BatfishException} will be thrown.</p>
+    */
+   static void validateJsonPathRegex(String jsonPathRegex) throws
+         BatfishException {
```
dhalperi: similar to the string construction comment above, it's best to break long lines semantically. That is, break them at places where the meaning or structure shifts, rather than in the middle of a construct. In this case:

```java
static void validateJsonPathRegex(String jsonPathRegex)
      throws BatfishException
```

would make more sense, because the `throws` is related to the `BatfishException`, not the function's argument list.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +323,151 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+      if (jsonPathRegex == null || jsonPathRegex.isEmpty()) {
+         throw new BatfishException("JsonPathRegex should not be empty or " +
+               "null.");
+      }
+      // Need to check what to do when encounter "/" and "/i", currently
+      // check size to filter them out
+      if (jsonPathRegex.startsWith("/")) {
```
dhalperi: this is a bit of a mess, maybe we can talk to see if we can clean it up.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,255 @@
...
+   public ExpectedException thrown = ExpectedException.none();
+
+   @Before
+   public void initMapper() {
+      mapper = new BatfishObjectMapper();
+   }
+
+   @Test
+   public void testNullValue() {
+      //Need to check if null input is allowed, otherwise verify before entering
```
dhalperi: Space after `//` in all cases, keep indent consistent on succeeding lines.

kidsbear:
Ari told me he is not sure about whether "/" and "/i" should be a valid jsonPathRegex, he currently assume they are valid, so I did the same thing here. Also, he said I cannot rely on the stringType field in "Variable.Type", so I added "String type check" in each block under switch(just like how it was done in python code). Unfortunately, the automatic check  failed, seems like it is caused by the "InvalidStateException", but I don't know how to fix it.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -333,12 +332,12 @@ private boolean answer(String questionTemplateName, String paramsLine,
     * type.
     * @throws BatfishException If the type encoded in input {@code value}
     * does not match the type specified in {@code variable}.
-    * @throws InvalidStateException If the input {@code value} is null.
+    * @throws IllegalArgumentException If the input {@code value} is null.
```
dhalperi: Sorry -- I meant `IllegalStateException`, not `InvalidStateException`; and if you use [`java.lang.IllegalStateException`](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html) instead of that `sun` InvalidStateException you won't have the compiler error you saw before.

dhalperi: Also, the exception should be thrown in the caller that is extracting `JsonNode` from the parameter map – not here in the `validateType` function.

This function should simply assume that the `value` is non-null and does not need to handle that case or declare an extra exception for the case that it is null.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
       return execute(wItemAs, outWriter);
    }
 
+   /**
+    * Validate the contents contained in json-encoded {@code value}
+    * matches the type required by {@code variable}.
+    * Call {@link Variable#getType()} on {@code variable} shows the expected
```
dhalperi: Why newline before `Call`? If it's a new paragraph, add a blank line and a `<p>`. If it's part of the same paragraph, merge into above.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+            }
+            break;
+         case IP:
+            if (!value.isTextual()) {
+               throw new BatfishException(errorMessage);
+            }
+            // TODO: Need to double check isInetAddress()
+            if (!InetAddresses.isInetAddress(value.textValue())) {
+               throw new BatfishException(
+                     String.format("The input: %s is not a valid ip address.",
```
dhalperi: Please capitalize acronyms throughout: IP, JSON, etc.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+            if (!value.isTextual()) {
+               throw new BatfishException(errorMessage);
+            }
+            try {
+               Pattern.compile(value.textValue());
+            } catch (PatternSyntaxException e) {
+               throw new BatfishException(
+                     String.format("Input String: %s is not a valid Java " +
+                           "regular  expression for variable type: %s.",
+                           value.textValue(), expectedType));
```
dhalperi: include the regex exception in the BatfishException

dhalperi: (and test this.)

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+            break;
+         case JSON_PATH:
+            validateJsonPath(value);
+            break;
+      }
+   }
+
+   /**
+    * Validate that {@code jsonPathRegex} contains a valid Java regular
+    * expression of a JsonPath(Starts with '/', ends with either '/' or
```
dhalperi: use proper formatting.

1. Space between `JsonPath` and `(`.
2. Use `{@code JsonPath}`, for example.
3. Use `"` for Java strings, as `'/i'` is not a legal Java token.
4. Newline before throws
5. Capitalize Java.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    * @throws BatfishException If the content of {@code jsonPathRegex} is not
+    * a valid java regular expression of a JsonPath.
+    */
+   static void validateJsonPathRegex(String jsonPathRegex)
+         throws BatfishException {
+      if (jsonPathRegex == null || jsonPathRegex.isEmpty()) {
+         throw new BatfishException(
+               "JsonPathRegex should not be empty or null.");
+      }
+      // TODO: Currently assume "/" and "/i" is allowed here.
```
dhalperi: This is not quite the right comment.

Maybe,

"As written, this function will accept the strings `/` and `/i` as complete expressions – resulting in an empty inner Java regular expression."

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    * expression of a JsonPath(Starts with '/', ends with either '/' or
+    * '/i', contains a valid Java regular expression between '/').
+    * @throws BatfishException If the content of {@code jsonPathRegex} is not
+    * a valid java regular expression of a JsonPath.
+    */
+   static void validateJsonPathRegex(String jsonPathRegex)
+         throws BatfishException {
+      if (jsonPathRegex == null || jsonPathRegex.isEmpty()) {
+         throw new BatfishException(
+               "JsonPathRegex should not be empty or null.");
```
dhalperi: Is the name of the invalid parameter in this message?

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +325,176 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+      }
+      // TODO: Currently assume "/" and "/i" is allowed here.
+      if (jsonPathRegex.startsWith("/")) {
```
dhalperi: This sequence of `if`s would be a lot cleaner and less nested using the following pattern:

```java
if (BAD CONDITION) {
   throw exception;
}

if (OTHER BAD CONDITION) {
   throw exception;
}

... good code ...
```

You have much less nesting if you "exit early" instead of having

```java
if (GOOD CONDITION) {
   if (OTHER GOOD CONDITION) {
      ... good code ...
   } else {
      throw exception
   }
} else {
   throw exception;
}
```

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,288 @@
...
+   public void testValidIntegerValue() throws IOException {
+      JsonNode comparatorNode = mapper.readTree("15");
+      Question.InstanceData.Variable variable
+            = new Question.InstanceData.Variable();
+      variable.setType(Question.InstanceData.Variable.Type.INTEGER);
+      try {
+         Client.validateType(comparatorNode, variable);
+      } catch (Exception e) {
+         throw new BatfishException("Unexpected exception");
+      }
```
dhalperi: what is this try .. catch construct in lines 134..138 doing? Why not just have line 135?

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,288 @@
...
+   }
+
+   @Test
+   public void testInvalidIntegerValue() throws IOException {
+      JsonNode invalidInteger = mapper.readTree("1.5");
+      Question.InstanceData.Variable variable
+            = new Question.InstanceData.Variable();
+      variable.setType(Question.InstanceData.Variable.Type.INTEGER);
```
dhalperi: there seems to be a lot of code duplication in this file .. the first 4 lines of every test are almost the same. Can you think of a way that utility functions could make this cleaner?

kidsbear:
Changed the way error message generated. Currently have no idea how to check the "nested Exception message".

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -342,18 +342,10 @@ static void validate(Map<String, String> parameters,
...
-                  String messageBody
-                        = String.format("The value: %s, is a type " +
-                        "of %s, doesn't match the type expected for %s: %s",
-                        parameterValue, value.getNodeType(), parameterName,
-                        variable.getType());
                   String errorMessage
-                        = messageHead.concat(e2.getMessage().isEmpty() ?
-                                             messageBody : e2.getMessage());
-                  throw new BatfishException(errorMessage);
+                        = String.format("Invalid value: %s for parameter: %s: ",
```
dhalperi: Please drop all the colons; extra punctuation just breaks up the flow of the text, makes it harder to read and less grammatically correct.

Instead, if you're concerned about it being hard to tell where the value begins and ends:

```java
Invalid value for parameter %s: [%s].   // parameter name then value, since we know the name is small doesn't have any special characters.
```

note that you don't need the final colon on the string either, it shows up naturally in the way that Java stack traces are rendered.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -380,86 +372,94 @@ static void validateType(JsonNode value, Variable variable)
       switch (expectedType) {
          case BOOLEAN:
             if (!value.isBoolean()) {
-               throw new BatfishException("");
+               throw new BatfishException(
+                     String.format("Value: %s, is not a valid Boolean " +
```
dhalperi: No punctuation needed here. Just "[%s] is not a valid JSON boolean value."

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -380,86 +372,94 @@ static void validateType(JsonNode value, Variable variable)
...
+                           "value.", value.toString()));
             }
             break;
          case COMPARATOR:
             if (!(COMPARATORS.contains(value.textValue()))) {
-               throw new BatfishException("");
+               throw new BatfishException(
+                     String.format("Value: %s, is not a valid comparator.",
+                           value.toString()));
```
dhalperi: Since the list of known comparators may not be obvious to the reader, we should print it.

"[%s] is not a known comparator. Valid options are %s." // value then list.

Why do you use `toString` here instead of `textValue`?

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -380,86 +372,94 @@ static void validateType(JsonNode value, Variable variable)
...
             }
             break;
          case INTEGER:
             if (!value.isInt()) {
-               throw new BatfishException("");
+               throw new BatfishException(
+                     String.format("Value: %s, is not a valid Integer.",
+                           value.toString()));
```
dhalperi: calling `.toString()` on something being passed to `String.format` is not necessary.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -380,86 +372,94 @@ static void validateType(JsonNode value, Variable variable)
...
             }
             break;
          case IP:
             if (!value.isTextual()) {
-               throw new BatfishException("");
+               throw new BatfishException(
+                     "IP address was expected to be a String.");
```
dhalperi: Combine these two checks and single error message

```java
if (!value.isTextual() || InetAddresses.isInetAddress...) {
   throw ...
}
```

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -380,86 +372,94 @@ static void validateType(JsonNode value, Variable variable)
...
-                                 value.textValue(), expectedType));
+                     String.format("Value: %s, is not a valid IP address.",
+                                 value.textValue()));
             }
             break;
          case IP_WILDCARD:
             if (!value.isTextual()) {
-               throw new BatfishException("");
+               throw new BatfishException(
+                     "IP_Wildcard was expected to be a String.");
```
dhalperi: As we discussed, figure out a way to write the variable types that make sense to the user. We can expect the user knows:

* The proper type names used in JSON: that is `expectedType.getName()`.
* English and also terms like Batfish and JSON.

But we cannot expect the user has read this code.

So, we should only use things we know the user knows.

For example:

"A Batfish ipWildcard must be a JSON string."  // plus make sure the actual JSON type plus value is printed somewhere.

or

"An IP wildcard address must be a JSON string." // same caveat.

kidsbear:
I'll try to come out with a better function name for it. Have no idea now.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+         catch (IOException e1) {
+            throw new BatfishException("Variable value is not valid JSON",
+                  e1);
+         }
+         if (value == null) {
+            throw new IllegalStateException(
+                  String.format("Failed to extract JsonNode " +
+                              "for parameter: %s, with variable value: %s.",
+                              parameterName, parameterValue));
+         }
```
dhalperi: extract lines 331–344 into a helper and test it. Ensure that the value (that is not valid JSON) is printed when throwing an error :)

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+         if (variable.getMinElements() != null) {
+            if (!value.isArray() ||
```
dhalperi: extract 346 – 365 into a helper and test it.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+               } catch (BatfishException e1) {
+                  String errorMessage
+                        = String.format("Invalid value for parameter " +
+                        "%s: %s", parameterName, node);
+                  throw new BatfishException(errorMessage, e1);
+               }
+            }
+         }
+         else {
+            try {
```
dhalperi: extract 367–373 into a helper, use it above at L356–365, and test it.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+               validateType(value, variable);
+            } catch (BatfishException e2) {
+               String errorMessage
+                     = String.format("Invalid value for parameter %s: %s",
```
dhalperi: Confirming: you print the value here, so you never print the value inside of `validateType`?

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+                              "for parameter: %s, with variable value: %s.",
+                              parameterName, parameterValue));
+         }
+         if (variable.getMinElements() != null) {
+            if (!value.isArray() ||
+                  value.size() < variable.getMinElements()) {
+               throw new BatfishException(
+                     String.format("Invalid value for parameter %s: %s." +
+                                 " Input must be a JSON array contains " +
+                                 "at least %s elements",
```
dhalperi: Good job on this one!

Simplified second sentence, and adding a missing word: "Expecting a JSON array of at least %d elements."

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    * @throws BatfishException if the content of {@code jsonPathRegex} is not
+    * a valid Java regular expression of a JsonPath.
+    */
+   static void validateJsonPathRegex(String jsonPathRegex)
+         throws BatfishException {
+      if (!jsonPathRegex.startsWith("/")) {
+         throw new BatfishException(
+               String.format("A Batfish %s must start with %s",
+                              Variable.Type.JSON_PATH_REGEX.getName(),
+                              "\"/\""));
```
dhalperi: why don't you inline these string constants?

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+      }
+      if (!(jsonPathRegex.endsWith("/") || jsonPathRegex.endsWith("/i"))) {
+         throw new BatfishException(
+               String.format("A Batfish %s must end in either %s or %s",
+                     Variable.Type.JSON_PATH_REGEX.getName(),
+                     "\"/\"", "\"/i\""));
```
dhalperi: and these string constants

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    * optional key 'suffix' which mapping to a boolean value).
+    *
+    * @throws BatfishException if {@code jsonPath} is not a valid jsonPath
+    * dictionary.
+    */
+   static void validateJsonPath(JsonNode jsonPath) throws BatfishException {
+      if (!jsonPath.isContainerNode()) {
+         throw new BatfishException(
+               String.format("%s is not a valid type for a Batfish %s. Type " +
+                           "expected: JSON dictionary", jsonPath.getNodeType(),
```
dhalperi: Is the return value of  `jsonPath.getNodeType()` something that a user that does not use Jackson know about?

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+               String.format("Missing 'path' element of %s",
+                     Variable.Type.JSON_PATH.getName()));
+      }
+      if (!jsonPath.get("path").isTextual()) {
+         throw new BatfishException(
+               String.format("'path' element of %s must be a JSON string",
+                     Variable.Type.JSON_PATH.getName()));
+      }
+      if (jsonPath.get("suffix") != null &&
+            (!jsonPath.get("suffix").isBoolean())) {
```
dhalperi: no need for extra parentheses here.

projects/batfish-client/src/org/batfish/client/Client.java
```diff
@@ -315,6 +307,290 @@ private boolean answer(String questionTemplateName, String paramsLine,
...
+    *
+    * @throws BatfishException when there exists a missing parameter: it is
+    * not optional in {@code variable}, but the user failed to provide it.
+    */
+   static void checkVariableState(Map<String, Variable> variables)
+         throws BatfishException{
+      for (Entry<String, Variable> e : variables.entrySet()) {
+         String variableName = e.getKey();
+         Variable variable = e.getValue();
+         if ((!variable.getOptional()) && variable.getValue() == null) {
```
dhalperi: delete the extra parentheses.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,461 @@
...
+   }
+
+   @Test
+   public void testValidJsonPath() throws IOException {
+      String validJsonPath = "{\"path\" : \"I am path.\", \"suffix\" : true}";
+      Client.validateJsonPath(mapper.readTree(validJsonPath));
+   }
+
+   // Tests for validateType method
+   public void validateTypeWithInvalidInput(String input, String expectedMessage,
```
dhalperi: why is this method public? It should probably be `private static`.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,461 @@
...
+      Client.validateType(IPNode, variable);
+   }
+
+   @Test
+   public void testInvalidJavaRegexValue() throws IOException {
+      String invalidJavaRegex = "\"...{\\\\Q8\\\\E}\"";
+      JsonNode inputNode = mapper.readTree(invalidJavaRegex);
+      Question.InstanceData.Variable.Type expectedType
+            = Question.InstanceData.Variable.Type.JAVA_REGEX;
+      String expectedMessage ="It is not a valid Java regular expression";
```
dhalperi: space after `=`.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,461 @@
...
+      Client.validateType(inputNode, variable);
+   }
+
+   @Test
+   public void testInvalidJsonPathValue() throws IOException {
+      String input = "{\"variable\" : \"I am variable\"}";
+      Question.InstanceData.Variable.Type expectedType
+            = Question.InstanceData.Variable.Type.JSON_PATH;
+      String expectedMessage =
+            String.format("Missing 'path' element of %s",
```
dhalperi: do `=` go at the end of the line or the beginning of the next one? Look through Batfish and see what the standard is.

We should be consistent, but here we do one thing at L285 and another at L286.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,461 @@
...
+            mapper.readTree("\"/.*/\"");
+      Question.InstanceData.Variable variable
+            = new Question.InstanceData.Variable();
+      variable.setType(Question.InstanceData.Variable.Type.JSON_PATH_REGEX);
+      Client.validateType(JsonPathRegexNode, variable);
+   }
+
+   @Test
+   public void testUnsatisfiedMinLengthValue() throws IOException {
+      String invalidIP = "\"0.0.0\"";
```
dhalperi: For simplicity, use a string here, so that we're not adding complexity mixing invalid IP with invalid length.

It's probably not valid to mix `minLength` with `IP` type anyway.

projects/batfish-client/test/org/batfish/client/ClientTest.java
```diff
@@ -0,0 +1,461 @@
...
+      variable.setMinLength(8);
+      thrown.expect(BatfishException.class);
+      thrown.expectMessage(equalTo(
+            "Must be at least 8 characters in length"));
+      Client.validateType(mapper.readTree(invalidIP), variable);
+   }
+
+   @Test
+   public void testSatisfiedMinLengthValue() throws IOException {
+      String validIP = "\"192.168.1.1\"";
```
dhalperi: same comment as above.

Issue #162, opened 2017-06-28
Rethink question preprocessing


dhalperi:
Right now the logic in `Batfish#preprocessQuestion` is complicated/brittle and also incomplete (specifically for the case of CompositeQuestion, which might have question-valued variables). We'd like to rewrite this code to be simpler, better tested, and potentially more powerful.

Specifically,

- [ ] add comprehensive tests.
- [ ] support composite question with question-valued variables.
- [ ] consider adding much more functionality than simple variable replacement like we have now.

dhalperi:
Closing as stale.

PR #164, opened 2017-06-29
Batfish: preprocess inner Questions

dhalperi:
Add a new Question type and, when a question is parameterized by a variable
of type question, preprocess that question too. Note that this is different
from an inner question, for which each question can only have one and is
shares variables from the other question. This is instead a nested question,
which is currently in a completely different parameter space (aka, in a
composite query).

This partially addresses #162, but does not yet improve testing or really
code clarity. (It does make errors slightly better.)


dhalperi:
Please take a look, @arifogel . So far, I've tested this manually and confirmed it works for simple cases like I've shown you earlier. I'm deferring adding new tests in the `ant` world -- it's so much simpler in the Maven world.

It was unclear to me precisely how validation should happen on the server side.

Issue #172, opened 2017-07-05
Audit use of Streams especially as regards files


dhalperi:
We've got some fairly big file descriptor leaks due to use of various `Stream<Path>` APIs rooted in `Files`. We need to audit the codebase for cleaning this up.

A sample fix:

```diff
@@ -780,9 +783,13 @@ public class Batfish extends PluginConsumer implements AutoCloseable, IBatfish {
          throw new CleanBatfishException(
                "Missing compiled vendor-independent configurations for this test-rig\n");
       }
-      else if (CommonUtil.list(path).count() == 0) {
-         throw new CleanBatfishException(
-               "Nothing to do: Set of vendor-independent configurations for this test-rig is empty\n");
+      else {
+         try (Stream<Path> paths = CommonUtil.list(path)) {
+            if (paths.count() == 0) {
+               throw new CleanBatfishException(
+                     "Nothing to do: Set of vendor-independent configurations for this test-rig is empty\n");
+            }
+         }
       }
    }
 
```

[Geeze I hope there's a more concise way to do this.]

PR #173, opened 2017-07-05
Batfish: fix leak in checkConfiguration

dhalperi:
This is for batfish/batfish#172


dhalperi:
Stuck with the streams API for now as fully materializing the list can be expensive if there are many things in it.

PR #180, opened 2017-07-07
Random fixups

dhalperi:
A few miscelleneous fixes or improvements... deleting dead code or variables or javadoc; simplifying a bunch of code with modern Java collections-isms; and .gitignore.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -449,8 +449,6 @@ public class Batfish extends PluginConsumer implements AutoCloseable, IBatfish {
 
    private BatfishLogger _logger;
 
-   private SortedMap<String, String> _questionMap;
-
    private Settings _settings;
 
    // this variable is used communicate with parent thread on how the job
@@ -484,7 +482,6 @@ public class Batfish extends PluginConsumer implements AutoCloseable, IBatfish {
       _terminatedWithException = false;
       _answererCreators = new HashMap<>();
       _testrigSettingsStack = new ArrayList<>();
-      _questionMap = new TreeMap<>();
    }
 
    private Answer analyze() {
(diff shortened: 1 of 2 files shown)


dhalperi:
@arifogel I manually merged some very minor crap, but you should probably take a look at this one. At least the computeIfAbsent commit is interesting.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -206,65 +206,63 @@
 
    public static void applyBaseDir(
          TestrigSettings settings, Path containerDir,
-         String testrig, String envName, String questionName) {
```
arifogel: I think this is leftover from before the last time we restructured testrig layout

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -630,17 +625,10 @@ public AnswerElement answerAclReachability(
       for (Entry<AclLine, Boolean> e : output.entrySet()) {
          AclLine line = e.getKey();
          String hostname = line.getHostname();
-         Map<String, List<AclLine>> byAclName = arrangedAclLines.get(hostname);
```
arifogel: I considered this pattern earlier and rejected it. I really don't remember why. Perhaps it was because I wanted to have a place to set breakpoints when a key was absent. In any case, that can be done ad hoc in the future if really necessary.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -450,8 +448,6 @@ else if (!settings.printParseTree()) {
 
    private BatfishLogger _logger;
 
-   private SortedMap<String, String> _questionMap;
```
arifogel: This was used for debugging. I believe I also had a future purpose in mind. But for now I really don't care if you remove it.

dhalperi:
Thanks for review!

PR #183, opened 2017-07-08
Todd

millstein:
A new JSON format for specifying node roles, as an explicit dictionary and/or a list of regexes.

A new question that can ask another question once per role.

First cuts at automatic node clustering and automatic outlier detection.


arifogel:
Looks like you need to update your refs

PR #184, opened 2017-07-08
facilitate more robust interfacing with the UI; missing dependencies for maven

ratulm:


tests/ui-focused/analysis-answers-before.ref
```diff
@@ -1,3 +1,3 @@
 {
-  "unusedstructures" : "Not answered"
+  "unusedstructures" : "{\n  \"answerElements\" : [\n    {\n      \"class\" : \"org.batfish.datamodel.answers.StringAnswerElement\",\n      \"answer\" : \"Not answered\"\n    }\n  ],\n  \"status\" : \"NOTFOUND\"\n}"
```
dhalperi: is this right? Should this be a string or an actual JSON object?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -937,115 +937,121 @@ public JSONArray listEnvironments(
    }
 
    /**
-    * Lists the questions under the specified container, testrig
+    * Lists the testrigs under the specified container
```
dhalperi: Hard to tell, but it looks like you just swapped two functions. If so, LGTM.

projects/coordinator/src/main/java/org/batfish/coordinator/authorizer/FileAuthorizer.java
```diff
@@ -25,17 +26,26 @@
...
       }
-      if (!Files.exists(_permsFile)) {
-         throw new BatfishException("Perms file not found: '"
-               + _permsFile.toAbsolutePath().toString() + "'");
+      catch (Exception e) {
+         throw new BatfishException(String.format("Could not initialize FileAuthorizer with "
+                 + " RootDir = %s UsersFile=%s. PermsFile=%s",
+                 Main.getSettings().getFileAuthorizerRootDir(),
+                 Main.getSettings().getFileAuthorizerUsersFile(),
+                 Main.getSettings().getFileAuthorizerPermsFile()));
```
dhalperi: Do you want to include the exception `e` in the result? Otherwise, seems weird to throw a FNF with a good message right before it won't get propagated out this catch.

tests/ui-focused/analysis-answers-before.ref
ratulm: the clients are currently written to expect strings, so we could embed arbitrary messages. not  the best decision in hindsight; the whole interface between clients and coordinator is a little hacky (e.g., no use of classes and packing responses by hand), which we need to fix at some point.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
ratulm: likely fixed an erroneous comment.

projects/coordinator/src/main/java/org/batfish/coordinator/authorizer/FileAuthorizer.java
ratulm: good point. fixed in https://github.com/batfish/batfish/commit/6c67f7c79b06ef681d1a0df1434cf17a99b76fbf

tests/ui-focused/analysis-answers-before.ref
dhalperi: Fair enough, should discuss.

PR #190, opened 2017-07-10
IntelliJ config: delete modules and *.iml

dhalperi:
The IML files are auto-generated by IntelliJ from
Maven, but differ somewhat arbitrarily from machine
to machine. This is why IDEA suggests not to version
them.

However, it seems to be the case that when they are
not checked in, IntelliJ barfs on missing IML files
from the modules.xml file. So let's delete that one,
too.

Instead, users will have to manually import modules
once when checking out the project. This is a pain,
but I can't find a better way to do it.


ratulm:
When I try to build using IntelliJ I get Error:Failed to load project configuration: cannot read file ..../ batfish/projects/.idea/misc.xml: ..../batfish/projects/.idea/misc.xml (No such file or directory).

Is that expected?

Issue #191, opened 2017-07-10
rearrange grammar files into natural directory structure


arifogel:
Currently the master grammar files are in `projects/batfish/src/main/antlr4/org/batfish/grammar/(name)/(name){Parser,Lexer}.g4`.
The imported (subordinate) grammars are all bunched together in `projects/batfish/src/main/antlr4/imports/(subordinate_grammar_file).g4`.
Ideally each imported grammar would reside in the same directory as the master grammar that imports it.

PR #193, opened 2017-07-11
antlr4: move imports to same directory

dhalperi:
Requires mildly more complex maven setup, but not too bad.

This fixes batfish/batfish#191.


dhalperi:
@arifogel I may be living a dream, but I think this actually makes the IntelliJ maven plugin useful again. Not for saving output and generating java source, but for editing a valid grammar and also using the built-in grammar validation and testing frameworks.

PR #196, opened 2017-07-11
Intellij formatting, remove unused line

arifogel:


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -364,7 +359,8 @@ else if (containerDir != null) {
     */
    static List<Path> listAllFiles(Path configsPath) {
       List<Path> configFilePaths;
-      try (Stream<Path> allFiles = Files.walk(configsPath, FileVisitOption.FOLLOW_LINKS)) {
+      try (Stream<Path> allFiles = Files.walk(configsPath,
+            FileVisitOption.FOLLOW_LINKS)) {
```
dhalperi: This is a place where 80 chars seems unnecessary.

Also, the break here is not semantic -- makes more sense to break at Files, at `.walk`, or at `configsPath`.

.idea/codeStyleSettings.xml
```diff
@@ -16,7 +16,7 @@
             <package name="" withSubpackages="true" static="false" />
           </value>
         </option>
-        <option name="ENABLE_JAVADOC_FORMATTING" value="false" />
```
dhalperi: is this not necessary? I found that reformatting questions would really screw up the //question_comment sections.

.idea/codeStyleSettings.xml
```diff
@@ -16,7 +16,7 @@
...
+        <option name="RIGHT_MARGIN" value="80" />
```
dhalperi: 80 is generally considered too small for java. 80 + 3-char indent more so. (Google uses 100+2).

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -629,10 +620,11 @@ public AnswerElement answerAclReachability(
       for (Entry<AclLine, Boolean> e : output.entrySet()) {
          AclLine line = e.getKey();
          String hostname = line.getHostname();
-         Map<String, List<AclLine>> byAclName =
-               arrangedAclLines.computeIfAbsent(hostname, k -> new TreeMap<>());
+         Map<String, List<AclLine>> byAclName = arrangedAclLines
+               .computeIfAbsent(hostname, k -> new TreeMap<>());
```
dhalperi: this break also seems bad semantically.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -862,11 +852,9 @@ private Answer compileEnvironmentConfigurations(
       Path indepConfigsDir = envSettings.getDeltaCompiledConfigurationsDir();
       if (deltaConfigurationsDir != null) {
          if (Files.exists(deltaConfigurationsDir)) {
-            answer.append(serializeVendorConfigs(
-                  envSettings.getEnvPath(),
+            answer.append(serializeVendorConfigs(envSettings.getEnvPath(),
                   vendorConfigsDir));
-            answer.append(serializeIndependentConfigs(
-                  vendorConfigsDir,
+            answer.append(serializeIndependentConfigs(vendorConfigsDir,
```
dhalperi: and these (the original was not much better):

```java
append(serializeIndependentConfigs(
      vendorConfigsDir, indepConfigsDir));
```

would seem much better, or

```java
append(
      serializeIndependentConfigs(vendorConfigsDir, indepConfigsDir));
```

assuming it fit.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -979,15 +965,16 @@ private InterfaceSet computeFlowSinks(
                i.getVrrpGroups().forEach((groupNum, vrrpGroup) -> {
                   Prefix prefix = vrrpGroup.getVirtualAddress();
                   Pair<Prefix, Integer> key = new Pair<>(prefix, groupNum);
-                  Set<Interface> candidates =
-                        vrrpGroups.computeIfAbsent(
-                              key, k -> Collections.newSetFromMap(new IdentityHashMap<>()));
+                  Set<Interface> candidates = vrrpGroups.computeIfAbsent(key,
+                        k -> Collections
+                              .newSetFromMap(new IdentityHashMap<>()));
```
dhalperi: there has to be a better place to break than at `Collections.`?

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2352,10 +2307,9 @@ public void initRemoteBgpNeighbors(
                      continue;
                   }
                   remoteAddresses.put(bgpNeighbor, remoteAddress);
-                  Set<BgpNeighbor> localAddressOwners =
-                        localAddresses.computeIfAbsent(
-                              localAddress,
-                              k -> Collections.newSetFromMap(new IdentityHashMap<>()));
+                  Set<BgpNeighbor> localAddressOwners = localAddresses
+                        .computeIfAbsent(localAddress, k -> Collections
+                              .newSetFromMap(new IdentityHashMap<>()));
```
dhalperi: here, previous seems much more readable to me.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2902,12 +2848,10 @@ void outputAnswerWithLog(Answer answer) {
                prettyAnswerString);
       }
       catch (Exception e) {
-         BatfishException be = new BatfishException(
-               "Error in sending answer",
+         BatfishException be = new BatfishException("Error in sending answer",
```
dhalperi: it feels like we're breaking as late as possible rather than semantically. I'd probably actually write

```java
BatfishException be = 
      new BatfishException("Error in sending answer", e);
```

for maximum clarity, assuming it can't be a 1-linear.

Issue #195, opened 2017-07-11
Fix ant->maven antlr4 build regressions


arifogel:
- [x] Change to an imported grammar file should trigger regeneration of grammar importing it
  Currently, only master parser/lexer grammar changes trigger a rebuild. Fixing this is **HIGHEST** priority

  - [x]  .tokens files should only be generated in target
  The .tokens files are now being generated in the src/main/antlr4/... directories. These should ideally not be produced in a src folder.

- [ ] Only recompile relevant java files when grammars are regenerated
  After any MASTER grammar file is changed, it appears that ALL java files are recompiled - rather than just the ones that have changed (or rely on ones that have changed). While this will certainly work, it seems heavy-handed. Please see behavior of old ant build.xml for a lighter-weight method.

- [ ] Parallel processing of grammars
Grammars are currently processed one at a time. When we used ant, all out-of-date lexers were generated in parallel, and then all out-of-date parsers were generated in parallel. While we could replicate this behavior, I'd like to do better:
  Each grammar should be its own task. Each parser task should depend on its corresponding lexer task. All such tasks should be executed in parallel to the extent feasible.

dhalperi:
Re: tokens -- I believe that Maven only produces them in the `target` folder:

```
➜  batfish git:(master) find . -name \*.tokens
./projects/batfish/target/generated-sources/antlr4/AssertionLexer.tokens
./projects/batfish/target/generated-sources/antlr4/AssertionParser.tokens
./projects/batfish/target/generated-sources/antlr4/BatfishTopologyLexer.tokens
./projects/batfish/target/generated-sources/antlr4/BatfishTopologyParser.tokens
./projects/batfish/target/generated-sources/antlr4/CiscoLexer.tokens
./projects/batfish/target/generated-sources/antlr4/CiscoParser.tokens
./projects/batfish/target/generated-sources/antlr4/EosRoutingTableLexer.tokens
./projects/batfish/target/generated-sources/antlr4/EosRoutingTableParser.tokens
./projects/batfish/target/generated-sources/antlr4/FlatJuniperLexer.tokens
./projects/batfish/target/generated-sources/antlr4/FlatJuniperParser.tokens
./projects/batfish/target/generated-sources/antlr4/FlatVyosLexer.tokens
./projects/batfish/target/generated-sources/antlr4/FlatVyosParser.tokens
./projects/batfish/target/generated-sources/antlr4/GNS3TopologyLexer.tokens
[...]

PR #197, opened 2017-07-12
antlr4-maven-plugin: recompile grammars when imports have changed

dhalperi:
The antlr4-maven-plugin already has dependency analysis that's supposed to do this, but it's got some bugs. I submitted several bug fixes upstream in https://github.com/antlr/antlr4/pull/1945.

This PR checks in a hacked version of the antlr4-maven-plugin with improved dependency analysis, as a stopgap until the fixes are accepted upstream.


dhalperi:
This is for #195

dhalperi:
Tho clearly, currently broken. Investigating.

projects/batfish/pom.xml
```diff
@@ -53,6 +53,7 @@
             <configuration>
               <includes>**/CiscoLexer.g4,**/CiscoParser.g4</includes>
               <libDirectory>${basedir}/src/main/antlr4/org/batfish/grammar/cisco</libDirectory>
+              <statusDirectory>${project.build.directory}/maven-status/antlr4-cisco</statusDirectory>
```
arifogel: Did you only need to do his for grammars with imports?

dhalperi: yes

arifogel: um tests are failing

dhalperi: Yeah something about how I'm installing the plugin is breaking vs working when installing dependencies.

PR #198, opened 2017-07-12
Support for Cisco 'ip nat pool' statement variants

xyzsa:
Parse two new variants of ip nat pool statements for CIsco IOS (see: http://www.cisco.com/c/en/us/td/docs/ios-xml/ios/ipaddr_nat/configuration/15-mt/nat-15-mt-book/iadnat-addr-consv.html). These worked till 0.22 release, but errors since then.

ip nat pool example-1-dynamic-nat-pool 192.168.0.128 192.168.0.254 prefix-length 24

ip nat pool example-2-dynamic-nat-pool 192.168.0.128 192.168.0.254 netmask 255.255.255.0


projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoParser.g4
```diff
@@ -505,6 +505,11 @@ ip_nat_pool
 :
    POOL name = variable PREFIX_LENGTH prefix_length = DEC NEWLINE
    ip_nat_pool_range*
+   | POOL name = variable first = IP_ADDRESS last = IP_ADDRESS
```
arifogel: If your tests pass I would imagine this is technically correct. However it is very hard to discern what is going on here. Can you please add parentheses around each alternative of the pipe for clarity?

Issue #98, opened 2017-04-12
add testrig creation timestamp


arifogel:
Testrigs should have a creation timestamp somewhere.

arifogel:
- Use a Data object to store time.
  - We should take the time right before parsing starts, and store it in the ParseVendorConfigurationAnswerElement.
  - We should take the time right before ConvertConfigurationAnswerElement is written to disk, and store it in the ConvertConfigurationAnswerElement
  - These times should be available in the InitInfoAnswerElement, but only if a flag (to be created for this ticket) is set. The reason for this is that we still need deterministic InitInfoAnswerElement for tests.

kidsbear:
Ya. Then the initInfoQuestioinPlugin will contain one more flag right? I am thinking should I create a inner structure to hold these additional flags? I think we might more parameters when implementing the ‘pretty print’ for verbose error.

arifogel:
@kidsbear Not InitInfoQuestionPlugin, but InitInfoQuestionPlugin.InitInfoQuestion (inner class).
I want to stress that this is not about pretty printing, but about the data held by the InitInfoAnswerElement. For now, I think there should just be a flag called 'includeTimestamps' or something like that. We can complicate the structure later if it becomes necessary.

dhalperi:
Looking at Wei's PR, I think this is the wrong approach.

* Why doesn't every question just have a timestamp when it was uploaded?
* Why doesn't every answer just have a timestamp when it was created? (Maybe more, e.g., when the worker started working on it, etc.)

If we take a single, principled, abstract approach we don't have to think too hard about how every since question/answer timestamp should be labeled or numbered.

As for the `includeTimestamps` flag -- this is a mistake. The timestamp should always be in the entities. The current testing framework is too brittle w.r.t. non-determinism. There are better solutions, e.g., using dependency injection of a fake clock to reduce non-determinism and make reliable tests.

These are longer-term efforts, but I think we're getting to where we can't avoid working on them.

arifogel:
Makes sense.

kidsbear:
This issue is in progress at #202 , currently add timestamps to the beginning of parsing step, and the end of the convert step. Since real clock is non-determinism, need to determine how to form a fake clock for testing and which instance should have a timestamp, and when should the timestamp been set before further work.

corinaminer:
Where did you guys leave off on this? I'm interested on getting those timestamps soon @arifogel @dhalperi

corinaminer:
[...]

PR #202, opened 2017-07-13
add timestamp to initInfoAnswerElement

kidsbear:
This is a naive version for #98


kidsbear:
Not finished. Currently should not pass the .ref test.

dhalperi:
Can you add a comment to the original issue that the work is in progress here, then close this PR?

Issue #212, opened 2017-07-14
list-questions is broken


virtuald:
This is a different kind of failure than #211 . 

```
batfish> list-questions 
exception: org.codehaus.jettison.json.JSONException: JSONObject["questionlist"] is not a JSONArray.
	at org.codehaus.jettison.json.JSONObject.getJSONArray(JSONObject.java:493)
	at org.batfish.client.BfCoordWorkHelper.listQuestions(BfCoordWorkHelper.java:957)
	at org.batfish.client.Client.listQuestions(Client.java:1753)
	at org.batfish.client.Client.processCommand(Client.java:2079)
	at org.batfish.client.Client.processCommand(Client.java:1985)
	at org.batfish.client.Client.runInteractive(Client.java:2319)
	at org.batfish.client.Client.run(Client.java:2272)
	at org.batfish.allinone.AllInOne.run(AllInOne.java:117)
	at org.batfish.allinone.Main.main(Main.java:6)
```

It appears that the returned question list is actually a JSON Object, not an array. And unfortunately, the object is empty, so even if it interpreted it correctly it wouldn't tell me what kinds of questions I can ask. :(

dhalperi:
Hey @virtuald,

These are nice bug reports, but you didn't really provide enough information to reproduce. Can you please say a bit more about how you get there?

Did you just run `allinone` and enter `list-questions`, or were there prior commands, etc?

virtuald:
1. Load a testrig
2. Run list-questions

arifogel:
list-questions show no results because you have not run load-questions. You
need a library of questions to load.

dhalperi:
@arifogel -- either way, the current behavior is a bug – it should not crash as such but should return an empty list. Wei's on it!

virtuald:
I would recommend if an empty list is returned that the client outputs something like "have you ran load-questions first?"

dhalperi:
After #216, this function returns a list instead of crashing. This restores the behavior before recent refactorings.

Further messaging improvements are welcome as pull requests!

Issue #211, opened 2017-07-14
get without arguments stack trace


virtuald:
```
batfish> get 
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.LinkedList.checkElementIndex(LinkedList.java:555)
	at java.util.LinkedList.get(LinkedList.java:476)
	at org.batfish.client.Client.get(Client.java:1224)
	at org.batfish.client.Client.processCommand(Client.java:2040)
	at org.batfish.client.Client.processCommand(Client.java:1985)
	at org.batfish.client.Client.runInteractive(Client.java:2319)
	at org.batfish.client.Client.run(Client.java:2272)
	at org.batfish.allinone.AllInOne.run(AllInOne.java:117)
	at org.batfish.allinone.Main.main(Main.java:6)
```

Maybe it could enumerate the things that one can get?

virtuald:
Speaking of which, `list-questions` doesn't work either:

```
batfish> list-questions
exception: org.codehaus.jettison.json.JSONException: JSONObject["questionlist"] is not a JSONArray.
	at org.codehaus.jettison.json.JSONObject.getJSONArray(JSONObject.java:493)
	at org.batfish.client.BfCoordWorkHelper.listQuestions(BfCoordWorkHelper.java:957)
	at org.batfish.client.Client.listQuestions(Client.java:1753)
	at org.batfish.client.Client.processCommand(Client.java:2079)
	at org.batfish.client.Client.processCommand(Client.java:1985)
	at org.batfish.client.Client.runInteractive(Client.java:2319)
	at org.batfish.client.Client.run(Client.java:2272)
	at org.batfish.allinone.AllInOne.run(AllInOne.java:117)
	at org.batfish.allinone.Main.main(Main.java:6)
```

virtuald:
... and there are other commands that don't check the number of parameters.

* set-loglevel
* set-environment
* test

... etc.

kidsbear:
Implemented by PR#217, close for now.

dhalperi:
@kidsbear Normally, don't close bugs until after the PR is merged.

PR #214, opened 2017-07-14
fix list-question, add size check for empty get command

kidsbear:
For #211, #212


projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -575,6 +574,10 @@ public Client(String[] args) throws Exception {
    private boolean addBatfishOption(
          String[] words, List<String> options,
          List<String> parameters) {
+      if (parameters.isEmpty()) {
```
dhalperi: please check that `words` is nonempty before accessing `words[0]` :)

kidsbear:
Split into two tickets.

PR #217, opened 2017-07-14
Issue 211 empty argument validation

kidsbear:
For issue #211


projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -792,6 +799,10 @@ private boolean answerType(
 
    private boolean cat(String[] words)
          throws IOException, FileNotFoundException {
+      if (words.length == 1) {
+         _logger.errorf(words[0] + " requires additional arguments\n");
```
dhalperi: Every one of these errors should be descriptive.

E.g., it should not be "cat requires additional parameters". It should be "usage: cat <filename>" or something like that.

dhalperi: @ratulm is there a precedent for how usage error messages are shown?

projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
```diff
@@ -787,4 +811,55 @@ public void testProvideOptionalParameterWithValue() throws IOException {
       Client.checkVariableState(variables);
    }
 
+   // Tests for processCommand method
+   @Test
+   public void testProcessEmptyArgCommand() throws Exception {
+      Client client = new Client(new String[]{"-runmode", "gendatamodel"});
```
dhalperi: test the actual messages, each in its own test.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -792,6 +799,10 @@ private boolean answerType(
...
+         return false;
+      }
```
dhalperi: the error here is not if the length is 1. The error here is if the length != 2. Make sure the length check rejects invalid commands with *too many* arguments also.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
ratulm: From the delAnalysis function:
      if (options.size() != 0 || parameters.size() != 1) {
         _logger.errorf("Invalid arguments: %s %s\n", options.toString(),
               parameters.toString());
         printUsage(Command.DEL_ANALYSIS);
         return false;
      }

Basically, an explanatory message of what was wrong, followed by usage of the command.

kidsbear: Do I need to explicitly check the number of options and parameters? If so, how can I determine what is the exact number of args for options and parameters required for each command? For example, in Command.buildUsageMap():
      descs.put(ADD_BATFISH_OPTION, new Pair<>(
            "<option-key> <option-value>",
            "Additional options to pass to Batfish"));
It looks like ADD_BATFISH_OPTION command requires two inputs? They are options or parameters? When we actually use ADD_BATFISH_OPTION, we do things like : add-batfish-option haltonparseerror .  It actually accept a single argument?

kidsbear:
@dhalperi I am not sure why this PR contains changes in the previous closed PR, is there any easy ways to remove them?

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -953,15 +953,13 @@ public JSONObject listAnalyses(String containerName) {
...
 
-         String[] questionList = new String[questions.length()];
+         String[] questionList = new String[questionArray.length()];
 
-         int i = 0;
-         for (Iterator it = questions.keys(); it.hasNext(); ) {
-            questionList[i] = (String) it.next();
-            ++i;
+         for (int index = 0; index < questionArray.length(); index++) {
+            questionList[index] = questionArray.getString(index);
```
dhalperi: please undo the changes to this file.

projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrTest.java
```diff
@@ -1,90 +0,0 @@
-package org.batfish.coordinator;
```
dhalperi: why delete this ?

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -855,15 +875,15 @@ private Path createTempFile(String filePrefix, String content) {
...
-         return false;
-      }
       if (options.size() != 0 || parameters.size() != 1) {
          _logger.errorf("Invalid arguments: %s %s\n", options.toString(),
                parameters.toString());
          printUsage(Command.DEL_ANALYSIS);
          return false;
       }
+      if (!isSetContainer(true)) {
+         return false;
```
kidsbear: The function call will print the error message.

PR #224, opened 2017-07-17
more robust parameter specification in java client

ratulm:


tests/basic/commands
```diff
@@ -19,17 +19,17 @@ test tests/basic/isisLoopbacks.ref get isisLoopbacks
...
 test tests/basic/routes.ref get routes
 test tests/basic/routes-diff.ref get routes differential=true
 test tests/basic/selfAdjacencies.ref get selfAdjacencies
-test tests/basic/traceroute-1-2.ref get traceroute ingressNode="as1core1" | dstIp="2.128.0.101"
-test tests/basic/traceroute-2-1.ref get traceroute ingressNode="host2" | dstIp="1.0.1.1"
-test tests/basic/multipath-host1.ref get reachability type="multipath" | ingressNodeRegex="host1" | srcIps=["2.128.0.0"] | dstIps=["3.0.1.2"] | ipProtocols=["TCP"] | srcPorts=[0] | dstPorts=[0]
-test tests/basic/multipath-host2.ref get reachability type="multipath" | ingressNodeRegex="host2" | srcIps=["2.128.0.0"] | dstIps=["1.0.1.1"] | ipProtocols=["UDP"] | srcPorts=[0] | dstPorts=[0]
+test tests/basic/traceroute-1-2.ref get traceroute ingressNode="as1core1", dstIp="2.128.0.101"
+test tests/basic/traceroute-2-1.ref get traceroute ingressNode="host2", dstIp="1.0.1.1"
+test tests/basic/multipath-host1.ref get reachability type="multipath", ingressNodeRegex="host1", srcIps=["2.128.0.0"], dstIps=["3.0.1.2"], ipProtocols=["TCP"], srcPorts=[0], dstPorts=[0]
```
dhalperi: sanity check: this will work if the arguments have commas in them? e.g., more than 1 `srcIp`?

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1898,26 +1903,23 @@ private void logOutput(FileWriter outWriter, String message) {
...
+            String key = (String)keys.next();
+            String value = jsonParamsObject.get(key).toString();
             _logger.debugf("key=%s value=%s\n", key, value);
 
             parameters.put(key, value);
          }
+         return parameters;
+      }
+      catch (JSONException e){
+         throw new BatfishException("Failed to parse paramters. (Are all key-value pairs separated by commas? Are all values valid JSON?)", e);
```
dhalperi: typo: parameters.

tests/basic/commands
ratulm: yes, i got away from string.split entirely.  the "spec" is that the parameters should be a comma-separated list of key=value pairs, and values should be valid JSON.

dhalperi: ack. another thing to check then, is that none of the keys will ever have hyphens (or other JSON-not-allowed tokens) in them? Or making them be JSON will require extra, perhaps odd, escaping.

dhalperi: (Consider adding a unit test or 10 :p.

Otherwise, looks good to me when you're happy with it.)

ratulm: i am actually casting things to JSON internally, so any non-JSON-ism will throw an error automatically.

i am easy to please :)  the current tests are testing this feature anyway, hence the modified tests.

Issue #233, opened 2017-07-18
testrig parsing: iptables parse error not reported on error log


virtuald:
You have to enable debug logging to see the error. Here's what I see without debug logging enabled:

```
batfish> init-testrig test_rigs/test
Uploaded testrig. Parsing now.
Status: FAILURE
org.batfish.common.BatfishException: Batfish job failed
   at org.batfish.main.Driver$1.run(Driver.java:406)
Caused by: org.batfish.common.BatfishException: Key not found for iptables: testrig/iptables/n7.v4
   at org.batfish.main.Batfish.serializeHostConfigs(Batfish.java:4123)
   at org.batfish.main.Batfish.serializeVendorConfigs(Batfish.java:4298)
   at org.batfish.main.Batfish.run(Batfish.java:3926)
   at org.batfish.main.Driver$1.run(Driver.java:374)
```

When I enable debug logging via `set-loglevel debug`, I get a nice error message informing me that testrig/iptables/n7.v4 is broken. 

```
...
Error context lines:
   1:      # This is in iptables-save format
   2:      # @see http://www.faqs.org/docs/iptables/iptables-save.html
   3:      *filter
   4:      :INPUT DROP [0:0]
   5:      :FORWARD DROP [0:0]
>>>6:      -A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
   7:      COMMIT
```

ratulm:
@wenfei-wu - did you fix this issue recently? do you mind taking a look?

kidsbear:
@ratulm I think it is a different issue because of inconsistency - for other parsing errors, we store the error information in AnswerElements instead of crash (except haltonprocesserror is set to true). However, current code throws an Exception for this error - https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L4125. I am trying to fix it, but I don't have a good idea in mind now. Any suggestions if you have time?

ratulm:
@arifogel should be able to help here.

ratulm:
addressed by 361b44ce189fa9d759cb30d2976d20813cfadb09

virtuald:
Ok, so the output is better, as my sample testrig gives the following:

```
Init'ed and set active container
Uploaded testrig. Parsing now.
Status: SUCCESS
PARSING SUMMARY
  h1.iptables: failed to parse
STATISTICS
  Parsing results:
    Parsed successfully: 1
    Failed to parse: 1
```
[...]

PR #237, opened 2017-07-19
Add error information to answer element instead of throw an exception

kidsbear:
For issue #233


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4118,16 +4147,11 @@ private void serializeHostConfigs(Path testRigPath, Path outputPath,
                   hostConfig.getIptablesFile());
             String relativePathStr = _testrigSettings.getBasePath()
                   .relativize(path).toString();
-            if (!iptablesConfigurations.containsKey(relativePathStr)) {
-               for (String key : iptablesConfigurations.keySet()) {
-                  _logger.errorf("key : %s\n", key);
-               }
-               throw new BatfishException(
-                     "Key not found for iptables: " + relativePathStr);
+            if (iptablesConfigurations.containsKey(relativePathStr)) {
```
kidsbear: I am not sure if the iptable file failed, should I set the parse_status of the host file to be failed? It was success.

projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
```diff
@@ -251,26 +251,10 @@ public void testValidJsonPath() throws IOException {
...
    }
 
-// ratul commented out this test
-//     parseParaValue should allow strings like "I am not valid" to parse.
-//     i can't think of what that new function will not parse
-//       TODO: remove test or give it some teeth
-//   // Tests for parseParaValue method
-//   @Test
-//   public void testParseInvalidJsonContent() {
-//      String invalidJsonContent = "I am not valid";
```
dhalperi: @ratulm I don't understand -- the java string `"I am not valid"` is not valid JSON, because the JSON string would be `"\"I am not valid\""` in Java.

(This follows from `"5"` being the JSON number 5 not and `"\"5\""` being the JSON string "5", and from `"false"` being the JSON boolean whereas `"\"false\""` is a JSON string of the word false.)

ratulm: so, where this is getting in the way is people typing   
    batfish> get questionType  a="b"
to specify parameter value b (string) for a. perhaps the validation needs to be differently structured.

kidsbear: @ratulm I think this function does the different thing. After we read the question json file, and stored variable name and its value into the map, this function just check the value of each variable is valid json value. example: 
{
    "variable" : invalid value
}
this should be invalid, since a string in json must inside "".
Sorry for the unclear function name, I named it based on the parameter name in origin code, it actually should be something like validateVariableValue(value).

ratulm:
can you make both the test and the following command work?
  batfish> answer reachability  actions=["drop"], ingressNodeRegex="n(7|8)"

kidsbear:
Will create a new PR for it~

projects/batfish/src/main/java/org/batfish/main/Batfish.java
dhalperi: I think this is a good question. I'm not sure I fully understand the parsing philosophy. @arifogel may have thoughts.

PR #239, opened 2017-07-20
Turn on Xlint:all and fix/supress all errors

dhalperi:


projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -958,9 +958,11 @@ public JSONObject listAnalyses(String containerName) {
 
          String[] questionList = new String[questions.length()];
 
+         @SuppressWarnings("unchecked")
```
arifogel: This is the wrong solution. You should simply declare it as Iterator<?> instead of Iterator. Specifically, the cast is pointless if this actually were an Iterator<String>, and also the suppression is a dangerous one that should only be used in rare cases.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -841,8 +841,7 @@ private void checkQuestionsDirExists() {
    }
 
    @Override
```
arifogel: Isn't this an Autocloseable method that should throw Exception? I didn't even realize you were allowed to change the throws declaration of an overridden method.

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
dhalperi: I don't quite understand the argument here. The library was written without any generics, but in practice all JSON keys are strings, and the library bears this out.

What's the problem with suppressing a cast you know is correct?

(There is no cast later FWICT.)

I did add a comment to the cast, and also simplified the code even further using `Iterators` library call.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
dhalperi: You should not claim to throw more than you do throw, and yes, it's compatible to throw fewer exceptions than the method you're overriding.

(In this particular case, the compile warning was that the method might throw InterruptedException, which is not semantically legal in a close method).

Issue #236, opened 2017-07-19
Coordinator should have a way to get info about a container


dhalperi:
Right now, we can initialize a container, delete it, or list things in it. But if we want to know whether it exists at all, the best way to do so is to simply try initializing / listing and then handling failures gracefully.

Instead, it would be nice to be able to `GET /container/<container name>` to see that it exists and maybe what testrigs are in it, etc.

dhalperi:
Supported in #241

PR #241, opened 2017-07-20
add get-container functionality

kidsbear:
For issue #236


ratulm:
Hey, Wei - I am curious if this functionality is not something that you could have provided through the existing listContainers() API call?

kidsbear:
Actually, I am not sure about this. The existing listContainers() API call requires the api key has the authority to list all containers under the folder, I am not sure if we also need the authority for get-container call. Also, if we only want to see if the container exist, we can check if the container is in the return list, I am not sure whether we need more information(testrig, etc).

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -555,6 +555,46 @@ public JSONArray getAnswer(
       }
    }
 
+   @POST
+   @Path(CoordConsts.SVC_RSC_GET_CONTAINER)
+   @Produces(MediaType.APPLICATION_JSON)
+   public JSONArray getContainer(
```
ratulm: getContainer() is perhaps not the right name for this function. Perhaps:
 -  getContainerInfo() if the plan down the line is to provide more information about the container (e.g., time of creation)
- doesContainerExists() is the intent is to simply let the clients check if the container with the provided name exists. (that is the purpose this function is currently serving, as no other information is provided.)

ratulm: getting back to your response, @kidsbear, i see the difference between the two functions. yes, being able to check if a containerName exists without having access to it is useful (and not supported by listContainers).

and by that token, the purpose of this function should stay as something close to doesContainerExist() as no other info should be provided with clients with API keys that don't have access to the container.

ratulm:
Sorry, I may have been unclear, but this behavior is unclear and not intuitive. 

Suggestions:
 - instead of existContainer, call things checkContainerExists
 - return just true or false accordingly

Makes sense?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
kidsbear: Now the command used is 'exist-container', will output a string whether indicates the container does not exist, or the creation time of the container

kidsbear:
Sure, will fix it later:)

dhalperi:
I don't buy it.i think getContainer is the right RESTful name for the
function. -- GET on a particular container to get its information.

Let's discuss tomorrow

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -575,6 +574,18 @@ public String getAnswer(
...
 
+   /**
+    * Return a {@link SortedSet SortedSet} contains all subdirectories under the path
+    * {@code containerDir}
+    */
+   public SortedSet<String> getContainer(Path containerDir) {
+      SortedSet<String> testrigs = new TreeSet<>(
+            CommonUtil.getSubdirectories(containerDir).stream()
+                  .map(dir -> dir.getFileName().toString())
+                  .collect(Collectors.toSet()));
```
ratulm: Technically, testrigs do not sit directly under the container directory. They sit under <containerdir>/BfConsts.RELPATH_TESTRIGS_DIR. (The constant is "" right now for backward compatibility but we'll make it something else at some point.)

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -448,6 +448,52 @@ private ClientBuilder getClientBuilder() throws Exception {
...
+               + " " + response + "\n");
+
+         if (response.getStatus() != Response.Status.OK.getStatusCode()) {
+            _logger.errorf("GetContainer: Did not get an OK response\n");
+            _logger.errorf(response.readEntity(String.class) + "\n");
+            return null;
+         }
+
+         String testrigs = response.readEntity(String.class) + "\n";
+         return testrigs;
```
ratulm: Consistent with server-side code, use 'containerInfo' as variable name instead of 'testrigs'?

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -385,6 +385,52 @@ private ClientBuilder getClientBuilder() throws Exception {
         .register(MultiPartFeature.class);
   }
 
+  /**
+   * Return true if successfully get the information of the container,
```
dhalperi: This does not seem up to date.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1322,6 +1322,23 @@ private boolean getAnswer(
     return parameters;
   }
 
+  /**
+   * Get information of the container(first element in {@code parameters}) return true if
```
dhalperi: Please clean up this javadoc. Space before `(` and I can't tell if this is 1 or 2 sentences, etc. Use punctuation.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -531,6 +531,20 @@ public String getAnswer(
     return answer;
   }
 
+  /**
+   * Return a {@link SortedSet SortedSet} contains all subdirectories under the path {@code
```
dhalperi: A Container is not a sorted set -- it should be a `Container`. Please make a proper java object that represents a container. One field it may have is `testrigs`, which could be an ordered list or set of testrigs.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -582,6 +582,59 @@ public JSONArray getAnswer(
...
+        return Response.status(Response.Status.NOT_FOUND)
+            .entity("Container '" + containerName + "' not found")
+            .type(MediaType.TEXT_PLAIN)
+            .build();
+      }
+
+      checkContainerAccessibility(apiKey, containerName);
+
+      SortedSet<String> containerInfo = Main.getWorkMgr().getContainer(containerDir);
+      return Response.ok(containerInfo.toString(), MediaType.TEXT_PLAIN).build();
```
dhalperi: This should be a JSON response of a container.

projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrTest.java
```diff
@@ -118,4 +118,39 @@ public void listSortedQuestionNames() throws IOException {
...
     assertThat(questions.toString(), equalTo("[access, initinfo, nodes]"));
   }
+
+  @Test
+  public void getEmptyContainer() throws IOException {
+    String containerName = "myContainer";
+    _folder.newFolder(containerName);
+    Main.mainInit(new String[] {"-containerslocation", _folder.getRoot().toString()});
+    Path containerDir = Paths.get(_folder.getRoot().toPath().resolve(containerName).toString());
+    SortedSet<String> containerInfo = _manager.getContainer(containerDir);
```
dhalperi: Most of the lines in these tests are identical. Can you pull all the setup of the Main and the containers to a helper function and simplify the tests dramatically?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -582,6 +582,59 @@ public JSONArray getAnswer(
...
+   *
+   * @param apiKey The API key of the client
+   * @param clientVersion The version of the client
+   * @param containerName The name of the container in which the question was asked
+   * @return A string representation of the container
+   */
+  @POST
+  @Path(CoordConsts.SVC_RSC_GET_CONTAINER)
+  @Produces(MediaType.TEXT_PLAIN)
+  public Response getContainer(
```
dhalperi: Where is the test of this function?

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -0,0 +1,62 @@
...
+  public void setTestrigs(SortedSet<String> testrigs) {
+    _testrigs = testrigs;
+  }
+
+  @JsonProperty(ANALYSIS_VAR)
+  public void setAnalysis(Path analysis) {
+    _analysis = analysis;
+  }
+
+  public String listTestrigs() {
```
dhalperi: drop this function – it's just a thin wrapper around `getTestrigs`, isn't it?

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1322,6 +1323,24 @@ private boolean getAnswer(
     return parameters;
   }
 
+  /**
+   * Get information of the container(first element in {@code parameters}).
+   * <p> </p>
+   * Returns true if successfully get container information, false otherwise
```
dhalperi:
```
Get information of the container (first element in {@code parameters}).

<p>Returns {@code true} if successfully get container information, {@code false} otherwise.
```

PR #243, opened 2017-07-21
Allow init container with name

kidsbear:
Allow client side to initialize a container with specific container name instead of providing prefix.

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -1513,7 +1513,7 @@ public class Client extends AbstractClient implements IClient {
          }
          else {
             if (parameters.isEmpty()) {
-               _logger.error("Container name is required when option '-setname' is set");
+               _logger.error("Container name is required when option '-setname' is set\n");
                printUsage(Command.INIT_CONTAINER);
                return false;
             }
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -1502,26 +1502,16 @@ public class Client extends AbstractClient implements IClient {
    }
 
    private boolean initContainer(List<String> options, List<String> parameters) {
-      if (!isValidArgument(options, parameters, 1, 0, 1, Command.INIT_CONTAINER)) {
-         return false;
-      }
-      if (!options.isEmpty()) {
-         if (!options.get(0).equals("-setname")) {
-            _logger.errorf("Invalid options: %s\n", options.toString());
-            printUsage(Command.INIT_CONTAINER);
+      if (options.contains("-setname")) {
+         if (!isValidArgument(options, parameters, 1, 1, 1, Command.INIT_CONTAINER)) {
             return false;
          }
-         else {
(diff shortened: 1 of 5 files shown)


kidsbear:
@ratulm We don't have Option class for client now right? For this particular one, I just compare the option string, should I create a new Option class for it?

ratulm:
what will the Option class do?

kidsbear:
Defines a bunch of constant options? For now we only have one option(-setname), if we are planning to add more options in the future, an Option class will make it easier to control?

ratulm:
i see. good idea but i'd say lets skip that generalization for now; we can get to it later if/when we add more options.

projects/batfish-client/src/main/java/org/batfish/client/Command.java
```diff
@@ -173,7 +173,7 @@
                   "<analysis-name> <question-directory>",
                   "Initialize a new analysis for the container"));
       descs.put(INIT_CONTAINER, new Pair<>(
-            "[<container-name-prefix>]",
+            "<-setname><container-name> or [<container-name-prefix>]",
```
dhalperi: this does not seem like the right syntax. 

`[-setname <container-name>] or [<container-name-prefix>]` seems closer. But probably it should be a multiline string with different options on different lines. Are there existing examples to mimic?

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1501,15 +1501,31 @@ private boolean help(List<String> options, List<String> parameters) {
...
-         _logger.errorf("Invalid arguments: %s\n", Arrays.toString(words));
-         printUsage(Command.INIT_CONTAINER);
+   private boolean initContainer(List<String> options, List<String> parameters) {
+      if (!isValidArgument(options, parameters, 1, 0, 1, Command.INIT_CONTAINER)) {
          return false;
       }
-      String containerPrefix = (words.length > 1) ? words[1]
-            : DEFAULT_CONTAINER_PREFIX;
-      _currContainerName = _workHelper.initContainer(null, containerPrefix);
+      if (!options.isEmpty()) {
```
dhalperi: It's okay, but I bet you could simplify this code a lot by ordering your `if`s better and unnesting. The current code is way more complex than it needs to be IMO ; hard to understand.

Example:

```java
if (options.contains("-setname")) {
   validate using =1 option, using =1 parameter
   _currContainerName = _workHelper.initCountainer(parameters.get(0), null));
}
else {
   validate using =0 options, 0 or 1 parameter.
   existing 2 lines of logic
}
final 3 lines of logs.
```

projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrTest.java
```diff
@@ -87,4 +88,41 @@ public void listSortedQuestionNames() throws IOException {
       assertThat(questions.size(), is(3));
       assertThat(questions.toString(), equalTo("[access, initinfo, nodes]"));
    }
+
+   // Tests for initContainer method
```
dhalperi: drop this comment -- such comments are unnecessary as the test name should make it clear what it's testing.

dhalperi:
Nice!

PR #249, opened 2017-07-22
Store input parameter values as JsonNode instead of String

kidsbear:


dhalperi:
+1 @ratulm 's comment – separate PRs is better. I don't see any reason to hold off on merging.

Thanks, Wei!

PR #258, opened 2017-07-26
Make explicit topology json

ratulm:


projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -4,7 +4,6 @@
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.collection.IsEmptyCollection.empty;
-import static org.junit.Assert.assertThat;
```
dhalperi: `import static` + just using `assertThat` (rather than `Assert.assertThat`) is generally preferred in unit test code. Just for sanity/cleanliness.

projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -75,10 +77,33 @@ private Batfish initBatfish() {
...
+      "\"node2\" : \"as2border1\"," +
+      "\"node2interface\" : \"GigabitEthernet0/0\"" +
+     "}" +
+    "]";
+
+    Path topologyFilePath = CommonUtil.createTempFileWithContent("testParseTopologyJson", topologyJson);
+    Batfish batfish = initBatfish();
+    Topology topology = batfish.parseTopology(topologyFilePath);
+    Assert.assertEquals(topology.getEdges().size(), 2);
+  }
```
dhalperi: test empty topology and bad topology.

PR #259, opened 2017-07-27
Fix 0.29 regressions

arifogel:


dhalperi:
I pushed a recompiled version of the jars that still passes, just so we can be absolutely sure it's compiled from the ref in the BUILDING.md file.

projects/question/src/main/java/org/batfish/question/jsonpath/NodesPathQuestionPlugin.java
```diff
@@ -1,289 +0,0 @@
-package org.batfish.question.jsonpath;
```
dhalperi: Did you delete this file intentionally?

arifogel: Yes. We don't use it.

arifogel: It is a projected version of JsonPathQuestionPlugin. It only existed for historical reasons, as JsonPathQuestionPlugin is more general. All references to it are gone.

arifogel: Well I was wrong. Back it goes.

PR #261, opened 2017-07-27
prepare for 0.29.3

arifogel:
- update poms, refs


arifogel:
This PR follows changes forcibly pushed into master. These changes only affect installed versions of batfish (via package scripts), and not core engine.

PR #265, opened 2017-07-28
RESTful API based on Jersey and Jackson

dhalperi:
Two example calls, to get information about all containers or a single
container.

Also adds a default exception mapper to return stack traces on failures, a redirect, and testing.

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/ExceptionMapper.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/ExceptionMapper.java
@@ -1,0 +1,22 @@
+package org.batfish.coordinator;
+
+import com.google.common.base.Throwables;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class ExceptionMapper implements javax.ws.rs.ext.ExceptionMapper<Exception> {
+  @Override
+  public Response toResponse(Exception exception) {
+    if (exception instanceof WebApplicationException) {
+      // An exception that already maps directly to a standard HTTP Response, likely thrown
+      // somewhere inside Jersey aka if the user typed a bad URI.
+      return ((WebApplicationException) exception).getResponse();
+    }
+
+    exception.printStackTrace();
+    String stackTrace = Throwables.getStackTraceAsString(exception);
+    return Response.status(500).entity(stackTrace).build();
+  }
+}
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="EntryPointsManager">
+    <writeAnnotations>
+      <writeAnnotation name="javax.ws.rs.core.Context" />
+    </writeAnnotations>
+  </component>
   <component name="MavenProjectsManager">
     <option name="originalFiles">
       <list>
--- a/projects/batfish-common-protocol/pom.xml
+++ b/projects/batfish-common-protocol/pom.xml
@@ -150,18 +150,26 @@
 
     <!-- Test dependencies. -->
     <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava-testlib</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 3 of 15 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -16,7 +16,7 @@
             <package name="" withSubpackages="true" static="false" />
```
dhalperi: separate out any config change and incidental format changes to a separate PR

virtuald:
It seems like you guys have spent a ton of effort creating various JSON-based REST APIs... why not use a high performance RPC toolkit like [gRPC](https://grpc.io/) instead? Would save you thousands of lines of code in server code and validation code -- and protobuf is easier to read/modify than JSON schemas et al.

dhalperi:
@virtuald that's a good question for a beer :), but here's a few answers...

* Batfish is not a system where the RPC latency is likely to be a bottleneck, so API usability matters more to me.
* REST APIs are ultimately more user friendly than anything else. You can write a client in any language, for example.
* If something like Apache Thrift had a REST adapter, I'd use it.
* And finally, there are many reasons why I don't want to use gRPC/Protobuf in particular for any enterprise software. Among them:
    * Google doesn't actually use their open-source versions (they run at internal HEAD),
    * Lack of API or binary stability in practice,
    * Very few officially-supported languages (no TypeScript!?),
    * other past experiences from my past open source and employment history.

virtuald:
To me, REST seems like a lot of extra boilerplate. But, I suspect this PR isn't the place to discuss it; and you have a rationale for it, so carry on! 👍

dhalperi:
@kidsbear I think this is ready for review now. Be merciless!

projects/coordinator/src/main/java/org/batfish/coordinator/ExceptionMapper.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class ExceptionMapper implements javax.ws.rs.ext.ExceptionMapper<Exception> {
+  @Override
+  public Response toResponse(Exception exception) {
+    if (exception instanceof WebApplicationException) {
+      // An exception that already maps directly to a standard HTTP Response, likely thrown
+      // somewhere inside Jersey aka if the user typed a bad URI.
+      return ((WebApplicationException) exception).getResponse();
```
kidsbear: This actually gives an empty entity body, can use statements like : 
return Response.status(((WebApplicationException) exception).getResponse().getStatus())
          .entity(Throwables.getStackTraceAsString(exception)).build();
to specify an entity body. Also, since our code throws BatfishException everywhere, I think it might be useful to provide handler for BatfishException, use status code other than 500. (maybe BAD_REQUEST? Since most of them are caused by invalid client inputs.)

kidsbear:
Great example of testing in Jersey, thanks Dan~ :)

dhalperi:
This is for #278.

projects/coordinator/src/main/java/org/batfish/coordinator/ExceptionMapper.java
dhalperi: I disagree – BatfishException is used for absolutely everything, E.g., error opening a file, which may well be a backend error. We need to rethink our exception handling if we're going to be able to distinguish user and backend errors. For now, 500 is a safe bet.

dhalperi:
Thanks @kidsbear

morfoh:
> @virtuald that's a good question for a beer :), but here's a few answers...
> 
>     * Batfish is not a system where the RPC latency is likely to be a bottleneck, so API usability matters more to me.
> 
>     * REST APIs are ultimately more user friendly than anything else. You can write a client in any language, for example.
> 
>     * If something like Apache Thrift had a REST adapter, I'd use it.
> 
>     * And finally, there are many reasons why I don't want to use gRPC/Protobuf in particular for any enterprise software. Among them:
>       
>       * Google doesn't actually use their open-source versions (they run at internal HEAD),
>       * Lack of API or binary stability in practice,
>       * Very few officially-supported languages (no TypeScript!?),
>       * other past experiences from my past open source and employment history.

No matter if REST or Protobuf. An API without documentation is no API imo! Obviously I came to that ticket because I am looking for an API documentation for people that do not want to waste time by reverse engineering Java and Python code.
Do you have something like that? Did you thought about adding an OpenAPI spec file?

Issue #207, opened 2017-07-13
More NullPointerExceptions


virtuald:
I'm sure I have a configuration error, but I ran into this (partial stack trace below):

```
Caused by: java.lang.NullPointerException
   at org.batfish.bdp.VirtualRouter.initStaticRib(VirtualRouter.java:699)
   at org.batfish.bdp.BdpDataPlanePlugin.lambda$computeFixedPoint$2(BdpDataPlanePlugin.java:324)
   at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
   at java.util.TreeMap$ValueSpliterator.forEachRemaining(TreeMap.java:2897)
   at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
   at java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:291)
   at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
   at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
   at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
   at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
   at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

If I add a check for a null nextHopIp at VirtualRouter:699, then I get by that error, but then there's another one...

```
at org.batfish.bdp.AbstractRib$ByteTrie.getLongestPrefixMatch(AbstractRib.java:37)
   at org.batfish.bdp.AbstractRib.longestPrefixMatch(AbstractRib.java:362)
   at org.batfish.bdp.VirtualRouter.activateStaticRoutes(VirtualRouter.java:188)
   at org.batfish.bdp.BdpDataPlanePlugin.lambda$8(BdpDataPlanePlugin.java:484)
   at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
   at java.util.TreeMap$ValueSpliterator.forEachRemaining(TreeMap.java:2897)
   at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
   at java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:291)
   at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
[...]

PR #266, opened 2017-07-28
Skipping the static routes with null nextHopIp in initStaticRib

hverma-upenn:
Fix for issue #207, details are in the comment on this commit

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
+++ b/projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
@@ -644,10 +644,12 @@ public class VirtualRouter extends ComparableStructure<String> {
         continue;
       }
       // interface route
-      if (sr.getNextHopIp().equals(Route.UNSET_ROUTE_NEXT_HOP_IP)) {
-        _staticInterfaceRib.mergeRoute(sr);
-      } else {
-        _staticRib.mergeRoute(sr);
+      if (sr.getNextHopIp() != null) {
+        if (sr.getNextHopIp().equals(Route.UNSET_ROUTE_NEXT_HOP_IP)) {
+          _staticInterfaceRib.mergeRoute(sr);
+        } else {
+          _staticRib.mergeRoute(sr);
+        }
       }
     }
   }


projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
```diff
@@ -644,10 +644,12 @@ public void initStaticRib() {
         continue;
       }
       // interface route
-      if (sr.getNextHopIp().equals(Route.UNSET_ROUTE_NEXT_HOP_IP)) {
-        _staticInterfaceRib.mergeRoute(sr);
-      } else {
-        _staticRib.mergeRoute(sr);
+      if (sr.getNextHopIp() != null) {
```
dhalperi: per our discussion:

* please add unit test
* please add `@Nullable` annotation to the getter `sr.getNextHopIp()` so that we know it can be null.

dhalperi:
Actually, it sounds like this is the wrong behavior. The actual issue is that something is gone wrong in setting up the data model, because that value should never be `null`. There's a Java value called `Route.UNSET_NEXT_HOP_IP` that is typically used in this case.

What we should do is add the null checks to the `StaticRoute` constructor that throw an exception if `null` is passed in instead of `...UNSET...`. And maybe that should also be applied to other arguments of the route constructors, as appropriate.

Here's a partial example diff:

```diff
diff --git a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java b/projects/batfish-
common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
index 41ef2d75a..0eb167a8a 100644
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
@@ -1,8 +1,11 @@
 package org.batfish.datamodel;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class StaticRoute extends AbstractRoute {
 
@@ -26,9 +29,17 @@ public class StaticRoute extends AbstractRoute {
       @JsonProperty(ADMINISTRATIVE_COST_VAR) int administrativeCost,
       @JsonProperty(TAG_VAR) int tag) {
     super(network);
+    _nextHopIp =
+        checkNotNull(
+            nextHopIp,
+            "A missing next hop IP address should be indicated by %s.",
+            Route.UNSET_ROUTE_NEXT_HOP_IP);
     _administrativeCost = administrativeCost;
-    _nextHopIp = nextHopIp;
-    _nextHopInterface = nextHopInterface;
+    _nextHopInterface =
+        checkNotNull(
+            nextHopInterface,
+            "A missing next hop interface should be indicated by %s.",
+            Route.UNSET_NEXT_HOP_INTERFACE);
     _tag = tag;
   }
 
```

Alternative:
* Don't push this work onto the caller, and instead if given `null`, then force the `UNSET` variant. That might actually be more convenient throughout the codebase... (@arifogel, what do you think?)

If we go the former, then also please take a look at the other parameters and use the IDE's nullability analysis to find and fix other violations. For example, IntelliJ finds 10 violations for `nextHopInterface` being `null`:

screen shot 2017-07-28 at 11 39 32 am

screen shot 2017-07-28 at 11 39 38 am

dhalperi:
Ari had also suggested a builder strategy that he's used in other places: look at BgpRoute.Builder. That's kinda like the alternative suggested above, because the defaults are configured in the Builder itself.

virtuald:
I've been using the approach of replacing null with UNSET_NEXT_HOP_INTERFACE in the constructor, and it seems to be ok.

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+/** Tests for {@link StaticRoute} */
+public class StaticRouteTest {
+
+   StaticRoute.Builder _srBuilder;
+
+   @Test public void checkAllAttrs() {
```
dhalperi: It looks to me like you have not properly set up your google-java-format plugin. Please do so and reformat all changed files. (See https://github.com/batfish/batfish/wiki/IntelliJ)

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -19,7 +19,7 @@
   private final int _tag;
 
   @JsonCreator
-  public StaticRoute(
+  private StaticRoute(
```
dhalperi: This may still need to be public for Jackson. I'm not sure. (Add a test)

dhalperi: If so, we should figure out a better way to balance builder vs constructor.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -110,4 +110,35 @@ protected final String protocolRouteString() {
...
+
+    public void setAdministrativeCost(int administrativeCost) {
+      this._administrativeCost = administrativeCost;
+    }
+
+    public String getNextHopInterface() {
+      return _nextHopInterface;
+    }
+
+    public void setNextHopInterface(String nextHopInterface) {
```
dhalperi: these should use the chaining method.

dhalperi: check for invalid values. (Specifically, `null` nextHopInterface should throw an exception).

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -110,4 +110,35 @@ protected final String protocolRouteString() {
...
+  public static class Builder extends AbstractRouteBuilder<StaticRoute> {
+
+    private int _administrativeCost;
+
+    private String _nextHopInterface;
+
+    @Override
+    public StaticRoute build() {
+      if (_nextHopIp == null) {
+        _nextHopIp = Route.UNSET_ROUTE_NEXT_HOP_IP;
```
dhalperi: this logic should be in the class that owns `_nextHopIp`. Additionally, these variables should be `private` – children should (without a very good reason) not be able to access properties of their parents.

projects/batfish/src/main/java/org/batfish/representation/aws_vpcs/InternetGateway.java
```diff
@@ -63,13 +63,13 @@ public Configuration toConfigurationNode(AwsVpcConfiguration awsVpcConfiguration
...
-          new StaticRoute(
-              vpc.getCidrBlock(),
-              vpcIfaceIp,
-              null,
-              Route.DEFAULT_STATIC_ROUTE_ADMIN,
-              Route.DEFAULT_STATIC_ROUTE_COST);
+      StaticRoute.Builder srBuilder = new StaticRoute.Builder();
+      srBuilder.setNetwork(vpc.getCidrBlock());
+      srBuilder.setNextHopIp(vpcIfaceIp);
+      srBuilder.setNextHopInterface(null);
```
dhalperi: Skip setting the invalid, unknown, or default properties.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -110,4 +110,35 @@ protected final String protocolRouteString() {
   public int routeCompare(AbstractRoute rhs) {
     return 0;
   }
+
+  public static class Builder extends AbstractRouteBuilder<StaticRoute> {
+
+    private int _administrativeCost;
+
+    private String _nextHopInterface;
```
dhalperi: make sure you also handle unset `_nextHopInterface` while you're here.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
dhalperi: https://stackoverflow.com/a/17165079/1715495 has how to make builders that can be chained even through inheritance.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRouteBuilder.java
```diff
@@ -1,56 +1,70 @@
...
-    return _metric;
+  public final S setAdmin(int admin) {
+    _admin = admin;
+    return getThis();
   }
 
-  public final Prefix getNetwork() {
-    return _network;
+  // To handle the class casting exception while returning S in chaining methods
+  public abstract S getThis();
```
hverma-upenn: done

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRouteBuilder.java
```diff
@@ -1,56 +1,70 @@
...
-  public final void setMetric(int metric) {
-    _metric = metric;
+  public final Ip getNextHopIp() {
+    return _nextHopIp;
   }
 
-  public final void setNetwork(Prefix network) {
-    _network = network;
+  public final S setNextHopIp(Ip nextHopIp) {
+    _nextHopIp = nextHopIp == null ? Route.UNSET_ROUTE_NEXT_HOP_IP : nextHopIp;
```
hverma-upenn: nextHopInterface in staticroute made consistent with this. Since Abstractroute does not declare  nextHopIp, I can think of to put this check only at this place

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
```diff
@@ -13,7 +13,7 @@
 
 public class BgpRoute extends AbstractRoute {
 
-  public static class Builder extends AbstractRouteBuilder<BgpRoute> {
+  public static class Builder extends AbstractRouteBuilder<Builder,BgpRoute> {
```
hverma-upenn: done,

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
```diff
@@ -50,13 +50,13 @@ public BgpRoute build() {
         throw new BatfishException("Missing originType");
       }
       return new BgpRoute(
-          _network,
-          _nextHopIp,
-          _admin,
+          this.getNetwork(),
```
hverma-upenn: done

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -32,6 +33,15 @@ public StaticRoute(
     _tag = tag;
   }
 
+  private StaticRoute(Builder builder) {
```
hverma-upenn: done

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Environment.java
```diff
@@ -33,7 +33,7 @@
 
   private AbstractRoute6 _originalRoute6;
 
-  private final AbstractRouteBuilder<?> _outputRoute;
+  private final AbstractRouteBuilder<?,?> _outputRoute;
```
hverma-upenn: done,

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,93 @@
...
+    assertThat(sr.getNextHopIp(), is(new Ip("192.168.1.1")));
+    assertThat(sr.getNetwork(), is(Prefix.ZERO));
+    assertThat(sr.getNextHopInterface(), is("192.168.1.2"));
+    assertThat(sr.getAdministrativeCost(), is(1));
+    assertThat(sr.getTag(), is(0));
+  }
+
+  @Test
+  public void checkNullNextHop() {
+    _srBuilder.setNetwork(Prefix.ZERO).setNextHopIp(null);
```
hverma-upenn: no, it does not crashes, I just wanted to check the value assigned to nextHopIp when it is set to Null

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,93 @@
...
+    StaticRoute sr = _srBuilder.build();
+    assertThat(sr.getNextHopIp(), is(Route.UNSET_ROUTE_NEXT_HOP_IP));
+  }
+
+  @Test
+  public void checkNullable() {
+    String errorMessage = "nextHopInterface cannot be null in StaticRoute";
```
hverma-upenn: done, not relevant, as now no exception gets thrown

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,93 @@
...
+      assertThat(parsedObj.getAdministrativeCost(), is(1));
+      assertThat(parsedObj.getTag(), is(0));
+
+    } catch (IOException e) {
+      throw new BatfishException(
+          "Cannot parse the json to StaticRoute object", e);
+    }
+  }
+
+  @Before
```
hverma-upenn: done

projects/batfish/src/main/java/org/batfish/representation/aws_vpcs/Route.java
```diff
@@ -65,47 +65,31 @@ public StaticRoute toStaticRoute(
       Subnet subnet,
       Configuration subnetCfgNode) {
     StaticRoute staticRoute;
```
hverma-upenn: done

projects/batfish/src/main/java/org/batfish/representation/aws_vpcs/Subnet.java
```diff
@@ -181,13 +181,14 @@ public Configuration toConfigurationNode(AwsVpcConfiguration awsVpcConfiguration
     vpcIface.getAllPrefixes().add(vpcIfacePrefix);
     vpcIface.setPrefix(vpcIfacePrefix);
     // add a static route on the vpc router for this subnet
+    StaticRoute.Builder srBuilder = new StaticRoute.Builder();
```
hverma-upenn: done

projects/batfish/src/main/java/org/batfish/representation/aws_vpcs/Subnet.java
```diff
@@ -181,13 +181,14 @@ public Configuration toConfigurationNode(AwsVpcConfiguration awsVpcConfiguration
...
     StaticRoute vpcToSubnetRoute =
-        new StaticRoute(
-            _cidrBlock,
-            subnetIfacePrefix.getAddress(),
-            null,
-            Route.DEFAULT_STATIC_ROUTE_ADMIN,
-            Route.DEFAULT_STATIC_ROUTE_COST);
+        srBuilder
```
hverma-upenn: done

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ConnectedRoute.java
```diff
@@ -35,6 +37,8 @@ public Integer getMetric() {
     return 0;
   }
 
+  // TODO(http://github.com/batfish/batfish/issues/207)
```
dhalperi: I don't think this one needs a TODO -- you didn't change the behavior.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/GeneratedRoute.java
```diff
@@ -112,7 +118,7 @@ public GeneratedRoute(
     _discard = discard;
     _generationPolicy = generationPolicy;
     _metric = metric;
-    _nextHopInterface = nextHopInterface;
+    _nextHopInterface = MoreObjects.firstNonNull(nextHopInterface, Route.UNSET_NEXT_HOP_INTERFACE);
     _nextHopIp = nextHopIp;
```
dhalperi: do you not need this for nextHopIp too?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/GeneratedRoute.java
```diff
@@ -161,6 +167,8 @@ public Integer getMetric() {
     return _metric;
   }
 
+  // TODO(http://github.com/batfish/batfish/issues/207)
```
dhalperi: same comment re TODO

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
dhalperi: (test shows this is not needed.)

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -19,16 +23,16 @@
   private final int _tag;
 
   @JsonCreator
-  public StaticRoute(
+  private StaticRoute(
       @JsonProperty(NETWORK_VAR) Prefix network,
       @JsonProperty(NEXT_HOP_IP_VAR) Ip nextHopIp,
```
dhalperi: nextHopIp too?

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,89 @@
...
+    StaticRoute sr = _srBuilder.build();
+    try {
+      String json = mapper.writeValueAsString(sr);
+      StaticRoute parsedObj = mapper.readValue(json, StaticRoute.class);
+      assertThat(parsedObj.getNextHopIp(), is(new Ip("192.168.1.1")));
+      assertThat(parsedObj.getNetwork(), is(Prefix.ZERO));
+      assertThat(parsedObj.getNextHopInterface(), is("192.168.1.2"));
+      assertThat(parsedObj.getAdministrativeCost(), is(1));
+      assertThat(parsedObj.getTag(), is(0));
+
```
dhalperi: drop newline.

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/StaticRouteTest.java
```diff
@@ -0,0 +1,89 @@
...
+    assertThat(sr.getNextHopIp(), is(Route.UNSET_ROUTE_NEXT_HOP_IP));
+    assertThat(sr.getAdministrativeCost(), is(Route.UNSET_ROUTE_ADMIN));
+    assertThat(sr.getTag(), is(Route.UNSET_ROUTE_TAG));
+  }
+
+  @Test
+  public void checkSerialization() {
+    BatfishObjectMapper mapper =
+        new BatfishObjectMapper(Thread.currentThread().getContextClassLoader());
+    _srBuilder
```
dhalperi: Throughout this file, drop `_srBuilder` and just build `StaticRoute` inline.

tests/basic/nodes-unit-tests.ref
```diff
@@ -4668,18 +4666,13 @@
...
                     "cseFailSwCoreNotifyExtended"
                   ],
-                  "tty" : [ ],
                   "upgrade" : [
                     "UpgradeJobStatusNotify",
                     "UpgradeOpNotifyOnCompletion"
                   ],
-                  "vlan-membership" : [ ],
-                  "vlancreate" : [ ],
-                  "vlandelete" : [ ],
```
dhalperi: I think the tests are failing because you deleted these empty arrays.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/GeneratedRoute.java
```diff
@@ -118,8 +120,8 @@ public GeneratedRoute(
     _discard = discard;
     _generationPolicy = generationPolicy;
     _metric = metric;
+    _nextHopIp = firstNonNull(nextHopIp, Route.UNSET_ROUTE_NEXT_HOP_IP);
     _nextHopInterface = MoreObjects.firstNonNull(nextHopInterface, Route.UNSET_NEXT_HOP_INTERFACE);
```
dhalperi: on 123 you use `firstNonNull` ; on 124 you use `MoreObjects.firstNonNull` -- pick one and be consistent throughout?

I prefer just `firstNonNull` – it's cleaner.

PR #273, opened 2017-07-31
New example: example2, which looks like a DC

ratulm:


dhalperi:
- Image of example2 network seems wrong (it's just the original example image).
- Should we call it something more descriptive than example2? example-dc?

ratulm:
image updated. i'll do the example renaming later, after we have a few more examples. there may be multiple dc examples.

dhalperi:
Images still seem messed up:

- https://github.com/batfish/batfish/pull/273/files#diff-ba68590522b991e3794e3fbabe109558 is old network in (test_rigs/example2-with-delta/example-network.png)
- There are two copies of new network, at https://github.com/batfish/batfish/pull/273/files#diff-2cf6689f4744946247ea7ab4c3725202 and https://github.com/batfish/batfish/pull/273/files#diff-9a156f5375d6af2b2e8c5fbd7d8e2d9f. (test_rigs/example2/example-network.png and test_rigs/example2/example-network2.png)

PR #276, opened 2017-08-01
Review of preliminary changes for Storage API ("Not to be Merged with Master")

hverma-upenn:
Putting together a preliminary form of Storage APIs, Containerdir path resolution has been replaced with Storage API call with the container name as an argument, along with few other basic APIs. Container Name is being exchanged between the coordinator and worker now. 

Requesting review to confirm if this is the expected approach(and what should be changed)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/storage/Storage.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/storage/Storage.java
@@ -1,12 +1,18 @@
 package org.batfish.storage;
 
 import java.nio.file.Path;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 import org.batfish.common.BatfishException;
 import org.batfish.common.Container;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.GenericConfigObject;
 import org.batfish.datamodel.Topology;
 import org.batfish.datamodel.answers.Answer;
+import org.batfish.datamodel.answers.ConvertConfigurationAnswerElement;
+import org.batfish.datamodel.answers.ParseVendorConfigurationAnswerElement;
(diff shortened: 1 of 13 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/storage/Storage.java
```diff
@@ -0,0 +1,14 @@
+package org.batfish.storage;
+
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+/** Common storage APIs */
+public interface Storage {
+
+  Path getContainer(String containerName);
```
dhalperi: The `Storage` abstractions do not know about files. They need to be flexible and adaptable to any backend storage system – database, file system, cloud storage, etc. Instead of `Path`, we should be returning things like `Container`.

Everywhere the Coordinator or Batfish worker assume files we need to insert a layer of indirection – so that things are addressed by (e.g.) container name , container name + testrig name, etc.

projects/batfish/src/main/java/org/batfish/config/Settings.java
```diff
@@ -372,6 +372,10 @@ public void setTopologyPath(Path path) {
 
   private static final String ARG_ANONYMIZE = "anonymize";
 
+  private static final String ARG_CONTAINERS_LOCATION = "containerslocation";
```
dhalperi: Container location should be coming from the Coordinator, not from the command line. It should probably come when the worker registers with the Coordinator:

- Worker: "I am a new worker"
- Coordinator: "Great! We use this storage: <x>"

(or something like that.)

hverma-upenn:
Storage.java includes other APIs needed for coordinator now and can be reviewed. Next version will contain the missing APIs which may be needed for batfish, batfish-client and question

hverma-upenn:
Get form of the APIs in Storage is the most comprehensive. Whether we need create and delete for all objects can be discussed

Issue #256, opened 2017-07-26
Need more robust .ref production, comparison


arifogel:
When we have a field as e.g.
```java
SortedMap<String, SortedSet<String>> _someField
```
as used for snmp server traps in the data model, if we store a mapping of some key to an empty set of strings, the whole mapping disappears during serialization. The current JSON comparison we perform in our end-to-end testing cannot tell the difference between the mapping for a key being absent, and the mapping for that key being to an empty set.

My guess is using `JsonInclude.NON_EMPTY` causes the empty set to be simplified into a null (possible bug in this context, as it is a map value and not a field). Then the mapping from key to null is potentially indistinguishable from an absent key depending on comparison method used.

dhalperi:
Posted a candidate fix: #280 

Based on a conversation with the Jackson authors: https://groups.google.com/forum/#!topic/jackson-user/WfZzlt5C2Ww

PR #280, opened 2017-08-01
BatfishObjectMapper: use ALWAYS content serialization

dhalperi:
ObjectMapper has two separate configurations for inclusion: value and
content. The default method for configuring the inclusion applies the
supplied inclusion strategy to both.  However, Batfish's default of
NON_NULL means that we could serialize {map: {key: []}} to {} (where the
class Foo has a single property Map<String, String[]>), because:

1. key's value is empty, so we drop key
2. the map is empty, so we drop the map.

By swapping from value=NON_NULL (which applies to properties like
Foo.map) and content=NON_NULL (which applies to the content of
collections that are inside of objects) to content=ALWAYS, we do not
drop key from the map and thus don't drop the map.

This fixes batfish/batfish#256.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/BatfishObjectMapper.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/BatfishObjectMapper.java
@@ -1,5 +1,6 @@
 package org.batfish.common.util;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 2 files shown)


dhalperi:
@hverma-upenn has a [PR](https://github.com/batfish/batfish/pull/266) where he changed the data model and got the same [`bgp: []` missing](https://travis-ci.org/batfish/batfish/builds/260039677#L2974) diff on Travis.

This PR fixes that issue, which I believe is why the original bug was filed. So I'll merge.

PR #284, opened 2017-08-02
a new question to infer roles from node names

millstein:
The current version relies on several heuristics for how to parse node names, identify possible role names within them, and choose among possibilities.  It works for a few networks but needs testing on a broader set of networks to see how it should evolve.  (It does not do the right thing for our example testrig, whose node names are somewhat unusual, but it can be updated to support that style if it turns out to be useful.)


millstein:
Thanks.  Let me play with your new testrig first.  I didn't make a test because it does the wrong thing on the example testrig, but I will add one anyway just so the usage is clear.  (But you just need to call answer() with no args and it should work.)  I also have a template question that I can check into the questions-internal repo.

ratulm:
Yes, please check in the template question, which should make things more clear. I was wondering, for instance, what nodeRegex does.

The new example is example2 under test_rigs. Its topology is attached. 
example-network
I was thinking of changing the device names there from things like border-01 to dc1-border-01, for more realism; that change may help your inference logic as well.

millstein:
nodeRegex does the same thing as in other questions -- it lets you specify the subset of nodes that you want the question to analyze (in this case, the subset of nodes that you want to infer roles for).

yes, more uniform naming schemes help -- currently the analysis just tries to identify one common naming scheme and then find the role within that naming scheme (though it could be updated to recursively analyze the other nodes for a new naming scheme, etc.).

PR #285, opened 2017-08-02
Api sample(Not for merge)

kidsbear:
Requesting a code review to see if this is the right way to do it, especially the use of uri strings in Container and Testrig classes. Thanks~


projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -389,27 +389,32 @@ private ClientBuilder getClientBuilder() throws Exception {
```
dhalperi: looks like your google-java-format plugin is messed up. Please ensure it's installed. If it's acting wonky, maybe disable and renable?

projects/coordinator/src/main/java/org/batfish/coordinator/ExceptionMapper.java
```diff
@@ -12,11 +12,13 @@ public Response toResponse(Exception exception) {
     if (exception instanceof WebApplicationException) {
       // An exception that already maps directly to a standard HTTP Response, likely thrown
       // somewhere inside Jersey aka if the user typed a bad URI.
-      return ((WebApplicationException) exception).getResponse();
+      return Response.status(((WebApplicationException) exception).getResponse().getStatus())
+          .entity(Throwables.getStackTraceAsString(exception))
```
dhalperi: please do not modify the existing response generically like this. For instance, we do not want stack trace for 404 errors.

Find where the exception is being generated and customize the entity there.

Additionally, in the future we may be throwing our own WebApplicationExceptions with good entities, and we don't want to override that to bad entities here.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1335,7 +1335,7 @@ private boolean getContainer(List<String> options, List<String> parameters) {
     String containerName = parameters.get(0);
     Container container = _workHelper.getContainer(containerName);
     if (container != null) {
-      _logger.output(container.getTestrigs() + "\n");
```
dhalperi: What will this print? Add a test?

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -5,36 +5,35 @@
...
-  private SortedSet<String> _testrigs;
+  private String _testrigsUri;
   private Path _analysis;
 
   @JsonCreator
-  public static Container of(
-      @JsonProperty(NAME_VAR) String name, @JsonProperty(TESTRIGS_VAR) SortedSet<String> testrigs) {
-    return new Container(name, testrigs);
+  public static Container makeContainer(
+      @JsonProperty(NAME_VAR) String name, @JsonProperty(TESTRIGS_URI_VAR) String testrigsUrl) {
```
dhalperi: The URI is not a property of the container. it's a property of the container resource. we may need to distinguish these things in the object hierarchy.

kidsbear: This is my main concern, if a container object contains a list of Testrig Objects, the nested structure may make a container object really large. (e.g Testrig object contains a list of questions, a list of configs, etc.) We might only need some basic information of a container in some function calls (Testrig names in it for example). Have such a structure will cause some unnecessary informations in the network traffic. However, have a URI representation of inner resource also have some problems. For example, list all file names in a testrig requires additional calls to our web service (one for configs, one for questions, etc.) Also, there are some other issues for building a RESTful API for our existing service, I'll save them for later.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
kidsbear: For now, it should print something like "Container{name=foo, testrigsUri=xxx}" (tested in container's toString method). This is not a proper message here, will change it after decide the representation of a Container.

PR #286, opened 2017-08-03
Minor optimizations to the Data Plane Computation

dhalperi:
1. Simplify implementation of PrefixSpace#getAddressBits using helper methods and the fact that BitSet is already an array of longs under the hood.

2. Since the `SortedSet<Integer>` inside `AsPath` are never changed, make them `ImmutableSortedSet` under the hood. The key advantage of this is to save on copies: `ImmutableSortedSet.copyOf(set)` is a no-op if `set` is already an `ImmutableSortedSet`.

Combined, the two optimizations shave about 20% off of one large network data plane computation (8m13s [493s] down to 6m41s [401s]).

It took me a long time to convince myself that the sets inside of `AsPath` aren't changed (and frankly, I may be wrong). We would do well to add some proper Java abstractions here; based on what we do to these AsPaths in practice, it may expose other opportunities for improvement.


ratulm:
Looks good to me. Probably best to have @arifogel's eyes on it too.

dhalperi:
Yes, will not merge until @arifogel gets back.

PR #290, opened 2017-08-04
added a test case for inferring roles

millstein:


tests/basic/commands
```diff
@@ -35,6 +35,10 @@ test tests/basic/undefinedReferences.ref get undefinedReferences
 test tests/basic/uniqueIpAssignments.ref get uniqueIpAssignments
 test tests/basic/unusedStructures.ref get unusedStructures
 
+# example2 testrig
+test tests/basic/init-example2.ref init-testrig test_rigs/example2
+test tests/basic/inferroles.ref get inferRoles
```
dhalperi: should this be lowercase `inferRoles`? To match the name above?

millstein: I'm not sure if this is what you're referring to, but currently there is a naming inconsistency -- we use camel (e.g., inferRoles) at the command line (and for the names of template questions), but the getName() method of a Question must be all lower case (e.g., inferroles) or else the Java client breaks (since it converts question-name arguments to lowercase before looking them up).  That bug took me a while to find.

millstein: Or maybe you just mean that inferroles.ref should be inferRoles.ref? That would be more consistent with the other tests.

millstein:
The question results in a NodeRoleSpecifier object, which is the format used when someone wants to manually specify roles (via a node_roles.json file in a testrig).  This allows specifying roles by some combination of manual mapping and regular expressions.  The NodeRoleSpecifier has a method to produce the full mapping from roles to nodes -- the InferRolesAnswerElement's prettyPrint method prints the full mapping that was inferred.

PR #291, opened 2017-08-04
fix exception when convert json to AnswerElement for two questions

kidsbear:
The origin code will throw exceptions when convert answer.json to AnswerElement(null value in bgpAdvertisement constructor, no @JsonProperty set for NamesStructureOutlierSet). Two simple tests added to illustrate (will throw exception without changes).


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java
```diff
@@ -13,6 +13,12 @@
 
   private static final String STRUCT_TYPE_VAR = "structType";
 
+  private static final String NAMED_STRUCTURE_VAR = "namedStructTypes";
```
kidsbear: I use the name in outlier.json, I am not sure if this is the right name for it.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java
```diff
@@ -36,11 +42,11 @@
...
   public NamedStructureOutlierSet(
-      String structType,
-      String name,
-      T namedStructure,
-      SortedSet<String> conformers,
-      SortedSet<String> outliers) {
+      @JsonProperty(STRUCT_TYPE_VAR) String structType,
+      @JsonProperty(NAME_VAR) String name,
+      @JsonProperty(NAMED_STRUCTURE_VAR) T namedStructure,
+      @JsonProperty(CONFORMERS_VAR) SortedSet<String> conformers,
```
kidsbear: Also, in my understanding, this might also be null? Or they are checked some place else?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java
ratulm: To keep things consistent, perhaps call this NAMED_STRUCT_TYPES_VAR?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java
ratulm: If they are created through json, then they can be null or whatever default value Jackson uses. Do you see this constructor being called from anywhere in the code? If not, then Jackson is perhaps the only valid caller of this constructor.

PR #292, opened 2017-08-04
Warning removal

dhalperi:
* Automated removal of unnecessary .toString() followed by autoformat
* Automated swap to Map.computeIfAbsent followed by autoformat
* Replace Map.getOrDefault(x, null) with Map.get(x)
* Manually mark every function that returns null explicitly as `@Nullable`
* Add maven dependency on jsr305 annotations for nullable
* Fix StaticRoute w.r.t. `@Nonnull` annotations and invalid null checks.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
@@ -16,8 +16,10 @@ public class StaticRoute extends AbstractRoute {
 
   private final int _administrativeCost;
 
+  @Nonnull
   private final String _nextHopInterface;
 
+  @Nonnull
   private final Ip _nextHopIp;
 
   private final int _tag;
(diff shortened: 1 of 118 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/StaticRoute.java
```diff
@@ -105,8 +99,8 @@ public int hashCode() {
     int result = 1;
     result = prime * result + _network.hashCode();
     result = prime * result + _administrativeCost;
-    result = prime * result + ((_nextHopInterface == null) ? 0 : _nextHopInterface.hashCode());
-    result = prime * result + ((_nextHopIp == null) ? 0 : _nextHopIp.hashCode());
+    result = prime * result + _nextHopInterface.hashCode();
```
hverma-upenn: If nextHopInterface is UNSET_NEXT_HOP_INTERFACE (which now replaces null), the hashcode is not zero, this may change the result value. I am not sure if that is desired

dhalperi: Hash codes are only valid within the context of a single JVM instance, so I don't think this can possibly affect anything.

PR #293, opened 2017-08-08
Basic api calls

kidsbear:
Few basic API calls at the top level, will modify Testrig representation as needed later.


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -1024,4 +1024,14 @@ public void uploadTestrig(String containerName, String testrigName, InputStream
     CommonUtil.deleteDirectory(unzipSubdir);
     CommonUtil.delete(zipFile);
   }
+
+  private Path getContainerPath(String containerName) {
+    return Main.getSettings().getContainersLocation().resolve(containerName).toAbsolutePath();
+  }
+
+  public boolean checkContainerExist(String containerName) {
```
arifogel: English grammar: Please rename to checkContainerExists

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +32,74 @@
...
+  /** Get information about the Batfish environment and supported methods. */
   @GET
-  @Path("/container/{id}")
-  public Response getContainer(
-      @DefaultValue(DEFAULT_API_KEY) @HeaderParam(SVC_KEY_API_KEY) String apiKey,
-      @PathParam("id") String id) {
-    if (!Main.getAuthorizer().isAccessibleContainer(apiKey, id, false)) {
-      // TODO: put a proper error entity here.
-      return Response.status(Status.NOT_FOUND).build();
+  @Path("/info")
```
arifogel: Why are we using string literals here? Should you use an RSC constant?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +32,74 @@
...
+  public Response getInfo() {
```
arifogel: I don't like the name of this method. It sounds too generic and not obviously different from e.g. `getStatus`. Please make it more informative. If for instance it is supposed to return information about an `Environment`, then it could be called `getEnvironmentInfo`.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +32,74 @@
...
+    _logger.info("WMS:getInfo\n");
+    try {
+      JSONObject map = new JSONObject();
```
arifogel: I thought the whole point of this file was to replace the codehaus JSON provider API. Why is this here again? Was there a complication?

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,121 @@
...
+    return Response.ok(testrigNames).build();
+  }
+
+  /** Redirect to /testrigs if the user does not supply a testrig ID. */
+  @GET
+  @Path("/testrig")
+  public Response redirectTestrig() {
+    validate();
+    UriBuilder ub = _uriInfo.getBaseUriBuilder();
+    ub.path(CoordConsts.SVC_CFG_WORK_MGR2).path("container").path(_name).path("testrigs");
```
arifogel: `"testrigs"` similarly should be an RSC constant

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
kidsbear: I actually copy these two API calls from our previous WorkMgrService. I think I might rename getStatus to something like 'getWorkQueueStatus' and remove getInfo for now. I think the purpose of 'getInfo' is to provide some instructions, but not fully implemented in the previous WorkMgrService.  I think it should be better to add this API call after I got most of other calls set. Sounds ok?

kidsbear:
Trying to start from another direction, see #322 , close for now

PR #294, opened 2017-08-08
Set a default policy for iptables rules

### LINKED ISSUE #252: default Iptables chain policy should be ACCEPT
I believe the default policy (if not specified) is ACCEPT, so it should be set to that, or `getIpAccessListLineAction` will throw an exception. 

https://github.com/batfish/batfish/blob/dc12e2fee4aaab6601e2c822cf11c10be72e2e8d/projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java#L24

Additionally, line 54 will throw a NullPointerException if the _policy is null. Better to remove the toString there.

However, if your intent is to tell a user "you really should set a default policy".... then that error message should be better, as the error message (with the toString removed) is "Unsupported ChainPolicy for mapping to LineAction: null", which isn't very useful. It should probably give context (a line number? filename? hostname? something).

Also, #233 applies here, as the error message only shows up in debug mode.

dspicuzzbbn:
- Fixes #252

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java
@@ -25,6 +25,7 @@ public class IptablesChain implements Serializable {
 
   public IptablesChain(String name) {
     _name = name;
+    _policy = ChainPolicy.ACCEPT;
     _rules = new LinkedList<>();
   }
 


projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java
```diff
@@ -19,7 +19,7 @@
 
   private String _name;
 
-  private ChainPolicy _policy;
```
dhalperi: For consistency, can you please move this to the constructor, where `_name` and `_rules` are initialized?

projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java
```diff
@@ -19,7 +19,7 @@
...
+  private ChainPolicy _policy = ChainPolicy.ACCEPT;
 
   private List<IptablesRule> _rules;
```
dhalperi: to test this, can you just delete the `ACCEPT` line from the existing testrig? Maybe just delete this line: https://github.com/batfish/batfish/blob/ea39b4ccf537a67df0f477cbc921d4b7d88d7af2/test_rigs/unit-tests/iptables/host1.iptables#L5

virtuald:
I'll try to get to the changes this week.

projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesChain.java
dspicuzzbbn: @dhalperi deleting that line does not cause a test failure

dspicuzzbbn: Never mind, I was only running the maven tests, I see now that travis does a bit more.

dhalperi:
Thanks @virtuald !

Issue #206, opened 2017-07-13
allinone NullPointerException


virtuald:
When running allinone.Main from eclipse, if I try to do a simple query it fails with a NullPointerException:

```
batfish> get #checkreachability 10.0.4.200 tcp
get #checkreachability 10.0.4.200 tcp
java.lang.NullPointerException
	at org.batfish.client.QuestionHelper.getReachabilityQuestion(QuestionHelper.java:103)
	at org.batfish.client.QuestionHelper.resolveMacro(QuestionHelper.java:168)
	at org.batfish.client.Client.answerType(Client.java:730)
	at org.batfish.client.Client.get(Client.java:1254)
	at org.batfish.client.Client.processCommand(Client.java:2040)
	at org.batfish.client.Client.processCommand(Client.java:1985)
	at org.batfish.client.Client.runInteractive(Client.java:2319)
	at org.batfish.client.Client.run(Client.java:2272)
	at org.batfish.allinone.AllInOne.run(AllInOne.java:117)
	at org.batfish.allinone.Main.main(Main.java:6)
```

I suspect I just don't have my classpath or some environment setup correctly, but it seems like it would be nice if there were a better error message if that were the case. :)

virtuald:
When not running from the shell environment, you need to provide the `-DbatfishQuestionPluginDir` system property to the JVM. In eclipse, I add this to the run configuration 'VM Arguments'.

```
-DbatfishQuestionPluginDir=${workspace_loc:/question}/target
```

It seems that if the question plugins are required (surely they are?), that batfish should exit with an error (and mention to set this property) if one is not configured.

ratulm:
Yes, most of the questions require that plugin. We'll make the error message more helpful. 

While you are update the WiKi for eclipse, feel free to add this point too. 

Thanks.

dhalperi:
Want to send a PR?

ratulm:
Fixed by 4d4e1396f64286a8579428f71a7f6e307c9485bb

PR #295, opened 2017-08-09
Plugin folders

### LINKED ISSUE #240: question jar not built during eclipse build
The question.jar that used to be built with ant in eclipse is required to be created in order to run, even if the other projects are being run directly from class files. We need question.jar to be built by m2e in eclipse when the question project is built / question project java files are modified.

dspicuzzbbn:
This is useful when using batfish from eclipse. Eclipse's build process doesn't automatically create JARs (as far as I can see), so you have to do a separate mvn build to create the jar. Instead of dealing with that mess, this changes the plugin process to look for jars first, and if they're not found then see if it's a folder.

This fixes #240 (which I hadn't seen previously) and is related to #206


arifogel:
@dspicuzzbbn I like this in principle. Would you be able to add tests with associated test plugins (that do not get built into jars).

dspicuzzbbn:
Are there existing tests that explicitly test a jar plugin?

arifogel:
@dspicuzzbbn I don't believe so, but @dhalperi can chime in. We actually discussed this yesterday. @dhalperi, should we just do the tests ourselves on this branch?

dspicuzzbbn:
If there are existing tests that test jar plugin functionality, I'm happy to update them to test a folder based plugin. However, I won't have time to create tests from scratch.

PR #296, opened 2017-08-09
driver thread should catch throwable so it can report all errors

### LINKED ISSUE #277: Batfish driver should catch throwable to report errors
Catching Exception isn't enough, sometimes other odd errors can be thrown. 

https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Driver.java#L396

dspicuzzbbn:
- Fixes #277


dhalperi:
THanks!

PR #300, opened 2017-08-09
add authenticationFilter

kidsbear:
Current version of web service in PR #293 assume authentications done before the http calls been processed. Not registered to the service for now, no default version for client.


projects/coordinator/src/main/java/org/batfish/coordinator/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.ext.Provider;
+import org.batfish.common.CoordConsts;
+import org.batfish.common.Version;
+import org.batfish.coordinator.Main;
+
+/**
+ * This filter verify the access permissions for a user based on apiKey and version provided in
+ * request header.
+ */
+@Provider public class AuthenticationFilter implements ContainerRequestFilter {
```
dhalperi: please enable the java formatting using the `google-java-format` plugin.

They had a [bug in the 1.3.0 version](https://github.com/google/google-java-format/issues/167#issuecomment-319532138) where it could disable itself – are you using the latest version?

projects/coordinator/src/main/java/org/batfish/coordinator/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override public void filter(ContainerRequestContext requestContext) {
+    String apiKey = CoordConsts.DEFAULT_API_KEY;
+    if (requestContext.getHeaders().containsKey(CoordConsts.SVC_KEY_API_KEY)) {
+      apiKey = requestContext.getHeaderString(CoordConsts.SVC_KEY_API_KEY);
+    }
+
+    String clientVersion = requestContext.getHeaderString(CoordConsts.SVC_KEY_VERSION);
+
+    // TODO: provides default version? Assume right version now
```
dhalperi: For now, please separate `VersionFilter` from `AuthenticationFilter` -- Version and API Key checks should be done separately.

(If we see this is a performance bottleneck, we can re-combine.. but I doubt it for now).

projects/coordinator/src/main/java/org/batfish/coordinator/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    if (clientVersion == null) {
+      clientVersion = Version.getVersion();
+    }
```
dhalperi: Simplify these 4 lines:

```java
String clientVersion = firstNonNull(requestContext.getHeaderString(Coord...), Version.getVersion());
```

projects/coordinator/src/main/java/org/batfish/coordinator/resources/AuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+      requestContext.abortWith(Response.status(Status.BAD_REQUEST)
+          .entity(e.getMessage())
+          .type(MediaType.APPLICATION_JSON)
+          .build());
+    }
+
+  }
+
+  private void checkStringParam(String paramStr, String parameterName) {
+    if (paramStr == null || paramStr.equals("")) {
```
dhalperi: static import `Strings.isNullOrEmpty`.

projects/coordinator/src/main/java/org/batfish/coordinator/ApiKeyAuthenticationFilter.java
```diff
@@ -0,0 +1,47 @@
...
+@Provider
+public class ApiKeyAuthenticationFilter implements ContainerRequestFilter {
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    String apiKey =
+        firstNonNull(
+            requestContext.getHeaderString(CoordConsts.SVC_KEY_API_KEY),
+            CoordConsts.DEFAULT_API_KEY);
+    try {
+      if (Strings.isNullOrEmpty(apiKey)) {
```
dhalperi: `if (apiKey.empty())` -- it can't be null.

projects/coordinator/src/main/java/org/batfish/coordinator/ApiKeyAuthenticationFilter.java
```diff
@@ -0,0 +1,47 @@
...
+        throw new IllegalArgumentException("apikey is missing or empty");
```
dhalperi: Just call `requestContext.abortWith` directly, then return. Here and below.

Don't need the try-catch to convert to a response.

dhalperi: This one (empty API key provided) should be a 401 Unauthorized.

projects/coordinator/src/main/java/org/batfish/coordinator/ApiKeyAuthenticationFilter.java
```diff
@@ -0,0 +1,47 @@
...
+          Response.status(Status.BAD_REQUEST)
+              .entity(e.getMessage())
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    }
+  }
+
+  private void checkApiKeyValidity(String apiKey) throws Exception {
+    if (!Main.getAuthorizer().isValidWorkApiKey(apiKey)) {
+      throw new AccessControlException("Invalid API key: " + apiKey);
```
dhalperi: this one (API Key provided but Invalid) should be a 403.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionAuthenticationFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import org.batfish.common.CoordConsts;
+import org.batfish.common.Version;
+
+/**
+ * This filter verify the access permissions for a user based on version provided in request header.
+ */
+public class VersionAuthenticationFilter implements ContainerRequestFilter {
```
dhalperi: Why is this called an `Authentication` filter? It doesn't authenticate the client.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionAuthenticationFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import org.batfish.common.CoordConsts;
+import org.batfish.common.Version;
+
+/**
+ * This filter verify the access permissions for a user based on version provided in request header.
```
dhalperi: this is not about permissions, it's about compatibility.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionAuthenticationFilter.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class VersionAuthenticationFilter implements ContainerRequestFilter {
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    String clientVersion = requestContext.getHeaderString(CoordConsts.SVC_KEY_VERSION);
+    try {
+      if (Strings.isNullOrEmpty(clientVersion)) {
+        throw new IllegalArgumentException("version is missing or empty");
```
dhalperi: call `abortWith` directly and return, don't bother with the throw+try+catch.

projects/coordinator/src/test/java/org/batfish/coordinator/ApiKeyAuthenticationFilterTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.glassfish.jersey.media.multipart.MultiPartFeature;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.test.JerseyTest;
+import org.glassfish.jersey.test.TestProperties;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/** Tests for {@link ApiKeyAuthenticationFilterTest} */
```
dhalperi: period at the end of the Javadoc.

projects/coordinator/src/test/java/org/batfish/coordinator/ApiKeyAuthenticationFilterTest.java
```diff
@@ -0,0 +1,77 @@
...
+    Main.mainInit(new String[] {"-containerslocation", _folder.getRoot().toString()});
+    Main.setLogger(logger);
+    Main.initAuthorizer();
+    Main.setWorkMgr(new WorkMgr(settings, logger));
+  }
+
+  @Override
+  protected Application configure() {
+    forceSet(TestProperties.CONTAINER_PORT, "0");
+    return new ResourceConfig(WorkMgrServiceV2.class)
```
dhalperi: for testing a filter don't use the `WorkMgrServiceV2.class` at all. Just make up a resource in this file.

projects/coordinator/src/test/java/org/batfish/coordinator/VersionAuthenticationFilterTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.glassfish.jersey.media.multipart.MultiPartFeature;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.jersey.test.JerseyTest;
+import org.glassfish.jersey.test.TestProperties;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/** Tests for {@link VersionAuthenticationFilter} */
```
dhalperi: period at the end of the javadoc.

projects/coordinator/src/main/java/org/batfish/coordinator/ApiKeyAuthenticationFilter.java
```diff
@@ -23,25 +19,19 @@ public void filter(ContainerRequestContext requestContext) {
...
       requestContext.abortWith(
-          Response.status(Status.BAD_REQUEST)
-              .entity(e.getMessage())
+          Response.status(Status.UNAUTHORIZED)
+              .entity("ApiKey is empty")
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    } else if (!Main.getAuthorizer().isValidWorkApiKey(apiKey)) {
+      requestContext.abortWith(
+          Response.status(Status.FORBIDDEN)
```
kidsbear: Actually, I am not sure whether unauthorized should be used here, since this filter checks whether the apikey is valid in general case, not about a specific resource. (This should be checked in the service class)

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
```diff
@@ -10,26 +10,29 @@
...
-      }
-
-      checkClientVersion(clientVersion);
-
-    } catch (Exception e) {
+    if (Strings.isNullOrEmpty(clientVersion)) {
       requestContext.abortWith(
-          Response.status(Status.BAD_REQUEST)
-              .entity(e.getMessage())
+          Response.status(Status.PRECONDITION_FAILED)
```
kidsbear: Not sure which status code to use here, seems like '426 Upgrade Required' specifically refers to the protocol in use, not the client version.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
```diff
@@ -0,0 +1,42 @@
...
+ * This filter verifies that the client's version is compatible with the version of Batfish service.
+ */
+public class VersionCompatibilityFilter implements ContainerRequestFilter {
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    String clientVersion = requestContext.getHeaderString(CoordConsts.SVC_KEY_VERSION);
+    if (Strings.isNullOrEmpty(clientVersion)) {
+      requestContext.abortWith(
+          Response.status(Status.PRECONDITION_FAILED)
```
kidsbear: Not sure which status code to use here, seems like '426 Upgrade Required' specifically refers to the protocol in use, not the client version.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
```diff
@@ -0,0 +1,42 @@
...
+              .entity("Version is missing or empty")
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    } else {
```
dhalperi: Probably, when the header is missing we should put a very specific error message.

`HTTP Header %s should contain a client version.`

If it's present but wrong, we probably don't need to worry about it.
and same for API key.

projects/coordinator/src/test/java/org/batfish/coordinator/VersionCompatibilityFilterTest.java
```diff
@@ -0,0 +1,83 @@
+package org.batfish.coordinator;
```
kidsbear: I don't understand "If it's present but wrong, we probably don't need to worry about it.
and same for API key". We only check whether the apiKey and version are provided in the header, without worrying about whether or not they are valid?

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
```diff
@@ -0,0 +1,39 @@
...
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    String clientVersion = requestContext.getHeaderString(CoordConsts.SVC_KEY_VERSION);
+    if (Strings.isNullOrEmpty(clientVersion)) {
+      requestContext.abortWith(
+          Response.status(Status.BAD_REQUEST)
+              .entity(
+                  String.format(
+                      "HTTP header %s should contain a client version",
+                      requestContext.getHeaders()))
```
dhalperi: this does not look like the name of the header that should contain the version.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
```diff
@@ -0,0 +1,39 @@
...
+              .type(MediaType.APPLICATION_JSON)
+              .build());
+    } else if (!Version.isCompatibleVersion("Service", "Client", clientVersion)) {
+      requestContext.abortWith(
+          Response.status(Status.BAD_REQUEST)
+              .entity(String.format("Illegal version '%s' for Client", clientVersion))
```
dhalperi: This error message is not actionable. How about

Client version %s is not compatible with server version %s.

projects/coordinator/src/main/java/org/batfish/coordinator/VersionCompatibilityFilter.java
kidsbear: The call requestContext.getHeaders() actually returns all the content in Http Header. In my opinion, a message like String.format("HTTP Header should contain a %s field with a non-empty value", CoordConsts.SVC_KEY_VERSION) should be ok here, will discuss tomorrow. Thank you~

PR #301, opened 2017-08-10
show diffs after demo failure. fixes #99

### LINKED ISSUE #99: diff not shown during travis build.sh when z3 absent
When the travis build script .travis/build.sh crashes during the demo portion on a machine without z3 installed, the diffs of other refs are not shown.

arifogel:


.travis/build.sh
```diff
@@ -43,7 +47,7 @@ for i in $($GNU_FIND -name *.testout); do
 done
 
 #exit with exit code 1 if any test failed
-if [ -n "$($GNU_FIND -name '*.testout')" ]; then
+if [ -n "$($GNU_FIND -name '*.testout')" -o -n "$DEMO_FAILED" ]; then
```
dhalperi: why is the `DEMO_FAILED` variable at all necessary now? Won't the testout check work?

arifogel: Ah I forgot that I chose not to delete the testout when failure occurs. You are right, and I will fix.

PR #304, opened 2017-08-11
refactor variables with name (xxx)_VAR that are used in JsonProperty

### LINKED ISSUE #303: need to refactor *_VAR constants for easier sorting
Currently, all Jackson property names are constants with names like `DESCRIPTION_VAR`, etc.
The placement of `_VAR` at the end of the constant name hinders sorting. Furthermore, `VAR` itself is not appropriately descriptive.
I suggest every such constant be refactored to `PROP_(name)` from `(name)_VAR`.
However, I am open to other naming suggestions.

kidsbear:
Strategy: Use Java regex to replace all '(xxx)_VAR' words under @JsonProperty annotation to PROP_(xxx), then search for '(xxx)_VAR' pattern, for each of them, check whether or not it was used inside a @JsonProperty, replace it if it is. (Actually, the only few exceptions are in org.batfish.grammar, org.batfish.z3 and Warnings.java). This closes #303

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -522,7 +522,7 @@ public class Client extends AbstractClient implements IClient {
     }
     JSONObject instanceJson;
     try {
-      instanceJson = questionJson.getJSONObject(BfConsts.INSTANCE_VAR);
+      instanceJson = questionJson.getJSONObject(BfConsts.PROP_INSTANCE);
     } catch (JSONException e) {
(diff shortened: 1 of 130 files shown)


arifogel:
@kidsbear This means you should also remove any `PROP_` constants defined in that file.

kidsbear:
@arifogel Do you mean remove all @Json* annotation in org.batfish.representation.cisco package, and rename all the refactored variable name back to (XXX)_VAR?

arifogel:
No I mean you should remove all annotations and all property constants. They never should have been in that file in the first place. You don't have to undo the rename, just remove those constants altogether.

PR #310, opened 2017-08-11
Add Findbugs configuration, but do not run it automatically yet

dhalperi:
Two reasons not to run it yet:

- It takes over 2 minutes to analyze Batfish module alone, probably because it doesn't yet exclude generated code.
- The checks aren't even close to passing.


dhalperi:
It can be run manually as follows:

    mvn compile findbugs:check

or

    mvn compile findbugs:check -pl batfish

to check only one module.

PR #311, opened 2017-08-11
Environment validation for core and client

arifogel:
DO NOT MERGE WITHOUT TESTS


arifogel:
After painstaking merge, and with some small fixes, I believe it is working now. Due to changes to batfish test initialization routine, this PR is high priority.

PR #313, opened 2017-08-11
Findbugs: remove redundant != null checks

dhalperi:
Since these are so complex to review, I separated each removal into its own commit for clarity.

Note that there are two nested redundant checks in Batfish.java in the same function.


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2332,37 +2332,35 @@ public void initRemoteOspfNeighbors(
                     if (remoteProc.getOspfNeighbors() == null) {
                       remoteProc.setOspfNeighbors(new TreeMap<>());
                     }
-                    if (remoteProc != null) {
```
arifogel: This line should be moved above 2332 rather than being removed.

PR #314, opened 2017-08-11
AWSVPC/Subnet: fix a mistaken variable assignment

dhalperi:
The VPN gateway address was never set; we assigned both internet and VPN
gateways addresses to the igwAddress.

Q: what did this fix, if anything? How should I add a test?


projects/batfish/src/main/java/org/batfish/representation/aws_vpcs/Subnet.java
```diff
@@ -246,7 +246,7 @@ public Configuration toConfigurationNode(AwsVpcConfiguration awsVpcConfiguration
       vgwSubnetIface.getAllPrefixes().add(vgwSubnetIfacePrefix);
       vgwConfigNode.getInterfaces().put(vgwSubnetIfaceName, vgwSubnetIface);
       vgwConfigNode.getDefaultVrf().getInterfaces().put(vgwSubnetIfaceName, vgwSubnetIface);
-      igwAddress = vgwSubnetIfacePrefix.getAddress();
```
dhalperi: Note that igwAddress is set at L220.

PR #318, opened 2017-08-14
Issue 209 fix

hverma-upenn:
NP Exception was thrown when topology contained a non-existent node. The exception was thrown at https://github.com/batfish/batfish/blob/7d0be6a761605dc55fb8b20b5c9140a92e8b2391/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L245 and also when executing a query(as mentioned in the issue) at https://github.com/batfish/batfish/blob/7d0be6a761605dc55fb8b20b5c9140a92e8b2391/projects/batfish/src/main/java/org/batfish/z3/Synthesizer.java#L923.
This fix solves the issue by failing the parsing and showing a message when the topology is inconsistent with the config/hosts


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4424,4 +4425,21 @@ private void writeJsonTopology() {
       throw new BatfishException("Failed to synthesize JSON topology", e);
     }
   }
+
+  public void checkTopology(Map<String, Configuration> configurations, Topology topology) {
+    for (Edge edge : topology.getEdges()) {
+      if (!configurations.containsKey(edge.getNode1())
+          || !configurations.containsKey(edge.getNode2())) {
+        throw new BatfishException("ERROR: topology contains a non-existent node\n");
```
dhalperi: Does the format of this message match the other exceptions we throw? E.g., Line 4425 does not seem to include an `ERROR:` prefix.

projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -176,6 +178,90 @@ public void testParseTopologyJson() throws IOException {
...
+        "["
+            + "{ "
+            + "\"node1\" : \"h1\","
+            + "\"node1interface\" : \"eth0\","
+            + "\"node2\" : \"h2\","
+            + "\"node2interface\" : \"e0\""
+            + "}"
+            + "]";
+    Path topologyFilePath =
+        CommonUtil.createTempFileWithContent("testParseTopologyJson", topologyJson);
```
dhalperi: Can you simplify this setup code, perhaps with some `private static` utility methods. Things that come to mind:

For building a topology, it's simpler & more readable to build a Java map yourself and then let Jackson do the work of serialization.

```java
List<Map<String, String>> topology = Collections.singletonList(
    ImmutableMap.<String, String>builder()
         .add("n1", "i1")
         .add...
         .build());
//use BatfishObjectMapper to serialize it to a file.
```
and that probably gets wrapped in a utility method.

Making a new host Configuration for a given host/ethernet pair definitely should go in a utility - you do it 5 times.

projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -176,6 +180,43 @@ public void testParseTopologyJson() throws IOException {
     assertEquals(topology.getEdges().size(), 2);
   }
 
+  @Test
+  public void testCheckValidTopology() throws IOException {
+    Batfish batfish = initBatfish();
```
dhalperi: I don't think you need this object any more. The only thing you use it for is to call `checkTopology`, and that's a static function. So you can just do `Batfish.checkTopology`.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4424,4 +4425,28 @@ private void writeJsonTopology() {
...
+      if (!configurations.containsKey(edge.getNode2())) {
+        throw new BatfishException(
+            "Topology contains a non-existent node '" + edge.getNode2() + "'");
+      }
+      //nodes are valid, now checking corresponding interfaces
+      Configuration config1 = configurations.get(edge.getNode1());
+      Configuration config2 = configurations.get(edge.getNode2());
+      if (!config1.getInterfaces().containsKey(edge.getInt1())) {
+        throw new BatfishException(
+            "Node '" + edge.getNode1() + "' doesn't has the interface '" + edge.getInt1() + "'");
```
dhalperi: Modeling after https://github.com/batfish/batfish/blob/90770469c7efa7739cf015df7fcc62c55611b3e2/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L3376 , maybe "Topology contains a non-existent %s on node %s"

projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -176,6 +180,40 @@ public void testParseTopologyJson() throws IOException {
     assertEquals(topology.getEdges().size(), 2);
   }
 
+  @Test
+  public void testCheckValidTopology() throws IOException {
+    Map<String, Configuration> configs = new HashMap<>();
+    configs.put("h1", createConfiguration("h1", "eth0"));
+    configs.put("h2", createConfiguration("h2", "e0"));
+    //test that checking a valid topology does not throw any exception
```
dhalperi: So for tests, I like to have it as unambiguous as possible what is the test. I might write the end of this function as:

```java
EdgeSet edges = new EdgeSet(Collections.singletonList(new Edge...));
Topology topology = new Topology(edges);

// Test that checkTopology does not throw
Batfish.checkTopology(configs, topology);
```

A couple of elements of this change:

* Absolutely no code after the comment other than the function under test.
* Blank line before the test to visually separate setup and test.
* Slightly simpler constructors.

projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
```diff
@@ -176,6 +180,40 @@ public void testParseTopologyJson() throws IOException {
...
+    Batfish.checkTopology(configs, new Topology(edges));
+  }
+
+  @Test
+  public void testCheckTopologyInvalidNode() throws IOException {
+    Map<String, Configuration> configs = new HashMap<>();
+    configs.put("h1", createConfiguration("h1", "eth0"));
+    EdgeSet edges = new EdgeSet();
+    edges.add(new Edge("h1", "eth0", "h2", "e0"));
+    _thrown.expect(BatfishException.class);
```
dhalperi: similarly, blank line before `_thrown`. Create `topology` before thrown is setup.

PR #319, opened 2017-08-14
Batfish: no longer AutoCloseable

dhalperi:


projects/batfish/src/main/java/org/batfish/main/Driver.java
```diff
@@ -419,7 +418,6 @@ public void run() {
         batfish.setTerminatedWithException(true);
       }
 
-      batfish.close();
       return !batfish.getTerminatedWithException();
```
dhalperi: (As I read it, this function does not rely on closed behavior)

PR #320, opened 2017-08-14
Batfish: use built-in bit twiddling

dhalperi:
There should be less magic in our code.

I ran the following test to confirm that the old implementation and the new
are equivalent:

```java
  @Test
  public void testMagic() {
    for (int numInterfaces = 1; numInterfaces < 33; ++ numInterfaces) {
      int numHostBits = 0;
      for (int shiftedValue = numInterfaces - 1;
          shiftedValue != 0;
          shiftedValue >>= 1, numHostBits++) {
      }
      assertThat(Integer.numberOfLeadingZeros(numInterfaces - 1), equalTo(32 - numHostBits));
    }
  }
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -1476,11 +1476,7 @@ public class Batfish extends PluginConsumer implements AutoCloseable, IBatfish {
         throw new BatfishException(
             "The following interface set contains less than two interfaces: " + interfaceSet);
       }
-      int numHostBits = 0;
-      for (int shiftedValue = numInterfaces - 1;
-          shiftedValue != 0;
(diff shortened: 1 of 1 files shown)


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -1499,7 +1495,7 @@ private void generateOspfConfigs(Path topologyPath, Path outputPath) {
         config.getInterfaces().put(ifaceName, iface);
         offset++;
       }
-      currentStartingIpAsLong += (1 << numHostBits);
+      currentStartingIpAsLong += (1 << (32 - subnetBits));
```
arifogel: Use Prefix.MAX_PREFIX_LENGTH instead of 32.

PR #321, opened 2017-08-14
assign testrigDir a non-empty value

kidsbear:
Assign "testrigs" to be default directory of testrigs(BfConsts.RELPATH_TESTRIGS_DIR). Notice that some current tests in WorkMgrTest and WorkMgrServiceTest are not useful(for example, get non-exist container should give an error instead of an empty container), will fix them in a later PR. (trying to make each PR small)

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
@@ -537,9 +537,6 @@ public class WorkMgr {
 
   /** Return a {@link Container container} contains all testrigs directories inside it */
   public Container getContainer(Path containerDir) {
-    if (!Files.exists(containerDir.resolve(BfConsts.RELPATH_TESTRIGS_DIR))) {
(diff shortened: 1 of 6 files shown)


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -536,6 +536,9 @@ public Container getContainer(String containerName) {
 
   /** Return a {@link Container container} contains all testrigs directories inside it */
   public Container getContainer(Path containerDir) {
+    if (!Files.exists(containerDir.resolve(BfConsts.RELPATH_TESTRIGS_DIR))) {
+      return Container.of(containerDir.toFile().getName(), new TreeSet<>());
+    }
```
dhalperi: Probably initializing a container should initialize any required subfolders (like testrigs).

kidsbear: I agree, it makes sense to me.

PR #322, opened 2017-08-14
add basic representation for objects

kidsbear:
Basic representation of Objects, need more modifications later. e.g. create object representation of 'environment' to replace String in Testrig object. Also, a Testrig object should contain more fields. Planning to do next: add basic functions in WorkMgr for those objects, (e.g. listContainers, getContainer, listTestrigs, etc.) Any suggestions are appreciated!


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/AccessLevel.java
```diff
@@ -0,0 +1,23 @@
+package org.batfish.datamodel.pojo;
+
+import org.batfish.common.BatfishException;
+
+public enum AccessLevel {
```
kidsbear: This enum class will be used to determine how much information about a pojo object will be returned. e.g. function calls will be in form of: getContainer(String containerName, AccessLevel accessLevel), etc.

projects/batfish-common-protocol/src/test/java/org/batfish/common/ContainerTest.java
```diff
@@ -1,39 +0,0 @@
-package org.batfish.common;
```
dhalperi: You're deleting this test without deleting `Container`?

kidsbear: It should be a rename, I moved Container class into org.batfish.common.pojo. I don't know why it shows as a deletion and a creation of the class.

dhalperi: You did not delete `Container` in this PR.

kidsbear: Sorry, you are right. And now it still shows as a deletion and creation, it should because I copied the file and delete the origin one separately instead of using a 'git mv' command.

PR #323, opened 2017-08-15
Compute Dataplane issue

hverma-upenn:
when any of the configuration(nodes) had 0 vrfs, computeDataPlane didn't return because computeFixedPoint started executing an infinite while loop at https://github.com/batfish/batfish/blob/7d0be6a761605dc55fb8b20b5c9140a92e8b2391/projects/batfish/src/main/java/org/batfish/bdp/BdpDataPlanePlugin.java#L497 since staticChanged could never be set to false in that case.
Also in previous implementation of the same while loop only the last call to vr.activateStaticRoutes() will have any effect on the value of staticChanged(previous one will be overwritten), so I moved reinitializing of staticChanged to outside the for loop
Also, I have tried to simplify the second while loop at https://github.com/batfish/batfish/blob/7d0be6a761605dc55fb8b20b5c9140a92e8b2391/projects/batfish/src/main/java/org/batfish/bdp/BdpDataPlanePlugin.java#L521 with a do... while loop as I think that seems more relevant


projects/batfish/src/test/java/org/batfish/bdp/BdpDataPlanePluginTest.java
```diff
@@ -0,0 +1,90 @@
...
+  @Rule public TemporaryFolder _folder = new TemporaryFolder();
+
+  private static Configuration createConfiguration(String hostname, String interfaceName) {
+    Configuration config = new Configuration(hostname);
+    config.setConfigurationFormat(ConfigurationFormat.HOST);
+    config.getInterfaces().put(interfaceName, new Interface(interfaceName, config));
+    return config;
+  }
+
+  private Batfish initBatfish(SortedMap<String, Configuration> configurations) throws IOException {
```
dhalperi: We seem to be duplicating this code a bunch. Should we have some sort of BatfishTestUtils class that provides this function as a static factory method to produce a `Batfish` for testing?

hverma-upenn: yes, that would be better I think. I will make it configurable through arguments to make it usable across various functions

hverma-upenn:
Separated out utility methods for BatfishTests and BdpDataPlanePluginTest

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
```diff
@@ -0,0 +1,94 @@
...
+        new Batfish(
+            settings,
+            CACHED_TESTRIGS,
+            CACHED_DATA_PLANES,
+            CACHED_ENVIRONMENT_BGP_TABLES,
+            CACHED_ENVIRONMENT_ROUTING_TABLES);
+    return batfish;
+  }
+
+  public static Configuration createConfiguration(String hostname, String interfaceName) {
```
dhalperi:
```java
public static Configuration createConfiguration(String nodename, ConfigurationFormat type, String interfaceNames...) {
...
}
```

Consider if this is sufficiently generic that we should put it in `Configuration` instead.

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import org.batfish.datamodel.Interface;
+import org.batfish.datamodel.collections.BgpAdvertisementsByVrf;
+import org.batfish.datamodel.collections.RoutesByVrf;
+import org.batfish.main.Batfish;
+import org.junit.rules.TemporaryFolder;
+
+public class BatfishTestUtils {
+
+  private static Batfish _batfish;
+  private static Batfish _batfishConfigurations;
```
dhalperi: drop these; return new instances every time instead. Of course individual tests or test suites may cache the return value ; but multiple different tests won't share static state.

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
dhalperi: Add simple javadoc to all public functions explaining what they do.

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
```diff
@@ -0,0 +1,94 @@
...
+  public static Batfish getBatfishWithConfigurations(
+      SortedMap<String, Configuration> configurations, TemporaryFolder tempFolder)
+      throws IOException {
+    if (_batfishConfigurations == null) {
+      _batfishConfigurations = initBatfish(configurations, tempFolder);
+    }
+    return _batfishConfigurations;
+  }
+
+  public static Batfish getBatfish() throws IOException {
```
dhalperi: I'd probably drop this one in favor of the other. I think it's confusing because the user can't take the output of `getBatfish` and easily transform it to get the output of `getBatfishWithConfigurations`.

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
```diff
@@ -0,0 +1,94 @@
+package org.batfish.common;
```
dhalperi: This is in the `batfish` module, not the common module -- it should not be in the `org.batfish.common` package. I would suggest the same package as `Batfish` class.

projects/batfish/src/test/java/org/batfish/common/BatfishTestUtils.java
dhalperi: Or make it call `getBatfishWithConfigurations` with empty parameters that make the expectation above true.

PR #325, opened 2017-08-15
Symbolic Control Plane Model

rabeckett:
Adds code to construct and query a symbolic model of the control plane. Specifically, this adds:

- A translation of configurations into an SMT formula (batfish project /smt)
- 9 new question plugins prefixed by "Smt" (e.g., SmtLocalConsistencyQuestionPlugin.java)
- A number of new, small example test networks (test_rigs/smt-examples)
- A large number of unit tests (tests/java-smt)

For the most part, this does **not** modify any existing Batfish files, with the exception of Batfish.java and IBatfish.java, where it adds several simple functions that delegate to the smt code to instrument the model with variables to answer queries.

I tried to stick to a similar set of parameters for the questions that are related to what Batfish can already answer. For example, the SmtReachability question should be almost exactly the same as the Batfish Reachability question. While they are different questions now, it may be possible to reuse the existing Batfish questions in the future.


jkhourybbn:
@rabeckett @ratulm what is the purpose of this smt/ logic? for example, is `smt-reachability` meant to replace or augment reachability? I was just looking at it and could not reproduce the `reachability` results using `smt-reachability` hence my question, but perhaps I am misinterpreting the output. Any docs/info would help.

ratulm:
This is a cut at beginning to integrate Minesweeper (https://batfish.github.io/minesweeper/) into the core code base. We have to work on making things more understandable here. If there are Minesweeper queries that are of particular interest to you, let us know. We can prioritize those.

jkhourybbn:
makes sense, will do

PR #328, opened 2017-08-16
move usage of JsonArray in WorkMgr to WorkMgrService

kidsbear:
make configureAnalysis in WorkMgr accepts a list of strings instead of JsonArray


projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrServiceTest.java
```diff
@@ -84,4 +91,57 @@ public void getNonEmptyContainer() throws Exception {
     expectedTestrigs.add("testrig2");
     assertThat(container.getTestrigs(), equalTo(expectedTestrigs));
   }
+
```
dhalperi: I would simplify these tests dramatically, or drop them. They're way too redundant (almost the same 50 lines of code in each file) and complex.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -153,11 +156,18 @@ public JSONArray configureAnalysis(
       checkClientVersion(clientVersion);
       checkContainerAccessibility(apiKey, containerName);
 
-      boolean newAnalysis = (newAnalysisStr == null || newAnalysisStr.equals("")) ? false : true;
+      boolean newAnalysis = !Strings.isNullOrEmpty(newAnalysisStr);
+      List<String> questionToDelete = new ArrayList<>();
```
dhalperi: *questions* ToDelete.

PR #330, opened 2017-08-16
add new Analysis type

kidsbear:
This is the simple version of Analysis, will use it for rewriting configureAnalysis function in WorkMgr. Some storage api calls needed here: createAnalysis, addAnalysisQuestion, delAnalysisQuestion, etc.


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * <p>Each {@link Analysis Analysis} contains a name and a mapping from question name to {@link
+ * Question question} for questions in the Analysis.
+ */
+public class Analysis {
+  private static final String PROP_NAME = "name";
+  private static final String PROP_QUESTIONS = "questions";
+
+  private String _name;
+  private @Nullable Map<String, Question> _questions;
```
kidsbear: I am not sure here, questions in Analysis should have a name(directory name), but our existing Question type does not has this field. Other thinking: 1. add a string field to Question class, but I think this will break our existing tests because we write the Question to .ref, also the name of the Question is not consistent, it should not work on today's testing strategy. 2. Create another "AnalysisQuestion" type which contains a name and a Question instance. Since it is so similar to Question, I not sure whether it is worth to do it.  Also, is Jackson fine with Maps?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public Analysis(@JsonProperty(PROP_NAME) String name) {
+    this(name, null);
```
kidsbear: Should I use empty hashmap here? I feel when we creating an empty Analysis with its name(same as creating empty folder today), it makes more sense to have an empty map of questions.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,94 @@
...
+ * Question question} for questions in the Analysis.
+ */
+public class Analysis {
+  private static final String PROP_NAME = "name";
+  private static final String PROP_QUESTIONS = "questions";
+
+  private String _name;
+  private Map<String, Question> _questions;
+
+  public Analysis(@JsonProperty(PROP_NAME) String name) {
```
hverma-upenn: do we need @JsonProperty when the constructor is not a @JsonCreator?

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/pojo/AnalysisTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Test
+  public void testConstructorWithName() {
+    Analysis a = new Analysis("analysis");
+    assertThat(a.getName(), equalTo("analysis"));
+    assertTrue(a.getQuestions().isEmpty());
+  }
+
+  @Test
+  public void testGetterSetter() {
```
hverma-upenn: I think this function is doing two things, testing the Constructor with names and questions and testing the setters and getters. The name testGetterSetter seams too general for the purpose

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/pojo/AnalysisTest.java
```diff
@@ -0,0 +1,125 @@
...
+    _thrown.expectMessage(equalTo("Question 'question' already exists for analysis 'analysis'"));
+    a.addQuestion("question", new TestQuestion("other-question"));
+  }
+
+  @Test
+  public void testDeleteQuestion() {
+    Analysis a = new Analysis("analysis");
+    Question question = new TestQuestion("question");
+    a.addQuestion("question", question);
+    assertThat(a.getQuestions().get("question"), equalTo(question));
```
hverma-upenn: we already tested addQuestion in the above case, I think we can skip asserting here again. It will fail in the above function anyways if there is something wrong

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @JsonProperty(PROP_QUESTIONS)
+  public void setQuestions(Map<String, String> questions) {
+    _questions = questions;
+  }
+
+  public void addQuestion(String questionName, String questionContent) {
+    if (this._questions.containsKey(questionName)) {
+      throw new BatfishException(
+          "Question '" + questionName + "' already exists for analysis '" + _name + "'");
```
dhalperi: Using String.format for these, also drop the `'`.

projects/batfish-common-protocol/src/main/resources/org/batfish/common/util/datamodel.thrift
```diff
@@ -0,0 +1,10 @@
+# The representation of datamodel for now
```
dhalperi: don't check these in, please.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,94 @@
...
+ * Question question} for questions in the Analysis.
+ */
+public class Analysis {
+  private static final String PROP_NAME = "name";
+  private static final String PROP_QUESTIONS = "questions";
+
+  private String _name;
+  private Map<String, String> _questions;
+
+  public Analysis(String name) {
```
dhalperi: delete this constructor. Classes should have only 1 public constructor.

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/pojo/AnalysisTest.java
```diff
@@ -0,0 +1,90 @@
...
+    assertTrue(a.getQuestions().isEmpty());
+  }
+
+  @Test
+  public void testBasicFunctions() {
+    Map<String, String> questions = Collections.singletonMap("question", "questionContent");
+    Analysis a = new Analysis("analysis", questions);
+    assertThat(a.getName(), equalTo("analysis"));
+    assertThat(a.getQuestions(), equalTo(questions));
+    a.setName("other-analysis");
```
dhalperi: these tests seem very redundant. Why do you test that the constructor worked here, you already tested it?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Analysis.java
```diff
@@ -0,0 +1,94 @@
...
+ * The {@link Analysis Analysis} is an Object representation of the analysis for BatFish service.
+ *
+ * <p>Each {@link Analysis Analysis} contains a name and a mapping from question name to {@link
+ * Question question} for questions in the Analysis.
+ */
+public class Analysis {
+  private static final String PROP_NAME = "name";
+  private static final String PROP_QUESTIONS = "questions";
+
+  private String _name;
```
dhalperi: make these final, drop the mutators.

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/pojo/AnalysisTest.java
```diff
@@ -0,0 +1,90 @@
...
+    assertThat(a.getQuestions().get("question"), equalTo("questionContent"));
+    _thrown.expect(BatfishException.class);
+    _thrown.expectMessage(equalTo("Question 'question' already exists for analysis 'analysis'"));
+    a.addQuestion("question", "questionContent");
+  }
+
+  @Test
+  public void testDeleteQuestion() {
+    Map<String, String> questions = new HashMap<>();
+    questions.put("question", "questionContent");
```
dhalperi: how come here you use new HashMap with put, but before you used Collections.singletonMap?

dhalperi: Use consistent style.

PR #336, opened 2017-08-17
Modify WorkMgrService and WorkMgr tests

kidsbear:
Modify getContainer call in WorkMgrService to return a Container object. Rewrite some tests. (Some of them are actually wrong tests, should pass names to listQuestion function instead of path)


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -623,11 +623,8 @@ public Response getContainer(
       checkContainerAccessibility(apiKey, containerName);
 
       Container container = Main.getWorkMgr().getContainer(containerDir);
-      container.setName(containerName);
-      BatfishObjectMapper mapper = new BatfishObjectMapper();
-      String containerString = mapper.writeValueAsString(container);
 
-      return Response.ok(containerString, MediaType.APPLICATION_JSON).build();
+      return Response.ok(container).build();
```
kidsbear: Use jackson to convert it into a Json String now, need to determine how many information to be returned in future.

dhalperi: Doesn't this use `jettison`, not Jackson? Because this is `WorkMgrService`.

kidsbear:
This PR now are mostly about simplify some tests. Also, it corrects the previous tests I wrote for listQuestions which are actually wrong.

PR #337, opened 2017-08-17
Ensure that not specifying a default iptables rule doesn't crash

dspicuzzbbn:
- Tests issue #252


dhalperi:
I think your previous comment was correct: removing the `FORWARD` rule is not a good test because `FORWARD` isn't used anywhere else. We should probably remove `INPUT` instead, since that's referred to by following lines.

dspicuzzbbn:
Hm, travis didn't send me a failure notification, I'll have to fix that.

I think for testing the crash issue, removing FORWARD is sufficient -- I wasn't running the tests that are being ran on travis (but I had assumed that the tests would be ran when I did a maven build). I have to catch a flight, but I'll fix the test crash later.

dspicuzzbbn:
Tests updated, I think it should pass now.

dhalperi:
Thanks!

PR #338, opened 2017-08-17
Allow iptables rules from hosts to be merged into routers

### LINKED ISSUE #221: feature: support for iptables on routers
We need to add batfish support for iptables not only for hosts, but for routers as well. What is the best way to support iptables for routers? There are many network configuration with linux routers (or quagga) where iptables is used on those routers instead of cisco ACLs for example. This feature will allow us to scale our testing.

For example, in a test configuration I have the following iptables rules on the ingress interface of an ospf router
```
# allow OSPF input
iptables -A INPUT -s 10.0.2.1 -p 89 -j ACCEPT
iptables -A INPUT -j DROP

# allow traffic forwarding from 10.0.4 subset
iptables -A FORWARD -s 10.0.4.0/24 -i eth0 -j ACCEPT
iptables -A FORWARD -i eth0 -j DROP
```
The router's is only doing ospf and config is very simple
```
interface eth0
  ip address 10.0.2.10/24
  
!
interface eth1
  ip address 10.0.3.1/24
  
!
router ospf
  router-id 10.0.2.10
  network 10.0.2.0/24 area 0
  network 10.0.3.0/24 area 0
!
router ospf6
  router-id 10.0.2.10
  interface eth0 area 0.0.0.0
  interface eth1 area 0.0.0.0
!
```

dspicuzzbbn:
- Writes host vendor configs to a separate directory
- Converts iptables ACLs to router ACLs
- Added tests
- Fixes #221

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IpAccessListLine.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IpAccessListLine.java
@@ -1,7 +1,9 @@
 package org.batfish.datamodel;
 
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import com.google.common.base.Strings;
 import com.kjetland.jackson.jsonSchema.annotations.JsonSchemaDescription;
+import java.util.Objects;
 
 @JsonSchemaDescription("A line in an IpAccessList")
 public final class IpAccessListLine extends HeaderSpace {
@@ -11,6 +13,10 @@ public final class IpAccessListLine extends HeaderSpace {
   private LineAction _action;
 
   private String _name;
+
+  private String _inInterfaceName;
+
+  private String _outInterfaceName;
 
   @Override
   public boolean equals(Object obj) {
@@ -24,6 +30,12 @@ public final class IpAccessListLine extends HeaderSpace {
     if (_action != other._action) {
       return false;
     }
+    if (!Objects.equals(_inInterfaceName, other._inInterfaceName)) {
+      return false;
+    }
+    if (!Objects.equals(_outInterfaceName, other._outInterfaceName)) {
+      return false;
+    }
     return true;
   }
 
@@ -33,9 +45,19 @@ public final class IpAccessListLine extends HeaderSpace {
     return _action;
   }
 
+  @JsonSchemaDescription("The input interface this rule is associated with")
+  public String getInInterfaceName() {
+    return _inInterfaceName;
+  }
+
   @JsonSchemaDescription("The name of this line in the list")
   public String getName() {
     return _name;
+  }
+
+  @JsonSchemaDescription("The output interface this rule is associated with")
+  public String getOutInterfaceName() {
+    return _outInterfaceName;
   }
 
   @Override
@@ -48,8 +70,16 @@ public final class IpAccessListLine extends HeaderSpace {
     _action = action;
   }
 
+  public void setInInterfaceName(String inInterfaceName) {
+    _inInterfaceName = inInterfaceName;
+  }
+
   public void setName(String name) {
     _name = name;
+  }
+
+  public void setOutInterfaceName(String outInterfaceName) {
+    _outInterfaceName = outInterfaceName;
   }
 
   @Override
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -1236,17 +1236,19 @@ public class Batfish extends PluginConsumer implements AutoCloseable, IBatfish {
             namesByPath.size());
     namesByPath.forEach(
         (inputPath, name) -> {
-          logger.debug(
-              "Reading and gunzipping: "
-                  + outputClassName
-                  + " '"
-                  + name
-                  + "' from '"
-                  + inputPath
-                  + "'");
-          byte[] data = fromGzipFile(inputPath);
-          logger.debug(" ...OK\n");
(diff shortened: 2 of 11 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IpAccessListLine.java
```diff
@@ -12,6 +14,10 @@
 
   private String _name;
 
+  private String _inInterfaceName;
```
arifogel: This should not go in IpAccessListLine. This heavily clashes with existing semantics. To filter by input interface, you simply apply the appropriate acl to the appropriate interface in the incoming direction. I understand you are trying to add support for iptables, but this is too hackish. IpAccessListLine is only supposed to filter by header bits (and classifiers).

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -1236,17 +1236,19 @@ private boolean dataPlaneDependenciesExist(TestrigSettings testrigSettings) {
...
-                  + outputClassName
-                  + " '"
-                  + name
-                  + "' from '"
-                  + inputPath
-                  + "'");
-          byte[] data = fromGzipFile(inputPath);
-          logger.debug(" ...OK\n");
-          dataByName.put(name, data);
+          if (!Files.isDirectory(inputPath)) {
```
arifogel: We have an open issue to scan directories recursively for input files. This filtering would then not be necessary, as the inputPath args supplied would be guaranteed not to be directories.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -1705,8 +1707,15 @@ public TestrigSettings getBaseTestrigSettings() {
       Path serializedVendorConfigPath, ConvertConfigurationAnswerElement answerElement) {
     Map<String, GenericConfigObject> vendorConfigurations =
         deserializeVendorConfigurations(serializedVendorConfigPath);
+    Map<String, GenericConfigObject> hostVendorConfigurations =
```
arifogel: This will interfere with resolution of the issue mentioned earlier to scan input directories recursively. Instead, we should have a new top-of-testrig-level folder

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -3169,6 +3178,97 @@ private void populateFlowHistory(FlowHistory flowHistory, String environmentName
     }
   }
 
+  // this is a hack to deal with router names that are named ethX
```
arifogel: I don't understand this comment. What does any of this code have to do with the name of the router? Did you mean interface names? If so, you should just use normalized name of the interface in the host config to be merged. Alternatively, use the appropriate normalization function for the vendor of the base configuration to be merged into.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -3169,6 +3178,97 @@ private void populateFlowHistory(FlowHistory flowHistory, String environmentName
...
+          throw new BatfishException("Merging iptables rules for " + hc.getName()
+              + ": only mangle tables are supported");
+        }
+
+        if (prerouting != null) {
+          for (Interface i: c.getInterfaces().values()) {
+
+            String dbgName = c.getHostname() + ":" + i.getName();
+
+            List<IpAccessListLine> newRules = prerouting.getLines().stream()
```
arifogel: Since you should not add input/output interface to IpAccessListLine, you may be able to accomplish your intended purpose here by creating a host-config-specific IpAccessListLine type class like we have for the other vendors.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4003,12 +4103,13 @@ private void serializeHostConfigs(
     // now, serialize
     _logger.info("\n*** SERIALIZING VENDOR CONFIGURATION STRUCTURES ***\n");
     resetTimer();
-    CommonUtil.createDirectories(outputPath);
+    Path hostsOutputPath = outputPath.resolve(BfConsts.RELPATH_HOST_CONFIGS_DIR);
```
arifogel: See earlier comment about where to put host configs meant to be merged.

projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesMatch.java
```diff
@@ -98,4 +98,12 @@ public IpWildcard toIpWildcard() {
 
     return subRanges;
   }
+
+  public String toInterfaceName() {
```
arifogel: This won't be necessary after earlier comments are addressed.

arifogel:
@dspicuzzbbn I made edits to the PR review with more context

projects/batfish/src/main/java/org/batfish/main/Batfish.java
dspicuzzbbn: Ok.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
dspicuzzbbn: Yes, that's what I meant.

The problem is that the normalization isn't applied uniformly -- it's applied to the cisco interface names, but not to host interface names.

My opinion is that the interface name normalization is a bad idea. Not only does it make it difficult to do matching, but it makes debugging more difficult also (like 'hm... why is my router interface Ethernet0 instead of eth0')

Issue #309, opened 2017-08-11
Delete EnvironmentCreationQuestionPlugin


dhalperi:
This is the old way of creating an environment; it should not be used any more in favor of the [`uploadenvironment` command](https://github.com/batfish/batfish/blob/5af643f910c8c927f13738d0770616c3df179508/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java#L1281).

This is primarily used by the Batfish Java client. In order to not break existing users, we should probably:

1. Implement an `init-environment` command (if it has not already been done).
2. Make `get environmentcreation` redirect to the `init-environment` command.
    Also we should log a warning.

ratulm:
#340

PR #340, opened 2017-08-17
Delete environment creation

ratulm:


ratulm:
#309

ratulm:
Invoking admin privileges here to reduce Dan's backlog. (Should be a safe change as it removes code used in only one place.)

PR #341, opened 2017-08-17
add ContainerResource and tests

kidsbear:
Add a subResource locator "ContainerResource" for WorkMgrServiceV2. 
Will change all function calls to WorkMgr once have new storage API call for that. e.g. listTestrig should should return a list of Testrigs instead of string.


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +27,39 @@
 @Path(CoordConsts.SVC_CFG_WORK_MGR2)
 @Produces(MediaType.APPLICATION_JSON)
 public class WorkMgrServiceV2 {
+
+  private BatfishLogger _logger = Main.getLogger();
+
+  @DefaultValue(CoordConsts.DEFAULT_API_KEY)
+  @HeaderParam(CoordConsts.SVC_KEY_API_KEY)
+  private String _apiKey;
```
kidsbear: Here I assume the key and version already been validated by filters, store apiKey for container accessibility check.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +27,39 @@
...
   /** Returns the list of {@link Container containers} that the given API key may access. */
   @GET
-  @Path("/containers")
-  public Response getContainers(
-      @DefaultValue(DEFAULT_API_KEY) @HeaderParam(SVC_KEY_API_KEY) String apiKey) {
-    List<Container> containers = Main.getWorkMgr().getContainers(apiKey);
+  @Path(CoordConsts.SVC_KEY_CONTAINERS)
+  public Response getContainers() {
+    _logger.info("WMS:getContainers\n");
+    List<Container> containers = Main.getWorkMgr().getContainers(_apiKey);
```
kidsbear: Should I return a sorted list for determinism?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +27,39 @@
...
-      @DefaultValue(DEFAULT_API_KEY) @HeaderParam(SVC_KEY_API_KEY) String apiKey,
-      @PathParam("id") String id) {
-    if (!Main.getAuthorizer().isAccessibleContainer(apiKey, id, false)) {
-      // TODO: put a proper error entity here.
-      return Response.status(Status.NOT_FOUND).build();
-    }
-
-    Container container = Main.getWorkMgr().getContainer(id);
-    return Response.ok(container).build();
+  public ContainerResource getResource(@PathParam("id") String id) {
```
kidsbear: Uses SubResource locator here, one issue here is the SubResource does not have @Path annotation, if I create tests for SubResource in it's own test class, I still need to do all initialization steps for the root resource(V2 here), and access them from the root resource, which will create a lot of redundant codes. Now I put tests together.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
dhalperi: It should be in the same order as `getContainers` returns it -- the contract should be on that function.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +27,39 @@
...
+  @DefaultValue(CoordConsts.DEFAULT_API_KEY)
+  @HeaderParam(CoordConsts.SVC_KEY_API_KEY)
+  private String _apiKey;
+
   /** Information on the URI of a request, injected by the server framework at runtime. */
   @Context private UriInfo _uriInfo;
 
   /** Returns the list of {@link Container containers} that the given API key may access. */
   @GET
-  @Path("/containers")
```
dhalperi: It looks to me like this will lose the `/`. Is  is not important?

Find the documentation, get back to me :)

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
dhalperi: Moving this to a property seems clean, but how can we set it during tests?

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,110 @@
...
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.core.UriBuilder;
+import javax.ws.rs.core.UriInfo;
+import org.batfish.common.BatfishLogger;
+import org.batfish.common.Container;
+import org.batfish.common.CoordConsts;
+import org.batfish.coordinator.Main;
+
+/**
+ * The ContainerResource is a RESTful service for servicing client API calls at container level.
```
dhalperi: Javadoc: anything that is a class should be in `{@link }` tags. (Maybe give the [style guide](https://google.github.io/styleguide/javaguide.html) a quick read).

This resource is not a service -- only the WorkMgrServiceV2 is a RESTful service. This is just a resource.

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -29,40 +27,39 @@
...
     return Response.status(Status.MOVED_PERMANENTLY)
-        .location(_uriInfo.getRequestUri().resolve("containers"))
+        .location(_uriInfo.getRequestUri().resolve(CoordConsts.SVC_KEY_CONTAINERS))
         .build();
   }
 
-  /** Returns information about the given {@link Container}, provided this user can access it. */
-  @GET
+  /** Relocate the request to ContainerResource. */
   @Path("/container/{id}")
```
dhalperi: The separation between `WorkMgrServiceV2` and `ContainerResource` seems a little weird to me. I would think a `Container`-specific resource would handle the `/{id}` link for example.

I'm not sure of the best approach here, what do you think?

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,110 @@
...
+  public Response getContainer() {
+    _logger.info("WMS: getContainer '" + _name + "'\n");
+    validate();
+    Container container = Main.getWorkMgr().getContainer(_name);
+    return Response.ok(container).build();
+  }
+
+  /** Initialize a new container with name: {@link #_name}. */
+  @POST
+  public Response initContainer() {
```
dhalperi: This relates back to my previous comment:
- I think there should be a `@POST` handler for the `/container` URL itself, and the body should be something with multiple optional fields: `name`, `prefix`, etc. That is to say, the `name` should not be in the URL for create -- instead, it should be a name or prefix request.



Aside, to match more what it's doing in an API sense, we should call the function createContainer, not initContainer.

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,110 @@
...
+    _logger.info("WMS: initContainer '" + _name + "'\n");
+    // TODO do we need to worry about containerPrefix here?
```
dhalperi: (see comment above)

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
kidsbear: We can set the value of API key or Version through the header of the request. [example](https://github.com/batfish/batfish/blob/master/projects/coordinator/src/test/java/org/batfish/coordinator/ApiKeyAuthenticationFilterTest.java#L73).

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
kidsbear: From Jersey documentation: A @Path value may or may not begin with a '/', it makes no difference. Likewise, by default, a @Path value may or may not end in a '/', it makes no difference, and thus request URLs that end or do not end in a '/' will both be matched.

PR #347, opened 2017-08-18
assign a value to testrigs path constant

kidsbear:
BfConst.RELPATH_TESTRIGS_DIR is empty now, the function call 'listTestrigs' will also include the analyses path in the result. Also, I modified initContainer to also create the path for testrigs and analyses(these were created when first time the client upload a testrig/analysis). Notice I didn't remove the existence check in listTestrigs and listAnalyses, they can't be empty from now on. (assume no direct operations on file system)

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
@@ -129,7 +129,10 @@ public class WorkMgr {
       Path containerDir =
           Main.getSettings().getContainersLocation().resolve(work.getWorkItem().getContainerName());
       String testrigName = work.getWorkItem().getTestrigName();
-      Path testrigBaseDir = containerDir.resolve(testrigName).toAbsolutePath();
+      Path testrigBaseDir =
+          containerDir
+              .resolve(Paths.get(BfConsts.RELPATH_TESTRIGS_DIR, testrigName))
+              .toAbsolutePath();
       task.put(BfConsts.ARG_CONTAINER_DIR, containerDir.toAbsolutePath().toString());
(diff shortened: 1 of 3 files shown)


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -706,6 +709,14 @@ public String initContainer(@Nullable String containerName, @Nullable String con
...
     }
+    Path testrigsDir = containerDir.resolve(BfConsts.RELPATH_TESTRIGS_DIR);
+    if (!testrigsDir.toFile().mkdir()) {
+      throw new BatfishException("failed to create directory '" + testrigsDir + "'");
+    }
+    Path analysesDir = containerDir.resolve(BfConsts.RELPATH_ANALYSES_DIR);
+    if (!analysesDir.toFile().mkdir()) {
+      throw new BatfishException("failed to create directory '" + analysesDir + "'");
+    }
     return containerName;
```
ratulm: What is the motivation for proactively creating these directories? Why not just let the first call to initialize a testrig or analysis take care of it?

kidsbear: I feel this makes more sense when we has Object representation of it. For example, a container object has a list of Analysis, I feel an empty list should map to an empty folder, and null map to non-exist folder. In this case, I think a null value doesn't make sense. Also, initialization of testrig now creates all needed paths, I think they should be consistency.

ratulm: Ok, that is a reasonable way to look at it. Thanks.

PR #348, opened 2017-08-19
assertions for jsonpath question

### LINKED ISSUE #145: provide assert question
Assert question on jsonpaths

ratulm:
Fixes #145 

I extended the JsonPath question itself such that you can specify assertions on top of the answers you get. Two operators ("eq" and "count") are supported currently, and we can add more later.

Examples for the Java client (from the tests):

get jsonpath innerQuestion: {"class": "org.batfish.question.NodesQuestionPlugin$NodesQuestion"}, paths: [{"path" : "$.nodes['as2core2']..mtu", "assertion": {"type" : "count", "expect": 6}}]

get jsonpath innerQuestion: {"class": "org.batfish.question.NodesQuestionPlugin$NodesQuestion"}, paths: [{"path" : "$.nodes['as2core2']..mtu", "assertion": {"type" : "eq", "expect": [1500, 1800, 1600, 1700, 1500, 1500]}}]


tests/basic/jsonpath-assertion-count.ref
```diff
@@ -0,0 +1,44 @@
+{
+  "answerElements" : [
+    {
+      "class" : "org.batfish.question.jsonpath.JsonPathQuestionPlugin$JsonPathAnswerElement",
+      "results" : {
+        "0" : {
+          "assertionResult" : false,
```
dhalperi: add a test where the assertion passes?

projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathAssertion.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String PROP_EXPECT = "expect";
+
+  private JsonPathAssertionType _assertionType;
+
+  private JsonNode _expect;
+
+  public boolean evaluate(ArrayNode suffixes) {
+    switch (getType()) {
+    case count:
+      if (_expect.isInt()) {
```
dhalperi: I prefer to write these with exit paths early, business logic late. It's easier for a reader to confirm properties like "we don't fall through the case":

```java
if (!_expect.isInt()) {
  throw...
}

return suffixes.size() == _expect.asInt();
```

projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathAssertion.java
```diff
@@ -0,0 +1,77 @@
...
+        boolean match = (suffixes.size() == _expect.asInt());
+        return match;
+      } else {
+        throw new BatfishException("Expected value of assertion type count ("
+            + _expect.toString() + ") is not an integer");
+      }
+    case equal:
```
dhalperi: add some Java-level unit tests... things that come to mind are verifying `evaluate` for true and false assertions, complex objects, and type mismatches.

(What happens if it's not an ArrayNode?)

projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java
```diff
@@ -159,7 +162,10 @@ public JsonPathAnswerElement answer() {
                 nodePathResult.setPath(nodesPath);
                 nodePathResult.setNumResults(numResults);
                 boolean includeSuffix = nodesPath.getSuffix();
-                if (!nodesPath.getSummary()) {
+                if (nodesPath.hasValidAssertion()) {
+                  boolean assertion = nodesPath.getAssertion().evaluate(suffixes);
+                  nodePathResult.setAssertionResult(assertion);
+                } else if (!nodesPath.getSummary()) {
```
dhalperi: why can't I make an assertion over the prefix as well? e.g., that node:mtu is right?

dhalperi: given that the default for JsonPath is to return the prefixes, this seems an odd choice.

tests/basic/jsonpath-assertion-count.ref
ratulm: done (which revealed another bug)

projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java
ratulm: Umm....the count assertion will let you check if the path exists, right?

ratulm: anyway, we can expand the assert operators when a use case come up.

ratulm:
@dhalperi this is ready for review again .... once we are done here, i'll modify (some of) the templates so that they can be used as assertions.

projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java
dhalperi: It's not about expanding the operators. Right now, the assertion operates only on the suffix (rather than, say, the entire JsonPath result) -- it's not possible to extend cleanly to including the prefix as well, not without adding something like prefixAssertion or wholeResultAssertion, e.g., – different operators entirely.

tests/basic/jsonpath-assertion-count.ref
dhalperi: Looking at the bugfix, it seems you just set summary false in the inner questions in the tests. so the bug was that the inner question had summary set to true?

That seems error-prone – how would we tell that the assertion count=0 succeeded vs summary was set incorrectly. Can you think of a way this could be done in code – rejecting questions that are likely to be malformed?

PR #350, opened 2017-08-21
Modify container

kidsbear:
Modify the Container object representation. With this representation, it should be able to implement functions like delAnalysis, listAnalysis, etc. Functions about testrigs will make more sense when we have Testrig object representation (listTestrigs, delTestrig, etc.), it will be the next thing to add~


projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -34,35 +42,37 @@ public String getName() {
...
-    _name = name;
+  @JsonProperty(PROP_ANALYSES)
+  public Map<String, Analysis> getAnalyses() {
+    return _analyses;
   }
 
-  @JsonProperty(PROP_TESTRIGS)
-  public void setTestrigs(SortedSet<String> testrigs) {
-    _testrigs = testrigs;
+  public void addAnalysis(String analysisName, @Nullable Analysis analysis) {
```
hverma-upenn: I think it would not be useful to pass a null analysis in this function, so "Nullable" does not seem necessary

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -518,21 +520,13 @@ public String getAnswer(
...
-    SortedSet<String> testrigs =
-        new TreeSet<>(
-            CommonUtil.getSubdirectories(containerDir.resolve(BfConsts.RELPATH_TESTRIGS_DIR))
-                .stream()
-                .map(dir -> dir.getFileName().toString())
-                .collect(Collectors.toSet()));
-
-    return Container.of(containerDir.toFile().getName(), testrigs);
+    SortedSet<String> testrigs = listTestrigs(containerName);
+    return new Container(containerName, new ArrayList<>(testrigs), new HashMap<>());
```
hverma-upenn: I think this can be a storage api which returns a container object, constructing a new Container object in this method may not be consistent

kidsbear: Right, I will modify this to use the storage API call when we have the implementation in storage API, now it is actually a temp function.

PR #351, opened 2017-08-21
configureAnalysis in WorkMgr to accept a map instead of inputStream

kidsbear:
Modify configureAnalysis function to accept a map of questions to add to analysis, move the logic of getting questions from inputStream to WorkMgrService.


projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrServiceTest.java
```diff
@@ -89,13 +91,33 @@ public void getNonEmptyContainer() throws Exception {
   }
 
   @Test
-  public void testDeleteQuestionsFromAnalysis() throws Exception {
+  public void testConfigureAnalysis() throws Exception {
```
kidsbear: These two tests are actually very long, but since we need to initialize analysis and add questions before testing deleting them, I make these tests together. I think we will split this function into three parts later (initAnalysis, addAnalysisQuestion, delAnalysisQuestion).

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -156,6 +160,28 @@ public JSONArray configureAnalysis(
...
+          }
+          String questionText;
+          try {
+            questionText = qJson.toString(1);
+          } catch (JSONException e) {
+            throw new BatfishException("Failed to convert question JSON to string", e);
+          }
+          questionsToAdd.put(qName, questionText);
+        }
+      }
```
dhalperi: 1. It looks like this block (Stream -> Map<String, String>) could be moved into a static utility method. This would be much clearer to read, documented, and could be tested in isolation without worrying about file system, etc.

2. Why can we not simply do something like use Jackson to read a Map<String, Object> from the stream?

kidsbear: I tried the second approach to use Jackson to convert the input raw question text into a Question object, it gives exception like this : "com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Missing type id when trying to resolve subtype of [simple type, class org.batfish.datamodel.questions.Question]: missing type id property 'class'". 
At Batfish size, we actually [preprocessQuestion](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L3213) before converting it into a Question object. So I just string to store the question context here.

kidsbear:
Since the function is not general enough, it requires the inputStream to contain a mapping from String to JSON String, should I just add these lines into WorkMgrService instead of creating a method in commonUtil? Also, should I keep the `writeStreamToJSONObject ` method in CommonUtil?

PR #352, opened 2017-08-22
add Environment object type

kidsbear:
Object representation of 'Environment', this representation should be enough for function calls related to environment in WorkMgr, (mostly uploadEnvironment). Use information of [initEnvFilenames](https://github.com/batfish/batfish/blob/master/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java#L68) as a reference of what we expect from client. For each of them, check how BatFish uses these informations inside environment. [example1](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L1756), [example2](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2903)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
```diff
@@ -0,0 +1,132 @@
...
+  public Environment(
+      @JsonProperty(PROP_NAME) String name,
+      @JsonProperty(PROP_EDGE_BLACKLIST) @Nullable List<Edge> edgeBlacklist,
+      @JsonProperty(PROP_INTERFACE_BLACKLIST) @Nullable List<NodeInterfacePair> interfaceBlacklist,
+      @JsonProperty(PROP_NODE_BLACKLIST) @Nullable List<String> nodeBlacklist,
+      @JsonProperty(PROP_ENVIRONMENT_BGP_TABLES) @Nullable Map<String, String> bgpTables,
+      @JsonProperty(PROP_ENVIRONMENT_ROUTING_TABLES) @Nullable Map<String, String> routingTables,
+      @JsonProperty(PROP_EXTERNAL_BGP_ANNOUNCEMENTS) @Nullable String externalBgpAnnouncements) {
+    this._name = name;
+    this._edgeBlacklist = edgeBlacklist == null ? new ArrayList<>() : edgeBlacklist;
```
kidsbear: Since these files are not required to be present, should we allow null values here?

ratulm: I think so. For these files, we may want to be able to distinguish between file not present and an empty file. The former means blacklist wasn't specified, and the latter means that no (link/node) was blacklisted.

PR #353, opened 2017-08-22
separated out some interfaces (ready to merge)

ratulm:
@arifogel trying to get plugins going for injecting functionality into coordinator. i am mirroring what happens for batfish. it is not done yet, but could use your help if getting it going. more in person.


projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
```diff
@@ -194,8 +194,16 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs)
...
+    Class<?> pluginClass = null;
+    try {
+      cl.loadClass(className);
+      pluginClass = Class.forName(className, true, cl);
+    } catch (ClassNotFoundException | NoClassDefFoundError e) {
+      //Ignoring this exception is a potentially dangerous hack
+      //but I don't quite know yet why some classes cannot be loaded and how to do things cleanly
+      System.err.println("Couldn't load class " + className + ". Skipping.");
+      return;
+    }
```
ratulm: Pay attention here

ratulm:
@arifogel ready to merge now. please take a look.

projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
```diff
@@ -167,8 +168,21 @@ public void run() {
   }
 
   private void runCoordinator() {
-
-    final String[] argArray = getArgArrayFromString(_settings.getCoordinatorArgs());
+    String coordinatorArgs =
+        String.format("%s", _settings.getCoordinatorArgs());
```
dhalperi: is the `String.format` call here not a no-op?

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -2606,6 +2610,66 @@ private boolean showVersion(List<String> options, List<String> parameters) {
...
 
+  private boolean syncTestrigsSyncNow(List<String> options, List<String> parameters) {
+    if (!isValidArgument(options, parameters, 1, 1, 1,
+            Command.SYNC_TESTRIGS_SYNC_NOW)) {
+      return false;
+    }
+    if (!isSetContainer(true)) {
+      return false;
+    }
+
```
dhalperi: do you have the `google-java-format` plugin installed that Ari set up to keep everything consistent? Looks like this file is fairly malformed w.r.t. that format.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -2606,6 +2610,66 @@ private boolean showVersion(List<String> options, List<String> parameters) {
...
+    String settingsStr = "{"
+            + String.join(" ", Arrays.copyOfRange(words, 2 + options.size(), words.length))
+            + "}";
+
+    Map<String, String> settings = null;
+
+    try {
+      BatfishObjectMapper mapper = new BatfishObjectMapper();
+      settings = mapper.readValue(new JSONObject(settingsStr).toString(),
```
dhalperi: what does `new JSONObject(string).toString()` do to the string?

projects/batfish-client/src/main/java/org/batfish/client/Command.java
```diff
@@ -275,6 +277,12 @@
     descs.put(SHOW_DELTA_TESTRIG, new Pair<>("", "Show delta testrig and environment"));
     descs.put(SHOW_TESTRIG, new Pair<>("", "Show base testrig and environment"));
     descs.put(SHOW_VERSION, new Pair<>("", "Show the version of Client and Service"));
+    descs.put(SYNC_TESTRIGS_SYNC_NOW, new Pair<>("-force <plugin-id>",
```
dhalperi: `[-force]`?

projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
dhalperi: No strong objection, but integrate with the logging framework rather than syserr.

projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/SyncTestrigsPlugin.java
```diff
@@ -0,0 +1,17 @@
+package org.batfish.common.plugin;
+
+import java.util.Map;
+
+public abstract class SyncTestrigsPlugin extends CoordinatorPlugin {
```
dhalperi: Can you add significant javadoc here explaining what this plugin is for, how it's used, etc?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -43,7 +45,7 @@
 import org.codehaus.jettison.json.JSONObject;
 import org.glassfish.jersey.uri.UriComponent;
 
-public class WorkMgr {
+public class WorkMgr extends AbstractCoordinator {
```
dhalperi: `WorkMgr` isn't a coordinator -- seems like something's wrong here.

projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
ratulm: Yes, it was. Good catch.

projects/batfish-client/src/main/java/org/batfish/client/Client.java
ratulm: Fixed (by you:)

projects/batfish-client/src/main/java/org/batfish/client/Client.java
ratulm: The parser that JSONObject uses is more liberal (and easier for humans to handle). We use that to parse the user supplied input and then the toString() function spits out a "proper" JSON string that mapper can parse. (We use this pattern in another place.)

projects/batfish-client/src/main/java/org/batfish/client/Command.java
ratulm: Yes.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -146,6 +146,7 @@
 import org.batfish.grammar.assertion.AssertionCombinedParser;
 import org.batfish.grammar.assertion.AssertionExtractor;
 import org.batfish.grammar.assertion.AssertionParser.AssertionContext;
+import org.batfish.grammar.cisco.CiscoParser.Color_setterContext;
```
dhalperi: Looks like you added an unused import -- this will make Ari sad later.

(The Java Findbugs plugin, once fully supported, will be able to catch this in Travis. The issue right now is that there are so many bugs :).

PR #354, opened 2017-08-22
Fixes #345: extends reachability to support checks for transit nodes

### LINKED ISSUE #345: path policy question
@ratulm To check that a class of traffic originating at ingressNode/iface to some destination always traverses an ordered set of waypoints, it seems that would be doable using the traceroute question. Specifically, use traceroute underneath to get all paths, and check that each contains the ordered waypoints. Makes sense?

dspicuzzbbn:
This adds support for the following:
   - transitNodes: set of transit nodes (packet must transit through all of them)
   - notTransitNodes: set of non-transit nodes (packet does not transit through any of them)

Now one can check that all traffic ingress|egress to a zone goes through a transit node for example by issuing the following query

 ```
get reachability type=standard, ingressNodeRegex=from, srcIps=[1.1.1.0/24], dstIps=[2.2.2.2], actions=[accept], notTransitNodes=[firewall]
```

From @jkhourybbn

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
@@ -156,7 +156,9 @@ public interface IBatfish extends IPluginConsumer {
       String ingressNodeRegexStr,
       String notIngressNodeRegexStr,
       String finalNodeRegexStr,
-      String notFinalNodeRegexStr);
+      String notFinalNodeRegexStr,
+      Set<String> transitNodes,
+      Set<String> notTransitNodes);
 
   void writeDataPlane(DataPlane dp, DataPlaneAnswerElement ae);
 
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -3611,13 +3611,13 @@ public class Batfish extends PluginConsumer implements IBatfish {
         ReachabilityQuerySynthesizer acceptQuery =
             new ReachabilityQuerySynthesizer(
                 Collections.singleton(ForwardingAction.ACCEPT), headerSpace,
-                Collections.<String>emptySet(), nodeVrfs);
+                Collections.<String>emptySet(), nodeVrfs, null, null);
         ReachabilityQuerySynthesizer notAcceptQuery =
             new ReachabilityQuerySynthesizer(
                 Collections.singleton(ForwardingAction.ACCEPT),
                 new HeaderSpace(),
                 Collections.<String>emptySet(),
-                nodeVrfs);
+                nodeVrfs, null, null);
         notAcceptQuery.setNegate(true);
         NodeVrfSet nodes = new NodeVrfSet();
         nodes.add(new Pair<>(node, vrf));
@@ -4167,7 +4167,9 @@ public class Batfish extends PluginConsumer implements IBatfish {
       String ingressNodeRegexStr,
       String notIngressNodeRegexStr,
       String finalNodeRegexStr,
-      String notFinalNodeRegexStr) {
+      String notFinalNodeRegexStr,
+      Set<String> transitNodes,
+      Set<String> notTransitNodes) {
     if (SystemUtils.IS_OS_MAC_OSX) {
       // TODO: remove when z3 parallelism bug on OSX is fixed
       _settings.setSequential(true);
@@ -4218,6 +4220,27 @@ public class Batfish extends PluginConsumer implements IBatfish {
               + "'");
     }
 
+    //check transit nodes
+    Set<String> allNodes = configurations.keySet();
+    if (transitNodes != null) {
+      for (String node: transitNodes) {
+        if (!allNodes.contains(node)) {
+          return new StringAnswerElement(String.format("Unknown transit node %s", node));
+        }
+      }
+    }
+    if (notTransitNodes != null) {
+      for (String node: notTransitNodes) {
+        if (!allNodes.contains(node)) {
+          return new StringAnswerElement(String.format("Unknown notTransit node %s", node));
+        }
+        if (transitNodes != null && transitNodes.contains(node)) {
+          return new StringAnswerElement(
+              String.format("Same node %s can not be in both transit and notTransit", node));
+        }
+      }
+    }
+
     // build query jobs
     List<NodJob> jobs = new ArrayList<>();
     for (String ingressNode : activeIngressNodes) {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: projects/batfish/src/main/java/org/batfish/main/Batfish.java [hunk 1]
```diff
@@ -3611,13 +3611,13 @@ public AnswerElement reducedReachability(HeaderSpace headerSpace) {
         ReachabilityQuerySynthesizer acceptQuery =
             new ReachabilityQuerySynthesizer(
                 Collections.singleton(ForwardingAction.ACCEPT), headerSpace,
-                Collections.<String>emptySet(), nodeVrfs);
+                Collections.<String>emptySet(), nodeVrfs, null, null);
```
dhalperi: Prefer `emptySet` instead of making the variable nullable.

projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
```diff
@@ -156,7 +156,9 @@ AnswerElement standard(
       String ingressNodeRegexStr,
       String notIngressNodeRegexStr,
       String finalNodeRegexStr,
-      String notFinalNodeRegexStr);
+      String notFinalNodeRegexStr,
+      Set<String> transitNodes,
+      Set<String> notTransitNodes);
```
dhalperi: If these are allowed to be `null`, annotate them `@Nullable`. But honestly, seems easier if they're not and we just require the caller to pass the empty set.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4167,7 +4167,9 @@ public AnswerElement standard(
       String ingressNodeRegexStr,
       String notIngressNodeRegexStr,
       String finalNodeRegexStr,
-      String notFinalNodeRegexStr) {
+      String notFinalNodeRegexStr,
+      Set<String> transitNodes,
+      Set<String> notTransitNodes) {
```
dhalperi: If these are allowed to be null, annotate `@Nullable`. Either way, when stored they should not be allowed to be null

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -4218,14 +4220,36 @@ public AnswerElement standard(
               + "'");
     }
 
+    //check transit nodes
+    Set<String> allNodes = configurations.keySet();
+    if (transitNodes != null) {
```
dhalperi: Can simplify using [`Sets.difference`](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Sets.html#difference-java.util.Set-java.util.Set-) as follows:

```java
Set<String> allNodes = configurations.keySet();
Set<String> invalidTransitNodes = Sets.difference(transitNodes, allNodes);
if (!invalidTransitNodes.isEmpty()) {
   ... 
}
Set<String> invalidNotTransitNodes = Sets.difference(notTransitNodes, allNodes);
if (!invalidNotTransitNodes.isEmpty()) {
   ... 
}
Set<String> illegalTransitNodes = Sets.intersection(transitNodes, notTransitNodes);
if (!illegalTransitNodes.isEmpty()) {
   ... 
}
```

projects/batfish/src/main/java/org/batfish/z3/ReachabilityQuerySynthesizer.java
```diff
@@ -163,6 +173,23 @@ public NodProgram getNodProgram(NodProgram baseProgram) throws Z3Exception {
     queryConditions.addConjunct(finalActions);
     queryConditions.addConjunct(SaneExpr.INSTANCE);
 
+    // check transit constraints (unordered)
+    if (_transitNodes != null && _transitNodes.size() > 0) {
```
dhalperi: rule of thumb I'm trying to put in: no variable that's not `@Nullable` can be `null`. So many ways to fix here:

- constructor parameter is not `@Nullable`
- constructor is `@Nullable`, but we default to empty in constructor so that `_transitNodes` is not `@Nullable`
- `_transitNodes` is `@Nullable`.

I lean towards first or second alternative. Because then we can drop the `if` altogether -- the list will always be present (maybe empty). AFAICT, the `.size() > 0` check is not needed, because it'll just be a `for` loop that terminates immediately.

projects/question/src/main/java/org/batfish/question/ReachabilityQuestionPlugin.java
```diff
@@ -18,6 +18,7 @@
 import org.batfish.datamodel.answers.AnswerElement;
 import org.batfish.datamodel.questions.IReachabilityQuestion;
 import org.batfish.datamodel.questions.Question;
+import scala.collection.generic.Sorted;
```
dhalperi: Looks like this was introduced accidentally, unused?

projects/question/src/main/java/org/batfish/question/ReachabilityQuestionPlugin.java
```diff
@@ -523,6 +552,16 @@ public void setFinalNodeRegex(String regex) {
...
+    @JsonProperty(PROP_TRANSIT_NODES)
+    public void setTransitNodes(SortedSet<String> transitNodes) {
+      _transitNodes = transitNodes;
+    }
+
+    @JsonProperty(PROP_NOT_TRANSIT_NODES)
+    public void setNotTransitNodes(SortedSet<String> notTransitNodes) {
+      _notTransitNodes = notTransitNodes;
+    }
+
```
dhalperi: Can you please add tests of these new features to the example tests? 

Probably, just go in [here](https://github.com/batfish/batfish/blob/master/tests/basic/commands#L35) similar to what you wrote in the PR desc, with new ref files.

jkhourybbn:
@dhalperi some comments. 

First in terms of adding a test case, I didn't do so because the existing test cases are by no means exhaustive -- for example, you don't have tests in there for standard reachability let alone for all params of the reachability query, and adding a test to one of your existing testrigs would require undue work at this point to make sure we don't use it in ways you didn't want. We could definitely consider adding one later but I don't think we should hold this PR as a result?

In terms of nulls, while I do not agree with adding empty pointers all over the place as a best practice to avoid nulls, happy to do that if this the approach you want to enforce.

jkhourybbn:
@dhalperi I made the stylistic changes per your comments. 

Regarding the test case, yes I am a big proponent of test cases. In this case, it seemed like it is the first standard reachability test in your suite hence my comment. We do plan to add tests going forward.

dhalperi:
Looks great! Will merge once Travis testing is green.

dhalperi:
Looks great! Will merge once Travis testing is green.

Thanks again, @dspicuzzbbn and @jkhourybbn !

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -3611,13 +3612,14 @@ public AnswerElement reducedReachability(HeaderSpace headerSpace) {
...
-                Collections.<String>emptySet(), nodeVrfs);
+                Collections.<String>emptySet(), nodeVrfs,
+                Collections.<String>emptySet(), Collections.<String>emptySet());
         ReachabilityQuerySynthesizer notAcceptQuery =
             new ReachabilityQuerySynthesizer(
                 Collections.singleton(ForwardingAction.ACCEPT),
                 new HeaderSpace(),
                 Collections.<String>emptySet(),
-                nodeVrfs);
+                nodeVrfs, null, null);
```
dhalperi: I'm guessing this line still needs to be updated for travis to pass, maybe?

jkhourybbn: Interesting, I ran the travis build script locally and it passed.

Issue #97, opened 2017-04-12
question answer summaries


arifogel:
Every question answer should have a one-line summary field.

PR #355, opened 2017-08-23
Answer summaries

ratulm:
this got messy in terms of commits because i made the mistake of working off of the assertions branch. in reality, i had just two commits. sorry about that. the "Files changed" view does the right thing, however, because assertions branch was merged into master.


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/AnswerSummary.java
```diff
@@ -0,0 +1,84 @@
+package org.batfish.datamodel.answers;
+
+import com.google.common.base.Strings;
+
+public class AnswerSummary {
```
dhalperi: I think this needs to be JSON Serializable -- given that `AnswerSummary` is now a `@JsonProperty` of `AnswerElement` entity.

dhalperi: I guess it in fact is, using the default Jackson inspection of objects. That's not the usual standard in Batfish, but ok with me.

ratulm: If you don't do @JsonProperty, Jackson will use field names based on getter and setter names. I started using @JsonProperty because I was worried about people changing getter and setter names, but it does create a lot of clutter and extra LoC. I now think a better way to deal is to have serialization and deserialization unit tests, which I added for this class. 

If you think @JsonProperty buys us something, I can add those annotations.

ratulm:
Fixes #97

PR #361, opened 2017-08-25
Various fixes to the cisco parser

dspicuzzbbn:


test_rigs/unit-tests/configs/cisco_misc
```diff
@@ -36,6 +36,8 @@ username abc
 username rancid attributes
  vpn-group-policy GroupPolicy_NOC-VPN
  group-lock value NOC-VPN
+username bob1 password 1 1234
+username bob2 privilege 1 password 1 1234
```
dhalperi: @arifogel : separately, we should make sure this information ends up in the data model, too.

PR #363, opened 2017-08-25
fix initial setting of origin type for juniper

arifogel:


projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
```diff
@@ -321,10 +322,11 @@ private BgpProcess createBgpProcess(RoutingInstance routingInstance) {
       Disjunction isBgp = new Disjunction();
       isBgp.getDisjuncts().add(new MatchProtocol(RoutingProtocol.BGP));
       isBgp.getDisjuncts().add(new MatchProtocol(RoutingProtocol.IBGP));
-      setOriginForNonBgp.setGuard(isBgp);
-      setOriginForNonBgp
-          .getTrueStatements()
+      BooleanExpr isNotBgp = new Not(isBgp);
+      setOriginForNonBgp.setGuard(isNotBgp);
```
dhalperi: reason for introducing `isNotBgp` instead of inlining in `setGuard`?

arifogel: Actually, could just use falseStatements instead of trueStatements

arifogel: For new BGP advertisements, i.e. those created from non-BGP routes, an origin code must be set. Juniper sets it to IGP by default. The code to do that was there, but there were two bugs:

  1. The condition was inverted.
  2. The If statement was never added to the policy.

PR #364, opened 2017-08-25
make Jersey client class field in WorkMgr

kidsbear:
The configuration for two Clients are different. I am not sure whether we can use one of them to satisfy both, I keep two Clients in WorkMgr now. No multithread here~


kidsbear:
You are right. I forgot it is a service. Since Jersey removed the guarantee of thread-safe, we need one client for each function call?

PR #365, opened 2017-08-25
Add AnalysisAnswer object type

kidsbear:
Object representation of AnalysisAnswer


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/AnalysisAnswer.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p>Each {@link AnalysisAnswer AnalysisAnswer} contains the corresponding analysis name, an
+ * environment the analysis {@link #_analysisName} were answered, and a list of <@link Answer
+ * answer> as the result.
+ */
+public class AnalysisAnswer {
+  private static final String PROP_ANALYSIS_NAME = "analysisName";
+  private static final String PROP_ENVIRONMENT = "environment";
+  private static final String PROP_ANSWERS = "answers";
+
+  private final String _analysisName;
```
dhalperi: Per in-person discussion, the whole object might not be needed. instead, `getAnalysisAnswers(String containerName, Scenario baseScenario, @Nullable Scenario deltaScenario, String analysisName)` can just return `List<Answer>` .. for example.

PR #367, opened 2017-08-25
Add TestrigQuestion object type

kidsbear:
The representation of testrig Question. Currently create testrig question by using 'get' or 'answer' command, each time it creates a new question with questionName prefix. I am not sure whether we will allow same testrig question to be answered in different environments, (now we create new question every time when it been asked in java client), if so, will need to modify it to have something like a mapping from environment to answer.
A question: Is there any preference of Question type over string representation of the question? I am using string in both Analysis and here for question.


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/Answer.java
```diff
@@ -101,4 +101,8 @@ public void setQuestion(Question question) {
   public void setStatus(AnswerStatus status) {
     _status = status;
   }
+
+  @Override public String toString() {
+    return prettyPrint();
+  }
```
dhalperi: this seems an unrelated change. Pull into its own PR with justification?

dhalperi: (also, please use the formatter to fix the format of this file.)

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/TestrigQuestion.java
```diff
@@ -0,0 +1,85 @@
...
+import org.batfish.datamodel.answers.Answer;
+
+/**
+ * The {@link TestrigQuestion TestrigQuestion} is an Object representation of the testrig question
+ * for BatFish service.
+ *
+ * <p>Each {@link TestrigQuestion TestrigQuestion} contains the name of the question, the question
+ * content, an environment, and the answer of the question in that environment.
+ */
+public class TestrigQuestion {
```
dhalperi: Why is this not just `Question`? What is `Testrig` about this?

Why does it have an `Answer` field?

kidsbear: Questions under testrig folder are created by commands like 'get' or 'answer'(need to load question json first) command, when we create the question, we also generate the answer for that question, so I put the answer as part of TestrigQuestion. Also, I named it TestrigQuestion because it is different from Questions under analysis, which only contains Question instance, does not requires the questions been answered.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/Answer.java
kidsbear: This changed is added for supporting the toString method for the TestrigQuestion, I can make a separate PR for it~

PR #368, opened 2017-08-25
Storage Analysis APIs

haverma:
Creating a new PR for storage analysis APIs as some irrelevant changes(from some other branch) made into the older PR,


projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+public class FileStorageImpl implements Storage {
+
+  private final Path _containersLocation;
+
+  public FileStorageImpl(Path containersLocation) throws BatfishException {
+    try {
+      if (containersLocation != null) {
+        _containersLocation = containersLocation;
+        _containersLocation.toFile().mkdir();
```
kidsbear: Do we need to check the existence of all parent directories? Or use mkdirs() function?

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Retrieve a container Path
+   *
+   * @param containerName Container name
+   * @return Path of the container
+   */
+  // deprecated
+  @Override
+  public Path getContainerPath(String containerName) {
```
kidsbear: Should be a private method?

hverma-upenn: This method was to be used to get container path using container name from storage apis for the places where container path was necessary. But I think this is not being used anywhere right now and is irrelevant with respect to this pull request, since we are only targeting analysis; I will remove this function

kidsbear: I thought we shouldn't provide any public functions that returns any file system path~ Maybe I was wrong~ :)

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  public Analysis updateAnalysis(String containerName, Analysis analysis) {
+    Analysis oldAnalysisObj = getAnalysis(containerName, analysis.getName());
+    Set<String> questionsToDelete =
+        Sets.difference(oldAnalysisObj.getQuestions().keySet(), analysis.getQuestions().keySet());
+    Set<String> questionsToAdd =
+        Sets.difference(analysis.getQuestions().keySet(), oldAnalysisObj.getQuestions().keySet());
+    Path questionsDir =
```
kidsbear: I think it would be nice to have all private function calls to get the directories. When the number of function grows in this class, using Path.resolve everywhere will make it hard to manage.

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
hverma-upenn: No, you are correct, it was for backward compatibility but right now noone is using it so it can be removed

projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    _storage.saveAnalysis(containerName, analysis);
+  }
+
+  @Test
+  public void testUpdateAnalysis() {
+    Analysis analysis =
+        new Analysis(
+            "testanalysisupdate", Collections.singletonMap("testquestion", "testquestionvalue"));
+    _storage.saveAnalysis("testcontainerupdate", analysis);
+    Analysis savedAnalysis = _storage.getAnalysis("testcontainerupdate", "testanalysisupdate");
```
kidsbear: Any reason for not using return value from previous call?

projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Test
+  public void testDeleteNonEmptyAnalysis() {
+    Analysis analysis =
+        new Analysis(
+            "testanalysis", Collections.singletonMap("testquestion", "testquestioncontent"));
+    _storage.saveAnalysis("testcontainer", analysis);
+    _thrown.expect(BatfishException.class);
+    _thrown.expectMessage(
+        String.format("%s' is not empty, deletion must be forced", "testanalysis"));
```
kidsbear: There should be a leading '? Suggest to use _thrown.expectMessage(equalsTo()) for comparing message equality, default only checks the actual message contains the parameter string as subString.

projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
```diff
@@ -55,6 +58,10 @@ public static PoolMgr getPoolMgr() {
     return _poolManager;
   }
 
+  public static Storage getStorage() {
+    return _storage;
```
kidsbear: Suggest to use checkState() function here for consistency.

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
hverma-upenn: Done

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
hverma-upenn: Done

projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileStorageImpl.java
hverma-upenn: Done, separated the functions to get various paths in a StorageUtils class

projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
hverma-upenn: Done

projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
hverma-upenn: I am not sure about this comment, as Stack Overflow https://stackoverflow.com/questions/2469911/how-do-i-assert-my-exception-message-with-junit-test-annotation and existing code in BatfishTest and other classes do the same

projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
hverma-upenn: Done

projects/batfish-common-protocol/src/main/java/org/batfish/storage/Storage.java
```diff
@@ -0,0 +1,16 @@
...
+/** Common storage APIs */
+public interface Storage {
+  //Storage APIs for Configure Analysis
+
+  Analysis getAnalysis(String containerName, String analysisName);
+
+  Analysis saveAnalysis(String containerName, Analysis analysis);
+
+  Analysis updateAnalysis(String containerName, Analysis analysis);
+
```
kidsbear: Should have a function "listAnalyses"?

hverma-upenn: Thanks for pointing that out

hverma-upenn:
Splitting the PR into smaller changes, this PR now just contains storage interface and implementations and relevant unit tests

PR #372, opened 2017-08-26
PrependAsPath: optimize algorithm and fix bug

dhalperi:
When prepending a path of N nodes, we used to insert each node at the front of existing lists.
For something like an array list, this is N full shifts of the array. Optimize this by
computing the entire prefix to be appended first, then doing a single array shift.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/statement/PrependAsPath.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/statement/PrependAsPath.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.google.common.collect.ImmutableSortedSet;
 import java.util.List;
 import java.util.SortedSet;
+import java.util.stream.Collectors;
 import org.batfish.datamodel.BgpRoute;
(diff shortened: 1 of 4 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/statement/PrependAsPath.java
```diff
@@ -48,20 +50,23 @@ public boolean equals(Object obj) {
...
     BgpRoute.Builder bgpRouteBuilder = (BgpRoute.Builder) environment.getOutputRoute();
     List<SortedSet<Integer>> asPath = bgpRouteBuilder.getAsPath();
-    List<SortedSet<Integer>> iAsPath = null;
+    asPath.addAll(0, newAsPaths);
+
     if (environment.getWriteToIntermediateBgpAttributes()) {
       BgpRoute.Builder ir = environment.getIntermediateBgpAttributes();
-      iAsPath = ir.getAsPath();
-    }
-    for (Integer as : toPrepend) {
```
arifogel: I think the original and proposed implementations are wrong here. What I wanted to do was simply concatenate original list to the prepend list.

dhalperi:
Fixed bug and added tests.

PR #378, opened 2017-08-28
Adding builder to Environment

haverma:
It will be needed to create Environment object incrementally while uploading environment as we don't have public setters and there is no zero argument constructor


kidsbear:
Actually, I am not sure we need this builder now. Since we don't modify an existing environment, I don't feel we need setters now.

PR #380, opened 2017-08-28
Storage environment APIs

haverma:
Branched off from "storage" branch, dependent on PR #368. Can be reviewed after PR #368 is merged so that this shows the changes only for environment storage APIs


projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
```diff
@@ -0,0 +1,287 @@
...
+    String envName = "testenv";
+    Environment environment = Environment.builder().setName(envName).build();
+    _storage.saveEnvironment(containerName, testrigName, environment);
+
+    assertTrue(_storage.deleteEnvironment(containerName, testrigName, envName, false));
+
+    _thrown.expect(BatfishException.class);
+    _thrown.expectMessage(
+        String.format(
+            String.format(
```
kidsbear: Two String.format?

projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileStorageImplTest.java
```diff
@@ -0,0 +1,287 @@
...
+    Environment environment = getTestEnvironment(envName);
+    _storage.saveEnvironment(containerName, testrigName, environment);
+    _thrown.expect(BatfishException.class);
+    _thrown.expectMessage(String.format("'%s' is not empty, deletion must be forced", envName));
+
+    //Should throw for non empty environment
+    _storage.deleteEnvironment(containerName, testrigName, envName, false);
+
+    //Deletion should be possible with force=true
+    assertTrue(_storage.deleteEnvironment(containerName, testrigName, envName, true));
```
kidsbear: This line is actually not reachable, will not run codes in same test after throws exception

PR #382, opened 2017-08-29
Todd - outlier detection

millstein:
Two kinds of outlier detection:
 - find nodes whose definition of a structure named N differs from the way that many nodes that have a structure named N define it
- find nodes that differ from many other nodes in whether they contain a structure named N

In both cases, the hypothesis is that the set of nodes in the largest equivalence class are correct and the other nodes are potentially incorrect.  The results of CompareSameName are used for both of these kinds of outlier detection.  All possible outliers are produced, but they are ranked by likelihood.`


millstein:
Thanks.  Should I merge, or you'll do it at some point?

PR #394, opened 2017-08-31
Adding getStorage in Coordinator Main

haverma:
Split off from storage analysis PR,  Contains changes of  PR #368 , can be merged after that


projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public StorageUtils(Path containerLocation) {
+    _containerLocation = containerLocation;
+  }
+
+  public Path resolvePath(Path basePath, String... dirs) {
+    if (dirs.length == 0) {
+      return basePath;
+    }
+    return basePath.resolve(String.join("/", dirs));
```
dhalperi: There's an OS independent constant somewhere in Java for `/`. Something like File.Seperator or Path.Separator or something along those lines. Use that instead.

projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
```diff
@@ -183,6 +191,10 @@ private static void initWorkManager() {
 
   }
 
+  public static void initStorage() {
+    _storage = new FileStorageImpl(_settings.getContainersLocation());
```
dhalperi: None of the code in a storage consumer should have to know that it's using a FileStorageImpl.

Instead, there should be some sort public factory method `Storage.getImplementation(Settings)` that returns the implementation based on the settings.

Probably, every StorageImpl should have a function like `create(Settings)` that extracts the storage-system-specific settings.

PR #402, opened 2017-09-02
modify the environment object

kidsbear:
Modify the environment, it should be a sample for other POJOs

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
@@ -3,52 +3,53 @@ package org.batfish.datamodel.pojo;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
-import java.util.List;
-import java.util.Map;
+import java.util.Date;
 import java.util.Objects;
 import javax.annotation.Nullable;
-import org.batfish.datamodel.Edge;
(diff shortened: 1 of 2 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
```diff
@@ -3,52 +3,53 @@
...
       @JsonProperty(PROP_NAME) String name,
-      @JsonProperty(PROP_EDGE_BLACKLIST) @Nullable List<Edge> edgeBlacklist,
-      @JsonProperty(PROP_INTERFACE_BLACKLIST) @Nullable List<NodeInterfacePair> interfaceBlacklist,
-      @JsonProperty(PROP_NODE_BLACKLIST) @Nullable List<String> nodeBlacklist,
-      @JsonProperty(PROP_ENVIRONMENT_BGP_TABLES) @Nullable Map<String, String> bgpTables,
-      @JsonProperty(PROP_ENVIRONMENT_ROUTING_TABLES) @Nullable Map<String, String> routingTables,
+      @JsonProperty(PROP_CREATED_AT) Date createdAt,
+      @JsonProperty(PROP_EDGE_BLACKLIST_COUNT) int edgeBlacklistCount,
+      @JsonProperty(PROP_INTERFACE_BLACKLIST_COUNT) int interfaceBlacklistCount,
+      @JsonProperty(PROP_NODE_BLACKLIST_COUNT) int nodeBlacklistCount,
```
haverma: We should have node Blacklist in the object itself as it is just a list of strings and may be loaded with the object in the memory

PR #411, opened 2017-09-05
More cadant, various fixes

arifogel:
Not sure why merged commits are showing. Anyway, here is the rest of the CMTS stuff, and some other fixes.
I believe you only need to look at 8d7ce68 and later.


dhalperi:
I believe the old commits showed up because you didn't start a fresh branch forked from `master`.

PR #412, opened 2017-09-06
Fix some missing structure references

dhalperi:
Questions:

* When do we use CiscoStructureType.IPV4_ACCESS_LIST vs IP_ACCESS_LIST? It looks like CiscoStructureType.IP_ACCESS_LIST_EXTENDED is never used, nor is CiscoStructureType.IP_ACCESS_LIST_STANDARD.
* (aka, should these be registered as IPV4 or IP)

* What determines which types of structures we "mark references to IPv4/6 ACLs that may not appear in data model". Why do we not mark all references, regardless of whether they appear in the data model? Only mildly redundant work.


dhalperi:
Note that referencing the NAT pool and ACL here does not actually have any affect, because we handle their references completely differently: https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L1974

I can't quite understand when we use which method.

PR #415, opened 2017-09-06
add createEnvironmentRequest type

kidsbear:
Add object type for creating environment and update environment function call. Will use this object and the Environment object in PR #402 to support API calls on Environment.


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/TableObject.java
```diff
@@ -0,0 +1,48 @@
+package org.batfish.datamodel.pojo;
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+
+/**
+ * {@link TableObject TableObject} is an Object representation of an bgpTable/routingTable file.
+ * {@link #_name} stores the name of the file, and {@link #_content} contains the file content.
+ */
+public class TableObject {
```
dhalperi: What about this object is specific to a Table vs any other kind of file?

kidsbear: It should be able to represent any actual files. I changed the name to FileObject~

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/CreateEnvironmentRequest.java
```diff
@@ -0,0 +1,110 @@
...
+  private final String _name;
+  private final List<Edge> _edgeBlacklist;
+  private final List<NodeInterfacePair> _interfaceBlacklist;
+  private final List<String> _nodeBlacklist;
+  private final List<TableObject> _bgpTables;
+  private final List<TableObject> _routingTables;
+  private final String _externalBgpAnnouncements;
+
+  @JsonCreator
+  public CreateEnvironmentRequest(
```
haverma: Since we don't have any setters and no zero arguments constructor, how will we create an Environmentrequest object if we need to create it incrementally ? This will be required as in service we are reading from the filestream and incrementally populating various Environment attributes

PR #419, opened 2017-09-07
[Findbugs] possible SMT bugfix where variables were not being modified correctly

dhalperi:
@rabeckett, any thoughts on this one?

From what both Findbugs and I can tell, neither of the two functions on the RHS here have side effects.


dhalperi:
travis weird outage; kicking PR to restart.

dhalperi:
@rabeckett - any thoughts?

dhalperi:
Thanks!

PR #422, opened 2017-09-08
PluginConsumer: fix a number of bugs

dhalperi:
1. Inability to handle serialized objects bigger than 2GB
2. Leaked streams

The fix for 1 involves not reading the data into an array before
passing it to the deserializer.

The fix for 2 involves making sure we close streams, even in exception cases.
Guava's Closer makes this easier, although it still throws IOException which
makes integration with Batfish's exception policies more challenging.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
@@ -1,5 +1,7 @@
 package org.batfish.common.plugin;
 
+import com.google.common.io.ByteStreams;
+import com.google.common.io.Closer;
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.io.xml.DomDriver;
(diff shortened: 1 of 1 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
```diff
@@ -90,17 +101,31 @@ public PluginConsumer(boolean serializeToText, List<Path> pluginDirs) {
...
+              outputClass.getCanonicalName(), inputFile),
+          e);
+    }
   }
 
   protected byte[] fromGzipFile(Path inputFile) {
     try {
-      FileInputStream fis = new FileInputStream(inputFile.toFile());
-      GZIPInputStream gis = new GZIPInputStream(fis);
-      byte[] data = IOUtils.toByteArray(gis);
```
dhalperi: This file stream was leaked -- IOUtils.toByteArray does not close the input stream.

PR #427, opened 2017-09-11
More progress in making Findbugs runnable on Batfish

dhalperi:
1. Enable it by default, but manually disable in all modules.
2. A large number of small fixes.


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java
```diff
@@ -581,6 +584,12 @@ public int compareTo(Flow rhs) {
 
   @Override
   public boolean equals(Object obj) {
+    if (obj == this) {
```
ratulm: s/obj/o/g for consistency with other files?

PR #429, opened 2017-09-11
PluginConsumer: when writing to files, serialize directly

dhalperi:
Rather than serializing to RAM first.

This is a follow-on to #422, but on the write side.


dhalperi:
(Serializing a relatively large data plane went from 91 to 75 seconds with this change.)

PR #431, opened 2017-09-12
Fix Travis checking of unit tests, fix unit tests

dhalperi:
Travis wasn't running unit tests any more after #383. Fix this by adding a new command `batfish_test_all`, then fix the recently-broken unit test.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -3456,7 +3456,7 @@ public class Batfish extends PluginConsumer implements IBatfish {
       inferredRoles = true;
       nodeRolesPath = settings.getInferredNodeRolesPath();
       if (!Files.exists(nodeRolesPath)) {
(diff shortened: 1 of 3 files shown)


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -3456,7 +3456,7 @@ public NodeRoleSpecifier getNodeRoleSpecifier(boolean inferred) {
       inferredRoles = true;
       nodeRolesPath = settings.getInferredNodeRolesPath();
       if (!Files.exists(nodeRolesPath)) {
-        result = new NodeRoleSpecifier();
+        return new NodeRoleSpecifier();
```
dhalperi: Before @millstein 's recent change, this code was just a `return` statement. I believe this is the desired behavior; otherwise the assignment on this line was unused (it got overwritten at L3462 below).

dhalperi:
Going to go ahead and merge as the build is currently broken. Please take a look at your leisure.

PR #433, opened 2017-09-12
modify getContainer and delContainer methods in BfCoordWorkHelper to use WorkMgrServiceV2

kidsbear:
Also introduces a subResource class - `ContainerResource`, only add a limit amount api calls in `ContainerResource` that can be supported now. Next step: introduce a new structure to support container creation, and provide api call `createContainer` in `WorkMgrServiceV2`.


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrServiceV2.java
```diff
@@ -28,40 +27,39 @@
...
 
   /** Returns the list of {@link Container containers} that the given API key may access. */
   @GET
-  @Path("/containers")
-  public Response getContainers(
-      @DefaultValue(DEFAULT_API_KEY) @HeaderParam(SVC_KEY_API_KEY) String apiKey) {
-    List<Container> containers = Main.getWorkMgr().getContainers(apiKey);
+  @Path(CoordConsts.SVC_KEY_CONTAINERS)
+  public Response getContainers() {
+    _logger.info("WMS:getContainers\n");
```
dhalperi: WMS2

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.ws.rs.core.UriInfo;
+import org.batfish.common.BatfishLogger;
+import org.batfish.common.Container;
+import org.batfish.coordinator.Main;
+
+/**
+ * The {@link ContainerResource} is a resource for servicing client API calls at container level.
+ *
+ * <p>The ContainerResource provides information about a specified container, and provides the
+ * ability to create/delete a specified container for authenticated clients.
```
dhalperi: does not have ability to create a container -- that goes somewhere else. (in the ContainersResource)

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,72 @@
...
+    validate();
+    if (Main.getWorkMgr().delContainer(_name)) {
+      return Response.noContent().build();
+    } else {
+      return Response.serverError().build();
+    }
+  }
+
+  /** Validates the container {@link #_name} exists and {@link #_apiKey} has accessibility to it. */
+  private void validate() {
```
dhalperi: checkAccessToContainer

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,72 @@
...
+    java.nio.file.Path containerDir = Main.getSettings().getContainersLocation().resolve(_name);
```
dhalperi: is there a WorkManager function you can use for this instead?

One proposal would be to use a function like

```java
@Nullable getContainer(String name);
```

then you throw the right exception type if it's null.

But that's just a random thought, might not work well in practice, think about it.

projects/coordinator/src/main/java/org/batfish/coordinator/resources/ContainerResource.java
```diff
@@ -0,0 +1,72 @@
...
+    if (!Files.exists(containerDir)) {
+      throw new NotFoundException(String.format("Container '%s' does not exist", _name));
```
dhalperi: another thing you could do is just

```java
List<Container> listContainers(String apiKey)
```

and check if the list contains the specified container.

It's okay to not return a Forbidden exception if the user doesn't have access to even list it. Security-wise, it's okay to not leak that a container exists with that name.

projects/coordinator/src/test/java/org/batfish/coordinator/WorkMgrServiceV2Test.java
```diff
@@ -54,32 +60,48 @@ protected Application configure() {
   }
 
   @Test
-  public void getContainers() throws Exception {
-    Response response = target("/v2/containers").request().get();
+  public void getContainers() {
+    Response response = target(BASE_CONTAINER_PATH).request().get();
```
dhalperi: target(V2_CONSTANT).path(CONTAINERS)
target(V2_CONSTANT).path(CONTAINER)
target(V2_CONSTANT).path(CONTAINER).path(ID)

(something like that)

and then you don't need to use Paths to join URLs -- that's what chained calls to target are for.

dhalperi: or you could make a private static function getContainers() that returns a Response and contains this whole line (with proper targeting)

and another one getContainer(String id)
...

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -152,22 +157,24 @@ public boolean delAnalysis(String containerName, String analysisName) {
 
   public boolean delContainer(String containerName) {
     try {
-      WebTarget webTarget = getTarget(CoordConsts.SVC_RSC_DEL_CONTAINER);
+      WebTarget webTarget =
+          getTargetV2(Lists.newArrayList(CoordConsts.SVC_KEY_CONTAINERS, containerName));
```
dhalperi: this should be `container/name`, not `containers/name`.

kidsbear: I think Github is using plural form for all resource nouns. [get-single-issue](https://developer.github.com/v3/issues/#get-a-single-issue), [get-single-PR](https://developer.github.com/v3/pulls/#get-a-single-pull-request), [delete-a-comment](https://developer.github.com/v3/issues/comments/#delete-a-comment)

PR #425, opened 2017-09-11
Findbugs support

ratulm:
This branch includes changes from #353, so perhaps do that first?

PR #434, opened 2017-09-13
Exceptions in jsonpath templates

ratulm:


projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -575,6 +575,35 @@ public String getObject(String containerName, String testrigName, String objectN
...
+
+      addTextMultiPart(multiPart, CoordConsts.SVC_KEY_API_KEY, _settings.getApiKey());
+
+      JSONObject jObj = postData(webTarget, multiPart);
+      if (jObj == null) {
+        return null;
+      }
+
+      if (!jObj.has(CoordConsts.SVC_KEY_QUESTION_LIST)) {
+        _logger.errorf("question list key not found in: %s\n", jObj.toString());
```
dhalperi: drop `.toString()` -- `%s` should already handle that conversion.

(And better yet, also handles `null` values; tho in this case that's not an issue)

projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgrService.java
```diff
@@ -36,6 +37,33 @@ public JSONArray getInfo() {
   }
 
   @GET
+  @Path(CoordConsts.SVC_RSC_POOL_GET_QUESTION_TEMPLATES)
```
dhalperi: why do the work manager and pool manager have overlapping APIs?

(code looks identical or at least very close to work mgr?)

projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
```diff
@@ -339,6 +339,14 @@ public static void deleteIfExists(Path path) {
     return differenceSet;
   }
 
+  public static boolean equalsWithNullCheck(Object o1, Object o2) {
```
dhalperi: drop -- this is just the built-in Java Objects.equals(o1, o2)

tests/basic/jsonpath-assertion-equals.ref
```diff
@@ -0,0 +1,81 @@
+{
```
dhalperi: somehow I doubt this stack trace was desired?

projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgrService.java
ratulm: It is identical. I'll explain the goals/rational in person tomorrow. There might be better ways of it. 

FYI, in this comment and the previous one, changes from PR #425 are bleeding in.

ratulm:
approved in person; so forcing close.

PR #435, opened 2017-09-13
add BfClientRequestFilter and test

kidsbear:
Use a client filter to append header authentication information to every HTTP request, enable authentication filter for WorkMgrServiceV2.

### DIFF
--- a/projects/batfish-client/pom.xml
+++ b/projects/batfish-client/pom.xml
@@ -23,7 +23,8 @@
               <goals><goal>analyze-only</goal></goals>
               <configuration>
                 <ignoredUnusedDeclaredDependencies>
-                  org.slf4j:slf4j-jdk14
+                  <ignoredUnusedDeclaredDependency>org.glassfish.jersey.test-framework.providers:jersey-test-framework-provider-grizzly2</ignoredUnusedDeclaredDependency>
+                  <ignoredUnusedDeclaredDependency>org.slf4j:slf4j-jdk14</ignoredUnusedDeclaredDependency>
                 </ignoredUnusedDeclaredDependencies>
               </configuration>
             </execution>
@@ -137,6 +138,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.glassfish.jersey.core</groupId>
+      <artifactId>jersey-server</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>jline</groupId>
       <artifactId>jline</artifactId>
     </dependency>
@@ -157,6 +163,16 @@
     <dependency>
       <groupId>org.hamcrest</groupId>
       <artifactId>hamcrest-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.test-framework</groupId>
+      <artifactId>jersey-test-framework-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.test-framework.providers</groupId>
+      <artifactId>jersey-test-framework-provider-grizzly2</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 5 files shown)


projects/batfish-client/src/main/java/org/batfish/client/BfClientRequestFilter.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.ws.rs.client.ClientRequestContext;
+import javax.ws.rs.client.ClientRequestFilter;
+import javax.ws.rs.ext.Provider;
+import org.batfish.common.CoordConsts;
+import org.batfish.common.Version;
+
+/** This filter adds default authentication information (apikey and version) to HTTP headers for
+ * every HTTP request. */
+@Provider
+public class BfClientRequestFilter implements ClientRequestFilter {
```
dhalperi: Why doesn't this use the client's API key?

This seems somehow like the wrong approach to me.

projects/batfish-client/pom.xml
```diff
@@ -137,6 +138,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.glassfish.jersey.core</groupId>
+      <artifactId>jersey-server</artifactId>
```
dhalperi: Why is this not a test scope dependency? Client ought not to depend on service in compile scope, we hope.

projects/batfish-client/src/main/java/org/batfish/client/BfClientRequestFilter.java
dhalperi: Pick a class name that reflects what this filter does.

projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
```diff
@@ -47,7 +47,7 @@ public BfCoordWorkHelper(String workMgr, BatfishLogger logger, Settings settings
     _logger = logger;
     _settings = settings;
     try {
-      _client = getClientBuilder().build();
+      _client = getClientBuilder().build().register(BfClientRequestFilter.class);
```
dhalperi: is this really the right place to put this? Why not in the getClientBuilder() command, for example?

projects/batfish-client/src/test/java/org/batfish/client/BfClientRequestFilterTest.java
```diff
@@ -0,0 +1,55 @@
...
+  protected Application configure() {
+    forceSet(TestProperties.CONTAINER_PORT, "0");
+    return new ResourceConfig(TestService.class);
+  }
+
+  @Test
+  public void testFilter() {
+    Client client = ClientBuilder.newClient();
+    Response response = client.target(getBaseUri()).path("/test").request().get();
+    Assert.assertEquals(500, response.getStatus());
```
dhalperi: separate unit tests for failure and success cases.

projects/batfish-client/src/test/java/org/batfish/client/BfClientRequestFilterTest.java
```diff
@@ -0,0 +1,55 @@
...
+    client.register(BfClientRequestFilter.class);
+    response = client.target(getBaseUri()).path("/test").request().get();
+    Assert.assertEquals(200, response.getStatus());
```
dhalperi: static import, so you can just write `assertEquals`.

projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
```diff
@@ -171,16 +171,51 @@ private static void startWorkManagerService(
     }
   }
 
+  private static void startWorkManagerServiceV2(
```
dhalperi: Cloning 35 lines of code is likely not the right answer.

The existing function was configured to be usable by both v1 and v2 (that's why it takes serviceClass and jsonFeature as arguments). Keep that pattern going if possible.

PR #436, opened 2017-09-13
register filters to WorkMgrServiceV2

kidsbear:
Register two custom filters to WorkMgrServiceV2.

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
@@ -6,7 +6,6 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.collect.Lists;
 import java.net.URI;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
```diff
@@ -176,11 +180,16 @@ private static void initWorkManager() {
...
     // Initialize and start the work manager service using the legacy API and Jettison.
     startWorkManagerService(
-        WorkMgrService.class, JettisonFeature.class, _settings.getServiceWorkPort());
+        WorkMgrService.class,
+        Lists.newArrayList(JettisonFeature.class),
+        _settings.getServiceWorkPort());
     // Initialize and start the work manager service using the v2 RESTful API and Jackson.
-    startWorkManagerService(
-        WorkMgrServiceV2.class, JacksonFeature.class, _settings.getServiceWorkV2Port());
+    List<Class<?>> features = new ArrayList<>();
```
dhalperi: why not use Lists.newArrayList here too, to simplify the construction?

PR #439, opened 2017-09-13
update Container Object

kidsbear:
Update the Container Object to follow the [Object/Storage APIs design guidelines](https://docs.google.com/document/d/1cUvWEpjURkhtmzdS1gWESXUgEUxdSUYOJMsgNgQa_Rk/edit?ts=59a49860).

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -1347,7 +1347,7 @@ public class Client extends AbstractClient implements IClient {
     String containerName = parameters.get(0);
     Container container = _workHelper.getContainer(containerName);
     if (container != null) {
-      _logger.output(container.getTestrigs() + "\n");
+      _logger.output(container + "\n");
       return true;
     }
     return false;
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
@@ -3,29 +3,35 @@ package org.batfish.common;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
-import java.nio.file.Path;
 import java.util.Objects;
-import java.util.SortedSet;
 
+/**
+ * The {@link Container Container} is an object representation of the container for BatFish service.
+ *
+ * <p>Each {@link Container Container} contains a name and a summary of the container {@link
+ * #_name}.
+ */
 public final class Container {
   private static final String PROP_NAME = "name";
-  private static final String PROP_TESTRIGS = "testrigs";
-  private static final String PROP_ANALYSIS = "analysis";
+  private static final String PROP_CREATED_AT = "createdAt";
+  private static final String PROP_TESTRIGS_COUNT = "testrigsCount";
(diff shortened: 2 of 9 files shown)


projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -1347,7 +1347,7 @@ private boolean getContainer(List<String> options, List<String> parameters) {
     String containerName = parameters.get(0);
     Container container = _workHelper.getContainer(containerName);
     if (container != null) {
-      _logger.output(container.getTestrigs() + "\n");
+      _logger.output(container + "\n");
```
kidsbear: Should I modify the code here to get the same output? In my point of view, getContainer should return information about a container, rather than the testrig names in the container (which sounds like a listTestrig call). 
The current output look like:
```
Client output:
Container{name=containerName, createdAt=2017-09-12 14:58:53, testrigsCount=0, analysesCount=0}

use browser:
{
"name": "containerName",
"createdAt": "2017-09-12 14:58:53",
"testrigsCount": 0,
"analysesCount": 0
}
```

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -3,66 +3,64 @@
...
-  private static final String PROP_TESTRIGS = "testrigs";
-  private static final String PROP_ANALYSIS = "analysis";
+  private static final String PROP_CREATED_AT = "createdAt";
+  private static final String PROP_TESTRIGS_COUNT = "testrigsCount";
+  private static final String PROP_ANALYSES_COUNT = "analysesCount";
 
   private String _name;
-  private SortedSet<String> _testrigs;
-  private Path _analysis;
+  private String _createdAt;
```
haverma: createdAt should be a date

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -3,66 +3,64 @@
...
-    return new Container(name, testrigs);
-  }
-
-  private Container(String name, SortedSet<String> testrigs) {
+      @JsonProperty(PROP_CREATED_AT) String createdAt,
+      @JsonProperty(PROP_TESTRIGS_COUNT) int testrigsCount,
+      @JsonProperty(PROP_ANALYSES_COUNT) int analysesCount) {
     this._name = name;
-    this._testrigs = testrigs;
+    this._createdAt = createdAt;
```
haverma: this keyword is unnecessary here

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
```diff
@@ -72,11 +70,14 @@ public boolean equals(Object o) {
       return false;
     }
     Container other = (Container) o;
-    return Objects.equals(_name, other._name) && Objects.equals(_testrigs, other._testrigs);
+    return Objects.equals(_name, other._name)
+        && Objects.equals(_createdAt, other._createdAt)
```
haverma: two equal containers can be created at different times(also I think just comparing counts of testrigs and analyses is not very useful, we would better call APIs(if they exist) to get lists and compare them).

projects/batfish-common-protocol/src/main/java/org/batfish/common/Container.java
kidsbear: Jakson serializes Date as a long number by default, so I store createdAt as a string.

PR #443, opened 2017-09-15
enhancements for testrig sync APIs

ratulm:


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -605,6 +605,21 @@ public BatfishLogger getLogger() {
     return _logger;
   }
 
+  @Override
+  public Set<String> getContainerNames() {
+    Path containersDir = Main.getSettings().getContainersLocation();
+    if (!Files.exists(containersDir)) {
+      containersDir.toFile().mkdirs();
+    }
+    SortedSet<String> authorizedContainers =
```
dhalperi: what is authorized about these containers?

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -605,6 +605,21 @@ public BatfishLogger getLogger() {
     return _logger;
   }
 
+  @Override
+  public Set<String> getContainerNames() {
```
dhalperi: is listContainers not good enough here?

ratulm: listContainers() needs an API key, and only returns containers that are accessible for that key.

PR #449, opened 2017-09-15
Batfish: use a cache of weak references for the data plane

dhalperi:
This will allow the garbage collector to clear the cache if memory pressure is on.

I tried soft references instead, which in principle should be collected less aggressively, but unfortunately they were not actually garbage collected in practice. This may be OS-related [swap on?] or JRE-related [garbage collector algorithm?], but tl;dr: I was not able to make soft references work.


dhalperi:
Done

PR #461, opened 2017-09-21
SMT improvements

rabeckett:
Adds a couple of improvements to the SMT code:

* The current reachability encoding detects loops as being reachable (e.g., in the example test_rig). I changed the reachability encoding to associate an integer id variable with each router. A router will have a higher id than the neighbors to which it forwards, which prevents the solver from considering loops as reachable. Surprisingly, this encoding seems to make z3 faster as well.

* Provides more useful feedback for reachability questions, including the protocol used to route the packet at each hop as well as the collection of source routers that are (un)reachable.


projects/batfish/src/main/java/org/batfish/smt/PropertyChecker.java
```diff
@@ -197,10 +205,45 @@ public static AnswerElement computeReachability(IBatfish batfish, HeaderLocation
...
+    if (!res.isVerified()) {
+      if (q.getEquivalence()) {
+        assert (reach2 != null);
+        for (String source : sourceRouters) {
+          BoolExpr sourceVar1 = reach.get(source);
+          BoolExpr sourceVar2 = reach2.get(source);
+          String val1 = evaluate(model, sourceVar1);
+          String val2 = evaluate(model, sourceVar2);
+          boolean notEqual =
+              ((val1 == null && val2 != null) || (val1 != null && !val1.equals(val2)));
```
dhalperi: there's a built in library function for this: `!Objects.equals(val1, val2)`

rabeckett: That's convenient, thanks.

rabeckett:
The batfish.common.Pair class requires that that the two types are comparable. The Model datatype z3 provides does not implement comparable.

ratulm:
Got it. 

By the way, Objects.hashCode(....) might help simplify your hashCode function if you'd like.

rabeckett:
Right now I'm using the default intellij "cmd-n generate hashcode function" command. Is Objects.hashcode considered better?

ratulm:
It does similar things I believe. Just less code.

PR #462, opened 2017-09-22
unify hosts/iptables for overlay and non-overlay

arifogel:
Includes PR #338


dhalperi:
High-level: could use some background discussion on what's going on here. I read #221 and #338 but didn't grok fully.

PR #463, opened 2017-09-22
Turn on Jackson case-insensitive enums, remove a bunch of useless code

dhalperi:
There's more to do, but this is a small sanity check for review.


ratulm:
Merged but left the branch intact (in case you want to keep adding to this).

dhalperi:
Shouldn't add to branches -- every time you're going to start new changes after being merged, start from master (aka `git fetch origin master ; git checkout origin/master -b feature-part-2` for example)

PR #468, opened 2017-09-25
Fix crash when a VRRP is configured but unusable

dhalperi:
If there's no source IP that could be used in VRRP election.


dhalperi:
Ari approved manually in person; merging.

PR #478, opened 2017-09-27
Fix XStream serialization

dspicuzzbbn:
- Flush isn't sufficient, need to close the output stream

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
@@ -331,7 +331,7 @@ public abstract class PluginConsumer implements IPluginConsumer {
         oos = new ObjectOutputStream(gos);
       }
       oos.writeObject(object);
-      oos.flush();
+      oos.close();
       gos.finish(); // close the GZIP file entry
     } catch (IOException e) {
       throw new BatfishException("Failed to convert object to gzip data", e);


dhalperi:
In general, if someone passes you a stream they've opened, you can't close it. So this can't be the right fix. But I'll look into it.

(E.g, what if I want to serialize 3 objects back to back?)

dspicuzzbbn:
The stream is created in that same function, how is it incorrect to close it in that function?

dspicuzzbbn:
Using this function as currently implemented, I believe xstream serialization will produce invalid XML if you try to serialize 3 objects back to back.

dhalperi:
> I believe xstream serialization will produce invalid XML 

Agree that Xstream serialization is currently broken.

> The stream is created in that same function, how is it incorrect to close it in that function?

Closing the `oos` closes the stream passed into the function:

```java
package org.batfish.common.plugin;

import com.thoughtworks.xstream.XStream;
import com.thoughtworks.xstream.io.xml.DomDriver;
import java.io.FileOutputStream;
import java.io.ObjectOutputStream;
import org.junit.Rule;
import org.junit.Test;
import org.junit.rules.TemporaryFolder;
import org.junit.runner.RunWith;
import org.junit.runners.JUnit4;

@RunWith(JUnit4.class)
public class PluginConsumerTest {
  @Rule public TemporaryFolder _folder = new TemporaryFolder();

  @Test
  public void test() throws Exception {
    FileOutputStream out = new FileOutputStream(_folder.newFile("foo"));
    out.write(5);

    XStream xstream = new XStream(new DomDriver("UTF-8"));
    ObjectOutputStream oos = xstream.createObjectOutputStream(out);
    oos.close();

    out.write(4);  // this throws an exception. Closing oos closes out.
  }
}
```

dspicuzzbbn:
Oh interesting, I wouldn't have expected it to close the other stream.

dhalperi:
Check out #479, which now has tests (that should have been there all along :/)

dspicuzzbbn:
Closing in favor of #479

PR #466, opened 2017-09-22
Misc updates for cisco router configurations

dspicuzzbbn:
Found on a cisco ASA

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/466)

PR #481, opened 2017-09-28
Update enable password handling

arifogel:
Includes part of #466


dhalperi:
Approved with a question.

PR #483, opened 2017-09-28
added support for file lines

ratulm:


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/DisplayHints.java
```diff
@@ -303,29 +303,33 @@ public DisplayHints(
...
-    // extraVarsInExtractions);
-    //    }
-
-    Set<String> entitiesInTextDesc = new HashSet<>();
+
+    // the names of entities and extraction vars should have no overlap
+    Set<String> commonNames = Sets.intersection(varsInExtractionHints, entities.keySet());
+    if (!commonNames.isEmpty()) {
+      throw new BatfishException(
+          "entities and extraction vars should have common names: " + commonNames);
```
dhalperi: is the text here right? seems to disagree with the comment.

dhalperi:
approved with question

PR #484, opened 2017-09-28
Number of updates for the SMT encoding

rabeckett:
A number of misc. updates to the SMT code, including:
* Encoding minimization
    * Avoid modeling OSPF variables on edges that cross AS boundaries
    * Infer default admin distance when never overridden by the user
    * Only model different network slices (e.g., IGP for iBGP) for nodes within the same AS
    * These take the all flow differential query for test_rigs/example down from about 40 sec to 14 sec on my machine.
* A collection of bug fixes
    * Fixes a bug modeling Route Reflectors where originator id was propagated across AS boundaries
    * Fixes semantics of connected routes to only forward when ARP determines there is a host on the other end.
    * Fixes a bug with the interaction of SetMetric and BGP

* Improves the output of reachability queries by generating a FlowHistory in the same way as Batfish. Gives the line of the ACL causing a drop etc. Updated refs and there is a new test case: acl1-reachability that demonstrates this.


rabeckett:
The admin distance logic is something that Batfish provides, and the encoding is already using that logic. What I mean is that we don't need to actually model the adminDist integer variable. For example, on import from a neighbor we just leave the variable off. Then when computing the best route, we add a constraint that fills it in with the default for that protocol when we know the user never changed the value from the default.

For OSPF what I mean is mainly about export variables. Even if an interface is turned on for OSPF, if there is no device on the other end running OSPF, then we don't need to know things like what the path length would be to this other device, if it is an inter-area or external type route etc.

Good point about different ASes. I just pushed a fix to change it to whether the interface is enabled. This was previously done with a collection of different constraints that would set the valid bit to false when a link was not enabled for OSPF, but now it just doesn't model the variables at all. Seems to also take the z3 solving time down to around 9sec now from the original 40.

ratulm:
40 --> 9 is a massive reduction. I wish z3 got smarter about eliminating redundant information. Merging and deleting branch.

PR #486, opened 2017-10-01
Adding endpoint in Batfish Service to get all questions as configured

haverma:


dhalperi:
Suggest @ratulm would be a good person to take a look and see if this fits his way of thinking.

haverma:
Closing for now, need to make it a POST to make it consistent with the other APIs

Issue #485, opened 2017-09-29
Bug with the iptables overlay and transit 


jkhourybbn:
__Related to #462__

Hi Ari/Ratul,

I am attaching a very simple test rig which has 2 zones
```
z1-host1 <--> z1-firewall <--> r1 <--> r2  <--> z2-firewall <--> z2-host1
```
The following query and output is anomalous. Even though I am fixing the dstIp in the query (verified it against the NoD program), z3 returns an answer that effectively ignores the dstIp constraints. This seems to happen only with transit.
```
batfish> get reachability notIngressNodeRegex="z1-.*", dstIps=["10.0.3.20"], notTransitNodes=["z1-firewall"]
Status: SUCCESS
Question: reachability actions=[ACCEPT] | dstIps=[10.0.3.20] | notIngressNodeRegex=z1-.* | notTransitNodes=[z1-firewall]

Flow: ingress:z2-firewall vrf:default 0.0.0.0->10.0.1.10 ICMP icmpType:255 icmpCode:255 packetLength:0 state:NEW
  environment:BASE
    ACCEPTED

Flow: ingress:z2-host1 vrf:default 10.0.4.0->10.0.4.1 UDP sport:0 dport:0 packetLength:0 state:NEW
  environment:BASE
    Hop 1: z2-host1:eth0 -> z2-firewall:Ethernet1 --- [ConnectedRoute<10.0.4.0/24,nhip:AUTO/NONE(-1l),nhint:eth0>_fnhip:null]
    ACCEPTED
```

I spent a couple hours chasing the potential problem in z3 but I am slow at z3 still. Can you take a look? We can add this unit test once we figure out the problem.

jkhourybbn:
Here is an even simpler version of the unit test where the problem persists (for easier debugging)
```
z1-host1 <--> z1-firewall <-->  z2-firewall <--> z2-host1
```
[testrigverysimple.tar.gz](https://github.com/batfish/batfish/files/1345851/testrigverysimple.tar.gz)

jkhourybbn:
It looks like the iptables rules are not being converted correctly, I will open a separate ticket for it

jkhourybbn:
resuming the discussion on #464

arifogel:
Since #464 is closed, is this bug fixed too? Can we close?

jkhourybbn:
Ari, lets keep this open and use it to discuss the NoD anomalies I am seeing. Ill see if I can recreate those now that iptables is fixed.

[...]

PR #488, opened 2017-10-02
iptables fixes

dspicuzzbbn:
This PR fixes two things:

* Allows not specifying the `:FORWARD ACCEPT [0:0]` style of defaults (iptables-save doesn't require it, so neither should batfish)
* Applies iptables rules without explicit interfaces specified to all interfaces, and adds a test for that


dspicuzzbbn:
This partially resolves #485, but I believe that there are further issues that need to be investigated by someone else.

dspicuzzbbn:
We've added another commit with a new iptables testrig that does more comprehensive tests.

tests/basic/example-iptables-protection.ref
```diff
@@ -0,0 +1,36 @@
+{
+  "answerElements" : [
+    {
+      "class" : "org.batfish.datamodel.FlowHistory"
+    }
```
arifogel: Is this supposed to be empty? What about the other reachability answers?

virtuald: @jkhourybbn?

jkhourybbn: Yes all three questions are supposed to be empty. A non empty result is a violation of the policy. The questions are crafted so they return no results.

PR #507, opened 2017-10-05
Example testrig: check in approximate node locations in the diagram

dhalperi:


dhalperi:
I don't see a strong reason not to save it in this file -- you can use the putObject API (as a dual of getObject).

PR #510, opened 2017-10-06
Truncate parser error output with really large rules

dspicuzzbbn:
- Should only be useful for development

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/510)


dspicuzzbbn:
.. and if travis fails, I didn't bother running the tests yet, but I found this useful during parser rule development.

arifogel:
This cannot be the default behavior, because it would hide changes during parsing testing where rules are large. If you want to make this an option in `org.batfish.config.Settings` that is disabled by default (something like `maxparsererrorlines` where `0` is unlimited), then we could integrate this.

dhalperi:
I think this is a great idea, seems really useful. Given that Dustin did the initial work and has found it useful in his own work, I think we can take on the work of adding the flag.

dhalperi:
Hey @sfraint , Can you take a look at this, too? I think you will find it educational and hopefully simple. Let's chat at your leisure.

dhalperi:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/510)*

arifogel:
Closed via #675

PR #511, opened 2017-10-06
Implement Cisco bgp neighbor local-as and add as-check to BgpSessionCheck

arifogel:


projects/batfish/src/test/resources/org/batfish/grammar/cisco/testrigs/bgp-local-as/configs/r1
```diff
@@ -0,0 +1,14 @@
...
+!
+interface Ethernet1
+ ip address 1.2.0.1/24
+!
+interface Loopback0
+ ip address 1.1.1.1/32
+!
+router bgp 1
+ router-id 1.1.1.1
+ neighbor 1.2.0.2 remote-as 2
```
dhalperi: Would be nice to add a comment here explaining that the local as number is `1` because Cisco family devices default to the process id.

projects/batfish/src/main/java/org/batfish/representation/cisco/BgpPeerGroup.java
```diff
@@ -376,6 +382,9 @@ private void inheritUnsetFields(BgpPeerGroup pg) {
       _inboundRoute6Map = pg.getInboundRoute6Map();
       _inboundRoute6MapLine = pg.getInboundRoute6MapLine();
     }
+    if (_localAs == null) {
+      _localAs = pg.getMaximumPaths();
```
dhalperi: this looks like a bug – copy and paste error?

projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
```diff
@@ -2022,7 +2022,8 @@ private AsPathAccessListLine toAsPathAccessListLine(AsPathSetElem elem) {
           _w.redFlag("No remote-as set for peer: " + lpg.getName());
           continue;
         }
-
+        Integer pgLocalAs = lpg.getLocalAs();
+        int localAs = pgLocalAs != null ? pgLocalAs : proc.getName();
```
dhalperi: simpler: `int localAs = firstNonNull(lpg.getLocalAs(), proc.getName());`

dhalperi: (import static com.google.common...?...MoreObjects.firstNonNull from guava)

PR #515, opened 2017-10-06
Implement cisco-family bgp remove-private-as 

arifogel:


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AsPath.java
```diff
@@ -15,6 +15,10 @@
 
   private static final long serialVersionUID = 1L;
 
+  public static boolean isPrivateAs(int as) {
+    return (as >= 64512 && as <= 65535);
```
dhalperi: Please add a comment with a link to the RFC. My javadoc proposal:

```java
/**
 * Returns true if the provided AS number is reserved for private use by RFC 6696:
 * https://tools.ietf.org/html/rfc6996#section-5
 */
```

My reasoning: makes it self-evident why this goes in data model (is vendor independent) and is educational for new people.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/statement/Statements.java
```diff
@@ -73,6 +79,41 @@ public Result execute(Environment environment) {
...
+            if (environment.getWriteToIntermediateBgpAttributes()) {
+              BgpRoute.Builder ir = environment.getIntermediateBgpAttributes();
+              List<SortedSet<Integer>> iAsPath =
+                  ir.getAsPath()
+                      .stream()
+                      .map(
+                          asSet ->
+                              asSet
+                                  .stream()
+                                  .filter(AsPath::isPrivateAs)
```
dhalperi: is this logic supposed to be negated from L93?

assuming not, and the same behavior is intended to be applied to both lists, extract this to a private static helper function and then use it twice.

projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java
```diff
@@ -83,4 +90,62 @@ public void testBgpLocalAs() throws IOException {
...
+    AtomicBoolean r2HasPrivate = new AtomicBoolean();
+    r2.getReceivedEbgpAdvertisements()
+        .stream()
+        .filter(a -> a.getNetwork().equals(r1Loopback))
+        .toArray(BgpAdvertisement[]::new)[0]
+        .getAsPath()
+        .getAsSets()
+        .stream()
+        .flatMap(asSet -> asSet.stream())
+        .forEach(
```
dhalperi: can this and below be simplified using something like [Stream.anyMatch](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-)?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/statement/Statements.java
```diff
@@ -73,6 +79,41 @@ public Result execute(Environment environment) {
...
+                      .stream()
+                      .map(
+                          asSet ->
+                              asSet
+                                  .stream()
+                                  .filter(AsPath::isNotPrivateAs)
+                                  .collect(
+                                      ImmutableSortedSet.toImmutableSortedSet(Integer::compare)))
+                      .filter(asSet -> !asSet.isEmpty())
+                      .collect(ImmutableList.toImmutableList());
```
dhalperi: I would extract everything from L101-111 in a static helper, so we can call the same function twice.

PR #519, opened 2017-10-07
cleanup cisco grammar wrt remove-private-as

arifogel:


projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4
```diff
@@ -823,7 +821,10 @@ router_bgp_stanza_tail
    | template_peer_policy_rb_stanza
    | template_peer_session_rb_stanza
    | vrf_block_rb_stanza
-   | { !_disableUnrecognized }? unrecognized_line
+   |
+   { !_disableUnrecognized }?
+
+   unrecognized_line
```
dhalperi: this seems like a bad change.

arifogel: The autoformatter does this. Unless and until an improved antlr4-ide autoformatter implementation is released, I prefer to stick with its results. Feel free to submit a PR to antlr4ide/antlr4ide if you this formatting bothers you just enough more than it bothers me ;)

PR #523, opened 2017-10-08
BDP Debugging overhaul

arifogel:
Includes #516

  - Move debugging constants to `Settings` so recompile is no longer necessary for debugging
  - Allow printing of detailed `AbstractRoute` instead of summary `Route` when dumping iterations (From #516)
  - Fix error in #516 causing incorrect results when printing detailed `AbstractRoute`


arifogel:
@dhalperi Your suggestions apply to every configuration setting. If we are going to do this, it should be done consistently. Punting for now.

Issue #512, opened 2017-10-06
report a smt-reachability bug with only static route


YuhuWang2002:
I create sample config as below:
[node2.txt](https://github.com/batfish/batfish/files/1361514/node2.txt)
[node1.txt](https://github.com/batfish/batfish/files/1361512/node1.txt)
[node3.txt](https://github.com/batfish/batfish/files/1361516/node3.txt)
[host3.txt](https://github.com/batfish/batfish/files/1361517/host3.txt)
[host1.txt](https://github.com/batfish/batfish/files/1361518/host1.txt)
Topology:
![topology](https://user-images.githubusercontent.com/21324650/31263008-d3f1403e-aa91-11e7-9136-3e21280f88b5.png)

I met some bugs:
1 : private static Ip ipVal(Model m, Expr e) {
   return new Ip(intVal(m,e));
   }
overfolw,
I change to: return new Ip(Long.parseLong(evaluate(m, e)));
2: static route lost by create Graph, I think it is a bug:
in graph.java line 225
`             if (isNextHop) {

                someIface = true;

                List<StaticRoute> srs = map.getOrDefault(hereName, new ArrayList<>());

                srs.add(sr);

                map.put(there.getName(), srs);

              }`
  I think it is      `map.put(hereName, srs);`
if it is not a bug, I cann't understand why put "there.getName"

base on fix these bugs, it is still have some strange phenomenon : 
My intention is to check "is the host1 can reach to host3?"
1, I run the command as below, it is work
get smt-reachability ingressNodeRegex="host1",finalNodeRegex="host3",dstIps=["192.168.8.0/24"]
Status: SUCCESS
Question: reachability actions=[ACCEPT] | dstIps=[192.168.8.0/24]

Flow: ingress:host1 vrf:default 0.0.0.0->192.168.8.0 HOPOPT packetLength:0 state:NEW
  environment:BASE
    Hop 1: host1:eth0 -> R1:Ethernet0/0 --- [StaticRoute<192.168.8.0/0,nhip:192.168.1.2/24,nhint:dynamic>]
    Hop 2: R1:Ethernet0/1 -> R2:Ethernet0/0 --- [StaticRoute<192.168.8.0/24,nhip:10.10.10.1/24,nhint:dynamic>]
    Hop 3: R2:Ethernet0/1 -> R3:Ethernet0/0 --- [StaticRoute<192.168.8.0/24,nhip:10.10.11.1/24,nhint:dynamic>]
[...]

PR #526, opened 2017-10-08
SMT reachability bug

rabeckett:
Fixes an SMT reachability bug that occurs when the final node is a host, related to #512 

It also changes the default destination IP inference behavior for hosts to avoid spurious counterexamples when the user does not specify a destination IP.

Added the example network from #512 as a test case.


ratulm:
Fixes #512. 

(I believe such a comment will automatically close the issue when we merge.)

dhalperi:
Re: auto-closing, my belief is that it has to be in the:

* git commit message,
* PR title
* PR description (aka 0th comment by the user).

https://help.github.com/articles/closing-issues-using-keywords/

(But I'm not certain that's true.)

PR #528, opened 2017-10-09
Reject iBGP advertisements containing own AS in path

arifogel:
Includes #523, #524


dhalperi:
I looked at this commit-by-commit in order to understand the "new" changes in this PR. I believe it's just the last 3 commits , in which case it looks good.

Approved after other PRs are in.

PR #539, opened 2017-10-11
enable initinfo for bgp table parsing

arifogel:


projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2120,78 +2120,108 @@ public void initBgpOriginationSpaceExplicit(Map<String, Configuration> configura
...
+    if (!summary) {
+      if (verboseError) {
+        SortedMap<String, Set<BatfishStackTrace>> errors = answerElement.getErrors();
+        parseAnswer
+            .getErrors()
+            .forEach(
+                (hostname, parseErrors) -> {
+                  errors.computeIfAbsent(hostname, k -> new HashSet<>()).add(parseErrors);
+                });
+        parseAnswer
```
dhalperi: Seems like this pattern of copying errors from an answerElement to a map / warnings from an answerElement to a map has opportunity for some code reuse.

arifogel:
I refactored in 2678275. This good?

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2769,6 +2715,41 @@ private ValidateEnvironmentAnswerElement loadValidateEnvironmentAnswerElement(
...
+                });
+      }
+      SortedMap<String, org.batfish.common.Warnings> warnings = answerElement.getWarnings();
+      convertAnswer
+          .getWarnings()
+          .forEach(
+              (hostname, convertWarnings) -> {
+                org.batfish.common.Warnings combined = warnings.get(hostname);
+                if (combined == null) {
+                  warnings.put(hostname, convertWarnings);
```
dhalperi: do you want this `put` to clone `convertWarnings`? seems like you might mutate warnings from one source if merging in another source later.

projects/batfish/src/main/java/org/batfish/main/Batfish.java
```diff
@@ -2121,56 +2122,16 @@ public void initBgpOriginationSpaceExplicit(Map<String, Configuration> configura
...
 
-  public InitInfoAnswerElement initInfoBgpAdvertisements(boolean summary, boolean verboseError) {
+  private InitInfoAnswerElement initInfo(
+      boolean summary, boolean verboseError, ParseAnswerElement parseAnswer) {
     InitInfoAnswerElement answerElement = new InitInfoAnswerElement();
-    ParseEnvironmentBgpTablesAnswerElement parseAnswer =
-        loadParseEnvironmentBgpTablesAnswerElement();
     if (!summary) {
       if (verboseError) {
         SortedMap<String, Set<BatfishStackTrace>> errors = answerElement.getErrors();
```
dhalperi: I was thinking it may be useful to have a private static helper for merging errors alone: a pattern that occurs 3x in this file (and maybe elsewhere)?

(Lines 2138,2144, and and 2723).

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseAnswerElement.java
```diff
@@ -0,0 +1,25 @@
...
+
+  SortedMap<String, Warnings> getWarnings();
+
+  void setErrors(SortedMap<String, BatfishException.BatfishStackTrace> errors);
+
+  void setParseStatus(SortedMap<String, ParseStatus> parseStatus);
+
+  void setParseTrees(SortedMap<String, ParseTreeSentences> parseTrees);
+
+  void setWarnings(SortedMap<String, Warnings> warnings);
```
dhalperi: dumb q: should all answer elements have optional errors and warnings, not just parsing-related?

arifogel: That would require a huge structural change. It might not be a bad idea, though. Definitely out of scope for this PR IMO.

arifogel:
@dhalperi Please see e63de5b which addresses your latest comments.

PR #540, opened 2017-10-11
Fix missing nodes in bdp debug output

arifogel:


projects/batfish/src/main/java/org/batfish/bdp/BdpDataPlanePlugin.java
```diff
@@ -934,6 +934,23 @@ private RouteSet computeOutputRoutes(Map<String, Node> nodes, Map<Ip, String> ip
...
                 routes.addAll(vr._mainRib.getRoutes());
+                for (AbstractRoute route : routes) {
+                  route.setNode(hostname);
+                  route.setVrf(vrName);
+                  Ip nextHopIp = route.getNextHopIp();
+                  if (route.getProtocol() == RoutingProtocol.CONNECTED
+                      || (route.getProtocol() == RoutingProtocol.STATIC
+                          && nextHopIp.equals(Route.UNSET_ROUTE_NEXT_HOP_IP))
+                      || Interface.NULL_INTERFACE_NAME.equals(route.getNextHopInterface())) {
+                    route.setNextHop(Configuration.NODE_NONE_NAME);
```
dhalperi: all this is just for debugging info about the routes?

arifogel: Yes. The debug output incorrectly had null values, because the appropriate fields were not populated in all cases.

PR #542, opened 2017-10-11
refactoring to support automatic processing of testrigs

ratulm:


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -1145,44 +1188,18 @@ public void uploadTestrig(String containerName, String testrigName, InputStream
...
-      if (isEnvFile(subFile)) {
-        target = defaultEnvironmentLeafDir.resolve(subFile.getFileName());
-      } else {
-        target = unzipDir.resolve(subFile.getFileName());
-      }
-      CommonUtil.moveByCopy(subFile, target);
+    try {
+      initTestrig(testrigDir, unzipDir, false);
+    } catch (Exception e) {
+      CommonUtil.deleteDirectory(unzipDir);
```
dhalperi: Slightly simpler:

```java
try {
   init Testrig
} catch (Exception e) {
   throw BE(..., e);
} finally {
   CommonUtil.delete
   CommonUtil.delete
}
```

This way the deletion happens in both success and failure cases naturally. We also get slightly better handling of exceptions in the delete functions themselves in the failure case.

(then you don't need the ones at L1202/03)

ratulm: ah, i had forgotten that the finally will executed even if the catch block throws an exception.

PR #545, opened 2017-10-11
Removing global questions and testing getQuestionTemplates

haverma:
getQuestionTemplates was also not checking apikey

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
@@ -807,38 +807,6 @@ public class WorkMgrService {
     }
   }
 
-  /**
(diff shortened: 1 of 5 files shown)


projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
```diff
@@ -848,6 +816,8 @@ public JSONArray getQuestionTemplates(@FormDataParam(CoordConsts.SVC_KEY_API_KEY
 
       checkStringParam(apiKey, "API key");
 
+      checkApiKeyValidity(apiKey);
```
dhalperi: One PR should have one purpose only. Please fix the existing API key issue in a separate PR.

haverma: done

PR #549, opened 2017-10-11
Fix IOS routing table static route attributes

### LINKED ISSUE #535: Batfish sets admin distance as 1 for Null0 route instead of 0 as set by Cisco IOS
ENV  lhr-leaf-01 vrf:default net:10.1.1.0/24 nhip:AUTO/NONE(-1l) nhint:Null0 nhnode:N/A admin:0 cost:0 tag:none prot:static
BAT  lhr-leaf-01 vrf:default net:10.1.1.0/24 nhip:AUTO/NONE(-1l) nhint:null_interface nhnode:N/A admin:1 cost:0 tag:none prot:static

arifogel:
- Fix #535 
- Generalize and clean up testrig junit tests

### DIFF
--- a/projects/batfish/src/test/java/org/batfish/bdp/BdpDataPlanePluginTest.java
+++ b/projects/batfish/src/test/java/org/batfish/bdp/BdpDataPlanePluginTest.java
@@ -149,9 +149,19 @@ public class BdpDataPlanePluginTest {
   public void testIbgpRejectOwnAs() throws IOException {
     String testrigName = "ibgp-reject-own-as";
     String[] configurationNames = new String[] {"r1", "r2a", "r2b"};
+    String[] bgpTableNames = new String[] {};
+    String[] hostFilenames = new String[] {};
+    String[] iptablesFilenames = new String[] {};
+    String[] routingTableNames = new String[] {};
     Batfish batfish =
         BatfishTestUtils.getBatfishFromTestrigResource(
-            TESTRIGS_PREFIX + testrigName, configurationNames, _folder);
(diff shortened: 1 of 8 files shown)


projects/batfish/src/test/java/org/batfish/main/BatfishTestUtils.java
```diff
@@ -209,4 +257,15 @@ public static Batfish getBatfish(
     }
     return initBatfish(configurations, tempFolder);
   }
+
+  private static SortedMap<String, String> readTestrigResources(
+      String testrigResourcePrefix, String subfolder, String[] filenames) {
```
dhalperi: suggest making filenames a `String...` so that you can use the function more manually in the future.

projects/batfish/src/test/java/org/batfish/bdp/BdpDataPlanePluginTest.java
```diff
@@ -149,9 +149,19 @@ public void testApplySourceNatInvalidAclThrows() {
   public void testIbgpRejectOwnAs() throws IOException {
     String testrigName = "ibgp-reject-own-as";
     String[] configurationNames = new String[] {"r1", "r2a", "r2b"};
+    String[] bgpTableNames = new String[] {};
+    String[] hostFilenames = new String[] {};
+    String[] iptablesFilenames = new String[] {};
+    String[] routingTableNames = new String[] {};
```
dhalperi: suggest general inlining of these variables throughout PR, (if you feel like it). If only used once, may make code more readable.

PR #550, opened 2017-10-12
fix mutual exclusivity of route question flags

arifogel:


arifogel:
fromEnvironment and detail are in fact exclusive.

PR #553, opened 2017-10-12
Auto processing testrigs

ratulm:
First version of the feature that helps automatically analyze new testrigs. Auto analysis includes parsing, followed by running all configured analysis for the container. 

There are two limitations to the current implementation:
 1/ works only in the scenario of 1 worker
 2/ no tests yet. I am actually not sure how to test delayed, asynchronous actions in the current framework.


ratulm:
Please approve soon. Needs to be part of the demo workflow (and thus tested heavily)

projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
```diff
@@ -811,10 +814,38 @@ public void initTestrig(Path testrigDir,  Path srcDir, boolean autoProcess) {
...
+        autoWorkQueue.add(analyzeWork);
+      }
+
+      // NB: This way of doing things only works when we have a single worker; otherwise workitems
+      // lower down the order can get fired before those higher in the order
+      // The right solution is to put workitem2 on the queue only after workitem1 has finished
+      // successfully. The rightest solution is for workers to be aware of dependencies so they
+      // don't try to execute tasks that depend on other tasks that are currently being executed.
+
+      for (WorkItem workItem : autoWorkQueue) {
```
dhalperi: Is there a way to bail here if there's more than 1 worker connected?

ratulm: Being tracked as #573

PR #555, opened 2017-10-12
Smt reachability crash

rabeckett:
I changed the implementation of the symbolic packet to use a bitvector rather than an integer. This allows for any netmask to be used in an ACL.

Also made it so it will pinpoint the device and interface for a policy that uses an unimplemented feature.


projects/batfish/src/main/java/org/batfish/smt/Encoder.java
```diff
@@ -802,12 +803,12 @@ private BoolExpr environmentBlockingClause(Model m) {
...
-      System.out.println("Variables: " + stats.getNumVariables());
-      System.out.println("Z3 Time: " + stats.getTime());
-    }
-    // System.out.println("Stats:\n" + _solver.getStatistics());
+    //if (ENABLE_DEBUGGING) {
+    System.out.println("Constraints: " + stats.getNumConstraints());
+    System.out.println("Variables: " + stats.getNumVariables());
+    System.out.println("Z3 Time: " + stats.getTime());
+    System.out.println("Stats: \n" + _solver.getStatistics());
+    //}
```
ratulm: Will this lead to printing things on the console unconditionally?

rabeckett: Yes it will!

PR #556, opened 2017-10-12
External announcements redo

ratulm:


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
```diff
@@ -423,27 +424,25 @@ public int getWeight() {
...
+        _dstVrf,
+        _localPreference,
+        _med,
+        _network,
+        _nextHopIp,
+        _originType,
+        _originatorIp,
+        _srcIp,
+        _srcNode,
+        _srcProtocol,
```
arifogel: Does Objects.hash of an Enum use Enum.hashCode or Enum.ordinal? I think to be safe you should just hash the ordinal. Using hash of Enum itself instead of its ordinal was the cause inconsistent refs for BDP (and other issues I believe) earlier.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
```diff
@@ -423,27 +424,25 @@ public int getWeight() {
...
+        _srcVrf,
+        _type,
```
arifogel: Enum hash vs. ordinal as above.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
```diff
@@ -45,7 +46,8 @@ public Environment(
       @JsonProperty(PROP_NODE_BLACKLIST) @Nullable Set<String> nodeBlacklist,
       @JsonProperty(PROP_ENVIRONMENT_BGP_TABLES) @Nullable Map<String, String> bgpTables,
       @JsonProperty(PROP_ENVIRONMENT_ROUTING_TABLES) @Nullable Map<String, String> routingTables,
-      @JsonProperty(PROP_EXTERNAL_BGP_ANNOUNCEMENTS) @Nullable String externalBgpAnnouncements) {
+      @JsonProperty(PROP_EXTERNAL_BGP_ANNOUNCEMENTS) @Nullable
+          Set<BgpAdvertisement> externalBgpAnnouncements) {
```
arifogel: Should be SortedSet for consistent refs and to ease debugging. Generally, any Jackson collection should be SortedSet, SortedMap, or List for consistent ordering.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
```diff
@@ -92,7 +94,7 @@ public String getTestrigName() {
   }
 
   @JsonProperty(PROP_EXTERNAL_BGP_ANNOUNCEMENTS)
-  public String getExternalBgpAnnouncements() {
+  public Set<BgpAdvertisement> getExternalBgpAnnouncements() {
```
arifogel: Should be SortedSet as above.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
dhalperi: > hash ... cause of inconsistent refs

Can you say more? Hash codes are not guaranteed to be the same across different implementations of JVM, versions of JVM, or even different instances of the JVM.

Nothing should not rely on hash code consistency.

projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/pojo/EnvironmentTest.java
```diff
@@ -20,49 +25,67 @@
...
-            "announcement");
-    assertThat(
-        e.toString(),
-        equalTo(
-            "Environment{envName=environment, testrigName=testrig, "
-                + "edgeBlacklist=[], interfaceBlacklist=[], "
-                + "nodeBlacklist=[], bgpTables={}, routingTables={}, "
-                + "externalBgpAnnouncements=announcement}"));
-  }
+  // does not seem like a useful test; commenting for now. remove?
```
dhalperi: do you not want Environment to have a useful toString function?

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
arifogel: We may need to remove dependence on hash code consistency in the future. However, for BDP currently, it is required.
All the types that are hashed in BDP at the moment are deterministic by specification (that is, we do not rely on `Object.hashCode`). This might actually not matter for this class at the moment if it is not hashed during BDP oscillation detection, so perhaps this change is not currently required. But it is a simple change nonetheless.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
arifogel: Actually all the `Sets` and `Maps` (not just lines you modified) in this file should be switched to sorted versions if you don't mind.

ratulm:
Ok, I believe I addressed all comments.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Environment.java
```diff
@@ -29,23 +30,25 @@
...
   private final String _envName;
   private final String _testrigName;
-  private final Set<Edge> _edgeBlacklist;
-  private final Set<NodeInterfacePair> _interfaceBlacklist;
-  private final Set<String> _nodeBlacklist;
+  private final SortedSet<Edge> _edgeBlacklist;
+  private final SortedSet<NodeInterfacePair> _interfaceBlacklist;
+  private final SortedSet<String> _nodeBlacklist;
   private final Map<String, String> _bgpTables;
   private final Map<String, String> _routingTables;
```
arifogel: Use `SortedMap` here for `_bgpTables` and `_routingTables`

ratulm:
ping on this. please approve soon. ryan needs to build on it for the demo.

PR #558, opened 2017-10-12
Improve toString for a few classes

dhalperi:
`MatchAsPath@6f` -> `MatchAsPath{expr=NamedAsPathSet{name=1}}`

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchAsPath.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchAsPath.java
@@ -1,10 +1,12 @@
 package org.batfish.datamodel.routing_policy.expr;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
 import org.batfish.datamodel.routing_policy.Environment;
 import org.batfish.datamodel.routing_policy.Result;
 
-public class MatchAsPath extends BooleanExpr {
+public final class MatchAsPath extends BooleanExpr {
 
   /** */
   private static final long serialVersionUID = 1L;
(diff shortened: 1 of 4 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchAsPath.java
```diff
@@ -1,10 +1,12 @@
 package org.batfish.datamodel.routing_policy.expr;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
 import org.batfish.datamodel.routing_policy.Environment;
 import org.batfish.datamodel.routing_policy.Result;
 
-public class MatchAsPath extends BooleanExpr {
+public final class MatchAsPath extends BooleanExpr {
```
haverma: why is it made final ? If that is because it is not (or supposed to be not) extended by anyone, shouldn't NamedAsPath be made final as well?

arifogel: I intend for all of the classes in this package instantiated in the batfish project to be `final`.

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchAsPath.java
```diff
@@ -63,4 +55,13 @@ public int hashCode() {
   public void setExpr(AsPathSetExpr expr) {
     _expr = expr;
   }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(MatchAsPath.class)
+        .omitNullValues()
+        .add("comment", getComment())
+        .add("expr", _expr)
```
arifogel: Can you use the `PROP_XXX` constants here? If they don't exist in this class, I'd suggest making them and referencing them here (and in the getters and setters).

projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/NamedAsPathSet.java
```diff
@@ -83,4 +75,12 @@ public boolean matches(Environment environment) {
   public void setName(String name) {
     _name = name;
   }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(NamedAsPathSet.class)
+        .omitNullValues()
+        .add("name", _name)
```
arifogel: Use `PROP_XXX` here too.

dhalperi:
Is this what you had in mind?

arifogel:
Yes.

PR #561, opened 2017-10-13
initinfo overhaul with particular focus on iptables

arifogel:
- enshrine 'iptables' as testrig subdir rather than using arbitrary path
  set in host config

- generalize initInfo question to use set of extraComponent enums rather
  than having boolean option for each variant

- move iptables initinfo to extracomponent within initinfo to prevent
  name clashes

- update all test iptables file references to remove leading paths

- much refactoring to achieve above ends

- update refs

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/561)


arifogel:
Recently updated to resolve conflicts.

dhalperi:
Suggest that you think hard about whether this is a good output format. Are there other types of errors to include (e.g., errors converting a particular to to VI config)? Should we include all file types in the same structure? etc.

Minor comments, but under the assumption this is the way you want it general LGTM.

---

Reviewed 50 of 50 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfo.java, line 36 at r1](https://reviewable.io:443/reviews/batfish/batfish/561#-Ky-Yj0Y6gSxaEqF2xXH:-Ky-Yj0Y6gSxaEqF2xXI:b-hnvsqb) ([raw file](https://github.com/batfish/batfish/blob/753ff839bb6ad28207e1d423a8fd96711bf4e587/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfo.java#L36)):*
> ```Java
>   }
> 
>   public SortedMap<String, ParseTreeSentences> getParseTrees() {
> ```

Looks like this is `@Nullable` ; it's not set in constructor.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfo.java, line 189 at r1](https://reviewable.io:443/reviews/batfish/batfish/561#-Ky-Yp7cCRussgDHToVl:-Ky-Yp7cCRussgDHToVm:b-csid8t) ([raw file](https://github.com/batfish/batfish/blob/753ff839bb6ad28207e1d423a8fd96711bf4e587/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfo.java#L189)):*
> ```Java
>     }
>     return sb.toString();
>   }
> ```

should parse trees be used here?

parse trees seem generally ill-fit to this class – they're gettable and settable but not otherwise used.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfoAnswerElement.java, line 30 at r1](https://reviewable.io:443/reviews/batfish/batfish/561#-Ky-ZEfBIjsqycT7TrTE:-Ky-ZEfBIjsqycT7TrTF:b-x4ck6v) ([raw file](https://github.com/batfish/batfish/blob/753ff839bb6ad28207e1d423a8fd96711bf4e587/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/InitInfoAnswerElement.java#L30)):*
> ```Java
>     final AtomicBoolean empty = new AtomicBoolean(true);
>     if (!_parseStatus.isEmpty()) {
>       sb.append(super.prettyPrint());
> ```

something seems fishy here. The super class contains lots of things; why is this the only case in which we print something? (warnings, errors)

---

*[test_rigs/parsing-errors-tests/commands, line 8 at r1](https://reviewable.io:443/reviews/batfish/batfish/561#-Ky-_sOpr1nkQRaNLzD7:-Ky-_sOpr1nkQRaNLzD8:b-2zw1fw) ([raw file](https://github.com/batfish/batfish/blob/753ff839bb6ad28207e1d423a8fd96711bf4e587/test_rigs/parsing-errors-tests/commands#L8)):*
> ```
> #parsing Iptables error
> test test_rigs/parsing-errors-tests/init-invalid-iptables.ref init-testrig test_rigs/example-parsing-error/invalid-iptables
> test test_rigs/parsing-errors-tests/invalid-ip-no-verbose.ref get initinfo components=["CONFIGS","IPTABLES"]
> ```

suggest just making all components the default, and instead adding a test where we restrict to `IPTABLES`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/561#-:-Ky-a7rtrPjefb8LUr81:bgajwc8)*

arifogel:
Closing until need arises.

PR #566, opened 2017-10-13
Query for determinism

rabeckett:
Adds an "smt-deterministic" question that implements #525 to check if there can be more than one stable network solution. If the can be, it will show the different forwarding behavior for the two cases. The question takes the usual headerspace, number of failures, baseEnvType etc as parameters. The output for the classic bi-stable BGP network:

init-testrig test_rigs/smt-examples/STABLE3
get smt-deterministic

Flow<ingressNode:(none) ingressVrf:default srcIp:0.0.0.0 dstIp:42.42.42.1 ipProtocol:HOPOPT dscp: 0 ecn:0 fragmentOffset:0 packetLength:0 state:NEW tag:SMT>

Delta forwarding case 1:
   R2,Serial1 --> R3,Serial0 -- BgpRoute<42.42.42.0/24,nhip:192.168.43.1,nhint:dynamic>
   R3,Serial1 --> R1,Serial1 -- BgpRoute<42.42.42.0/24,nhip:192.168.44.2,nhint:dynamic>

Delta forwarding case 2:
   R2,Serial0 --> R1,Serial0 -- BgpRoute<42.42.42.0/24,nhip:192.168.42.2,nhint:dynamic>
   R3,Serial0 --> R2,Serial1 -- BgpRoute<42.42.42.0/24,nhip:192.168.43.2,nhint:dynamic>

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
@@ -188,4 +188,6 @@ public interface IBatfish extends IPluginConsumer {
   AnswerElement smtLoadBalance(HeaderLocationQuestion q, int threshold);
 
   AnswerElement smtLocalConsistency(Pattern routerRegex, boolean strict, boolean fullModel);
+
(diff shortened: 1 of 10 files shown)


ratulm:
nice!

PR #569, opened 2017-10-13
parse nxos (no) system interface

arifogel:


arifogel:
Err somehow lost ref changes. Will re-add review request after fixed.

PR #583, opened 2017-10-15
remove polling from BatfishJobExecutor

arifogel:
- Multiple orders of magnitude speedup for small jobs
- Use semaphore instead of polling to process completed jobs
- Don't allow jobs to throw Exception, simplifying handling
- Handle previously-unhandled exceptions possibly thrown during HostConfiguration deserialization during ParseVendorConfiguration job

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/583)

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/job/BatfishJob.java
+++ b/projects/batfish/src/main/java/org/batfish/job/BatfishJob.java
@@ -23,8 +23,7 @@ public abstract class BatfishJob<R extends BatfishJobResult<?, ?>> implements Ca
     try {
       R r = callBatfishJob();
       return r;
-    }
-    finally {
+    } finally {
       _semaphore.release();
     }
   }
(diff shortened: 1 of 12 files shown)


projects/batfish/src/main/java/org/batfish/job/BatfishJob.java
```diff
@@ -1,17 +1,36 @@
...
+      return r;
+    } finally {
+      _semaphore.release();
+    }
+  }
+
+  protected abstract R callBatfishJob();
+
+  public void setSemaphore(Semaphore semaphore) {
+    _semaphore = semaphore;
```
dhalperi: I don't think this is the right (set of) abstractions to be using here. I suspect that the built-in Callable is enough if we use the futures correctly. Let's discuss in person.

PR #594, opened 2017-10-17
Smt bgp network fix

rabeckett:
Improves the Minesweeper encoding of the BGP network statement so that BGP exports a route if there is *any* route known that matches the network, rather than just the best FIB entry (as per our discussion). Also implements null-routed static routes properly to test this behavior.

I added two test cases -- one using a static route and one with ospf. In both cases, BGP exports a route despite having a more specific /32 connected route.


ratulm:
How did you handle some of the intricacies that you were worried about? Are Quagga and Cisco being modeled differently? Should they be?

rabeckett:
Most of the intricacies are handled. When it sees an export filter for a BGP network command, it will first check statically if it knows there is a connected route or static route that will cause BGP to export a route. If there is no such route, then it will add a constraint that checks if there is an OSPF route with the appropriate prefix and length. The relevant logic is [here](https://github.com/batfish/batfish/blob/3f267bfe65e522ddfb129786d2050c45df711928/projects/batfish/src/main/java/org/batfish/smt/TransferFunctionSSA.java#L235)

Quagga and Cisco are not modeled differently right now. Do you know if Batfish generates a different AST for Quagga at the moment? Is there a way to tell Batfish that a configuration is Quagga -- otherwise, from the syntax it will probably assume it is Cisco?

ratulm:
It uses the parser and generates the same AST, but there is a getConfigurationFormat() function in the Configuration object that tells you what format Batfish thinks it was. It remembers the specific vendor format even if the parsers are shared (e.g., force10 vs Cisco). That said, it does not have a Quagga detector at all. If you know of specific ways to detect Quagga, you may add another format type to https://github.com/batfish/batfish/blob/3f267bfe65e522ddfb129786d2050c45df711928/projects/batfish/src/main/java/org/batfish/grammar/VendorConfigurationFormatDetector.java#L258 and use appropriate Cisco/Quagga logic. Either way, OK to merge and close this PR.

rabeckett:
FYI: I'm not authorized to merge this pull request

ratulm:
Merged. We'll fix permissions soon.

PR #595, opened 2017-10-17
Cleaned up and tested how OSPF inter-area summaries are computed

progwriter:


projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
```diff
@@ -217,11 +219,50 @@ public void computeFib() {
...
+   * @param route The route in question, whose metric is considered
+   * @param areaPrefix The Ip prefix of the OSPF area
+   * @param currentMetric The current summary metric for the area
+   * @param areaNum Area number.
+   * @return the newly computed summary metric.
+   */
+  @Nullable
+  static Long computeUpdatedOspfSummaryMetric(
+      OspfAreaRoute route, Prefix areaPrefix, @Nullable Long currentMetric, long areaNum) {
+    Prefix contributingRoutePrefix = route.getNetwork();
```
dhalperi: I generally prefer to structure my code with "exit early". e.g.,

```java
if (areaNum == route.getArea()) {
    // Routes from the same area have no effect.
    return currentMetric;
}

Prefix contributingRoutePrefix = route.getNetwork();
if (!areaPrefix.containsPrefix(contributingRoutePrefix)) {
    // Routes from outside of teh area prefix have no effect.
    return currentMetric;
}

...
```

This also makes the code generally less nested, hence easier to read.

dhalperi: (obviously you decide whether these are different `if` or the same `if`, etc.)

projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
```diff
@@ -217,11 +219,50 @@ public void computeFib() {
...
+    // Otherwise just take the best between route and current metrics
+    /*
+     * NOTE: The min function is used to compute the metric according to RFC 1583.
+     * HOWEVER, Cisco claims to have switched to using max function in IOS v.12.0 and later,
+     * as described in RFC 2328.
+     * (see https://www.cisco.com/c/en/us/support/docs/ip/open-shortest-path-first-ospf/7039-1.html#t29)
+     * Not sure if special handling is necessary by the OSPF process
+     */
+    return Math.min(currentMetric, contributingRouteMetric);
+  }
```
dhalperi: Seems like we should switch to the newer behavior here, or be version-aware. Keeping `min` seems like the worst of most worlds :)

dhalperi: (IOS 12.0 came out in 2001?)

projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
```diff
@@ -217,11 +219,50 @@ public void computeFib() {
...
+
+  boolean computeInterAreaSummaries() {
     OspfProcess proc = _vrf.getOspfProcess();
     boolean changed = false;
+    // Ensure we have a running OSPF process on the VRF
     if (proc != null) {
```
dhalperi: same comment re exit early: if proc == null just bail here, much easier to read.

projects/batfish/src/main/java/org/batfish/bdp/VirtualRouter.java
```diff
@@ -217,53 +219,95 @@ public void computeFib() {
...
+    if (currentMetric == null) {
+      return contributingRouteMetric;
+    }
+    // Take the best metric between the route's and current available
+    /*
+     * NOTE: Best was determined using the min function according to RFC 1583.
+     * However, OSPFv2 uses max function.
+     * Cisco claims to have switched to using max in IOS v.12.0 and later,
+     * as described in RFC 2328.
+     * (see https://www.cisco.com/c/en/us/support/docs/ip/open-shortest-path-first-ospf/7039-1.html#t29)
```
dhalperi: Cisco apparently defaults to "compatible rfc1583", so `min` is correct here. Please add a TODO here (and file a bug) for adding parsing support to handle `no compatible rfc1583` and plumbing it through to the dataplane (`useOldRfc` parameter on this function).

Also suggest renaming the variable from "old rfc" to something more meaningful.

PR #597, opened 2017-10-19
Multipath BGP AS-path comparison modes

### LINKED ISSUE #250: Handle AS path nuance for BGP multipath
There are several options regarding treatment of AS path when selecting BGP multipath routes after the best path is selected for a given network.

  1. Only consider advertisements sharing the exact same AS-path as the best path (Cisco default).
  2. Only consider advertisements sharing the same first AS and same AS-path length as the best path (Juniper default)
  3. Only consider advertisements sharing the same AS-path length as the best path (Juniper multipath multiple-as, Cisco multipath-relax).

Currently, option 3 is always assumed during BDP route computation. 
We should support options 1 and 2, which are the most common in real networks.

The following changes are required:
  - [ ] Parsing of AS-path-related configuration directives
  - [ ] Data-model update to support AS-path-related configuration directives
  - [ ] Updates to BDP to handle options 1 and 2.

arifogel:
Fixes #250


projects/batfish/src/main/java/org/batfish/bdp/BgpMultipathRib.java
```diff
@@ -47,6 +62,57 @@ public int comparePreference(BgpRoute lhs, BgpRoute rhs) {
     }
 
     /*
+     * AS path size is same. Now compare to best asPath (if available). Note we do not necessarily
+     * guarantee existing rhs route passes below test.
```
dhalperi: which below test are you talking about?

projects/batfish/src/main/java/org/batfish/bdp/BgpMultipathRib.java
```diff
@@ -47,6 +62,57 @@ public int comparePreference(BgpRoute lhs, BgpRoute rhs) {
...
+            return -1;
+          }
+          break;
+
+        case PATH_LENGTH:
+          // Skip since already compared
+          break;
+        default:
+          throw new BatfishException(
+              "Unsupported " + MultipathEquivalentAsPathMatchMode.class.getName());
```
dhalperi: Include the name of the class? `_multipathEquivalentAsPathMatchMode.getClass().getName()`?

projects/batfish/src/test/java/org/batfish/bdp/BdpDataPlanePluginTest.java
```diff
@@ -163,6 +167,147 @@ public void testApplySourceNatInvalidAclThrows() {
     BdpDataPlanePlugin.applySourceNat(flow, singletonList(nat));
   }
 
+  private void testBgpAsPathMultipath_helper(
```
dhalperi: generally no underscores in function names in Java. Just `Helper` at the end.

projects/batfish/src/test/java/org/batfish/main/BatfishTestUtils.java
```diff
@@ -102,11 +108,16 @@ private static Batfish initBatfishFromTestrigText(
...
-        settings.getBaseTestrigSettings().getSerializeVendorPath(),
-        settings.getBaseTestrigSettings().getSerializeIndependentPath());
+    try {
+      batfish.serializeVendorConfigs(
+          testrigPath, settings.getBaseTestrigSettings().getSerializeVendorPath());
+      batfish.serializeIndependentConfigs(
+          settings.getBaseTestrigSettings().getSerializeVendorPath(),
+          settings.getBaseTestrigSettings().getSerializeIndependentPath());
+    } catch (CompositeBatfishException e) {
+      System.err.print(e.getAnswerElement().prettyPrint());
```
dhalperi: Drop this. Put it in the exception.

PR #600, opened 2017-10-22
Allow plugin directory specifications to include a search package

dspicuzzbbn:
- Rationale is that some plugins may have dependencies that are problematic to load
  for lots of different reasons, and if there are a lot of classes then it can
  take a long time to scan/load/reflect all of the classes
- Added to the end of the path instead of as an argument to allow per-path search
  paths
- Default search path is empty, preserving the current behavior

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
@@ -150,7 +150,7 @@ public abstract class PluginConsumer implements IPluginConsumer {
     return ret;
   }
 
-  private boolean loadPluginJar(Path path) {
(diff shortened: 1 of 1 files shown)


arifogel:
We are going a whole different route with plug-ins in the next release (see PR #603). Specifically, we will be using Java [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) with Google's [AutoService](https://github.com/google/auto/tree/master/service).
Basically you should declare any new `Plugin` concrete classes as an implementer of `org.batfish.common.plugin.Plugin` via the following annotation:
`@AutoService(Plugin.class)`
You will also need to provide a `Plugin` service file in any new jars you create. The format is described in the [ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) documentation.

Please let us know if you have any issues using the new method once the new release is out.

PR #602, opened 2017-10-23
Removing polling in bf job executor and refactoring for unit testing

haverma:
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/602)


projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+/** Tests for {@link BatfishJobExecutor}. */
+public class BatfishJobExecutorTest {
+
+  @Rule public ExpectedException _thrown = ExpectedException.none();
+  BatfishLogger _logger;
+
+  @Before
+  public void setup() {
+    _logger = new BatfishLogger("info", false);
```
arifogel: Isn't there a `BfConsts` constant you can use here instead of `"info"`?

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+    BfTestAnswerElement ae = new BfTestAnswerElement();
+    BatfishJobExecutor<BfTestJob, BfTestAnswerElement, BfTestResult, Set<String>> executor =
+        new BatfishJobExecutor<>(
+            settings, _logger, settings.getHaltOnParseError(), "Test Job Executor");
+    long startTime = System.currentTimeMillis();
+    executor.executeJobs(jobs, output, ae);
+    long elapsedTime = System.currentTimeMillis() - startTime;
+
+    // tasks executed in parallel should take less than sequential time in most normal cases
+    assertThat(elapsedTime, lessThanOrEqualTo(2000L));
```
arifogel: It is not clear to me that sequential time would take more than 2000 milliseconds. The improvement you have implemented is over polling, not over sequential execution. Polling has been totally eliminated, so the test assertion is of dubious value here IMO.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+    // initializing executor
+    BatfishJobExecutor<BfTestJob, BfTestAnswerElement, BfTestResult, Set<String>> executor =
+        new BatfishJobExecutor<>(
+            settings, _logger, settings.getHaltOnParseError(), "Test Job Executor");
+    executor.initializeJobsStats(Lists.newArrayList(new BfTestJob(settings, "result1")));
+
+    // Simulating failure of a job and handling the result
+    BfTestResult bfTestResult =
+        new BfTestResult(
+            1000L, _logger.getHistory(), new BatfishException("Test Job Failure Message"));
```
arifogel: What is the `1000L` here for? Should it be extracted into a constant?

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+    List<BatfishException> failureCauses = new ArrayList<>();
+    BfTestAnswerElement ae = new BfTestAnswerElement();
+    executor.updateJobsStats();
+    executor.handleJobResult(bfTestResult, output, ae, failureCauses);
+
+    // checking the log of the executor for the job finished
+    assertEquals(
+        _logger.getHistory().toString(400),
+        String.format(
+            "Job terminated successfully with result: %s after elapsed time: %s - %d/%d "
```
arifogel: This test is highly sensitive to the text produced by `BatfishJobExecutor`, making it difficult to maintain. You should design your test in such a way that this sensitivity does not exist, or refactor the message producing code so it can be called here and in the executor.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+        new BfTestResult(
+            1000L, _logger.getHistory(), new BatfishException("Test Job Failure Message"));
+    Set<String> output = new HashSet<>();
+    List<BatfishException> failureCauses = new ArrayList<>();
+    BfTestAnswerElement ae = new BfTestAnswerElement();
+    executor.updateJobsStats();
+    executor.handleJobResult(bfTestResult, output, ae, failureCauses);
+
+    //checking that correct failure message is written in the log
+    assertEquals(
```
arifogel: This test is highly sensitive to the text produced by `BatfishJobExecutor`, making it difficult to maintain. You should design your test in such a way that this sensitivity does not exist, or refactor the message producing code so it can be called here and in the executor.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    // initializing executor
+    BatfishJobExecutor<BfTestJob, BfTestAnswerElement, BfTestResult, Set<String>> executor =
+        new BatfishJobExecutor<>(
+            settings, _logger, settings.getHaltOnParseError(), "Test Job Executor");
+    List<BatfishException> failureCauses = new ArrayList<>();
+
+    // checking if the exception thrown has correct class and message
+    _thrown.expect(CompositeBatfishException.class);
+    _thrown.expectMessage("Fatal exception due to failure of at least one job");
```
arifogel: In general I'd prefer that you do not expect messages from `Exception` classes that we control. Instead, subclass the `Exception` and simply expect the appropriate type. If there is really only one usage of the exception in question, then no test of the message is necessary.

arifogel:
This PR fixes #583. The latter should not be merged.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
arifogel: I should add that this test *is* actually useful to compare to polling. But then you should adjust the comment accordingly.

haverma: The BfTestJob is not using any parsing code, it is just returning a string after a configurable delay. So the tests are independent of any parsing logic and they are only testing the  BfJObExecutor

arifogel: Ah I misunderstood the test. I withdraw my objection.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
-            new BatfishException("Fatal exception due to failure of at least one job"),
-            failureCauses);
-      }
+      handleProcessingError(jobs, failureCauses);
     } else if (!_logger.isActive(BatfishLogger.LEVEL_INFO)) {
       _logger.info("All jobs executed successfully\n");
     }
   }
+
+  ExecutorService getExecutorService() {
```
dhalperi: this should be called `create` or `make` -- `get` implies it returns an existing one.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+    ExecutorService pool;
+    if (!_settings.getSequential()) {
```
dhalperi: Make the code more readable -- flip the order of the `if` statements so you don't need the negation.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+      List<BatfishException> failureCauses) {
+    boolean processingError = false;
+    String time = CommonUtil.getTime(result.getElapsedTime());
+    Throwable failureCause = result.getFailureCause();
+    if (failureCause == null) {
+      result.applyTo(output, _logger, answerElement);
+      _logger.infof(
+          "Job terminated successfully with result: %s after elapsed time: %s - %d/%d "
+              + "(%.1f%%) complete\n",
+          result.toString(), time, _finishedJobs, _totalJobs, _finishedPercent);
```
dhalperi: drop `.toString()` on value being passed to printf.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+   * @param answerElement Answer Element for the execution of jobs
+   * @param failureCauses List of failure causes for all jobs
+   * @return True if there is a processing error else False
+   */
+  boolean handleJobResult(
+      JobResultT result,
+      OutputT output,
+      AnswerElementT answerElement,
+      List<BatfishException> failureCauses) {
+    boolean processingError = false;
```
dhalperi: drop this useless local variable; just `return false` or `true` in the places you need to return.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+              time,
+              result.getHistory().toString(BatfishLogger.getLogLevel(_settings.getLogLevel())));
+      BatfishException bfc = new BatfishException(failureMessage, failureCause);
+      if (_settings.getExitOnFirstError()) {
+        result.appendHistory(_logger);
+        throw bfc;
+      } else {
+        processingError = true;
+        result.appendHistory(_logger);
+        _logger.error(failureMessage + ":\n\t" + ExceptionUtils.getStackTrace(failureCause));
```
dhalperi: `errorf`

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+          result.applyTo(output, _logger, answerElement);
+        }
+      }
+    }
+    return processingError;
+  }
+
+  void handleProcessingError(List<JobT> jobs, List<BatfishException> failureCauses) {
+    int numJobs = jobs.size();
+    int numFailed = numJobs - failureCauses.size();
```
dhalperi: is this right? Why isn't this just `failureCauses.size()`?

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+  }
+
+  void initializeJobsStats(List<JobT> jobs) {
+    _finishedJobs = 0;
+    _totalJobs = jobs.size();
+    _completed = Driver.newBatch(_settings, _description, _totalJobs);
+    _finishedPercent = 0;
+  }
+
+  void updateJobsStats() {
```
dhalperi: this is a bad name for this function. How about `markJobCompleted`?

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
     _description = description;
   }
 
+  /**
```
dhalperi: (comments at the top of diff apply to code above them -- GitHub won't let me comment up there.)

Add class-level javadoc with a simple, non-jargony explanation of what this class does.

Delete unused constants (e.g., JOB_POLLING_PERIOD)?

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
-        } else {
-          continue;
-        }
-      }
-      if (!futures.isEmpty()) {
+
+    for (int i = 0; i < jobs.size(); i++) {
+      try {
+        // getting the next completed task
+        Future<JobResultT> currentDoneFuture = completionService.take();
```
dhalperi: The code inside a try should be minimal. UpdateJobStats, result getting, etc., should be outside the `try` clause.

Honestly, I'd probably just

```java
try {
    for (...) {
       result = completionService.take().get();
       ...
    }
} catch (...) {
} finally {
    pool.shutdownNow();
}
```

and handle the various exceptions acceptably.

* InterruptedException: this means the task didn't finish.
* ExecutionException: the task failed.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+        boolean jobResultError = handleJobResult(result, output, answerElement, failureCauses);
+        if (jobResultError) {
+          processingError = true;
         }
+
+      } catch (InterruptedException e) {
+        throw new BatfishException("interrupted while waiting for jobs to complete", e);
       }
     }
     pool.shutdown();
```
dhalperi: need to shut down the `pool` even in case of exception.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+      _logger.infof("%d jobs succeeded; %d jobs failed\n", numSucceeded, numFailed);
+    }
+    if (_haltOnProcessingError) {
+      throw new CompositeBatfishException(
+          new BatfishException("Fatal exception due to failure of at least one job"),
+          failureCauses);
+    }
+  }
+
+  void initializeJobsStats(List<JobT> jobs) {
```
dhalperi: I don't quite understand why `BatfishJobExecutor` is a seemingly reusable class with state that is reinitialized per job rather than something that is created in a static factory method per job and used only once.

The current form seems hard to use right, doesn't actually support multiple concurrent uses, multithreading, etc.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: Updatejobstats, resultgetting/processing has to be done in each iteration so this approach will need them inside the try catch, I can put the try catch inside the for... loop but that will need the pool.shutdown() method to be called in all catch blocks and also after the looping is done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: done

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: done

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: done

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: done

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
     _description = description;
   }
 
+  /**
+   * Executes jobs in a thread pool or in a single thread executor depending on settings
+   *
+   * @param jobs Jobs to be executed
+   * @param output Data Structure which will contain the output of the jobs
```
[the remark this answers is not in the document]
haverma: updated description in the class java doc and here as well. OutputT is the type of the output argument

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+   * @param answerElement Answer Element containing the detail of the jobs executed
```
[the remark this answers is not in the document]
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done, but not sure if this is the best way to do it

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done, but related to above comment

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: changed name

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+  ExecutorService getExecutorService() {
+    ExecutorService pool;
+    if (!_settings.getSequential()) {
+      int maxConcurrentThreads = Runtime.getRuntime().availableProcessors();
+      int numConcurrentThreads = Math.min(maxConcurrentThreads, _settings.getJobs());
+      pool = Executors.newFixedThreadPool(numConcurrentThreads);
+    } else {
+      pool = Executors.newSingleThreadExecutor();
+    }
+    return pool;
```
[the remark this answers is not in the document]
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+  }
+
+  /**
+   * Handles the Job result appropriately in case of success/error, and updates logs
```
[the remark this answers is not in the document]
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +50,143 @@ public BatfishJobExecutor(
...
+   *
+   * @param result Result of Job
+   * @param output Output comprising results of all jobs
+   * @param answerElement Answer Element for the execution of jobs
+   * @param failureCauses List of failure causes for all jobs
+   * @return True if there is a processing error else False
```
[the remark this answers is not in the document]
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: done

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: changed, this was being wrongly done which I failed to notice

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: changed

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: changed

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +57,162 @@ public BatfishJobExecutor(
...
+        return true;
+      }
+    }
+  }
+
+  void handleProcessingError(List<JobT> jobs, List<BatfishException> failureCauses) {
+    int numJobs = jobs.size();
+    int numFailed = failureCauses.size();
+    int numSucceeded = numJobs - numFailed;
+    if (!_logger.isActive(BatfishLogger.LEVEL_INFO)) {
```
dhalperi: I don't understand this block -- we check if LEVEL_INFO is *disabled* and then, if it's disabled, log at INFO level.

What's going on?

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
```diff
@@ -41,105 +57,162 @@ public BatfishJobExecutor(
...
+      _logger.infof("%d jobs succeeded; %d jobs failed\n", numSucceeded, numFailed);
+    }
+    if (_haltOnProcessingError) {
+      throw new HandleProcessingErrorException(failureCauses);
+    }
+  }
+
+  void initializeJobsStats(List<JobT> jobs) {
```
dhalperi: any functions that are not used outside of this class (e.g., in a unit test) should be `private`.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,214 @@
...
+    BfTestAnswerElement ae = new BfTestAnswerElement();
+    BatfishJobExecutor<BfTestJob, BfTestAnswerElement, BfTestResult, Set<String>> executor =
+        new BatfishJobExecutor<>(
+            settings, _logger, settings.getHaltOnParseError(), TEST_EXECUTOR_DESC);
+    long startTime = System.currentTimeMillis();
+    executor.executeJobs(jobs, output, ae);
+    long elapsedTime = System.currentTimeMillis() - startTime;
+
+    // Without polling and correct parallel execution, following should be true
+    assertThat(elapsedTime, lessThanOrEqualTo(2 * TEST_JOB_DELAY));
```
dhalperi: This type of test should not exist – it will be flaky. There's no guarantee that the user's machine won't be under high load, go into garbage collection, or get swapped out during the test running.

If you want to test a virtual clock, great. (E.g., inject a fake system clock and move it forward whenever the user calls sleep)

But we should not have any unit tests that insist some code finishes in a fixed wall-clock time. Instead, we should have benchmarks and a dashboard that shows how the benchmark performance changes over time.

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
```diff
@@ -0,0 +1,214 @@
...
+      super(settings);
+      _testValue = testValue;
+    }
+
+    @Override
+    public BfTestResult call() throws Exception {
+      long startTime = System.currentTimeMillis();
+      long elapsedTime;
+      try {
+        TimeUnit.SECONDS.sleep(1);
```
dhalperi: no real sleeps in tests. Can use a virtual clock if you are testing something that is accounting for the total time slept. (But I don't think that's how this was being used)

arifogel:
Dismissing my review and deferring to @dhalperi, as his remaining objections make sense and I have no issues on top of his.

projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java
haverma: this is being used in the unit test in testing of handleJobResult

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: removed

projects/batfish/src/test/java/org/batfish/job/BatfishJobExecutorTest.java
haverma: removed

PR #603, opened 2017-10-23
switch plugins to use AutoService

arifogel:
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/603)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
@@ -1,12 +1,12 @@
 package org.batfish.common.plugin;
 
+import com.google.common.collect.Lists;
 import com.google.common.io.ByteStreams;
 import com.google.common.io.Closer;
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.io.xml.DomDriver;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
-import java.io.File;
 import java.io.FileInputStream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
@@ -16,33 +16,19 @@ import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.io.PushbackInputStream;
 import java.io.Serializable;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Modifier;
-import java.net.URL;
-import java.net.URLClassLoader;
-import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.SimpleFileVisitor;
-import java.nio.file.attribute.BasicFileAttributes;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
+import java.util.LinkedHashSet;
+import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 import org.apache.commons.io.IOUtils;
 import org.batfish.common.BatfishException;
-import org.batfish.common.BfConsts;
 import org.batfish.common.util.BatfishObjectInputStream;
 
 public abstract class PluginConsumer implements IPluginConsumer {
-
-  private static final String CLASS_EXTENSION = ".class";
 
   /**
    * A byte-array containing the first 4 bytes of the header for a file that is the output of java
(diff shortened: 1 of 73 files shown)


projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
```diff
@@ -150,153 +124,12 @@ private boolean isJavaSerializationData(PushbackInputStream stream) throws IOExc
...
-            throw new BatfishException(
-                "Error loading plugin folder: '" + pluginDir.toString() + "'", e);
-          }
-        }
-      }
+    Set<Plugin> plugins =
+        new LinkedHashSet<>(
+            Lists.newArrayList(ServiceLoader.load(Plugin.class, _currentClassLoader)));
+    for (Plugin plugin : plugins) {
+      plugin.initialize(this);
```
dhalperi: I'd probably wrap this in a try/catch and report an error loading any of these but continue loading others?

arifogel: Hmm I don't know. I don't think it really makes sense to keep running when a plugin has failed. How about catch, wrap, and throw?

arifogel:
Rebased and resolved conflict with OutliersQuestionPlugin.java (added AutoService annotation, imports on top of master version)

projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/PluginConsumer.java
arifogel: Addressed via catching of ServiceConfigurationException, throwing CompositeBatfishException after all are done loading.

dhalperi:
Reviewed 70 of 71 files at r1, 3 of 3 files at r2.

---

*[projects/batfish/batfish, line 5 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJyRUMg99x3gCNa0oo:-KxJyRUMg99x3gCNa0op:bnw6bsb) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish/batfish#L5)):*
> ```
> BATFISH_PATH="$(dirname $BATFISH)"
> ALLINONE_PATH="${BATFISH_PATH}/../allinone"
> ALLINONE_JAR="$ALLINONE_PATH/target/allinone-bundle-${BATFISH_VERSION}.jar"
> ```

What's the process you use to decide to include `${X}` vs just `$X`?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java, line 8 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJzR2dBU7IA40FssnR:-KxJzR2dBU7IA40FssnS:b2xbo1f) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java#L8)):*
> ```Java
> 
>   @Override
>   public final int compareTo(Plugin o) {
> ```

Should this really be `final`? Would plugin equality not need to be tested later for extensions that care about other fields?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/603)*

arifogel:

---

*[projects/batfish/batfish, line 5 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJyRUMg99x3gCNa0oo:-KxKMCI2XqQRFknz9_69:bs3an9e) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish/batfish#L5)):*

What's the process you use to decide to include `${X}` vs just `$X`?
</blockquote></details>

Really it is only necessary when there is a separator character (e.g. "-") in the name. But I like to use it all the time. Sometimes I am weak and forget.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java, line 8 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJzR2dBU7IA40FssnR:-KxKMTjLw8zbKOFF-r5A:b-iwnwci) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java#L8)):*

Should this really be `final`? Would plugin equality not need to be tested later for extensions that care about other fields?
</blockquote></details>

The hope is no. Can you provide a plausible counterexample?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/603)*

dhalperi:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java, line 8 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJzR2dBU7IA40FssnR:-KxKNlXImf6LFJE3lkhQ:b-sg0mv4) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java#L8)):*

The hope is no. Can you provide a plausible counterexample?
</blockquote></details>

Perhaps a better question would be why these need to be comparable at all; what's this actually used for?

If it's for putting them in some sorted order, probably makes more sense to have a delegate comparator of the names than to actually make `Plugin` itself comparable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/603)*

arifogel:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java, line 8 at r2](https://reviewable.io:443/reviews/batfish/batfish/603#-KxJzR2dBU7IA40FssnR:-KxKPwwyp9iTh2iAD45z:b-1und3v) ([raw file](https://github.com/batfish/batfish/blob/1931d897a51b9bc8435eedc1126a67805bfbcaa6/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/Plugin.java#L8)):*

Perhaps a better question would be why these need to be comparable at all; what's this actually used for?

If it's for putting them in some sorted order, probably makes more sense to have a delegate comparator of the names than to actually make `Plugin` itself comparable.
</blockquote></details>

Yes it is just for sorting. From my perspective, if it is ok for `equals` and `hashCode` being defined as they are and final, then it is not a jump to accept `compareTo` existing as implemented and being final. If someone wants to sort in a different order or using different fields, they can use a different comparator.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/603)*

PR #493, opened 2017-10-04
updating client to fetch questions from batfish service

haverma:
also, a small change in initOrAddAnalysis as it is using loadQuestions()

PR #605, opened 2017-10-24
Revert "updating client to fetch questions from batfish service"

arifogel:
Reverts batfish/batfish#493

This changes the output of the `load-questions` command. The output after the change includes only lower-case names of questions. This behavior should be prevented, and a unit test added to make sure the name that comes out is the actual `instanceName` of the loaded question.


dhalperi:
Harsh, can you look into this and figure out what needs to be done?

I suspect it's just something in the client output code.

PR #608, opened 2017-10-24
network abstraction

rabeckett:
Fairly large overhaul and addition to the SMT codebase that allows for network symmetry reductions to improve performance. For reference, the changes at a high level are:

* I made a number of refactorings and general code improvements to the existing SMT code
* The SMT module is also now broken down into a number of smaller modules including: a bdd module that contains representations and conversions from route-maps and acls to binary decision diagrams, and an abstraction module, which can be used to generate a stream of equivalence classes (pairs of new, abstracted configurations and a sub-headerspace).
* The BDD code is used to implement a Role question to infer the roles of devices based on policy equivalence. The two parameters are "nodeRegex" to select devices, and equivType=node|interface|policy
* Abstraction can be used with SMT by adding the useAbstraction=True flag to any query. To make it easier to compare performance I also added a benchmark=True flag to SMT queries that will break down how much time is being spent and where for both abstract and non-abstract queries. The abstraction produced should work with the failures flag 

Note: this adds a dependency on the javabdd library, which is included in the projects/lib folder

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/608)


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/smt/HeaderQuestion.java
```diff
@@ -42,7 +43,7 @@
 
   private static final String PROP_NOT_SRC_IPS = "notSrcIps";
 
-  private static final String NOT_SRC_PORTS_VAR = "notSrcPorts";
+  private static final String PROP_NOT_SRC_PORTS_VAR = "notSrcPorts";
```
ratulm: nix _VAR from the constant name?

rabeckett:
Sounds good.

ratulm:
Ok to merge after tests pass.

PR #611, opened 2017-10-25
Updating the PR to remove lower casing of question names

### LINKED ISSUE #606: Re-implement PR #493 without breaking changes
PR #493 introduced breaking changes, and so was reverted in #605.
It should be re-implemented without causing the breaking change described in #605, and with appropriate tests to prevent recurrence.

haverma:
Fixes #606


projects/batfish-client/src/main/java/org/batfish/client/Client.java
```diff
@@ -2000,6 +2133,53 @@ private void logOutput(FileWriter outWriter, String message) {
...
+   * @param ae {@link LoadQuestionAnswerElement} where info has to be updated
+   */
+  static void updateLoadedQuestionsInfo(
+      String questionName,
+      String questionContent,
+      Map<String, String> loadedQuestions,
+      LoadQuestionAnswerElement ae) {
+    //adding question name in added list if not present else add in replaced list
+    if (loadedQuestions.containsKey(questionName.toLowerCase())) {
+      ae.getReplaced().add(questionName);
```
haverma: Adding questionName without changing the case

dhalperi:
Please start this PR from master, using a revert of #605. I expect ~2 commits: 1. revert 605, 2. make the fixes changes and add a test.

projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
```diff
@@ -438,12 +445,47 @@ public void testGetDeltaValidParas() throws Exception {
...
   }
 
   @Test
   public void testGetQuestionInvalidParas() throws Exception {
-    testInvalidInput(GET_QUESTION, new String[] {});
+    testInvalidInput(GET_QUESTION, new String[] {}, new String[] {});
+  }
+
+  @Test
+  public void testGetQuestionName() throws JSONException {
```
haverma: Test to check question names being loaded

PR #612, opened 2017-10-25
Load questions lowercase answer element fix 

haverma:
FIxes issue #606

### DIFF
--- a/projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
+++ b/projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
@@ -913,7 +913,8 @@ public class ClientTest {
 
     // Reading the answer written by load-questions
     Answer answerLoadQuestions =
-        mapper.readValue(client.getLogger().getHistory().toString(220), Answer.class);
+        mapper.readValue(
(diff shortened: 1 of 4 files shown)


projects/batfish-client/src/test/java/org/batfish/client/ClientTest.java
```diff
@@ -894,6 +895,33 @@ public void testLoadQuestionFromTextInvalid() throws Exception {
...
+    Path questionJsonPath = _folder.newFile("testquestion.json").toPath();
+    CommonUtil.writeFile(questionJsonPath, testQuestion.toString());
+    client._logger = new BatfishLogger("output", false);
+    client.processCommand(
+        new String[] {LOAD_QUESTIONS.commandName(), questionJsonPath.getParent().toString()}, null);
+    BatfishObjectMapper mapper = new BatfishObjectMapper();
+
+    // Reading the answer written by load-questions
+    Answer answerLoadQuestions =
+        mapper.readValue(client.getLogger().getHistory().toString(220), Answer.class);
```
dhalperi: what is happening here? 220?

haverma: I am getting the output level logs written by load-questions as that has the answer element, which I want to compare. I will change it to the constant from BfLogger

PR #613, opened 2017-10-25
pom.xml: turn on Maven dependency convergence

dhalperi:
From the docs:

> This rule requires that dependency version numbers converge. If a
> project has two dependencies, A and B, both depending on the same
> artifact, C, this rule will fail the build if A depends on a different
> version of C then the version of C depended on by B.

This will generally improve build safety. Also resolve two existing
dependency convergence issues.


arifogel:
What are the existing dependency convergence issues?

PR #624, opened 2017-10-30
Parsing recovery

### LINKED ISSUE #610: Better recovery for parsers and lexers
Instead of barfing, throw stuff into unrecognized lines.

arifogel:
- Fixes #610
- For line-based configs, invalid lines are thrown out and recorded as unrecognized lines.
- Failure should not be possible for such grammars except in severe pathological circumstances
- Recovery can be disabled altogether with `-disableunrecognized` flag, yielding legacy-ish behavior
- Recovery can trigger graceful failure with `-urf false` flag, which has no effect when `-disableunrecognized` is set.
- Currently should work for following grammars:
  - Cisco
  - FlatJuniper
  - FlatVyos
  - Iptables
  - EosRoutingTable
  - IosRoutingTable
  - NxosRoutingTable

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/624)


dhalperi:
Pretty tough to review as-is.

Expect to see unit tests of the code in various new classes involved in recovery (not end-to-end, unit tests of the functions themselves).

---

Reviewed 41 of 64 files at r1.
Review status: 41 of 64 files reviewed at latest revision, 27 unresolved discussions.

---

*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiEwrVxUOjwnWWmnQL:-KxiEwrVxUOjwnWWmnQM:bdvso3p) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*
> ```XML
>               <goals><goal>add-test-source</goal></goals>
>               <configuration>
>                 <sources><source>${project.build.directory}/generated-test-sources/antlr4</source> </sources>
> ```

Drop space?

---

*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiF2GRcEisWHMQS73n:-KxiF2GRcEisWHMQS73o:bj60vo6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*
> ```XML
>               <goals><goal>add-test-source</goal></goals>
>               <configuration>
>                 <sources><source>${project.build.directory}/generated-test-sources/antlr4</source> </sources>
> ```

We should version the plugin in the top-level module, but only add `antlr4` test sources in the module(s) that have them. just `batfish`?

---

*[projects/suppressions.xml, line 7 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFa5BTCCUAKuRcWlQ:-KxiFa5BTCCUAKuRcWlR:b-4kgrma) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/suppressions.xml#L7)):*
> ```XML
>     "http://www.puppycrawl.com/dtds/suppressions_1_0.dtd">
> 
> <suppressions>
> ```

Explicitly listing the `main/src` and `test/src` seems a cleaner/simpler solution. See, e.g., https://github.com/spotify/dockerfile-maven/pull/60/files#diff-600376dffeb79835ede4a0b285078036

---

*[projects/batfish/pom.xml, line 113 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFx0E9P1jZeidmr9e:-KxiFx0E9P1jZeidmr9f:b-w8m9ra) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/pom.xml#L113)):*
> ```XML
>           
>           <execution>
>             <id>antlr4-test-simple</id>
> ```

Add `<phase>generate-test-sources</phase>`, I think [link](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html). That way the test sources will not be generated unless tests are to be run.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/CompositeBatfishException.java, line 58 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiGJalRhiGSvUsz6Ag:-KxiGJalRhiGSvUsz6Ah:bgajxnq) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/common/CompositeBatfishException.java#L58)):*
> ```Java
>   }
> 
>   public BatfishException asSingleException() {
> ```

Note to self: understand why this is here.

Add javadoc?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 336 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiGrLrjDwqT6RGD5-9:-KxiGrLrjDwqT6RGD5-A:bjikws8) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L336)):*
> ```Java
>   public static <E, T extends E, E1 extends T, E2 extends T, S extends Set<E>>
>       Set<E> immutableDifference(Set<E1> set1, Set<E2> set2, Supplier<S> setConstructor) {
>     return Collections.unmodifiableSet(difference(set1, set2, setConstructor));
> ```

Any reason not to use [Guava `Sets`](https://github.com/google/guava/wiki/CollectionUtilitiesExplained#sets) for these? It already returns an unmodifiable *view*, lazily evaluated. You could also use `ImmutableSet.copyOf(Sets.difference(set1, set2))` if there's worry that set1/set2 might change underneath you.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 1 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKA9wDOcOozDBiwzs:-KxiKA9wDOcOozDBiwzt:b1uci2f) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L1)):*
> ```Java
> package org.batfish.grammar;
> ```

I'm not precisely sure of the "mode" in which to review this. Clearly I am missing a lot of ANTLR background.

Should I expect implementation comments in this file to give me an overview of the necessary context? Should I expect the reader to be an expert on ANTLR?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 23 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiI1bbqjzIDJp30_2K:-KxiI1bcJi8Rcgz0OhGv:b-7iawb4) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L23)):*
> ```Java
> import org.antlr.v4.runtime.tree.ParseTree;
> 
> public class BatfishANTRLErrorStrategy extends DefaultErrorStrategy {
> ```

Top-level javadoc would be helpful as to what this class does from the public API context. What types of errors can it recover and how does it do so?

Also a (non-javadoc) block comment as to how it does it would be useful for those reading the code.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 29 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiHY3h6TboAMpUwVgE:-KxiHY3h6TboAMpUwVgF:b-21quyf) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L29)):*
> ```Java
>   private int _separatorToken;
> 
>   public BatfishANTRLErrorStrategy(
> ```

Javadoc - what are these parameters for?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 30 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiIAA-F6H9BnpXW7VJ:-KxiIAA-F6H9BnpXW7VK:b-ysg0a9) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L30)):*
> ```Java
> 
>   public BatfishANTRLErrorStrategy(
>       int separatorToken, String minimumRequiredSeparatorText, String text) {
> ```

Inferring from the code, it looks like the first two parameters are specific to the strategy and the third is the actual text being parsed? I'd probably separate these into two different classes / functions for clarity and reusability.

```java
F = new BatfishErrorStrategy(separatorToken, minimumRequiredSeparatorText);
E = F.forText(text);
```

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 32 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiHJBC5Y19YpXR4iaF:-KxiHJBC5Y19YpXR4iaG:bdleg9s) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L32)):*
> ```Java
>       int separatorToken, String minimumRequiredSeparatorText, String text) {
>     _lines =
>         Collections.unmodifiableList(
> ```

ImmutableList.copyOf?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 53 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJWiqykxkJCT2G_uM:-KxiJWiqykxkJCT2G_uN:b-9pgwjd) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L53)):*
> ```Java
>       String separator = separatorToken.getText();
> 
>       // Get the line number and separator text from the separator token
> ```

Comment seems wrong -- copy and paste of the one 5 lines up?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 59 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJboilb3HiVe8L14w:-KxiJboilb3HiVe8L14x:b3kre) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L59)):*
> ```Java
>       recognizer.consume();
> 
>       nextToken = recognizer.getInputStream().LA(1);
> ```

?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 91 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJir4XeLGLFh_Hcip:-KxiJir4XeLGLFh_Hciq:b9agr0f) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L91)):*
> ```Java
>   }
> 
>   private Token recover(Parser recognizer) {
> ```

what does this do? Expected pre-post conditions? etc.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 127 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKZziSvKIwKWVe_uq:-KxiKZziSvKIwKWVe_ur:bsapmea) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L127)):*
> ```Java
>   }
> 
>   public void recoverInCurrentNode(Parser recognizer) {
> ```

javadoc on all public functions

(Without looking elsewhere to see how it's used:) does this need to be public?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java, line 73 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKq9fRrU4UfX9vFrk:-KxiKq9fRrU4UfX9vFrl:bvs0jod) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java#L73)):*
> ```Java
>     _parser.initErrorListener(this);
>     _parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
>     if (!settings.getDisableUnrecognized()) {
> ```

? - don't get this.

comment?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishGrammarErrorListener.java, line 36 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL-Gx58g_vq3wE-42:-KxiL-Gx58g_vq3wE-43:b-uudwgy) ([raw file](https://github.com/batfish/batfish/blob/53015193909cd1113fa939acc8b6bda67ac752c1/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishGrammarErrorListener.java#L36)):*
> ```Java
>   public void reportContextSensitivity(
>       Parser arg0, DFA arg1, int arg2, int arg3, int arg4, ATNConfigSet arg5) {}
> ```

also delete blank line?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 15 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL6SbAPno65WP67ZC:-KxiL6ScyQvDAnYZpgK1:bo3npbu) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L15)):*
> ```Java
>   private BatfishCombinedParser<?, ?> _parser;
> 
>   private BatfishLexerRecoveryStrategy _recoveryStrategy;
> ```

`@Nullable`

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 25 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL5X2x1tetjgGc7tl:-KxiL5X2x1tetjgGc7tm:bo3npbu) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L25)):*
> ```Java
>   }
> 
>   public BatfishLexerRecoveryStrategy getRecoveryStrategy() {
> ```

`@Nullable`

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 62 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL7ns_rztDgOCADzH:-KxiL7ns_rztDgOCADzI:bdnfbrl) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L62)):*
> ```Java
>   }
> 
>   public void setRecoveryStrategy(BatfishLexerRecoveryStrategy recoveryStrategy) {
> ```

is this `@Nullable`?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java, line 12 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLmNLO7pxEOGYthto:-KxiLmNLO7pxEOGYthtp:b5fubzl) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java#L12)):*
> ```Java
> import org.batfish.common.util.CommonUtil;
> 
> public class BatfishLexerRecoveryStrategy {
> ```

javadoc

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java, line 29 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLJDpIcBounIBXx00:-KxiLJDpIcBounIBXx01:bq4tn93) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java#L29)):*
> ```Java
> 
>   private static Set<Integer> whitespaceAndNewlines() {
>     return CommonUtil.immutableUnion(whitespace(), newlines(), HashSet::new);
> ```

`ImmutableSet.builder().addAll(whitespace()).addAll(newlines()).build();` ? (guessed at that syntax btw)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserATNSimulator.java, line 8 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLw9IrShuUuhjcPzq:-KxiLw9J6SUMs2y7CHLo:b2g0fkp) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserATNSimulator.java#L8)):*
> ```Java
> import org.antlr.v4.runtime.atn.ParserATNSimulator;
> 
> public class BatfishParserATNSimulator extends ParserATNSimulator {
> ```

Javadoc. What does this do that's different than the default?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishRecognitionException.java, line 8 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiM2Rh89iJgAJOxeDH:-KxiM2Rh89iJgAJOxeDI:b-dqocmw) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishRecognitionException.java#L8)):*
> ```Java
> import org.antlr.v4.runtime.Recognizer;
> 
> public class BatfishRecognitionException extends RecognitionException {
> ```

Looks like this adds nothing. Why?

`final class`?

Javadoc?

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4, line 858 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiMNTPNGQ-uK0IfZ9U:-KxiMNTQx9It1qSTz61q:b-6n3d7h) ([raw file](https://github.com/batfish/batfish/blob/53015193909cd1113fa939acc8b6bda67ac752c1/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4#L858)):*
> ```ANTLR
>    { !_disableUnrecognized }?
> 
>    unrecognized_line
> ```

we now support all lines here? or?

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4, line 120 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiNM3l_7hr5rpB_nIW:-KxiNM3l_7hr5rpB_nIX:ba7mnst) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4#L120)):*
> ```ANTLR
> s_vlans
> :
>    VLANS
> ```

I expected to see more changes to Juniper parsers to go with this change. What am I missing?

---

*[test_rigs/parsing-errors-tests/recovery-no-verbose.ref, line 13 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiMkBXGPhbnbSH74QA:-KxiMkBXGPhbnbSH74QB:b-rdkz2e) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/test_rigs/parsing-errors-tests/recovery-no-verbose.ref#L13)):*
> ```
>         "as1r1" : {
>           "Red flags" : {
>             "1" : "MISCELLANEOUS: Unrecognized Line: 4:  ip address 10.12.11.1 255.255.255.0aaaaa:wq\nLINES BELOW LINE  ip address 10.12.11.1 255.255.255.0aaaaa:wq ARE UNLIKELY TO BE PROCESSED CORRECTLY",
> ```

"are unlikely to" vs "may not be"?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624#-:-KxiNbWzxh6Wdo8-7cTc:b-lkrtnn)*

arifogel:
Fair complaint. After adding javadocs I will be in a better position to address and add more granular tests.

---

Review status: 41 of 64 files reviewed at latest revision, 27 unresolved discussions.

---

*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiEwrVxUOjwnWWmnQL:-KxiP7HU46thNR3eTyUN:b-49f9a5) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*

Drop space?
</blockquote></details>

Will address.

---

*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiF2GRcEisWHMQS73n:-KxiOsOafsvGVxbv_cY0:b-tm8iil) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*

We should version the plugin in the top-level module, but only add `antlr4` test sources in the module(s) that have them. just `batfish`?
</blockquote></details>

Sure OK. Yes. Will address.

---

*[projects/suppressions.xml, line 7 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFa5BTCCUAKuRcWlQ:-KxiOfswvomy15lmP4_F:bei1psl) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/suppressions.xml#L7)):*

Explicitly listing the `main/src` and `test/src` seems a cleaner/simpler solution. See, e.g., https://github.com/spotify/dockerfile-maven/pull/60/files#diff-600376dffeb79835ede4a0b285078036
</blockquote></details>

I tried something similar but it didn't work. There is a slight difference at your link, though, so I can try that and get back to you.

---

*[projects/batfish/pom.xml, line 113 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFx0E9P1jZeidmr9e:-KxiPE9-RbJjKaqyzGAi:b-qspj2w) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/pom.xml#L113)):*

Add `<phase>generate-test-sources</phase>`, I think [link](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html). That way the test sources will not be generated unless tests are to be run.
</blockquote></details>

Makes sense. Will address.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/CompositeBatfishException.java, line 58 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiGJalRhiGSvUsz6Ag:-KxiPSXZy45k_bvdob8y:bx0bwvq) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/common/CompositeBatfishException.java#L58)):*

Note to self: understand why this is here.

Add javadoc?
</blockquote></details>

Actually not sure if I ended using this. But it is still a useful helpful function. I will add javadoc, and that should become clear.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 336 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiGrLrjDwqT6RGD5-9:-KxiPrGI5_20wW9aTjIN:b-tk1uvq) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L336)):*

Any reason not to use [Guava `Sets`](https://github.com/google/guava/wiki/CollectionUtilitiesExplained#sets) for these? It already returns an unmodifiable *view*, lazily evaluated. You could also use `ImmutableSet.copyOf(Sets.difference(set1, set2))` if there's worry that set1/set2 might change underneath you.
</blockquote></details>

My version allows you to specify the underlying set implementation. If there is a way to do that with Guava library functions, I'm happy to switch. I will investigate.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 1 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKA9wDOcOozDBiwzs:-KxiQCI1BkJ5nVb1lUP-:befwu4o) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L1)):*

I'm not precisely sure of the "mode" in which to review this. Clearly I am missing a lot of ANTLR background.

Should I expect implementation comments in this file to give me an overview of the necessary context? Should I expect the reader to be an expert on ANTLR?
</blockquote></details>

I'll see what I can come up with in terms of documentation, but I hope not to explain the whole ANTLR recognition process in docs in this file.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 23 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiI1bbqjzIDJp30_2K:-KxiQseYJJR5BnAfTuqo:bx9exb6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L23)):*

Top-level javadoc would be helpful as to what this class does from the public API context. What types of errors can it recover and how does it do so?

Also a (non-javadoc) block comment as to how it does it would be useful for those reading the code.
</blockquote></details>

I'll see what I can do here.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 29 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiHY3h6TboAMpUwVgE:-KxiQxG1xbaPttZ3hkC1:b-bqm6h6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L29)):*

Javadoc - what are these parameters for?
</blockquote></details>

Will add.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 30 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiIAA-F6H9BnpXW7VJ:-KxiRHr-3h-9Pm_LenbG:b-10qz80) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L30)):*

Inferring from the code, it looks like the first two parameters are specific to the strategy and the third is the actual text being parsed? I'd probably separate these into two different classes / functions for clarity and reusability.

```java
F = new BatfishErrorStrategy(separatorToken, minimumRequiredSeparatorText);
E = F.forText(text);
```


</blockquote></details>

The text is used by the strategy to create `ErrorNode`s including the whole line being thrown out. Without access to the raw text, it would be complicated, error prone, and expensive to recover this text.
I'm not sure if this addresses your comment at all, but in any case I am unclear on the pattern you are suggesting.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 32 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiHJBC5Y19YpXR4iaF:-KxiRBkbQt5NOQijfN67:b-7y7oa2) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L32)):*

ImmutableList.copyOf?
</blockquote></details>

Why copy it? No one else has a reference to it, so it seems like wasted work.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 53 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJWiqykxkJCT2G_uM:-KxiS20-U2ud34KWnf54:b2a3p0e) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L53)):*

Comment seems wrong -- copy and paste of the one 5 lines up?
</blockquote></details>

You are correct. Will address.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 59 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJboilb3HiVe8L14w:-KxiS8YvdbJmDwOy2FyW:bgvs7ge) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L59)):*

?
</blockquote></details>

Are you asking for a comment? If so I will provide.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 91 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiJir4XeLGLFh_Hcip:-KxiSCdGfobdL-fJmWlC:bb361hm) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L91)):*

what does this do? Expected pre-post conditions? etc.
</blockquote></details>

Will document.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 127 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKZziSvKIwKWVe_uq:-KxiSGCHf3z_YmsLWEOT:bvc3bi5) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L127)):*

javadoc on all public functions

(Without looking elsewhere to see how it's used:) does this need to be public?
</blockquote></details>

It is used elsewhere.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java, line 73 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiKq9fRrU4UfX9vFrk:-KxiSUN2VXhGtMSgu5iG:bqxfg7p) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java#L73)):*

? - don't get this.

comment?
</blockquote></details>

If we are not doing recovery, none of the recovery infrastructure needs to be supplied (it creates overhead).

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishGrammarErrorListener.java, line 36 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL-Gx58g_vq3wE-42:-KxiSbu68Yt6TgmHedyi:b-5wvppg) ([raw file](https://github.com/batfish/batfish/blob/53015193909cd1113fa939acc8b6bda67ac752c1/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishGrammarErrorListener.java#L36)):*

also delete blank line?
</blockquote></details>

What blank line? I'll take a look in code in case the review system is hiding for me.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 15 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL6SbAPno65WP67ZC:-KxiStFsXFqwydEEHfFI:bhy9ot6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L15)):*

`@Nullable`
</blockquote></details>

Will do.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 25 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL5X2x1tetjgGc7tl:-KxiSwvz7YjDNxp3R2Q5:bhy9ot6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L25)):*

`@Nullable`
</blockquote></details>

Will do.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java, line 62 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiL7ns_rztDgOCADzH:-KxiSxgEclYal8d3lR4-:bxlfbhd) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexer.java#L62)):*

is this `@Nullable`?
</blockquote></details>

That's a philosophical question. If you set it, presumably you want it to do something. But since the field is not required to be non-null, I don't suppose there is any specific harm for calling it `Nullable` here.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java, line 12 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLmNLO7pxEOGYthto:-KxiTGzDSwmCF4uQeA3l:bhy9ot6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java#L12)):*

javadoc
</blockquote></details>

Will do.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java, line 29 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLJDpIcBounIBXx00:-KxiTKz0n5yOh86RFgGv:bk5sdjl) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishLexerRecoveryStrategy.java#L29)):*

`ImmutableSet.builder().addAll(whitespace()).addAll(newlines()).build();` ? (guessed at that syntax btw)
</blockquote></details>

Will address contingent on decision for my earlier comment about Guava set functions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserATNSimulator.java, line 8 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiLw9IrShuUuhjcPzq:-KxiTWBP7rh1t7TYg4zL:bhy9md4) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserATNSimulator.java#L8)):*

Javadoc. What does this do that's different than the default?
</blockquote></details>

Will add

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishRecognitionException.java, line 8 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiM2Rh89iJgAJOxeDH:-KxiTYZQWpwDeuF_yUto:b-vgxrax) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishRecognitionException.java#L8)):*

Looks like this adds nothing. Why?

`final class`?

Javadoc?
</blockquote></details>

It's just meant to be a general Recognition exception to be caught by the `Parser`. Unfortunately this one is hard to understand without seeing generated parser code. I will add javadoc.

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4, line 858 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiMNTPNGQ-uK0IfZ9U:-KxiUHFzQ4LiUl2zRDmX:b1upv20) ([raw file](https://github.com/batfish/batfish/blob/53015193909cd1113fa939acc8b6bda67ac752c1/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4#L858)):*

we now support all lines here? or?
</blockquote></details>

No special handling of unrecognized lines needed, as this whole PR generalizes it for all parsers of this kind.

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4, line 120 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiNM3l_7hr5rpB_nIW:-KxiURtFkxB9tXIOzoOU:bvjaan0) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4#L120)):*

I expected to see more changes to Juniper parsers to go with this change. What am I missing?
</blockquote></details>

The intentional blank was supposed to go after the `VLANS` keyword. But the only reason I would have put a blank there is if an `apply-groups` was also valid there. The `apply` rule allows a blank, or `apply-groups`, or `apply-groups-except`, so it covers all needs here.

---

*[test_rigs/parsing-errors-tests/recovery-no-verbose.ref, line 13 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiMkBXGPhbnbSH74QA:-KxiUje5e9WEILBdWdam:b-shkse3) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/test_rigs/parsing-errors-tests/recovery-no-verbose.ref#L13)):*

"are unlikely to" vs "may not be"?
</blockquote></details>

After this PR should be "may not be". I wil change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624#-:-KxiUprCZcPXeQfNv8JC:b-1dcgs2)*

corinaminer:
Review status: 41 of 64 files reviewed at latest revision, 4 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 1 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiPxGHxqdsFFbTYb5u:-KxiPxGHxqdsFFbTYb5v:bb5n0db) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L1)):*
> ```Java
> package org.batfish.grammar;
> ```

Class name has ANTRL instead of ANTLR.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 119 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiRwXSwoQq_kpm7j4t:-KxiRwXSwoQq_kpm7j4u:b-kmd50g) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L119)):*
> ```Java
>       recognizer.consume();
>       return createErrorNode(recognizer, parent, lineIndex, separator);
>     }
> ```

These if statements are opaque to me. Would definitely help to have a block comment outlining the three cases and what we want to do with the parse tree in each case.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 197 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiU9IvjbVRszXBUUuC:-KxiU9IvjbVRszXBUUuD:bopucvk) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L197)):*
> ```Java
>            */
>           throw new InputMismatchException(recognizer);
>         }
> ```

I don't understand why this block in the switch is necessary, since the block after `case ATNState.STAR_LOOP_BACK:` is identical.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

arifogel:
Review status: 41 of 64 files reviewed at latest revision, 4 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 1 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiPxGHxqdsFFbTYb5u:-KxiWBLvLM1GnEj8cYq5:b-844qnc) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L1)):*

Class name has ANTRL instead of ANTLR.
</blockquote></details>

Good catch. Will fix.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 119 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiRwXSwoQq_kpm7j4t:-KxiWEcU-AvmSR_raM4A:bvenj09) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L119)):*

These if statements are opaque to me. Would definitely help to have a block comment outlining the three cases and what we want to do with the parse tree in each case.
</blockquote></details>

Will document and add tests for each case.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java, line 197 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiU9IvjbVRszXBUUuC:-KxiWJWEOVn9oL0RBs-4:b-qdtkb6) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTRLErrorStrategy.java#L197)):*

I don't understand why this block in the switch is necessary, since the block after `case ATNState.STAR_LOOP_BACK:` is identical.
</blockquote></details>

Until the very last change, these had different content. But now that they are the same, they can be merged. Good catch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

arifogel:
Review status: 41 of 64 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish/pom.xml, line 113 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFx0E9P1jZeidmr9e:-KxjRV3WN3mHmAgv2bWm:b-c08w7y) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/pom.xml#L113)):*

Makes sense. Will address.
</blockquote></details>

So I tried it. While I think it does bind it to the `generate-test-sources` phase, you should know that `generate-test-sources` is run during e.g. `mvn install -DskipTests` anyway. That is, tests are generated and compiled, but just not run.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

arifogel:
Review status: 41 of 64 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 336 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiGrLrjDwqT6RGD5-9:-KxjXZPSp2mD5UNqogan:b9x46fp) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L336)):*

My version allows you to specify the underlying set implementation. If there is a way to do that with Guava library functions, I'm happy to switch. I will investigate.
</blockquote></details>

I think there is a case for at least removing references to all these utility functions. I will defer to a future PR, and remove references and modifications to utility functions in this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

dhalperi:
*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiEwrVxUOjwnWWmnQL:-KxnPSqLSrL4aFwHn7Ha:b9fb2x9) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*

Will address.
</blockquote></details>

Did not address.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

arifogel:
Review status: 24 of 63 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/pom.xml, line 200 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiEwrVxUOjwnWWmnQL:-KxsXGSJZoNXv8oJjaUF:b-j7j4by) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/pom.xml#L200)):*

Did not address.
</blockquote></details>

Whoops got it now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

dhalperi:
Reviewed 5 of 71 files at r1, 42 of 48 files at r2, 2 of 2 files at r3.
Review status: 62 of 63 files reviewed at latest revision, 14 unresolved discussions.

---

*[projects/batfish/pom.xml, line 113 at r1](https://reviewable.io:443/reviews/batfish/batfish/624#-KxiFx0E9P1jZeidmr9e:-KxsPkUE_6pW3AXbT8s7:b-bmd90i) ([raw file](https://github.com/batfish/batfish/blob/4f018b632c8d92381fdc5c8fa1aead2f63248f4b/projects/batfish/pom.xml#L113)):*

So I tried it. While I think it does bind it to the `generate-test-sources` phase, you should know that `generate-test-sources` is run during e.g. `mvn install -DskipTests` anyway. That is, tests are generated and compiled, but just not run.
</blockquote></details>

Yep, that's expected. There's a difference between not running tests (`-DskipTests`) and not running the test phase at all (`-Dmaven.test.skip` or `-Dtest.skip`, [I think]). The latter is almost never what you want.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 28 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsQHF4CdGyrUZay90t:-KxsQHF4CdGyrUZay90u:b-s8bita) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L28)):*
> ```Java
>  * Intended as a replacement for default ANTLR parser error recovery strategy. The basic idea here
>  * is to throw out any lines containing invalidities, and try to parse as if those lines were never
>  * there. Meanwhile, each invalid lines should show up in the parse tree as {@link ErrorNode} at an
> ```

typo: `invalid line`, not lines.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 36 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsQNoaBVDyo35PwNh4:-KxsQNoaBVDyo35PwNh5:b-hzwm19) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L36)):*
> ```Java
>   /**
>    * Construct a factory for making instances of {@link BatfishANTLRErrorStrategy} with supplied
>    * {@link separatorToken} and {@link minimRequiredSeparatorText}.
> ```

these should be `@code`, I think, as those are not valid javadoc references.

typo in `minimum`.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 99 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsR0WW8mmy6e7yayKR:-KxsR0WW8mmy6e7yayKS:b7nom4g) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L99)):*
> ```Java
>    * @param recognizer The {@link Parser} to whom to delegate creation of each {@link ErrorNode}
>    */
>   private void consumeBlocksUntilWanted(Parser recognizer) {
> ```

I'm having a hard time connecting this function's doc to its name. Is the intent to turn something like

```
mtu abcdef 1234a 1500
```

into

````
mtu 1500
```

(skipping `abcdef` and `1234a` which are invalid numbers) ?

Or is the goal to skip the entire line?

The javadoc text in terms of "tokens" makes me think it's the first one, but then text in terms of "lines" makes me think maybe the latter, but the function name "consume blocks" makes me question both.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 166 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsYrYUpW_Sccsq5kkU:-KxsYrYUpW_Sccsq5kkV:bfper44) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L166)):*
> ```Java
>    *     created @{link ErrorNode}.
>    */
>   private Token recover(Parser recognizer) {
> ```

should this be called `parser`? I ran into issues trying to find Recognizer#consume, when I realized the function was actually limited to Parsers :)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 174 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsVL5_mnPwJUFmqn4K:-KxsVL5_mnPwJUFmqn4L:bik40qi) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L174)):*
> ```Java
>     IntervalSet followSet = new IntervalSet();
>     followSet.add(_separatorToken);
>     consumeUntil(recognizer, followSet);
> ```

I think this could be clearer as just:

```java
// Consume anything left on this line.
consumeUntil(recognizer, IntervalSet.of(_separatorToken));
```

or even using a `consumeUntilEndOfLine(recognizer)` helper method. FWICT you only ever use `consumeUntil` in this way, and you do so three times in the file.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 179 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-Kxs_zOiZiN5UuiaBMt5:-KxsZA0WvdFSl1-EIPFu:b-3doq9b) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L179)):*
> ```Java
>     Token separatorToken = recognizer.getCurrentToken();
>     int lineIndex = separatorToken.getLine() - 1;
>     String separator = separatorToken.getText();
> ```

Maybe I'm missing something, but why do we need to extract the separator string from a parse token -- don't we already know the separator string?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 194 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-Kxsa4EtMb6PxpQu8PAa:-Kxsa4EtMb6PxpQu8PAb:b-w65aga) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L194)):*
> ```Java
>       recognizer.consume();
>       return createErrorNode(recognizer, parent, lineIndex, separator);
>     }
> ```

I might restructure this in a simpler way:

1. At the top of the function, or as early as possible*, check for the case that we have a parent on the same line and throw.
    (*FWICT, this can be done before any of the work above this `if`, other than setting up parent and ctx variables. I may be wrong, tho – I only recursed a few levels deep in ANTLR functions.).

2. This code can get simplified:

    ```java
    if (parent != null) {
      // Delete this node from its parent, to be replaced by the error node we are about to create.
      List<ParseTree> parentChildren = parent.children;
      parentChildren.remove(parentChildren.size() - 1);
    }
    
    // Eat the separator token, enabling the parser to resume from the next line. (? is this right?)
    recognizer.consume();
    return createError(recognizer, parent, lineIndex, separator);
    ```

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 247 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsdOrzx0PLuzT6WrJn:-KxsdOrzx0PLuzT6WrJo:bdj4yh6) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L247)):*
> ```Java
>   public void sync(Parser recognizer) throws RecognitionException {
>     /*
>      * Copied from super
> ```

Which code here is not copied from super? If it's a clean break, just add a comment where the new code is?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/TestGrammarSettings.java, line 1 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsfLDxWUBmZDmtRw8s:-KxsfLDxWUBmZDmtRw8t:b-9jrht) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/TestGrammarSettings.java#L1)):*
> ```Java
> package org.batfish.grammar;
> ```

Put this in `src/test`, not `src/main`

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/TestGrammarSettings.java, line 3 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsfQhszwZBnbvDHP7H:-KxsfQhszwZBnbvDHP7I:b5fubzl) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/TestGrammarSettings.java#L3)):*
> ```Java
> package org.batfish.grammar;
> 
> public class TestGrammarSettings implements GrammarSettings {
> ```

javadoc

---

*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 78 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxshK45_b-gWLZTiNeY:-KxshK45_b-gWLZTiNeZ:b-altuu4) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L78)):*
> ```Java
> 
>     assertThat(extractor.getNumSets(), equalTo(8));
>     assertThat(extractor.getNumErrorNodes(), equalTo(8));
> ```

Since the test is so opaque (magic numbers manually computed from text in a different file), I might add a sanity check on the total number of lines in the file, e.g.,

```
int totalSetsLines = recoveryTest.split('\n').stream().filter(s -> s.trim().startsWith('set')....count.
assertThat(error.getNumSets() + error.getNumErrorNodes(), equalTo(totalSetsLines));
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*
<!-- Sent from Reviewable.io -->

dhalperi:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

arifogel:
Review status: 59 of 63 files reviewed at latest revision, 13 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 28 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsQHF4CdGyrUZay90t:-Kxt5ak7AC0o4Hzd_Mnu:b-886gsx) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L28)):*

typo: `invalid line`, not lines.
</blockquote></details>

Fixed

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 36 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsQNoaBVDyo35PwNh4:-Kxt5dcqP8un45Jk8qoI:b-886gsx) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L36)):*

these should be `@code`, I think, as those are not valid javadoc references.

typo in `minimum`.
</blockquote></details>

Fixed

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 99 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsR0WW8mmy6e7yayKR:-Kxt5hr9Fk8Jxh9GDfQF:b-qif248) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L99)):*

I'm having a hard time connecting this function's doc to its name. Is the intent to turn something like

```
mtu abcdef 1234a 1500
```

into

````
mtu 1500
```

(skipping `abcdef` and `1234a` which are invalid numbers) ?

Or is the goal to skip the entire line?

The javadoc text in terms of "tokens" makes me think it's the first one, but then text in terms of "lines" makes me think maybe the latter, but the function name "consume blocks" makes me question both.
</blockquote></details>

Adjusted javadoc for clarity.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 166 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsYrYUpW_Sccsq5kkU:-Kxt5miIjkLk7d8OkgJN:b-7uqqfs) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L166)):*

should this be called `parser`? I ran into issues trying to find Recognizer#consume, when I realized the function was actually limited to Parsers :)
</blockquote></details>

Changed as suggested.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 174 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsVL5_mnPwJUFmqn4K:-Kxt5qAlsCYlfbmBwzlI:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L174)):*

I think this could be clearer as just:

```java
// Consume anything left on this line.
consumeUntil(recognizer, IntervalSet.of(_separatorToken));
```

or even using a `consumeUntilEndOfLine(recognizer)` helper method. FWICT you only ever use `consumeUntil` in this way, and you do so three times in the file.
</blockquote></details>

Done.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 179 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-Kxs_zOiZiN5UuiaBMt5:-Kxst3vHMaK7vKz-dvQu:bhmf0f9) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L179)):*

Maybe I'm missing something, but why do we need to extract the separator string from a parse token -- don't we already know the separator string?
</blockquote></details>

We want to have all the text of the file in the parse tree. The separator string is a MINIMAL version of the separator token, but that token could have longer instantiations. We want to get the text for the specific instantiation at this point.
For reference, the minimum for most `NEWLINE` tokens we've defined is `"\n"`, but `"\r\n"` is also `lexed` as a single NEWLINE token.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 194 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-Kxsa4EtMb6PxpQu8PAa:-Kxsz5ro0GG8HlRRz2RT:b-im08lw) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L194)):*

I might restructure this in a simpler way:

1. At the top of the function, or as early as possible*, check for the case that we have a parent on the same line and throw.
    (*FWICT, this can be done before any of the work above this `if`, other than setting up parent and ctx variables. I may be wrong, tho – I only recursed a few levels deep in ANTLR functions.).

2. This code can get simplified:

    ```java
    if (parent != null) {
      // Delete this node from its parent, to be replaced by the error node we are about to create.
      List<ParseTree> parentChildren = parent.children;
      parentChildren.remove(parentChildren.size() - 1);
    }
    
    // Eat the separator token, enabling the parser to resume from the next line. (? is this right?)
    recognizer.consume();
    return createError(recognizer, parent, lineIndex, separator);
    ```
</blockquote></details>

Bug here. In the first base case (`parent == null`), `ctx` instead of `parent` should be passed to `createError`. Since tests were not crashing with NPE, I think I need to add another one that triggers this branch. Although it is possible that the logic in `sync` prevents this case from ever occurring. Need to explore.
In any case, I still refactored to address your comment.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java, line 247 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsdOrzx0PLuzT6WrJn:-Kxt68UZV4-9IpTinG1B:bb6un1j) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishANTLRErrorStrategy.java#L247)):*

Which code here is not copied from super? If it's a clean break, just add a comment where the new code is?
</blockquote></details>

Done

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/grammar/TestGrammarSettings.java, line 1 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsfLDxWUBmZDmtRw8s:-Kxt6DRmLROB_3xbi9it:bb6un1j) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/test/java/org/batfish/grammar/TestGrammarSettings.java#L1)):*

Put this in `src/test`, not `src/main`
</blockquote></details>

Done

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/grammar/TestGrammarSettings.java, line 3 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxsfQhszwZBnbvDHP7H:-Kxt6EJ9d0-7EZwAKAj4:b-we7dfl) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish-common-protocol/src/test/java/org/batfish/grammar/TestGrammarSettings.java#L3)):*

javadoc
</blockquote></details>

Done, and also added javadocs to `GrammarSettings`

---

*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 78 at r3](https://reviewable.io:443/reviews/batfish/batfish/624#-KxshK45_b-gWLZTiNeY:-Kxt6LUn1hmEEwtxI1Jl:bgfgdig) ([raw file](https://github.com/batfish/batfish/blob/114aa796ad4c3228c3145bfcbbfe0509a5b0f007/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L78)):*

Since the test is so opaque (magic numbers manually computed from text in a different file), I might add a sanity check on the total number of lines in the file, e.g.,

```
int totalSetsLines = recoveryTest.split('\n').stream().filter(s -> s.trim().startsWith('set')....count.
assertThat(error.getNumSets() + error.getNumErrorNodes(), equalTo(totalSetsLines));
```
</blockquote></details>

Addressed kind of. The number of lines referenced in the parse tree is only guaranteed to be less than or equal to the number of lines in the input text.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*
<!-- Sent from Reviewable.io -->

dhalperi:
Reviewed 2 of 49 files at r2, 1 of 1 files at r4, 5 of 5 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624)*

dhalperi:
:lgtm: At this point only unresolved comment is your own as to whether you're going to add a test to catch the potential NPE issue with recovery at the top level.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/624#-:-Kxten7JtssJNkNAitJ3:b-uz31qk)*

PR #628, opened 2017-10-31
removed advertisement set

ratulm:
Third installment of #559


arifogel:
I resolved conflicts and relaxed some declared types to `Set` from `LinkedHashSet`. Otherwise LGTM.

PR #636, opened 2017-11-01
support asa ospf network command that uses mask instead of wildcard

### LINKED ISSUE #470: Possible wildcard mask/netmask issue when parsing Cisco ASA configs
Hi batfish folks,
                        I've got a topology in GNS3 where I am configuring OSPF on Cisco CSR (v16.3) routers and ASA (v9.7) firewalls. With the CSR routers, the "network" command is as follows:

`network <network> <wildcard-mask> area <area-id>`

such that a subnet mask of 255.255.255.0 is represented as 0.0.0.255. Hence, a command to configure OSPF for 192.168.1.0/24 would be:

`network 192.168.1.0 0.0.0.255 area 0`

With the ASAs, the command to configure the same network above is:

`network 192.168.1.0 255.255.255.0 area 0 `

such that the "netmask" is used as opposed to the wildcard-mask. 

When I run the ASA configs through batfish using the netmask, the "ospfEnabled" parameter is set to "false": below is such a configuration which results exhibits this behavior:
[gns3-correct_batfish-ospfEnabled-false.txt](https://github.com/batfish/batfish/files/1331169/gns3-correct_batfish-ospfEnabled-false.txt)

Alternatively, when I use the wild-card mask for the ASAs (which is an invalid command for the ASAs), "ospfEnabled" is set to "true": below is such a configuration:
 
[gns3-wrong_batfish-ospfEnabled-true.txt](https://github.com/batfish/batfish/files/1331172/gns3-wrong_batfish-ospfEnabled-true.txt)

Can you kindly provide feedback?

PS. Adding @jkhourybbn and @dspicuzzbbn to the thread.

Thanks,
Ali

arifogel:
Fixes #470

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/636)


dhalperi:
:lgtm:

---

Reviewed 12 of 12 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/636#-:-Kxp2Zwi6j0GVTdcShXo:bnfp4nl)*

PR #640, opened 2017-11-03
removed outlier detection for 'no node should contain a Foo named Bar' hypotheses

millstein:
We don't check any more for outliers where a small number of nodes contain a structure of a given name, while many other nodes do not contain such a structure.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/640)


dhalperi:
:lgtm:

---

Reviewed 1 of 9 files at r2, 1 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/640#-:-KyHcq3L7lttelHwp1eY:bnfp4nl)*

PR #665, opened 2017-11-11
Removed freezing for RIBs

### LINKED ISSUE #646: Replace RIB freeze with memoization pattern
Follow up on the discussion from #634 and move away from explicit RIB freezing in favor of transparent/cached route collection

progwriter:
Now RIBs keep an up-to-date collection of all routes. Fixes #646

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/665)


projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java
```diff
@@ -430,18 +440,7 @@ final boolean containsRoute(R route) {
...
-    return _tree.getRoutes();
-  }
-
-  /**
-   * Freeze the RIB. Prevents addition (merging) of new routes. Also computes and caches the set of
-   * all routes for quick subsequent access.
-   */
-  public void freeze() {
-    _finalRoutes = Collections.unmodifiableSet(getRoutes());
+    return _allRoutes;
```
dhalperi: What is the expected semantics of the return value here? 

Questions:

1. Should clients be able to modify the return value?
2. Should clients see updated values if this AbstractRib is modified later?

progwriter: 1. They should not.
2. Good question. I believe the answer here is also no since we use getRoutes to iterate over a particular snapshot of the RIB.

dhalperi:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java, line 443 at r1](https://reviewable.io:443/reviews/batfish/batfish/665#-KycPpzGBsasSEqE2BYS-r1-443:-KyqRKlTox7zc1PCm0ik:baozix8) ([raw file](https://github.com/batfish/batfish/blob/cb944d5e752ebfab48fc514d59a72e29220cb830/projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java#L443)):*

1. They should not.
2. Good question. I believe the answer here is also no since we use getRoutes to iterate over a particular snapshot of the RIB.
</blockquote></details>

So it sounds like we either need to return a `Collections.unmodifiableList` or an `ImmutableList`.

Or figure out the combination of memoizing + access pattern that makes this work out?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/665)*

projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java
```diff
@@ -396,12 +404,12 @@ private boolean hasSameRoutes(@Nullable AbstractRib<?>.RibTreeNode other) {
...
   private RibTree _tree;
 
-  @VisibleForTesting Set<R> _finalRoutes;
+  private Set<R> _allRoutes;
 
   public AbstractRib(VirtualRouter owner) {
     _tree = new RibTree();
     _owner = owner;
-    _finalRoutes = null;
+    _allRoutes = new TreeSet<>();
```
dhalperi: why not `ImmutableSet.of();`?

dhalperi:
Minor comments, but :lgtm: 

---

Reviewed 3 of 4 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java, line 213 at r2](https://reviewable.io:443/reviews/batfish/batfish/665#-KyxWXZN4jjL1yD9V_qz:-KyxWXZN4jjL1yD9V_r-:b-3kdj29) ([raw file](https://github.com/batfish/batfish/blob/6c630a50c25957b1a7b75c47c79d478038f33825/projects/batfish/src/main/java/org/batfish/bdp/AbstractRib.java#L213)):*
> ```Java
> 
>     /**
>      * Add the route to the set of routes at this node
> ```

I feel like this function adds questionable value over the 5 places it was used, but okay :)

---

*[projects/batfish/src/test/java/org/batfish/bdp/AbstractRibTest.java, line 212 at r2](https://reviewable.io:443/reviews/batfish/batfish/665#-KyxWng7TqWFHqHv_LE2:-KyxWng7TqWFHqHv_LE3:b-leks72) ([raw file](https://github.com/batfish/batfish/blob/6c630a50c25957b1a7b75c47c79d478038f33825/projects/batfish/src/test/java/org/batfish/bdp/AbstractRibTest.java#L212)):*
> ```Java
>     // Add completely new route and check that the size increases
>     rib.mergeRoute(new OspfIntraAreaRoute(new Prefix("2.2.2.2/32"), null, 100, 30, 1));
>     assertThat(rib.getRoutes(), hasSize(2));
> ```

We should test the invariants we agreed upon in the PR discussion.

1. if I get routes, I can't modify the result.
2. if I get routes, then modify the RIB, the routes I previously got did not change.

and probably also

3. if I get routes, then don't modify and get again, I should get actually the same object. (`assertThat(rib.getRoutes(), is(prev))` for example).

(each in its own unit test).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/665#-:-KyxXP8JRgIzNE8JMGQJ:b-gi9nc3)*

dhalperi:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/665)*

dhalperi:
I played with this last night and it seemed to shave something like a minute (out of 6.5 mins) off of building a real data plane. Nice.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/665#-:-Kz01omd5Wick8xsdCna:b5brk2c)*

dhalperi:
Reviewed 4 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/665)*

PR #668, opened 2017-11-14
Add testrig metadata file with timestamp

corinaminer:
The method `WorkMgr.initTestrig` now creates a metadata.json file in the testrig's base folder. That file looks like:
```
{
	"timestamp": 1510696220687
}
```
where 1510696220687 is the time upon testrig initialization (in milliseconds since 1970).

Let me know if I should make changes to better match existing conventions (e.g. how I'm getting the timestamp, how and when I'm creating the file, where and how I'm storing it).

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/668)


ratulm:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/668)*

ratulm:
Looks good. One minor comment: "timestamp" is perhaps too generic as we may add other types of timestamps later. Make it creationtimestamp or uploadtimestamp?

dhalperi:
For (new) timestamps in web APIs, we should use the web standards. Use
RFC3339 timestamps./

PR #675, opened 2017-11-18
Add option to truncate parse tree print

sfraint:
Integrating [Dustin's change](https://github.com/batfish/batfish/pull/510)

Added option to specify a maximum number of characters to display for parse tree pretty print (defaults to 0, meaning no maximum) and unit test for new printWithCharacterLimit function

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/675)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserErrorListener.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserErrorListener.java
@@ -72,6 +72,10 @@ public class BatfishParserErrorListener extends BatfishGrammarErrorListener {
     sb.append("Offending Token: " + offendingTokenText + "\n");
     sb.append("Error parsing top (leftmost) parser rule in stack: '" + ruleStack + "'.\n");
     String ctxParseTree = ParseTreePrettyPrinter.print(ctx, _combinedParser);
+    // don't overflow a reasonable terminal if there's a runaway rule
+    if (ctxParseTree.length() > 10000) {
+      ctxParseTree = ctxParseTree.substring(0, 10000) + "\n... and lots more";
+    }
     sb.append("Parse tree of current rule:\n" + ctxParseTree + "\n");
     sb.append("Unconsumed tokens:\n");
     int endTokenIndex = tokens.size();
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserErrorListener.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishParserErrorListener.java
@@ -72,11 +72,10 @@ public class BatfishParserErrorListener extends BatfishGrammarErrorListener {
     sb.append("Offending Token: " + offendingTokenText + "\n");
     sb.append("Error parsing top (leftmost) parser rule in stack: '" + ruleStack + "'.\n");
     String ctxParseTree = ParseTreePrettyPrinter.print(ctx, _combinedParser);
-    // don't overflow a reasonable terminal if there's a runaway rule
-    if (ctxParseTree.length() > 10000) {
-      ctxParseTree = ctxParseTree.substring(0, 10000) + "\n... and lots more";
-    }
-    sb.append("Parse tree of current rule:\n" + ctxParseTree + "\n");
+
+    sb.append("Parse tree of current rule:\n");
+    sb.append(ctxParseTree);
+    sb.append("\n");
     sb.append("Unconsumed tokens:\n");
     int endTokenIndex = tokens.size();
     for (int i = startTokenIndex; i < endTokenIndex; i++) {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/GrammarSettings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/GrammarSettings.java
@@ -28,6 +28,13 @@ public interface GrammarSettings {
   int getMaxParserContextTokens();
 
   /**
+   * The maximum number of characters display of a parse tree when prettyPrinting.
+   *
+   * @return The max number of characters to display
+   */
+  int getMaxParseTreePrintLength();
+
+  /**
    * Controls whether parse trees are stored in parse job results.
    *
    * @return true iff parse trees should be stored in parse job results
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java
@@ -2,6 +2,7 @@ package org.batfish.grammar;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.ListIterator;
 import org.antlr.v4.runtime.Lexer;
 import org.antlr.v4.runtime.Parser;
 import org.antlr.v4.runtime.ParserRuleContext;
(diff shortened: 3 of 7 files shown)


arifogel:
Mostly good. Consider applying recommended change.

---

Reviewed 1 of 5 files at r2, 6 of 6 files at r3.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java, line 46 at r3](https://reviewable.io:443/reviews/batfish/batfish/675#-KzBsNNTVc9NWxrbIfZK:-KzBsNNUQOOGwi5fMig3:bg5hbqe) ([raw file](https://github.com/batfish/batfish/blob/93c79234d95a41f681a86ba63ee2983f525a8c01/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java#L46)):*
> ```Java
>   }
> 
>   public static String printWithCharacterLimit(List<String> strings, int maxStringLength) {
> ```

This algorithm uses complicated control flow. Perhaps some version of the following could make this easier to follow (for some definition of easier):
```
    AtomicInteger currentSum = new AtomicInteger();
    List<Integer> runningSums = strings.stream().map(s -> currentSum.addAndGet(s.length())).collect(Collectors.toList());
    int lastAllowedElement = Math.abs(Arrays.binarySearch(runningSums.toArray(new Integer[]{}), maxStringLength));
    String output = String.join("\n",strings.subList(0, lastAllowedElement));
    int numOmittedLines = strings.size() - lastAllowedElement;
   // add message about omitted lines
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/675#-:-KzBxNJiA3HTY1kMIe-B:btwdbay)*

dhalperi:
Reviewed 1 of 5 files at r2, 2 of 6 files at r3.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java, line 46 at r3](https://reviewable.io:443/reviews/batfish/batfish/675#-KzBsNNTVc9NWxrbIfZK:-KzPlmYk1VvkuNsG3pYl:bc44t3i) ([raw file](https://github.com/batfish/batfish/blob/93c79234d95a41f681a86ba63ee2983f525a8c01/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java#L46)):*

This algorithm uses complicated control flow. Perhaps some version of the following could make this easier to follow (for some definition of easier):
```
    AtomicInteger currentSum = new AtomicInteger();
    List<Integer> runningSums = strings.stream().map(s -> currentSum.addAndGet(s.length())).collect(Collectors.toList());
    int lastAllowedElement = Math.abs(Arrays.binarySearch(runningSums.toArray(new Integer[]{}), maxStringLength));
    String output = String.join("\n",strings.subList(0, lastAllowedElement));
    int numOmittedLines = strings.size() - lastAllowedElement;
   // add message about omitted lines
```

</blockquote></details>

Ari's right that this code seems complex.

IMO something fairly close to Spencer's current approach is cleaner, as there oughtn't need to be a pre-step to count the size of all lines.

Some abstract recommendations I'd make:

- Focus on keeping the code simple and understandable over meeting the character count exactly. A little slop is okay.
- Use [`List#listIterator()`](https://docs.oracle.com/javase/7/docs/api/java/util/List.html#listIterator%28%29) to simplify index tracking.
- Simplify checking if the string is too long with something like this: `if (maxStringCount <= 0) { maxStringCount = Integer.MAX_VALUE; }`. I'm not worried about the practical differences between 2 GiB and infinity.
- I bet a `while` loop would naturally simplify the code here, because you could check whether you stopped before the last string after the for loop simply and can include the breaking condition as part of the loop condition.

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/grammar/ParseTreePrettyPrintTest.java, line 11 at r3](https://reviewable.io:443/reviews/batfish/batfish/675#-KzPnyoN2svOtmfnGaM8:-KzPnyoN2svOtmfnGaM9:b-tspdj3) ([raw file](https://github.com/batfish/batfish/blob/93c79234d95a41f681a86ba63ee2983f525a8c01/projects/batfish-common-protocol/src/test/java/org/batfish/grammar/ParseTreePrettyPrintTest.java#L11)):*
> ```Java
> import org.junit.Test;
> 
> public class ParseTreePrettyPrintTest {
> ```

make the name of the test class literally `ClassBeingTestedTest`. In this case, `ParseTreePrettyPrinter` is the class being tested, I think.

---

*[projects/batfish/src/main/java/org/batfish/config/Settings.java, line 1354 at r3](https://reviewable.io:443/reviews/batfish/batfish/675#-KzPoAJ8ogZAAgj33nQU:-KzPoAJ8ogZAAgj33nQV:b-c330vv) ([raw file](https://github.com/batfish/batfish/blob/93c79234d95a41f681a86ba63ee2983f525a8c01/projects/batfish/src/main/java/org/batfish/config/Settings.java#L1354)):*
> ```Java
>     addOption(
>         ARG_MAX_PARSE_TREE_PRINT_LENGTH,
>         "max number of characters to print for parsetree pretty print",
> ```

comment that any number <= 0 is treated as no limit?x

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/675)*

dhalperi:
Reviewed 2 of 7 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java, line 47 at r4](https://reviewable.io:443/reviews/batfish/batfish/675#-KzQaDOyNpXhdLhNsJ-G:-KzQaDOyNpXhdLhNsJ-H:b-nl78kn) ([raw file](https://github.com/batfish/batfish/blob/29be25e1bc8994758655b941923b219f4f9a2abb/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java#L47)):*
> ```Java
>   }
> 
>   public static String printWithCharacterLimit(List<String> strings, int maxStringLength) {
> ```

```java
// visible for testing
static printWithCharacterLimit...
```

This function does not need to be exposed to users, I think, so this will remove it from `public` visibility to default (package-private) visibility.

The comment explains why it's not private.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java, line 49 at r4](https://reviewable.io:443/reviews/batfish/batfish/675#-KzQaz7Q8z50HiO6Celh:-KzQaz7Q8z50HiO6Celi:b-56lupi) ([raw file](https://github.com/batfish/batfish/blob/29be25e1bc8994758655b941923b219f4f9a2abb/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java#L49)):*
> ```Java
>   public static String printWithCharacterLimit(List<String> strings, int maxStringLength) {
>     StringBuilder sb = new StringBuilder();
>     // Handle base case by modifying maxStringLength
> ```

In general, comments that tell you what 1 line of code does are not helpful if the code is readable. The fact that the code below modifies `maxStringLength` is ... somewhat obvious to a reader I hope?

I might have done this:

```java
if (maxStringLength <= 0) {
   // A limit <= 0 is treated as effectively no limit.
   maxStringLength = Integer.MAX_VALUE;
}
```

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java, line 55 at r4](https://reviewable.io:443/reviews/batfish/batfish/675#-KzQcfLgCZNsC9_5N3NV:-KzQcfLgCZNsC9_5N3NW:b-i9j6n9) ([raw file](https://github.com/batfish/batfish/blob/29be25e1bc8994758655b941923b219f4f9a2abb/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/ParseTreePrettyPrinter.java#L55)):*
> ```Java
> 
>     ListIterator<String> iter = strings.listIterator();
>     // Keep going while there are more strings and we have not exceeded the max length
> ```

similar remark about the utility of this comment. Just drop it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/675)*

dhalperi:
:lgtm:

To wrap up, can you please squash all your changes together on top of Dustin's original change so that the branch only has 2 commits? That way we can preserve authorship without polluting history.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/675#-:-KzQcx2KROX6mlhDZ-bu:b-rxp7mn)*

dhalperi:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/675#-:-KzQnjqDBglAa-RnxjuV:bnfp4nl)*

PR #678, opened 2017-11-21
Infrastructure for container-wide adhoc questions

corinaminer:
__Definitely should be reviewed commit by commit.__ Can separate each API call into its own PR if that's appropriate; currently they're separated by commit.

Goal: Reformulate adhoc questions as container-level rather than testrig-level.

Introduces API calls that work with container-level `explore` and `compare` directories. The container-level directories have container-wide adhoc questions (differential in `compare`, non-differential in `explore`), while each testrig folder contains its own `explore` and `compare` directories to hold those questions' answers, similar to the current setup for analyses. Example file structure inside container:
image

In time we should move towards using only the new API calls for adhoc questions. For now all the old API calls are preserved. Batfish should not need to be changed again in order to start using the new ones.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/678)

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
@@ -135,6 +135,78 @@ public class BfCoordWorkHelper {
     }
   }
 
+  public boolean configureCompare(
+      String containerName,
+      @Nullable String addQuestionsFileName,
+      @Nullable String delQuestionsStr) {
+    try {
+      WebTarget webTarget = getTarget(CoordConsts.SVC_RSC_CONFIGURE_COMPARE);
+
+      MultiPart multiPart = new MultiPart();
+      multiPart.setMediaType(MediaType.MULTIPART_FORM_DATA_TYPE);
+
+      addTextMultiPart(multiPart, CoordConsts.SVC_KEY_API_KEY, _settings.getApiKey());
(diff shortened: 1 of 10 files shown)


ratulm:
@corinaminer, is the design intent documented someplace? I find it hard to judge new functionality without it. Some of this looks counter-intuitive to me otherwise. E.g., why would you need "configureExplore" functionality? I expected the set of questions inside explore to be "organically" populated as users interact with the system. (Its easier to validate the design intent first and then validate if the code correctly implements it. Otherwise, I have to infer intent from the code, which is erroneous.)

dhalperi:
Reviewed 10 of 10 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/batfish/batfish/678#-L-4N0mFWig1jbnj9AUW:-L-4N0mGR0eusk6xy9v1:b332oop) (no related file):*
My belief is that based on follow-up design discussions, this approach will be changed in favor of one that uses existing APIs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/678)*

corinaminer:
[Link to design doc](https://docs.google.com/document/d/1FeWmViKuCz2wnLT4S9pHZfx7gD8ofnNftGkmlquq41U/edit)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/678#-:-L-4WR2zmndDYxoYdr8b:b4vfqle)*

corinaminer:
Replaced by #690.

PR #679, opened 2017-11-21
global outlier ranking

millstein:
A new question (perRoleOutliers) that allows us to globally rank the results of per-role outlier detection across roles.  Currently the z statistic is used for this purpose, but it can be easily changed.  I also refactored and simplified the representation of outlier results.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/679)


dhalperi:
Reviewed 5 of 8 files at r1.
Review status: 5 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java, line 58 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-Kzzami-bKV_as9s91hn:-Kzzami-bKV_as9s91ho:bsvwuxd) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java#L58)):*
> ```Java
>   @Override
>   // sort in reverse order of zScore, which is a measure of how likely it is that
>   // our hypothesis is correct
> ```

update comment, typically put above `@Override` annotation instead of below.,

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java, line 62 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-KzzajVV5HRoJZPJNsXz:-KzzajVV5HRoJZPJNsY-:b-by03ic) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java#L62)):*
> ```Java
>     int oScore = Double.compare(other.outlierScore(), this.outlierScore());
>     if (oScore != 0) {
>       return oScore;
> ```

should this be called `oComp` to match the others, or `oScoreComp`?

---

*[projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java, line 26 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-KzzbwC5BLo5CQVaIlPE:-KzzbwC5BLo5CQVaIlPF:bfr4qwu) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java#L26)):*
> ```Java
> 
> @AutoService(Plugin.class)
> public class PerRoleOutliersQuestionPlugin extends QuestionPlugin {
> ```

High-level: how does this differ from per-role with outliers as the inner question? Just thinking about what those means semantically, not really looking at the code yet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/679)*

dhalperi:
Reviewed 2 of 8 files at r1.
Review status: 7 of 8 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/679)*

millstein:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java, line 58 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-Kzzami-bKV_as9s91hn:-L--hYIlyIiUdxUlT9P-:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java#L58)):*

update comment, typically put above `@Override` annotation instead of below.,
</blockquote></details>

Done.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java, line 62 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-KzzajVV5HRoJZPJNsXz:-L--hZbZEoxMmPUQ5l3l:b-n5hiez) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureOutlierSet.java#L62)):*

should this be called `oComp` to match the others, or `oScoreComp`?
</blockquote></details>

I changed it to scoreComp.  

---

*[projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java, line 26 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-KzzbwC5BLo5CQVaIlPE:-L--hm8RyaI5Utg-LUMs:b-ms9cnt) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java#L26)):*

High-level: how does this differ from per-role with outliers as the inner question? Just thinking about what those means semantically, not really looking at the code yet.
</blockquote></details>

The per-role question simply runs the inner question once per role and prints the results.  So what that will do for outlier detection as the inner question is, for each role, identify and rank outliers, and then print each of these sets of ranked outliers (in alphabetical order of roles).  

From recent discussions and experiments it seems that doing a global ranking of outliers, across all roles, will provide a better user experience, highlighting the most likely errors first.  That's what this question does.  Another way to do that would be to add a per-role option directly to the Outliers question, but it seemed cleaner to me to make it a separate question, especially given that this question calls per-role with an inner question of Outliers.  Possibly the generic per-role question idea should be revisited, however, and instead this functionality should simply be baked into the Outliers question.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/679)*

dhalperi:
:lgtm:

---

Reviewed 1 of 8 files at r1, 1 of 1 files at r2.

---

*[projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java, line 26 at r1](https://reviewable.io:443/reviews/batfish/batfish/679#-KzzbwC5BLo5CQVaIlPE:-L-4Cn2EO8LJ1S5ID-8s:bb6mt9b) ([raw file](https://github.com/batfish/batfish/blob/23e8cdfc315c518ffa23a289b0880b3db265d5c7/projects/question/src/main/java/org/batfish/question/PerRoleOutliersQuestionPlugin.java#L26)):*
<details><summary><i>Previously, millstein wrote…</i></summary><blockquote>

The per-role question simply runs the inner question once per role and prints the results.  So what that will do for outlier detection as the inner question is, for each role, identify and rank outliers, and then print each of these sets of ranked outliers (in alphabetical order of roles).  

From recent discussions and experiments it seems that doing a global ranking of outliers, across all roles, will provide a better user experience, highlighting the most likely errors first.  That's what this question does.  Another way to do that would be to add a per-role option directly to the Outliers question, but it seemed cleaner to me to make it a separate question, especially given that this question calls per-role with an inner question of Outliers.  Possibly the generic per-role question idea should be revisited, however, and instead this functionality should simply be baked into the Outliers question.


</blockquote></details>

:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/679#-:-L-4CppetUXUs-IV7iPV:bnfp4nl)*

PR #720, opened 2017-12-08
better status for analysis

ratulm:
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/720)

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java
@@ -499,9 +499,10 @@ public class WorkMgrService {
         if (work != null) {
           BatfishObjectMapper mapper = new BatfishObjectMapper();
           String taskStr = mapper.writeValueAsString(work.getLastTaskCheckResult());
-          response.put(CoordConsts.SVC_KEY_WORKID, work.getWorkItem().getId())
-                  .put(CoordConsts.SVC_KEY_WORKSTATUS, work.getStatus().toString())
-                  .put(CoordConsts.SVC_KEY_TASKSTATUS, taskStr);
+          response
+              .put(CoordConsts.SVC_KEY_WORKID, work.getWorkItem().getId())
+              .put(CoordConsts.SVC_KEY_WORKSTATUS, work.getStatus().toString())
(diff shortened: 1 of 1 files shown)


corinaminer:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/720#-:-L-rI76vlkzH_4-MyOUT:bnfp4nl)*

ratulm:
I realized there is a problem with this approach. If some of the questions in the analysis have been answered and not others, those answered are also not returned (as was happening previously). The new commit fixes that. 

@corinaminer please take another look.

corinaminer:
If an analysis is running with no answers yet, this version of `getAnalysisAnswers` will return both the `WorkItem` information and a `NOTFOUND` answer, right? I was thinking it wouldn't bother returning the answer part in that case, but this works too.

---

Reviewed 1 of 1 files at r2.

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java, line 500 at r2](https://reviewable.io:443/reviews/batfish/batfish/720#-L-sAMJSSx8Xvxfn2cC2:-L-sAMJSSx8Xvxfn2cC3:b-imb9op) ([raw file](https://github.com/batfish/batfish/blob/a7a8f85948465ced2a2a49c63f57f3a02ac201d2/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgrService.java#L500)):*
> ```Java
>         QueuedWork work = Main.getWorkMgr().getMatchingWork(workItem, QueueType.INCOMPLETE);
>         if (work != null) {
>           BatfishObjectMapper mapper = new BatfishObjectMapper();
> ```

Could we define this mapper up where we define `response`, since another mapper is needed to serialize `answersStr`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/720#-:-L-sBFH6O9yLQGVXrSXm:be6uw46)*

corinaminer:
:lgtm:
I can't think of any reason why sending back a `NOTFOUND` answer along with `WorkItem` data would be a problem, so if you can't either, this seems fine.

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/720#-:-L-sVaiqXWsGSMb0DoC0:b-t7d2bf)*

PR #721, opened 2017-12-08
NodeRoleSpecifier: add a missing deep copy

dhalperi:
The act of pretty printing the answer would change the role map from
the user's input. This seem like undesired behavior.

Note that the failing test aimed to test the case when roles were provided
and role inference was off. But the ref did not match the input roles
(and now does).

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/721)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NodeRoleSpecifier.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NodeRoleSpecifier.java
@@ -3,6 +3,7 @@ package org.batfish.datamodel;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.SortedSet;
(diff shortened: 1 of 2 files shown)


dhalperi:
Hey @millstein , I'm not sure this is right but I can't quite figure out the logic in the roles question.

The main symptom I was running into was that pretty printing the answer or not actually changed the structured answer, which seemed wrong. But this may not be the right fix.

millstein:
I don't see how pretty printing was changing _roleMap.  According to Java's documentation putAll exactly behaves like doing a put for each element of the given collection, so your change seems equivalent semantically.  Either way, the underlying _roleMap should be unaffected.

https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#putAll(java.util.Map)

It would help to see exactly the failing test and how it was run.  Happy to chat as well.  Thanks.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/721#-:-L-qtgM5n6JiHeV2ZmMh:b-sj9yq)*

dhalperi:
To be more precise, `roleMap` was not changing, but the contents of `roleMap` -- the `Set`s in the values -- were the same actual `Set` objects in both `roleMap` and `nodesRoleMap`. So `roleMap.get("foo")` is the same object as `nodesRoleMap.get("foo")` and thus `nodesRoleMap.get("foo").put("bar")` would change `_roleMap.get("foo")`.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/721#-:-L-r6xS60e84ovq95ysd:bb9w7z7)*

dhalperi:
We converged on agreeing this is a good change, restoring intended behavior.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/721#-:-L-rDQjNNDQrIeBZyd0F:b-wtkr29)*

dhalperi:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/721)*

PR #747, opened 2017-12-12
mark AWS instances as hosts

### LINKED ISSUE #733: AWS: create hosts for instances for which host.json is not supplied
(The following description is more observational than from code, so may be slightly wrong.)

- When we parse AWS configs, we create nodes for all the instances we see in the configs.
- These hosts do not appear in the output env_topology, so the default type inference logic assigns a `SWITCH` device type to them.
- We should create some minimal config with whatever host-specific info we know, at least so that device type gets set to `HOST`/`SERVER`.

ratulm:
Fixes #733

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/747)


dhalperi:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/747#-:-L0C7p-9gvIuRGQKbFlY:bnfp4nl)*

PR #748, opened 2017-12-13
fix route map continue serialization

arifogel:
- route-maps that have continue statements have their clauses converted to `RoutingPoilcy` instead of `If` to prevent exponential blow-up
- actual policy matching behavior untested and likely buggy
- nevertheless, dumping non-summary `nodes` question for node with  large route-map with continue statements(s) now succeeds

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/748)


dhalperi:
*[a discussion](https://reviewable.io:443/reviews/batfish/batfish/748#-L0CHYQiKRxyNlD7tHq6:-L0CHYQiKRxyNlD7tHq7:b4at3su) (no related file):*
Let's make a simple test case with a deep set of continuations, numbered referneces, and non undefined other structures that we can look at.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/748)*

dhalperi:
Inspect of output refs is sensible, if troubling. (`set community no-export additive` doesn't seem to have any effect? also we used a prefix-list that doesn't exist, so that also has no effect).

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/748#-:-L0CGjSWdYlKUryohbiX:b-ay5j0a)*

dhalperi:
*[a discussion](https://reviewable.io:443/reviews/batfish/batfish/748#-L0CHYQiKRxyNlD7tHq6:-L0CPfclL8tOkf5nUxEk:b-896fix) (no related file):*

Let's make a simple test case with a deep set of continuations, numbered referneces, and non undefined other structures that we can look at.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/748)*

dhalperi:
:lgtm:

---

Reviewed 4 of 4 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/748#-:-L0CPf-WsYdiP6-63CGa:bnfp4nl)*

PR #786, opened 2017-12-19
adding some logs for insights

haverma:
Starting with some logs to capture information about

 1. A new container, its API key
 2. Testrigs associated with a container
 3. List of analyses
 4. Question associated with analyses
 5. Parameters for analysis questions
 6. Computation time per analysis question
 7. Analysis Questions needing data plane

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/786)


haverma:
Closing to use a better way to check when question parsing has failed, rather than checking AnswerStatus

PR #793, opened 2017-12-21
save environment topology; retrieved saved topology; for init-environ…

### LINKED ISSUE #724: Save environment-specific topologies
Right now, topology is computed and stored at the testrig level.

Then for environments, the topology is computed on-demand from the testrig-level topology and the environment-specific changes to the environment (blacklists, etc) but never stored.

FR: If we saved these environment-specific topologies with the environment, then we wouldn't need to recompute them and they'd be accessible to external clients as well.

ratulm:
…ment, hooks into validate environment

Fixes #724 

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/793)

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -1602,18 +1602,10 @@ public class Client extends AbstractClient implements IClient {
     _logger.infof("to %s->%s\n", _currDeltaTestrig, _currDeltaEnv);
     _logger.output("\n");
 
-    WorkItem wItemGenDdp =
-        WorkItemBuilder.getWorkItemCompileDeltaEnvironment(
-            _currContainerName, _currDeltaTestrig, _currEnv, _currDeltaEnv);
-    if (!execute(wItemGenDdp, outWriter)) {
-      return false;
-    }
-
-    WorkItem wItemValidateEnvironment =
-        WorkItemBuilder.getWorkItemValidateEnvironment(
+    WorkItem wItemProcessEnv =
+        WorkItemBuilder.getWorkItemProcessEnvironment(
             _currContainerName, _currDeltaTestrig, _currDeltaEnv);
-
-    if (!execute(wItemValidateEnvironment, outWriter)) {
+    if (!execute(wItemProcessEnv, outWriter)) {
       return false;
     }
 
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
@@ -50,6 +50,8 @@ public interface IBatfish extends IPluginConsumer {
 
   void checkEnvironmentExists();
 
+  Topology computeEnvironmentTopology(Map<String, Configuration> configurations);
+
   Set<NodeInterfacePair> computeFlowSinks(
       Map<String, Configuration> configurations, boolean differentialContext, Topology topology);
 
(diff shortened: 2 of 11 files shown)


ratulm:
This became a bit of a rabit hole because of the lack of discipline about environments that are created by the user. Opened two new issues #794, #795 to handle leftover work.

dhalperi:
Reviewed 11 of 11 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java, line 72 at r1](https://reviewable.io:443/reviews/batfish/batfish/793#-L0uMa0pmhThSgWVeuCC:-L0uMa0pmhThSgWVeuCD:b4lgh51) ([raw file](https://github.com/batfish/batfish/blob/852a31b49ee6d089b3382ac6fd85a18b9a13c724/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java#L72)):*
> ```Java
>    * topology object.
>    */
>   public Topology pruneTopology(
> ```

Given that it destroys the input Topology, make make it a `void` function and rename to `prune`? That would make the behavior more clear.

Alternately, make `pruneTopology` private and then create the new topology from `getEdges` at the beginning of the function, and then `prune` that one before returning. Then you don't need to destroy the input.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/793)*

ratulm:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java, line 72 at r1](https://reviewable.io:443/reviews/batfish/batfish/793#-L0uMa0pmhThSgWVeuCC:-L0ucpRsJ4rClybztiMr:b2zoe07) ([raw file](https://github.com/batfish/batfish/blob/852a31b49ee6d089b3382ac6fd85a18b9a13c724/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java#L72)):*

Given that it destroys the input Topology, make make it a `void` function and rename to `prune`? That would make the behavior more clear.

Alternately, make `pruneTopology` private and then create the new topology from `getEdges` at the beginning of the function, and then `prune` that one before returning. Then you don't need to destroy the input.
</blockquote></details>

i considered both of those ideas but didn't go with them to save overhead. but better safe than sorry. went with the first one now; this stuff isn't called too often. related cleanup included for safe operations.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/793)*

dhalperi:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java, line 72 at r1](https://reviewable.io:443/reviews/batfish/batfish/793#-L0uMa0pmhThSgWVeuCC:-L0ujhTn2kcmTsW1uoQY:b7ytvty) ([raw file](https://github.com/batfish/batfish/blob/852a31b49ee6d089b3382ac6fd85a18b9a13c724/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Topology.java#L72)):*

i considered both of those ideas but didn't go with them to save overhead. but better safe than sorry. went with the first one now; this stuff isn't called too often. related cleanup included for safe operations.
</blockquote></details>

I am reading this like you are sending another commit, but no push. Am I reading it wrong?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/793)*

ratulm:
oh, sorry, got distracted. pushed now.

dhalperi:
:lgtm:

---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/793#-:-L0ul_zTLKwS2JkONTud:bnfp4nl)*

PR #806, opened 2017-12-28
allowing questionName to specified for templates

ratulm:
the design carries forward a weakness from other places that 'questionName' cannot be a parameter in the template itself.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/806)

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -552,8 +552,8 @@ public class Client extends AbstractClient implements IClient {
     }
     String questionName = DEFAULT_QUESTION_PREFIX + "_" + UUID.randomUUID();
     if (parameters.containsKey("questionName")) {
-     questionName = parameters.get("questionName").asText();
-     parameters.remove("questionName");
+      questionName = parameters.get("questionName").asText();
+      parameters.remove("questionName");
     }
     JSONObject instanceJson;
     try {
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -539,7 +539,6 @@ public class Client extends AbstractClient implements IClient {
 
   private boolean answer(
       String questionTemplateName, String paramsLine, boolean isDelta, FileWriter outWriter) {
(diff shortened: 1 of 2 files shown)


dhalperi:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish-client/src/main/java/org/batfish/client/Command.java, line 114 at r1](https://reviewable.io:443/reviews/batfish/batfish/806#-L1PdcWCGFN8yYUh9Bu9:-L1PdcWCGFN8yYUh9BuA:b3cof7t) ([raw file](https://github.com/batfish/batfish/blob/aee6712f9102bb46d0b5fa084270f3c7d5341c24/projects/batfish-client/src/main/java/org/batfish/client/Command.java#L114)):*
> ```Java
>         ANSWER,
>         new Pair<>(
>             "<template-name>  [questionName=name] [param1=value1 [param2=value2] ...]",
> ```

can this not be implemented using an option instead of a parameter, like `initContainer`? E.g., `-name foo` instead of `name=foo`? Then I suspect it wouldn't interfere with parameters directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/806)*

ratulm:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish-client/src/main/java/org/batfish/client/Command.java, line 114 at r1](https://reviewable.io:443/reviews/batfish/batfish/806#-L1PdcWCGFN8yYUh9Bu9:-L1PeOic4qDnicMk2W8O:b1pzq3e) ([raw file](https://github.com/batfish/batfish/blob/aee6712f9102bb46d0b5fa084270f3c7d5341c24/projects/batfish-client/src/main/java/org/batfish/client/Command.java#L114)):*

can this not be implemented using an option instead of a parameter, like `initContainer`? E.g., `-name foo` instead of `name=foo`? Then I suspect it wouldn't interfere with parameters directly.
</blockquote></details>

options are implemented as boolean flags. your suggestion is a good one but requires changes to options processing in Client. i believe the same constraint on not using that parameter in templates is needed for pybatfish.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/806)*

dhalperi:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/806)*

PR #843, opened 2018-01-10
Fixes to configuration of Maven ANTRL plugin

agember:
The Eclipse m2e plug-in fails to parse the batfish-common-protocol and batfish maven configurations, because the excludes and includes directives should be sets (not strings) according to the [plug-in specification](http://www.antlr.org/api/maven-plugin/latest/antlr4-mojo.html#excludes). The fix is based on [this Stack Overflow post](https://stackoverflow.com/questions/20735122/learning-maven-unable-to-parse-configuration-for-jar-plugin).

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/843)


dhalperi:
Thanks @agember!

Note that this project [requires Maven 3.3.9](https://github.com/batfish/batfish/blob/master/projects/pom.xml#L392). Starting in 3.3.9, declaring properties in this concise way is valid.

Can you confirm that you're using latest m2e and that it's using at least Maven 3.3.9?

Happy to take this change, but there's no way to enforce we won't break older Maven versions again...

agember:
Upgrading m2e to version 1.7 (from 1.6) fixed the issue. (I was already using Maven 3.3.9.)

PR #856, opened 2018-01-16
adding spans for checktask status in coordinator

haverma:
screen shot 2018-01-16 at 12 19 03 pm

It becomes clearer now why the client is getting the response much later after the job actually completes. The interval between coord's subsequent checktaskstatus calls and also client's get work status calls cause the delay in getting the final answer to generate dataplane.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/856)

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
@@ -270,7 +270,13 @@ public class WorkMgr extends AbstractCoordinator {
     task.setStatus(TaskStatus.UnreachableOrBadResponse);
 
     Client client = null;
-    try {
+    SpanContext queueWorkSpan = work.getWorkItem().getSourceSpan();
+    try (ActiveSpan checkTaskSpan =
+        GlobalTracer.get()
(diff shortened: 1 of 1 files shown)


dhalperi:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/856)*

dhalperi:
:lgtm:

Travis is apparently backlogged on PRs, so waiting for this to finish before merge.

- it would be nicer if the coordinator polls were inside another span, like client "Waiting for work status". But maybe that's hard because of the asynchrony/background? If so, fine as-is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/856#-:-L304mzCC_LzUKvxBfra:b-yn2n6r)*

haverma:
Yes, I cannot think of a way to bunch them together as they are called periodically by Executors.newScheduledThreadPool

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/856#-:-L30700bXKB7RIMmnCA-:b-rprwzm)*

PR #885, opened 2018-01-30
Add API calls to get parsing or dataplaning work

corinaminer:
Adds API calls `getParsingWork` and `getDataplaningWork`. Both take in a testrig name and return the most recent parsing or dataplaning work on that testrig, complete or incomplete, in the same format you would get it from `listIncompleteWork`. Makes progress information more accessible.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/885)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/CoordConsts.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/CoordConsts.java
@@ -96,6 +96,7 @@ public class CoordConsts {
   public static final String SVC_KEY_TESTRIG_NAME = "testrigname";
   public static final String SVC_KEY_VERSION = "version";
   public static final String SVC_KEY_WORK_LIST = "worklist";
+  public static final String SVC_KEY_WORK_TYPE = "worktype";
   public static final String SVC_KEY_WORKID = "workid";
   public static final String SVC_KEY_WORKITEM = "workitem";
   public static final String SVC_KEY_WORKSPACE_NAME = "workspace";
@@ -115,7 +116,9 @@ public class CoordConsts {
   public static final String SVC_RSC_GET_ANSWER = "getanswer";
   public static final String SVC_RSC_GET_CONFIGURATION = "getconfiguration";
   public static final String SVC_RSC_GET_CONTAINER = "getcontainer";
+  public static final String SVC_RSC_GET_DATAPLANING_WORK = "getdataplaningwork";
   public static final String SVC_RSC_GET_OBJECT = "getobject";
+  public static final String SVC_RSC_GET_PARSING_WORK = "getparsingwork";
   public static final String SVC_RSC_GET_QUESTION_TEMPLATES = "getquestiontemplates";
   public static final String SVC_RSC_GET_WORKSTATUS = "getworkstatus";
   public static final String SVC_RSC_GETSTATUS = "getstatus";
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
@@ -1151,8 +1151,9 @@ public class WorkMgr extends AbstractCoordinator {
     return environments;
   }
 
-  public List<QueuedWork> listIncompleteWork(String containerName) {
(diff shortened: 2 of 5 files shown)


ratulm:
Two higher-level thoughts/questions:
  - Combine the two API calls into one that takes the type as input since the code is so similar? Actually, I am thinking, why not just expand the listIncompleteWork API call? 
  - What happens when two PARSING works are found in the queue? Right now, you'll return the first one found. If you are going to return one, then returning the last one makes sense. Or, just return all of them and let the client do what they want?

---

Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/885#-:-L483Xpx6G8b2LvOnDmy:b-uk9ccn)*

corinaminer:
- Combine the two API calls: that seems reasonable.
- Just expand `listIncompleteWork`: I think the tasks seem different enough to merit separate API calls. If we use`listIncompleteWork` for this new functionality, then a call for most recent parsing work would return one of two things, which both seem unintuitive:
    - `{workList: stringified List<WorkStatus>}`, as usual, but the list would only have one element
    - `{work: stringified WorkStatus}`, a deviation from the normal structure returned from that function

Is there a reason to limit number of API calls at the expense of clarity? Your call; let me know what you think and I'll revise accordingly.

- Two PARSING works found in the queue: This is a great point, I didn't think about which direction to iterate through the work queues. If there are two parsing work items for the same testrig in the completed work queue, we want the more recent one; but if they're both in the incomplete work queue, we want the older one (queued earlier). Which end of the queue is the "front"?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/885#-:-L48DYDIpwU86lOHYT_y:buk6euq)*

ratulm:
:lgtm: (if green)

---

Reviewed 5 of 5 files at r2.

---

*[projects/coordinator/src/test/java/org/batfish/coordinator/WorkQueueMgrTest.java, line 318 at r2](https://reviewable.io:443/reviews/batfish/batfish/885#-L49oo7HLHOvSlXemJfe:-L49oo7HLHOvSlXemJff:b-4xp30m) ([raw file](https://github.com/batfish/batfish/blob/e7b1ae630e21f373a340747bb0d032aa79f4926b/projects/coordinator/src/test/java/org/batfish/coordinator/WorkQueueMgrTest.java#L318)):*
> ```Java
>         new QueuedWork(
>             new WorkItem(CONTAINER, "testrig"),
>             new WorkDetails("testrig2", "env", WorkType.UNKNOWN));
> ```

There is an inconsistency between the testrig name in WorkItem() and WorkDetails(). 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/885#-:-L49p-BefSeXyxZ2RFDf:bopuwz4)*

ratulm:
Travis failure is because of @Nullable annotation after @FormDataParam annotation. Putting the former before the latter should fix things. (I ran into this issue just now.)

corinaminer:
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/coordinator/src/test/java/org/batfish/coordinator/WorkQueueMgrTest.java, line 318 at r2](https://reviewable.io:443/reviews/batfish/batfish/885#-L49oo7HLHOvSlXemJfe:-L4CUwcjSuNU2PKy3a_X:b8gj3mz) ([raw file](https://github.com/batfish/batfish/blob/e7b1ae630e21f373a340747bb0d032aa79f4926b/projects/coordinator/src/test/java/org/batfish/coordinator/WorkQueueMgrTest.java#L318)):*

There is an inconsistency between the testrig name in WorkItem() and WorkDetails(). 
</blockquote></details>

Good catch, fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/885)*

ratulm:
:lgtm:

---

Reviewed 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/885#-:-L4CW9M25H__ieC9D3Df:bnfp4nl)*

Issue #873, opened 2018-01-24
Add OSPF override-metric support


arifogel:
- [ ] Lines such as the following should make their way into the data model.

    Juniper:
    ```
    set protocols ospf area 0.1.2.3 area-range 192.168.0.0/24 override-metric 123`
    ```

    Cisco IOS, Arista:
    ```
    router ospf 1
      area 0.1.2.3 range 10.0.0.0/16 cost 100
    ```

- [ ] We should also ensure that semantics are respected in OSPF data plane computation.

PR #886, opened 2018-01-30
Cisco OSPF reference bandwidth

dhalperi:
Draft PR to set up larger work on #873

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/886)


arifogel:
Mostly good. Seems like a prime candidate for junit tests.

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/886#-:-L48O3tKo-AZbTiI2Soq:b-20kpgl)*

arifogel:
:lgtm:

---

Reviewed 12 of 13 files at r2, 2 of 2 files at r3.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/OspfProcess.java, line 87 at r2](https://reviewable.io:443/reviews/batfish/batfish/886#-L49hYqRCLK7IC_GRAxL:-L4D41C3H_qofIo80QHW:bo92v56) ([raw file](https://github.com/batfish/batfish/blob/b093414647ea5462bc9382650db0ea9a089b6684/projects/batfish/src/main/java/org/batfish/representation/cisco/OspfProcess.java#L87)):*

@arifogel which other vendors use the Cisco parser?

Before the recent revision, the code would have crashed on Arista devices (since the default throws), but it was not caught by any existing test.

Should I make the default just use IOS default?
</blockquote></details>

I don't remember offhand which other vendors use that parse, but it's easy enough to check the switch case in `ParseVendorConfigurationJob`.
I don't think it's a good idea to have a default. I'd rather it throw by default.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/886#-:-L4D619-lhG0WITLRgOF:bnfp4nl)*

dhalperi:
Review status: 19 of 20 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/OspfProcess.java, line 87 at r2](https://reviewable.io:443/reviews/batfish/batfish/886#-L49hYqRCLK7IC_GRAxL:-L4DrfZ_Tfl_a73CSZes:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/b093414647ea5462bc9382650db0ea9a089b6684/projects/batfish/src/main/java/org/batfish/representation/cisco/OspfProcess.java#L87)):*

I don't remember offhand which other vendors use that parse, but it's easy enough to check the switch case in `ParseVendorConfigurationJob`.
I don't think it's a good idea to have a default. I'd rather it throw by default.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/886)*

dhalperi:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/886)*

PR #892, opened 2018-02-01
Added -dataplaneengine parameter as part of task arguments

progwriter:
When running Batfish in distributed mode (i.e., manually invoking coordinator and worker in separate processes), worker arguments can get ignored when a new task is started.

I believe this is because of re-initialization of settings upon new task being executed (see https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Driver.java#L639-L642)

It is unclear to me what overall behavior is desired. Should `_mainSettings` be selectively overriden or selectively chosen from?

This PR picks the latter for the `-dataplaneengine` parameter.
Also, some additional logging statements were added for sanity checking.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/892)

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/config/Settings.java
+++ b/projects/batfish/src/main/java/org/batfish/config/Settings.java
@@ -730,6 +730,119 @@ public final class Settings extends BaseSettings implements BdpSettings, Grammar
     parseCommandLine(args);
   }
 
+  /**
+   * Creates a new settings object based on an existing settings object, then parses args.
+   *
+   * <p>This ignores some of the fields that are not applicable for task-level settings (e.g., the
+   * runMode).
+   *
+   * @param other the {@link Settings} object to copy values from
+   */
+  public Settings(Settings other, String[] args) {
+    this(new String[] {});
+    _activeTestrigSettings = other._activeTestrigSettings;
+    _analysisName = other._analysisName;
+    _analyze = other._analyze;
+    _answer = other._answer;
+    _answerJsonPath = other._answerJsonPath;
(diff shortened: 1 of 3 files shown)


ratulm:
selectively overriden ... the attempt is to preserve all settings. how are you doing things?

progwriter:
Hm. Could it because I'm not using allinone? When I'm invoking both, I use different entry points: one is `org.batfish.coordinator.Main` the other one is `org.batfish.main.Driver`

ratulm:
if you want to pass an argument to batfish that is not abstracted by
coordinator (most are not) use -batfishargs argument to coordinator. (is
that what you were missing? or, am i missing something?)

progwriter:
Attempting to implement the desired behavior of selectively overriden settings. Doesn't seem like a very elegant solution though.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/892#-:-L4IJI3MZHWv8CrvUbjA:b9t6sf6)*

ratulm:
instead of a whitelist of settings, which is hard to maintain, why not do a blacklist of settings that shouldn't be carried over?  (that seems closer to what you want. the only blacklist entry at the moment is the running mode.)

ratulm:
:lgtm: (when green)

---

Reviewed 1 of 2 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/892#-:-L4iquPMh4vQKVFoLKQ_:bopeokz)*

progwriter:
Closing because #956 is a better (more general) implementation of this feature.

PR #895, opened 2018-02-02
Use output stream when writing task log files

progwriter:
Currently workers can crash when logging very large answer strings, because we are doing an in-memory utf-8 encoding.

This fixes the crash by switching the encoding to be done using streams

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/895)


dhalperi:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/895#-:-L4Noms0qh5FE0-BLwoG:bnfp4nl)*

PR #896, opened 2018-02-05
Moved signal handling from worker to client

progwriter:
Moved `sigint` handling from the worker to the client. 
I believe this more closely matches the expected `ctrl+c` behavior: if the mode is interactive (or client only): the signal is ignored and request to kill work is sent.

If the mode is coordinator or worker, the interrupt will kill the process.
Tested locally with intellij.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/896)


ratulm:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/896)*

PR #903, opened 2018-02-06
Add analysis metadata

corinaminer:
Adds analysis metadata (containerName/analyses/analysisName/metadata.json) with a creation timestamp and a boolean `suggested` property. `configureAnalysis` now has an optional boolean argument `suggested`. When `configureAnalysis` is called:
- with `suggested` set to a non-null value:
    - if analysis already has a metadata file, update its `suggested` property
    - otherwise create a metadata file with the given `suggested` value
- with `suggested = null`:
    - if creating a new analysis, create a metadata file with `suggested = false`
    - otherwise do not check or change metadata

Creation timestamps are not currently used but seem like a reasonable thing to include.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/903)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/CoordConsts.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/CoordConsts.java
@@ -96,6 +96,7 @@ public class CoordConsts {
   public static final String SVC_KEY_RESULT = "result";
   public static final String SVC_KEY_SETTINGS = "settings";
   public static final String SVC_KEY_SUCCESS = "success";
+  public static final String SVC_KEY_SUGGESTED = "suggested";
   public static final String SVC_KEY_TASKSTATUS = "taskstatus";
   public static final String SVC_KEY_TESTRIG_INFO = "testriginfo";
   public static final String SVC_KEY_TESTRIG_LIST = "testriglist";
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AnalysisMetadata.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AnalysisMetadata.java
@@ -1,0 +1,36 @@
+package org.batfish.datamodel;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+
+public class AnalysisMetadata {
+
+  private static final String PROP_CREATIONTIMESTAMP = "creationTimestamp";
+  private static final String PROP_SUGGESTED = "suggested";
+
+  private Instant _creationTimestamp;
+  private boolean _suggested;
+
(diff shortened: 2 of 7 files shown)


corinaminer:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java, line 798 at r1](https://reviewable.io:443/reviews/batfish/batfish/903#-L4hFHHgCgbnMBjJQ5ZX:-L4hFHHgCgbnMBjJQ5ZY:b-b7692p) ([raw file](https://github.com/batfish/batfish/blob/e3c6a6e4b810a96d89853ebb7b145db6c6d14a30/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java#L798)):*
> ```Java
> 
>   // this function should build on others but some overrides are getting in the way
>   // TODO: cleanup later
> ```

Seems like what's "getting in the way" here is that the other getpath methods aren't static, so this method and `getpathTestrigMetadata` can't reference them. Any reason not to make all getpath methods static?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/903)*

ratulm:
Reviewed 5 of 7 files at r1.
Review status: 5 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java, line 798 at r1](https://reviewable.io:443/reviews/batfish/batfish/903#-L4hFHHgCgbnMBjJQ5ZX:-L4hfV1tsrHzs3NaydeC:b-2xl5va) ([raw file](https://github.com/batfish/batfish/blob/e3c6a6e4b810a96d89853ebb7b145db6c6d14a30/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java#L798)):*

Seems like what's "getting in the way" here is that the other getpath methods aren't static, so this method and `getpathTestrigMetadata` can't reference them. Any reason not to make all getpath methods static?
</blockquote></details>

IIRC, I ran into an issue with that given WorkMgr implements an interface (ICoordinator) and interfaces cannot have static methods. I think the cleanup here would be a new class that implements all methods (which makes sense anyway) or at least taking out the inheritance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/903)*

ratulm:
:lgtm:

---

Reviewed 1 of 7 files at r1.
Review status: 6 of 7 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/903#-:-L4hhIMpriFfovEgqx9v:bnfp4nl)*

ratulm:
Reviewed 1 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/903)*

PR #913, opened 2018-02-07
Pool manager HttpClient

haverma:
1.)Creating Jersey http client in the constructor and not for every refreshworkerstatus call
2.)Initializing tracer before pool manager to enable registering ClientTracingFeature in createHttpClientBuilder() function

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/913)

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/Main.java
@@ -323,10 +323,10 @@ public class Main {
   private static void mainRun() {
     try {
       initAuthorizer();
-      initPoolManager();
       if (_settings.getTracingEnable() && !GlobalTracer.isRegistered()) {
         initTracer();
       }
+      initPoolManager();
       initWorkManager();
     } catch (Exception e) {
       System.err.println(
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java
@@ -15,6 +15,7 @@ import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.batfish.common.BatfishException;
 import org.batfish.common.BatfishLogger;
 import org.batfish.common.BfConsts;
 import org.batfish.common.util.CommonUtil;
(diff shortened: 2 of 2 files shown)


dhalperi:
Reviewed 2 of 2 files at r1.

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java, line 189 at r1](https://reviewable.io:443/reviews/batfish/batfish/913#-L4mq81s_JHrJJUGt7pm:-L4mq81s_JHrJJUGt7pn:bsnkwom) ([raw file](https://github.com/batfish/batfish/blob/8b6bcf43b0910949cf58f72aad98e8a654cd122a/projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java#L189)):*
> ```Java
>       if (client != null) {
>         client.close();
>       }
> ```

I'm no longer clear of the client lifecycle now -- when will the client be shut down?

If never, are we guaranteed that the client will never get into a bad state?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/913)*

haverma:
Commented inline, 

---

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java, line 189 at r1](https://reviewable.io:443/reviews/batfish/batfish/913#-L4mq81s_JHrJJUGt7pm:-L4oL2lscllbB62AC530:bm2ivqx) ([raw file](https://github.com/batfish/batfish/blob/8b6bcf43b0910949cf58f72aad98e8a654cd122a/projects/coordinator/src/main/java/org/batfish/coordinator/PoolMgr.java#L189)):*

I'm no longer clear of the client lifecycle now -- when will the client be shut down?

If never, are we guaranteed that the client will never get into a bad state?
</blockquote></details>

The client will never be shut down now, we were creating a new client for every request, which is resource extensive.
I didn't find any proof that client can't go into a bad state,  but my idea was that if we are not changing any property of the client, the object should retain its state,  https://stackoverflow.com/questions/34566278/how-to-correctly-share-jax-rs-2-0-client

However, another article (https://blogs.oracle.com/japod/how-to-use-jersey-client-efficiently) states that even creating a new webtarget effectively creates a new Client so I think this PR will not be useful. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/913#-:-L4oMZYnC93aqWllHM4i:b-toeap8)*

haverma:
Closing for now, will reopen after finalizing a way to optimize it(may be by sharing web targets as well)

PR #946, opened 2018-02-15
If parsing or dataplaning fails, put err msg in metadata

corinaminer:
The idea is that if a testrig has failed to parse or dataplane, the metadata should contain the error message that caused the failure (e.g., "No valid configurations found"). This is implemented by adding an `errMessage` property to `Task`. Details:
- `Batfish._terminatedWithException`, a boolean, has been replaced with `Batfish._terminatingExceptionMessage`, a String that is null if no error occurs
- In `Driver.runBatfish()`, when an exception occurs in the running thread, the Batfish instance stores the error message as `_terminatingExceptionMessage`
- `runBatfish()` returns the error message (if any, otherwise null) instead of a boolean
- `Driver.runBatfishThroughService` sets its task's error message to the output of `runBatfish()` if any error occurred
- `WorkQueueMgr.processTaskCheckResult()` will put the given task's error message (if any) into the environment metadata
- Metadata classes have been updated to accommodate an `errMessage` property in EnvironmentMetadata

To me it looks like the biggest question about this setup is whether it captures all the exceptions we want it to, because an exception outside of the runBatfish thread will not be recorded in a `Task`. There is also a question of what exactly to include in the error message: in some cases `e.getMessage()` is too verbose.

Also, it won't help with issues that prevent testrig initialization (like having multiple top-level folders) because no folder is created for the testrig, so no metadata exists for it.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/946)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Task.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Task.java
@@ -88,6 +88,8 @@ public class Task {
 
   private static final String PROP_BATCHES = "batches";
 
+  private static final String PROP_ERR_MESSAGE = "errMessage";
+
   private static final String PROP_OBTAINED = "obtained";
 
   private static final String PROP_STATUS = "status";
@@ -97,6 +99,8 @@ public class Task {
   private final String[] _args;
 
   private final List<Batch> _batches;
+
+  private String _errMessage;
 
   private final Date _obtained;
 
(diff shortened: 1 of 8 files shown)


ratulm:
:lgtm:

The code looks good to the eye; I assume that it works :)

I have some thoughts on handling testrig initialization errors  .....

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/946#-:-L5Q2YbPz-gmHhBIPnro:bnpp5yy)*

PR #947, opened 2018-02-15
EdgePath: stop returning Configuration in data model

dhalperi:
The Configuration for every node includes too much information -- EdgePath just blows
it up linearly.

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/947)


dhalperi:
I think Todd uses this question most -- Todd, any issues?

millstein:
I don't actually use the question (not sure who does?) but I created it a while back as a generalization of things like MTU check.  Note that this change doesn't affect the default behavior of EdgePath, which uses "summary" style.  It only affects "verbose" mode -- now you won't be able to get back information about the nodes other than their names.  But if in practice people only want verbose interface information, then providing the full node information is overkill (and takes up a lot of unnecessary space), and you can always do a second jsonpath query to get desired node information.  So fine with me.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/947#-:-L5OxdYz6Y6y2jGlgDz3:b-71bmv8)*

millstein:
Oh I see, verbose *is* the default behavior of the edge path question (but not of the neighbors question).  Yes, probably makes sense to not include full node information by default.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/947#-:-L5P-fRbXsSnJ_XeUYY3:b3scplg)*

dhalperi:
Thanks Todd!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/947#-:-L5PMcG9OPvFsrTsaccy:b7co8fo)*

dhalperi:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/947)*

PR #955, opened 2018-02-17
adding a fake domain name for AWS hosts. pushing domain insider the b…

ratulm:
...builder

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/955)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
@@ -54,13 +54,14 @@ public final class Configuration extends ComparableStructure<String> {
     @Override
     public Configuration build() {
       String name = _hostname != null ? _hostname : generateName();
-      Configuration configuration = new Configuration(name, _configurationFormat, _domainName);
+      Configuration configuration = new Configuration(name, _configurationFormat);
       if (_defaultCrossZoneAction != null) {
         configuration.setDefaultCrossZoneAction(_defaultCrossZoneAction);
       }
       if (_defaultInboundAction != null) {
         configuration.setDefaultInboundAction(_defaultInboundAction);
       }
+      configuration.setDomainName(_domainName);
       return configuration;
     }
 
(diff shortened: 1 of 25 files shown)


dhalperi:
One question for Ari, but otherwise LGTM.

---

Reviewed 25 of 25 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 57 at r1](https://reviewable.io:443/reviews/batfish/batfish/955#-L5_6izf3U_R9Nlt-TIo:-L5_6izf3U_R9Nlt-TIp:b-4qmbhh) ([raw file](https://github.com/batfish/batfish/blob/57368cbaeab234cd3043ece82fa35f0ba994b6f5/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L57)):*
> ```Java
>     public Configuration build() {
>       String name = _hostname != null ? _hostname : generateName();
>       Configuration configuration = new Configuration(name, _configurationFormat, _domainName);
> ```

It's not obvious to me that `domainName` needs to be pushed into the constructor like this. Is it really impossible for a device not to have a domain name?

Pushing it here has far-reaching impacts, as we saw.

@arifogel ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/955#-:-L5_78P8dvQA6SwUT1gb:bvlb6yd)*

arifogel:
Reviewed 25 of 25 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 57 at r1](https://beta.reviewable.io/reviews/batfish/batfish/955#-L5_6izf3U_R9Nlt-TIo:-L5jJdYN8oKrh5ZT8uMi:b-t8lk28) ([raw file](https://github.com/batfish/batfish/blob/57368cbaeab234cd3043ece82fa35f0ba994b6f5/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L57)):*

It's not obvious to me that `domainName` needs to be pushed into the constructor like this. Is it really impossible for a device not to have a domain name?

Pushing it here has far-reaching impacts, as we saw.

@arifogel ?
</blockquote></details>

For reference, `ConfigurationFormat`is in the constructor because it is always known immediately and always needed. Furthermore, it can never be null, and is never supposed to be modified. None of this is true about domain-name. This field does not belong in the constructor, and it seems the only time you ever use it there is in a fake matter in AWS configs where it might be updated later anyway.
Otherwise, builder changes are fine. Please revert the constructor modification.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/955)*

arifogel:
:lgtm:

---

Reviewed 4 of 14 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/955#-:-L5kPIJyBSMdKTNvbJtR:bnfp4nl)*

PR #968, opened 2018-02-21
Graph: delete seemingly unnecessary code

dhalperi:
No tests broke?

My reading of the current version of the code is that the [OSPF costs are initialized](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2969) before the [vendor-independent configurations](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L1584) are even written to disk.

What was this code for? Is it necessary?

---
This change is [Reviewable](https://reviewable.io/reviews/batfish/batfish/968)

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/symbolic/Graph.java
+++ b/projects/batfish/src/main/java/org/batfish/symbolic/Graph.java
@@ -79,7 +79,6 @@ public class Graph {
   }
 
   public static final String BGP_COMMON_FILTER_LIST_NAME = "BGP_COMMON_EXPORT_POLICY";
-  private static final int DEFAULT_CISCO_VLAN_OSPF_COST = 1;
   private static final String NULL_INTERFACE_NAME = "null_interface";
   private IBatfish _batfish;
   private Set<String> _routers;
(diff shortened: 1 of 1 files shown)


anothermattbrown:
Ryan wrote: I believe when I wrote that code, Batfish did not initialize the OSPF costs of interfaces unless you actually computed the data plane. Since Minesweeper didn’t compute the data plane, the costs would not get initialized, so I had to do it in the Minesweeper code. If that has changed, then there is no need for that code.

It looks like the costs are being set, e.g. in Interface.build(). Also, there is not reference to setOspfCost from BdpDataPlane. Seems safe to remove.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/968#-:-L5ty1rWQYcxtkVS9XDf:blt6u1l)*

dhalperi:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/968)*

PR #984, opened 2018-02-25
Produce a set of "role dimensions" for each node.

millstein:
As part of role inference, which currently tries to find a single best notion of role according to our metric, we now also generate a list of "role dimensions" for each node.  Right now these are alphabetic tags that are pulled from the node's name.  For example, a node with the name foo-bar32-baz0fa will have three role dimensions: foo, bar, and baz.  We record them as a map from an integer index to each dimension's string value (this is done rather than a list since some nodes will not have all dimensions).


dhalperi:
Cool!

---

Reviewed 9 of 9 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java, line 35 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LD_p2IGlqxfWaKdI-:-L6LD_p2IGlqxfWaKdI0:b-7bznzy) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L35)):*
> ```Java
>   // During role inference we compute the possible role dimensions of a node based on its name.
>   // Later these dimensions will be stored in each node's Configuration object.
>   private static SortedMap<String, NavigableMap<Integer, String>> _roleDimensions = new TreeMap<>();
> ```

Why is this static? Seems like it should be local to an instance of the class so that it is tied to the current execution of the plugin and not persisted further.

---

*[tests/basic/nodes.ref, line 1476 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LE5EzTyT7i6WsKoU5:-L6LE5EzTyT7i6WsKoU6:berm9by) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/tests/basic/nodes.ref#L1476)):*
> ```
>           "roles" : [
>             "border"
>           ],
> ```

how do dimensions and roles line up? I guess I expected to see `[as, border]` say. Or `[as1, border]`. Or something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/984#-:-L6LEZEaPDAh0JPBqxhW:b-89q77x)*

millstein:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java, line 35 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LD_p2IGlqxfWaKdI-:-L6N-MBTYb3LbPzabm2K:bmw713e) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L35)):*

Why is this static? Seems like it should be local to an instance of the class so that it is tied to the current execution of the plugin and not persisted further.
</blockquote></details>

Currently roles are inferred (as I understand it) during test_rig initialization (see inferNodeRoles in Batfish, called from serializeIndependentConfigs) but the configurations are not actually updated with their roles until the environment is applied (see processNodeRoles in Batfish, called from applyEnvironment).  I tried directly updating configs with the role dimensions right when roles were inferred, but that didn't work, I guess because the configs are overwritten when the environment is applied.  So instead I made the role dimensions static so I could grab them later during environment processing.  

A hack, I agree.  One option would be to store the InferRoles object somewhere, in which case the role dimensions could be a regular instance field of the class.  Or store the role dimensions themselves somewhere.  What do you suggest?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/984)*

millstein:

---

*[tests/basic/nodes.ref, line 1476 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LE5EzTyT7i6WsKoU5:-L6N0vljBzWeee9RGCQI:bhr87cu) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/tests/basic/nodes.ref#L1476)):*

how do dimensions and roles line up? I guess I expected to see `[as, border]` say. Or `[as1, border]`. Or something.
</blockquote></details>

Currently our existing notion of roles is unchanged.  So the "role" is still the unique best dimension or combination of dimensions as inferred by InferRoles.  The roleDimensions field separately keeps track of all dimensions.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/984)*

millstein:

---

*[tests/basic/nodes.ref, line 1476 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LE5EzTyT7i6WsKoU5:-L6N1Fa-Jww4_eThowzI:b-wb2zxx) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/tests/basic/nodes.ref#L1476)):*
<details><summary><i>Previously, millstein wrote…</i></summary><blockquote>

Currently our existing notion of roles is unchanged.  So the "role" is still the unique best dimension or combination of dimensions as inferred by InferRoles.  The roleDimensions field separately keeps track of all dimensions.  
</blockquote></details>

Also the roles can be user-provided which is the case here -- there is a node_roles.json file in the testrig.  So the "roles" field contains the user-provided roles in that case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/984)*

dhalperi:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java, line 35 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LD_p2IGlqxfWaKdI-:-L6bpIdP3JVDN2I31HKU:btihvzo) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L35)):*
<details><summary><i>Previously, millstein wrote…</i></summary><blockquote>

Currently roles are inferred (as I understand it) during test_rig initialization (see inferNodeRoles in Batfish, called from serializeIndependentConfigs) but the configurations are not actually updated with their roles until the environment is applied (see processNodeRoles in Batfish, called from applyEnvironment).  I tried directly updating configs with the role dimensions right when roles were inferred, but that didn't work, I guess because the configs are overwritten when the environment is applied.  So instead I made the role dimensions static so I could grab them later during environment processing.  

A hack, I agree.  One option would be to store the InferRoles object somewhere, in which case the role dimensions could be a regular instance field of the class.  Or store the role dimensions themselves somewhere.  What do you suggest?
</blockquote></details>

Sorry for the delay; not sure i have a great answer here yet.

The underlying issue is that we need to promote roles from per-snapshot to per-network, since we believe that roles will be consistent across multiple snapshots.

I believe that would likely alleviate the issues.

---

*[tests/basic/nodes.ref, line 1476 at r1](https://reviewable.io:443/reviews/batfish/batfish/984#-L6LE5EzTyT7i6WsKoU5:-L6boRSqDPRVBIKWjBJE:b-d54mz3) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/tests/basic/nodes.ref#L1476)):*
<details><summary><i>Previously, millstein wrote…</i></summary><blockquote>

Also the roles can be user-provided which is the case here -- there is a node_roles.json file in the testrig.  So the "roles" field contains the user-provided roles in that case.
</blockquote></details>

Hmm. There's something I'm missing here:

When a user supplied roles via node_roles.json how do those relate to role dimensions? Would we not just say there's 0 dimensions, or 1 dimension, or something else?

It looks like for the sake of dimensions we're simply inferring a regex anyway independent of the provided roles, which will likely not line up?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/984)*

millstein:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java, line 35 at r1](https://beta.reviewable.io/reviews/batfish/batfish/984#-L6LD_p2IGlqxfWaKdI-:-L6bul6t57Tz-5_RvDNZ:bpgq1vc) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L35)):*

Sorry for the delay; not sure i have a great answer here yet.

The underlying issue is that we need to promote roles from per-snapshot to per-network, since we believe that roles will be consistent across multiple snapshots.

I believe that would likely alleviate the issues.
</blockquote></details>

Makes sense.  For now I'm forging ahead with more functionality; when it makes sense we can discuss this issue and how best to resolve it.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/984)*

millstein:

---

*[tests/basic/nodes.ref, line 1476 at r1](https://beta.reviewable.io/reviews/batfish/batfish/984#-L6LE5EzTyT7i6WsKoU5:-L6bvC659NfjVgGwRI43:bwb9v8) ([raw file](https://github.com/batfish/batfish/blob/70d0cb8dd17d892a17484591cc5766083eb0b15e/tests/basic/nodes.ref#L1476)):*

Hmm. There's something I'm missing here:

When a user supplied roles via node_roles.json how do those relate to role dimensions? Would we not just say there's 0 dimensions, or 1 dimension, or something else?

It looks like for the sake of dimensions we're simply inferring a regex anyway independent of the provided roles, which will likely not line up?
</blockquote></details>

Inferred roles will line up with inferred role dimensions, but user-provided roles may not line up with inferred role dimensions.  We could also allow users to provide role dimensions explicitly; currently there's no way to do that.  But I'm not sure that it matters (at least for now) if they don't line up -- if we can use our inferred role dimensions to find interesting things out about the network, then they are useful even if they don't correlate with the user-defined roles.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/984)*

PR #994, opened 2018-02-27
Cisco_aaa: add support for aaa accounting update newinfo

dhalperi:
@haverma , want to make the first pass? Then Ari can come through to verify.


haverma:
It can be inferred from the rule that we are aiming that if PERIODIC exists then DEC has to be follow. So can we add a failing unit test where PERIODIC exists without a DEC ? Otherwise looks good

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/994#-:-L6NlwssdDgBCwlZ_ZbO:b-t6j08p)*

dhalperi:
We don't typically add parser rejection tests here; it's not like cisco's docs are so precise we can be confident about what's not accepted.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/994#-:-L6Nsmqf8P6oS6wrm2l5:b-245y4h)*

arifogel:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/994#-:-L6OGXHJ1wRsySaN4JeB:bnfp4nl)*

PR #1021, opened 2018-03-04
remove Interface references from OspfArea

arifogel:
- various updates to related matchers
- some steps towards immutability in related data model classes
- update affected tests
- update code previously relying on references

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
@@ -748,7 +748,7 @@ public class CommonUtil {
                       }
                       OspfArea remoteArea = remoteProc.getAreas().get(areaNum);
                       if (remoteArea != null
-                          && remoteArea.getInterfaceNames().contains(remoteIfaceName)) {
+                          && remoteArea.getInterfaces().contains(remoteIfaceName)) {
                         Ip remoteIp = remoteIface.getAddress().getIp();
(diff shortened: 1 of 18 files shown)


dhalperi:
:lgtm: other than one comment you can choose whether to address. (you can dismiss and ship if you want to ignore)

---

Reviewed 15 of 18 files at r1, 3 of 3 files at r2.

---

*[projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java, line 155 at r2](https://reviewable.io:443/reviews/batfish/batfish/1021#-L6ovK20EM23Ytk68Alv:-L6ovK20EM23Ytk68Alw:btyofrp) ([raw file](https://github.com/batfish/batfish/blob/7cc74b1e9fbccc096455a3a43100732fa04f791b/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L155)):*
I suspect this error is a lot worse now if it fails. something like

> expected: a configuration with a default vrf with an ospf process with an ospf area with a summary with isadvertised equal to false.
> 
> found: a configuration with a default vrf with an ospf process with an ospf area with a summary with isadvertised equal to true.

Prefer to do the assert over the actual thing being computed: the summary. Any other type of error isn't a bug in the code being tested, it's a bug somewhere else.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1021#-:-L6ovmr0J4ZbCW6bvlTo:bl0dleb)*

PR #1032, opened 2018-03-06
In configureAnalysis, do question deletions before adding questions

corinaminer:
This fixes a bug where trying to update a check without changing its name would cause a `Question 'foo' already exists for analysis 'bar'` error. The bug occurred because the code that deletes questions was after the code that adds questions. This PR just moves the question deletion section to before the question addition section.

### DIFF
--- a/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
+++ b/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java
@@ -488,6 +488,15 @@ public class WorkMgr extends AbstractCoordinator {
     }
 
     Path questionsDir = aDir.resolve(BfConsts.RELPATH_QUESTIONS_DIR);
+    /** Delete questions */
+    for (String qName : questionsToDelete) {
+      Path qDir = questionsDir.resolve(qName);
+      if (!Files.exists(qDir)) {
+        throw new BatfishException("Question " + qName + " does not exist for analysis " + aName);
+      }
+      CommonUtil.deleteDirectory(qDir);
+    }
+
     for (Entry<String, String> entry : questionsToAdd.entrySet()) {
       Path qDir = questionsDir.resolve(entry.getKey());
       if (Files.exists(qDir)) {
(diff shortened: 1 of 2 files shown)


dhalperi:
Reviewed 1 of 1 files at r1.

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java, line 494 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1032#-L6xaDccEHQpEMHfP-uQ:-L6xaDccEHQpEMHfP-uR:b8swi2y) ([raw file](https://github.com/batfish/batfish/blob/294bbaaa21493ef31f298cdfa02786d21bc468fb/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java#L494)):*
> ```Java
>     for (String qName : questionsToDelete) {
>       Path qDir = questionsDir.resolve(qName);
>       if (!Files.exists(qDir)) {
> ```

This is probably fine as-is given existing behavior, but I'll leave an informational comment.

In principle, it would be better to check that questionsToDelete all exist and (questionsToAdd-questionsToDelete) all do not exist before deleting or adding any. That way the analysis is less likely to be in an unknown state.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1032)*

corinaminer:

---

*[projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java, line 494 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1032#-L6xaDccEHQpEMHfP-uQ:-L6xawjF6UzQuETE5moI:bhyysbv) ([raw file](https://github.com/batfish/batfish/blob/294bbaaa21493ef31f298cdfa02786d21bc468fb/projects/coordinator/src/main/java/org/batfish/coordinator/WorkMgr.java#L494)):*

This is probably fine as-is given existing behavior, but I'll leave an informational comment.

In principle, it would be better to check that questionsToDelete all exist and (questionsToAdd-questionsToDelete) all do not exist before deleting or adding any. That way the analysis is less likely to be in an unknown state.
</blockquote></details>

that's super true and not that hard so i'm going to add it

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1032)*

ratulm:
:lgtm:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1032#-:-L6xu5pe6PlIg11YAp14:bnfp4nl)*

PR #1036, opened 2018-03-08
Cisco interface ospf

dhalperi:
Stop ignoring `bandwidth N kbps` in cisco interfaces. You can see it works in the nodes ref tests.

Had to update the `if_null_block` with the other `bandwidth` commands that are used transitively inside of interface to compensate. Most of these were not (well) tested, so I added parsing tests.

Also, if something in `if_null_block` is unrecognized, apparently the parser recovery jumps somewhere unpredictable and typically loses the entire rest of the file. (E.g., for `tx-queue 0 / bandwidth percent foo`, when `bandwidth percent` was temporarily unrecognized).


arifogel:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1036#-:-L729MDV39Z-gcKXLQlF:bnfp4nl)*

PR #1064, opened 2018-03-17
Save warning and continue parsing aws configs

haverma:
- don't abort parsing other AWS configs in case some config throw error
- log warning and continue


ratulm:
I realize that you are fixing an immediate problem, but I think a better way to deal with the situation where the data does not match what we expect or handle is to surface that problem to the user instead of having it buried in a log file. We do that through by embedding the warning in the Warnings object, from where it can be picked up by initiInfo.

So, my recommendation would be: instead of throwing exception in those inner toConfigurationNode functions, we should add an appropriate warning using something like awsConfiguration.getWarnings().redFlag(....) and do as much work as possible. redFlag is just one type of warning, there are other types; for the specific case you are solving, you may want to use a different type since that aberration in the data is an expected one.  Let me know if this recommendation does not make sense.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1064#-:-L7lyPK_ehstY3JBmJ7x:b-g55po8)*

haverma:
I think I misunderstood earlier and your suggestion makes perfect sense, I will make modifications in this PR. I guess for cases where a crucial AWS construct(like a subnet) which is referred is not present we should not do further work and not add that node to the vendor independent Configuration nodes as my understanding is that every instance/database instance should be in a subnet in a VPC. Will that be a reasonable thing to assume ?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1064#-:-L7mb38nzftWIMspNkBv:b-wmv1or)*

ratulm:
Good question. Unless it causes other problems, my inclination would be to add the node anyway. It might end up appearing as an isolated node in the visualization and impact answers but it will draw user's attention to the data anomaly. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1064#-:-L7oc7lWIKflwW2397zW:b-57cw24)*

haverma:
Changes dependent on PR batfish/batfish#1076, can be reviewed if that is merged
Used warnings where exceptions were thrown in AWS configs

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1064#-:-L7xZP5fqYr9ufDV67uB:bhhco05)*

ratulm:
Is this one ready to review now?

haverma:
No, I am just merging in changes from master and will update this one, as there were some changes in the other PR

---

Review status: 0 of 16 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1064#-:-L81OnZMHkwTa6zgctBw:bxvg646)*

haverma:
updated, now it can be reviewed

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1064#-:-L81cuVoBYSFWtKcLSHv:bu4neo3)*

ratulm:
:lgtm:

---

Reviewed 3 of 16 files at r2, 1 of 5 files at r3.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1064#-:-L838sK80aej53Ykkoac:bnfp4nl)*

PR #1067, opened 2018-03-18
pom.xml: switch to Jackson BOM

dhalperi:
The Bill of Materials is a single thing you import that
automatically defines the dependency version for all related
projects. Seems significantly simpler to maintain, since we
won't accidentally add new versions of every thing we want to
depend on.

### DIFF
--- a/projects/pom.xml
+++ b/projects/pom.xml
@@ -489,51 +489,11 @@
 
       <!-- External modules. -->
       <dependency>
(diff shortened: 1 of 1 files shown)


ratulm:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1067#-:-L7rjuzssqkqHixQkg3O:bnfp4nl)*

PR #1076, opened 2018-03-19
Enable recording warnings per AWS config node

haverma:
`ConvertConfigurationResult` class supported a map of `Configurations` but accepted only a single `Warnings` object, changed that to a `Map` of `<Hostname, Warnings>` so that we can store a separate `Warnings` object for each configuration. It worked for physical `Configuration` nodes as in that case the `configurations` map contains just a single node at a time, whereas for AWS configs it contains all AWS nodes parsed.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
@@ -165,6 +165,17 @@ public class Warnings implements Serializable {
     _unimplementedRecord = unimplementedRecord;
   }
 
+  public Warnings(Warnings other) {
+    this();
+    _pedanticAsError = other._pedanticAsError;
+    _pedanticRecord = other._pedanticRecord;
+    _printParseTree = other._printParseTree;
+    _redFlagAsError = other._redFlagAsError;
+    _redFlagRecord = other._redFlagRecord;
+    _unimplementedAsError = other._unimplementedAsError;
+    _unimplementedRecord = other._unimplementedRecord;
+  }
+
   public boolean getPedanticAsError() {
     return _pedanticAsError;
   }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
@@ -165,17 +165,6 @@ public class Warnings implements Serializable {
     _unimplementedRecord = unimplementedRecord;
   }
 
-  public Warnings(Warnings other) {
-    this();
-    _pedanticAsError = other._pedanticAsError;
-    _pedanticRecord = other._pedanticRecord;
-    _printParseTree = other._printParseTree;
-    _redFlagAsError = other._redFlagAsError;
-    _redFlagRecord = other._redFlagRecord;
-    _unimplementedAsError = other._unimplementedAsError;
-    _unimplementedRecord = other._unimplementedRecord;
-  }
-
   public boolean getPedanticAsError() {
     return _pedanticAsError;
   }
--- a/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
+++ b/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
@@ -19,14 +19,10 @@ public class ConvertConfigurationJob extends BatfishJob<ConvertConfigurationResu
 
   private String _name;
 
-  private Warnings _warnings;
-
-  public ConvertConfigurationJob(
-      Settings settings, Object configObject, String name, Warnings warnings) {
+  public ConvertConfigurationJob(Settings settings, Object configObject, String name) {
     super(settings);
     _configObject = configObject;
     _name = name;
-    _warnings = warnings;
   }
 
   @Override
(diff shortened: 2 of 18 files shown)


dhalperi:
Review status: 0 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationResult.java, line 21 at r1](https://reviewable.io:443/reviews/batfish/batfish/1076#-L7yqTMPy35mxEjoHaFP:-L7yqTMQv0gw5HbI2xVZ:b9x5hv9) ([raw file](https://github.com/batfish/batfish/blob/8ba7bedc324199a7303cbf4f2fb31fd7dcc71d44/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationResult.java#L21)):*
> ```Java
>   private String _name;
> 
>   private Map<String, Warnings> _warningsByHost;
> ```

This is not going to be good enough. We need filename, not just hostname, etc.

What's the goal? Can you point me at the design doc?

There's already work started in this area....

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1076)*

dhalperi:
After discussion, :LGTM:. A few cleanups remaining, but this seems like a solution that generally makes sense as a stopgap.

---

Reviewed 16 of 16 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java, line 168 at r1](https://reviewable.io:443/reviews/batfish/batfish/1076#-L8-haWVPeDZln5EXJHf:-L8-haWWNSqOZycOHI1F:be2e20w) ([raw file](https://github.com/batfish/batfish/blob/8ba7bedc324199a7303cbf4f2fb31fd7dcc71d44/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java#L168)):*
> ```Java
>   }
> 
>   public Warnings(Warnings other) {
> ```

I don't think this is a good method to add. It doesn't really clone warnings, it just copies settings from the other warnings.

The logic you really want is in Batfish.buildWarnings(). Move that logic as a public static helper method (`public static Warnings buildWarnings(Settings settings)`) somewhere and then use it in place of the existing builder.

Since CCJ has access to `Settings`, it should be able to call this helper to build a new warnings from Settings rather than copying from an existing warning.

---

*[projects/batfish/src/main/java/org/batfish/representation/aws/Region.java, line 265 at r1](https://reviewable.io:443/reviews/batfish/batfish/1076#-L8-iz3KvTEkh1sYs9Pa:-L8-iz3KvTEkh1sYs9Pb:b-29bru0) ([raw file](https://github.com/batfish/batfish/blob/8ba7bedc324199a7303cbf4f2fb31fd7dcc71d44/projects/batfish/src/main/java/org/batfish/representation/aws/Region.java#L265)):*
> ```Java
>     updateAllocatedIps();
>     for (Vpc vpc : getVpcs().values()) {
>       Warnings warnings = new Warnings(awsConfiguration.getWarnings());
> ```

to be replaced by `buildWarnings(settings)` or something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1076#-:-L8-jHZS2e7d0U1tk8cn:bmbqd98)*

haverma:
addressed comments, and did some refactoring to remove warnings instance member from `ConvertConfigurationJob` and `AwsConfiguration`

---

Review status: 12 of 17 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/batfish/src/main/java/org/batfish/representation/aws/Region.java, line 265 at r1](https://reviewable.io:443/reviews/batfish/batfish/1076#-L8-iz3KvTEkh1sYs9Pa:-L81AedOn2IsqcGuBmSO:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/8ba7bedc324199a7303cbf4f2fb31fd7dcc71d44/projects/batfish/src/main/java/org/batfish/representation/aws/Region.java#L265)):*

to be replaced by `buildWarnings(settings)` or something.
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java, line 168 at r1](https://reviewable.io:443/reviews/batfish/batfish/1076#-L8-haWVPeDZln5EXJHf:-L81APPfTiDYsKty22XD:b-cwowlr) ([raw file](https://github.com/batfish/batfish/blob/8ba7bedc324199a7303cbf4f2fb31fd7dcc71d44/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java#L168)):*

I don't think this is a good method to add. It doesn't really clone warnings, it just copies settings from the other warnings.

The logic you really want is in Batfish.buildWarnings(). Move that logic as a public static helper method (`public static Warnings buildWarnings(Settings settings)`) somewhere and then use it in place of the existing builder.

Since CCJ has access to `Settings`, it should be able to call this helper to build a new warnings from Settings rather than copying from an existing warning.
</blockquote></details>

Removed, made Batfish.buildWarning(Settings settings) as public static

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1076#-:-L81Ai4xMKFBAqflngpa:b-i6sm79)*

dhalperi:
:lgtm:

---

Reviewed 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1076#-:-L81Fis-LVhCWN-gjq61:bnfp4nl)*

PR #1097, opened 2018-03-23
Fixing AWS refs for elasticsearch and RDS

haverma:
- Changed the VPC ID in the RDS security group to match RDS's VPC
- Changed Elasticsearch's security to a security group in the same VPC as Elasticsearch
- Removed a security group with a duplicate SG-ID which was only used by RDS as RDS is not using it now


dhalperi:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/batfish/batfish/1097#-:-L8Jw4ue92y-LBxax7n-:bnfp4nl)*

PR #1107, opened 2018-03-28
Add ReachabilityQuestion constraining whether result must be srcNatted

arifogel:
- If `srcNatted == true `, then final header must have different source IP than original header.
- If `srcNatted == false `, then final header must have same source IP as original header.
- If `srcNatted == null `, then no additional constraints are imposed on the header.


ratulm:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1107#-:-L8fH0ssB1tLMyY29G3G:bnfp4nl)*

PR #1115, opened 2018-03-30
Cisco_interface: select inlining of null_rest_of_line

dhalperi:
As expected, most CPU time is spent in null_rest_of_line. Inlining it in the very large
null rules in interface actually saves a lot of parsing overhead. In most other cases,
it doesn't matter.

Reduces parsing of one large network by about 10%-15%. This would be even better if we
removed more clauses from if_null, if_null_inner, if_null_block, if_null_single and
so had fewer paths to these ~NEWLINE* blocks.


sfraint:
Good find!
:lgtm: 

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1115#-:-L96UXp2CCCdSt5ZYb7S:b82vdw4)*

PR #1151, opened 2018-04-11
Allowing access-groups to also accept mac-acls

haverma:
Cisco does not mentions anywhere that only IP access lists are allowed with access-groups (https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/qos/command/qos-cr-book/qos-m1.html#wp7738095440),
So we should support mac-acl as well

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
@@ -2645,7 +2645,7 @@ public class CiscoControlPlaneExtractor extends CiscoParserBaseListener
       line = ctx.num.getLine();
     }
     _configuration.referenceStructure(
-        CiscoStructureType.IP_ACCESS_LIST, name, CiscoStructureUsage.CLASS_MAP_ACCESS_GROUP, line);
+        CiscoStructureType.ACCESS_LIST, name, CiscoStructureUsage.CLASS_MAP_ACCESS_GROUP, line);
     _configuration.getClassMapAccessGroups().add(name);
   }
 
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
@@ -1048,6 +1048,18 @@ public final class CiscoConfiguration extends VendorConfiguration {
         cable != null ? cable.getDocsisPolicyRules() : null);
   }
 
+  private void markIpOrMacAcls(CiscoStructureUsage usage) {
+    markStructure(
+        CiscoStructureType.ACCESS_LIST,
+        usage,
+        Arrays.asList(
+            _extendedAccessLists,
+            _standardAccessLists,
+            _extendedIpv6AccessLists,
+            _macAccessLists,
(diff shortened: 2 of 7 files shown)


arifogel:
Reviewed 4 of 5 files at r1, 1 of 1 files at r2.

---

*[a discussion](https://beta.reviewable.io/reviews/batfish/batfish/1151#-L9pyJYBC_wxgM0XTeyF:-L9pyJYBC_wxgM0XTeyG:bmboyhg) (no related file):*
Please add junit tests for the new behavior. You can add to `CiscoGrammarTest`. I believe there are existing tests there that check for undefined/unused structures that you can use as templates.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1151)*

saparikh:
I need to look into this some more, but my read of the doc listed above (specifically this section here) makes me believe that only IP access-lists are supported for QoS policies, at least for the ISR and ASR routers:

```

Cisco ASR 1000 Series Aggregation Services Routers
Cisco ASR 1000 Series Routers do not support more than 16 match statements per class map. An interface with more than 16 match statements rejects the service policy.

Examples
The following example shows how to specify a class map named acl144 and to configure the ACL numbered 144 to be used as the match criterion for that class:

Device(config)# class-map acl144 
Device(config-cmap)# match access-group 144
The following example shows how to define a class map named c1 and configure the ACL numbered 144 to be used as the match criterion for that class:
Device(config)# class-map type inspect match-all c1
Device(config-cmap)# match access-group 144
Cisco Performance Monitor in Cisco IOS Release 15.1(3)T and 12.2(58)SE
The following example shows how to configure a service policy for the Performance Monitor in policy inline configuration mode. The policy specifies that packets traversing Ethernet interface 0/0 must match ACL144.

Device(config)# interface ethernet 0/0 
Device(config-if)# service-policy type performance-monitor inline input
Device(config-if-spolicy-inline)# match access-group name ACL144 
Device(config-if-spolicy-inline)# exit

Related Commands
Command | Description
-- | --
access-list (IP extended) | Defines an extended IP access list.
access-list (IP standard) | Defines a standard IP access list.
class-map | Creates a class map to be used for matching packets to a specified class.
match access-group | Configures the match criteria for a class map on the basis of the specified ACL.
match class-map | Uses a traffic class as a classification policy.
match input-interface | Configures a class map to use the specified input interface as a match criterion.
match mpls experimental | Configures a class map to use the specified EXP field value as a match criterion.
match protocol | Configures the match criteria for a class map on the basis of the specified protocol.
service-policy type performance-monitor | Associates a Performance Monitor policy with an interface.

```

But on the Catalyst switches running IOS this doc suggests that MAC ACLs are valid:

https://www.cisco.com/c/en/us/support/docs/switches/catalyst-4500-series-switches/200179-Use-MAC-ACL-for-Layer-2-Control-Frames-o.html

arifogel:
If this is true, then we need to check the format to know which `StructureUsage` is applicable at post-processing and conversion time.

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1151#-:-L9qqFtl3wQB9iAxJpB9:bz4k0jv)*

arifogel:
... and`StructureType`

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1151#-:-L9qqcX01NY3Ice8rKyn:brzychf)*

PR #1157, opened 2018-04-12
do not pair interfaces that have the same address

ratulm:


arifogel:
Reviewed 1 of 1 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 1178 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1157#-L9v26xm54XpTZV2d8gs:-L9v26xnB4T0QtMyk_AU:bhkhm2o) ([raw file](https://github.com/batfish/batfish/blob/485eef816dd4f78ff0a57ee60cd1651bfa563e51/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L1178)):*
> ```Java
>           }
>         });
>     for (Set<Interface> bucket : prefixInterfaces.values()) {
> ```

I believe this will introduce a bug. Interfaces are keyed by name. Two interfaces with the same name (on different routers) will never end up in this set. That is why I previously used NodeInterfacePair. Please revert.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1157)*

ratulm:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 1178 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1157#-L9v26xm54XpTZV2d8gs:-LA17w75BLRo9dKJck9H:bu5xbio) ([raw file](https://github.com/batfish/batfish/blob/485eef816dd4f78ff0a57ee60cd1651bfa563e51/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L1178)):*

I believe this will introduce a bug. Interfaces are keyed by name. Two interfaces with the same name (on different routers) will never end up in this set. That is why I previously used NodeInterfacePair. Please revert.
</blockquote></details>

simply reverting does not solve the problem of course. using list instead of set solves the problem?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1157)*

arifogel:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 1178 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1157#-L9v26xm54XpTZV2d8gs:-LAENTKAC9YQudJALavI:bd0eblx) ([raw file](https://github.com/batfish/batfish/blob/485eef816dd4f78ff0a57ee60cd1651bfa563e51/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L1178)):*

simply reverting does not solve the problem of course. using list instead of set solves the problem?
</blockquote></details>

I mean just revert to using `NodeInterfacePair`. But yeah I think `List`is actually the best solution.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1157#-:-LAENvqm8RSbYolncyL9:bnfp4nl)*

PR #1167, opened 2018-04-14
Support for dynamic BGP sessions

progwriter:
Support for dynamic BGP peering.

Key points:
- When constructing the data model, allow the listening end of the peering to have `localIp` of `Ip.AUTO` *if* `update-source` is not used (at least for cisco-flavor devices)
- Move towards using a bgp topology instead of `_remoteNeighbor`
     - Instead of having a single `remoteBgpNeighbor` inside `BgpNeighbor` , compute a topology where `BgpNeighbor`s are vertices and `BgpSession`s are edges, make that topology immutable
     - Topology computation supports keeping invalid neighbors (e.g., for hygiene checks) and doing reachability checks (for dataplane computation)
     - Single hop eBGP being single hop is also enforced now.
- Update data plane computation to use said topology
- Changes to `NeighborsQuestion` and `BgpSessionStatus` questions to utilize the above BGP topology.

- Side effects:
    - Flattened the answer hierarchy for `BgpSessionStatus`, seemed pointless to have nested dictionaries there.
    - Speed up data plane computation by re-initing bgp topology only twice: once after IGP computation; and once after a single iteration of non-monotonic computation.
    - Because we are now more stringent about BGP reachability, example network (and the delta version) had to be modified with new permit ACLs to match the desired refs.
    - Many of the ref changes have to do with new ACLs


saparikh:
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 694 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1167#-LAEjyXm1UKfCD0PW0ct:-LAEw6n32TF0Jv5FNLBU:b-9gj9k3) ([raw file](https://github.com/batfish/batfish/blob/ba47b07275bf9bb003c6e6bf7269b75ec07556a2/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L694)):*

@saparikh Shouldn't you also require that this connection occur via a `ConnectedRoute`?
</blockquote></details>

Need to test this, but I can envision a scenario (it would be a bad idea, IMO) where you are peering to secondary IP addresses that are configured on the interfaces and they are not in the same subnet, so each device has a static route pointing to the next-hop interface for that IP address. Something like this:

Router 1:
interface Eth0
  ip address 10.3.1.1/31
  ip address 10.2.1.2/32 sec

router bgp 1
  neighbor 10.1.1.1 remote-as 2

ip route 10.1.1.1 255.255.255.255 Eth0

Router2:
interface Eth0
  ip address 10.3.1.0/31
  ip address 10.1.1.1/32 sec

router bgp 2
  neighbor 10.2.1.2 remote-as 1

Technically, this would be an eBGP single-hop connection. But like I said it would be a really bad idea and a nightmare to troubleshoot.

Another potential scenario could be:

Router 1:
interface Loopback 0
  ip address 10.2.1.2/32
interface Eth0
  ip unnumbered Loopback0

router bgp 1
  neighbor 10.1.1.1 remote-as 2

ip route 10.1.1.1 255.255.255.255 Eth0

Router2:
interface Loopback 0
  ip address 10.1.1.1/32
interface Eth0
  ip unnumbered Loopback 0

router bgp 2
  neighbor 10.2.1.2 remote-as 1

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1167)*

arifogel:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 694 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1167#-LAEjyXm1UKfCD0PW0ct:-LAF-bJN6v7l1NvUdZ6y:b7vpnpi) ([raw file](https://github.com/batfish/batfish/blob/ba47b07275bf9bb003c6e6bf7269b75ec07556a2/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L694)):*
<details><summary><i>Previously, saparikh (Samir Parikh) wrote…</i></summary><blockquote>

Need to test this, but I can envision a scenario (it would be a bad idea, IMO) where you are peering to secondary IP addresses that are configured on the interfaces and they are not in the same subnet, so each device has a static route pointing to the next-hop interface for that IP address. Something like this:

Router 1:
interface Eth0
  ip address 10.3.1.1/31
  ip address 10.2.1.2/32 sec

router bgp 1
  neighbor 10.1.1.1 remote-as 2

ip route 10.1.1.1 255.255.255.255 Eth0

Router2:
interface Eth0
  ip address 10.3.1.0/31
  ip address 10.1.1.1/32 sec

router bgp 2
  neighbor 10.2.1.2 remote-as 1

Technically, this would be an eBGP single-hop connection. But like I said it would be a really bad idea and a nightmare to troubleshoot.

Another potential scenario could be:

Router 1:
interface Loopback 0
  ip address 10.2.1.2/32
interface Eth0
  ip unnumbered Loopback0

router bgp 1
  neighbor 10.1.1.1 remote-as 2

ip route 10.1.1.1 255.255.255.255 Eth0

Router2:
interface Loopback 0
  ip address 10.1.1.1/32
interface Eth0
  ip unnumbered Loopback 0

router bgp 2
  neighbor 10.2.1.2 remote-as 1


</blockquote></details>

Let's defer for now. Please file issue and reference here.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1167)*

dhalperi:
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoLexer.tokens, line 1 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1167#-LAElZ3H84NIq-Ui9jDN:-LAGDftr2FLjrQ0_EKFJ:bhmkqp3) ([raw file](https://github.com/batfish/batfish/blob/ba47b07275bf9bb003c6e6bf7269b75ec07556a2/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoLexer.tokens#L1)):*

Accidental add. Token files are not in the .gitignore, we should probably add them.
</blockquote></details>

Token files are deliberately not in the ignore. Some stupid tools create them, and if present they will screw up antlr.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1167)*

arifogel:
:lgtm:

---

Reviewed 8 of 8 files at r2.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1167#-:-LAGJe9QB1OxLCuN9O7M:bnfp4nl)*

PR #1181, opened 2018-04-17
Hotfixing the bdp computation

progwriter:
compute advertisements to outside can fail sometimes when applying routing policy; hotfixing for now, better solution in the works.


progwriter:
Closing because a proper fix is in #1201

PR #1203, opened 2018-04-18
serializable ipspace

arifogel:


dhalperi:
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 459 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1203#-LAPfbO27EFwD7RYrOYu:-LAPfbO27EFwD7RYrOYv:bbuz1km) ([raw file](https://github.com/batfish/batfish/blob/a57240a146752879fa4d2a509ec3880b56d7467f/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L459)):*
> ```Java
>   public NavigableMap<String, IpSpace> getIpSpaces() {
>     return _ipSpaces;
>   }
> ```

Rather than a custom deserializer, looks to me like we should just use a class that holds an IpSpace. It would get serialized as something like:

```
{
    "ip":  "1.2.3.4"
}
```

with different field names for the different types of IpSpace. Then it might have an unserialized getter getIpSpace that returns the right thing.

What am I missing?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1203)*

arifogel:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 459 at r1](https://beta.reviewable.io/reviews/batfish/batfish/1203#-LAPfbO27EFwD7RYrOYu:-LAPr3UvEOMXDa_j_LSt:bmi4tvo) ([raw file](https://github.com/batfish/batfish/blob/a57240a146752879fa4d2a509ec3880b56d7467f/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L459)):*

Won't work for `AclIpSpace`, which recursively contains `IpSpace`.
</blockquote></details>

Furthermore, every serializable client of IpSpace would instead have to directly use your wrapper.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1203)*

dhalperi:
Passes a very rote check.

---

Reviewed 36 of 36 files at r2.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixIpSpace.java, line 8 at r2](https://beta.reviewable.io/reviews/batfish/batfish/1203#-LAQaRZv3LGuYM6-6Aie:-LAQaRZv3LGuYM6-6Aif:b-69soam) ([raw file](https://github.com/batfish/batfish/blob/9b46af998ea8ebec626bdddb447cbd43a0ee21fc/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixIpSpace.java#L8)):*
> ```Java
> import org.batfish.datamodel.visitors.GenericIpSpaceVisitor;
> 
> public class PrefixIpSpace extends IpSpace {
> ```

many of these should be final unless you intend them to be extended.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1203#-:-LAQawir3WeXeRWftjUX:b7fviyx)*

PR #1219, opened 2018-04-23
Roleplay: Supporting multiple role dimensions

ratulm:
This PR adds support for multiple role dimensions, some of which may be automatically inferred and some may be customized by the user. The implementation involved several changes:
- New classes in org.batfish.role
- Removed the NodeRoleSpecifier class (as that does not support multiple role types). The closest kin of that class is the new NodeRoleDimension class.
- Remove roles from Configuration. Roles do not belong there because they are not embedded in the configuration and keeping them there makes it hard to work manage them out of band
- Changes to NodesSpecifier were needed as a result of change above 
 - All questions that handle roles have a roleDimension parameter that specifies which role dimension to operate upon


millstein:
Review status: 0 of 73 files reviewed at latest revision, all discussions resolved.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/NodesSpecifier.java, line 109 at r3](https://beta.reviewable.io/reviews/batfish/batfish/1219#-LAn6-izAhmfxuJVvchg:-LAn6-izAhmfxuJVvchh:b3av5o8) ([raw file](https://github.com/batfish/batfish/blob/a3427ee5b2434bd11f91d8513d1232c9da19f2a1/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/NodesSpecifier.java#L109)):*
> ```Java
>         return getMatchingNodesByName(nodes);
>       case ROLE:
>         NodeRoleDimension roleDimension = batfish.getNodeRoleDimension(_roleDimension);
> ```

Will this call do the right thing even if _roleDimension is null?  What about all the other places where _roleDimension is used?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219)*

millstein:
Review status: 0 of 73 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/NodesSpecifier.java, line 135 at r3](https://beta.reviewable.io/reviews/batfish/batfish/1219#-LAn6PxFANSY1HyLKrvX:-LAn6PxFANSY1HyLKrvY:b-703wd7) ([raw file](https://github.com/batfish/batfish/blob/a3427ee5b2434bd11f91d8513d1232c9da19f2a1/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/NodesSpecifier.java#L135)):*
> ```Java
> 
>   @JsonIgnore
>   public String getRoleDimension() {
> ```

Consider having this return Optional<String> so it's explicit to callers that the result may be null and they need to deal with that.  Seems less fragile than propagating the potential for null dereference.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219)*

millstein:
Review status: 0 of 73 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/question/src/main/java/org/batfish/question/InferPoliciesQuestionPlugin.java, line 393 at r3](https://beta.reviewable.io/reviews/batfish/batfish/1219#-LAnAEcd0uvGNe3JdXRb:-LAnAEcd0uvGNe3JdXRc:bblxm7m) ([raw file](https://github.com/batfish/batfish/blob/a3427ee5b2434bd11f91d8513d1232c9da19f2a1/projects/question/src/main/java/org/batfish/question/InferPoliciesQuestionPlugin.java#L393)):*
> ```Java
>               NamedStructureRoleConsistencyQuestion policy =
>                   new NamedStructureRoleConsistencyQuestion(
>                       name, hypothesis, question.getRoleDimension());
> ```

One difference now is that here and in other questions we have a String that identifies an existing role dimension, so that means you can't use these questions with your own NodeRoleDimension that you may want to explore.  (Previously people could pass in their own NodeRoleSpecifier.)  I.e., to explore a new dimension you'd have to first add it to the container's set of role dimensions (is there a way to do that?).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219)*

millstein:
Review status: 0 of 73 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java, line 408 at r3](https://beta.reviewable.io/reviews/batfish/batfish/1219#-LAnBeC_9k_0KRDOF312:-LAnBeC_9k_0KRDOF313:bt42ja8) ([raw file](https://github.com/batfish/batfish/blob/a3427ee5b2434bd11f91d8513d1232c9da19f2a1/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java#L408)):*
> ```Java
>     nodeRolesMap.add(
>         new NodeRoleDimension(
>             NodeRoleDimension.AUTO_DIMENSION_PRIMARY, inferredRoles, Type.AUTO, null));
> ```

Instead of null, in the last argument here you should make a singleton list containing regex and pass that as the last argument.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219)*

millstein:
:lgtm:

---

Review status: 0 of 73 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219#-:-LAnaZjcDjW3HccL05lG:bnfp4nl)*

millstein:
Reviewed 8 of 20 files at r1, 70 of 74 files at r3, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/batfish/batfish/1219)*

PR #1259, opened 2018-04-28
Bgpsessionanalyzer

ratulm:
Some refactoring to allow code reuse and a layer of analysis on the answer


progwriter:
Reviewed 5 of 7 files at r1, 2 of 2 files at r2.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 127 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDdJJfCAVzYeO7NCuV:-LBDdJJfCAVzYeO7NCuW:bjhwpy5) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L127)):*
> ```Java
> 
>   public Object fromRow(ObjectNode o) throws JsonProcessingException {
>     throw new BatfishException("This base method should not be called!");
> ```

Seems like it's worth making this an abstract method (& class) instead of throwing.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 173 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDfJF_Da2UxNyl7-w0:-LBDfJF_Da2UxNyl7-w1:b-ewdf62) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L173)):*
> ```Java
> 
>   public ObjectNode toRow(Object object) {
>     throw new BatfishException("This base method should not be called!");
> ```

Similarly, abstract method here.

---

*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionInfo.java, line 215 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDeCRz-p2ptbjLvv-P:-LBDeCRz-p2ptbjLvv-Q:btqft6t) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionInfo.java#L215)):*
> ```Java
>     public BgpSessionInfo build() {
>       BgpSessionInfo bgpSessionInfo =
>           new BgpSessionInfo(null, null, null, null, null, null, null, null, null);
> ```

This seems like asking for trouble (NPEs). Please pass values in the constructor. This way you can make members of BgpSessionInfo `final` as well, which goes well with our general push for immutability.
Finally, if you also order these alphabetically, it'll be easier to maintain.

---

*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java, line 115 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDeiQmBHLTOEMHSbuY:-LBDeiQmBHLTOEMHSbuZ:b-jb90d8) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java#L115)):*
> ```Java
>   @Override
>   public ObjectNode toRow(Object o) {
>     return toRowStatic((BgpSessionInfo) o);
> ```

Is there a safe(r) way to handle a type mismatch here? Otherwise we might just get a casting exception.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1259)*

ratulm:
Review status: 3 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 127 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDdJJfCAVzYeO7NCuV:-LBDs-mPEdI_s_dzZ-Mv:b48irta) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L127)):*

Seems like it's worth making this an abstract method (& class) instead of throwing.
</blockquote></details>

The test class was instantiating it so I got lazy. Handled now.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 173 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDfJF_Da2UxNyl7-w0:-LBDs5S17hKae3xADbl5:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L173)):*

Similarly, abstract method here.
</blockquote></details>

Done.

---

*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionInfo.java, line 215 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDeCRz-p2ptbjLvv-P:-LBDsAFK1KfOynZNWWDX:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionInfo.java#L215)):*

This seems like asking for trouble (NPEs). Please pass values in the constructor. This way you can make members of BgpSessionInfo `final` as well, which goes well with our general push for immutability.
Finally, if you also order these alphabetically, it'll be easier to maintain.
</blockquote></details>

Done.

---

*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java, line 115 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDeiQmBHLTOEMHSbuY:-LBDsCX78nC4eL_jM1-v:b-mxq2el) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java#L115)):*

Is there a safe(r) way to handle a type mismatch here? Otherwise we might just get a casting exception.
</blockquote></details>

What are you thinking?  Something awful has happened if this function gets an object of another type, and even if we check explicitly, I wouldn't know what to do except for throwing up my hands (which the runtime is doing for me anyway). 

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1259)*

progwriter:
:lgtm:

---

Reviewed 7 of 7 files at r3.

---

*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java, line 115 at r1](https://reviewable.io/reviews/batfish/batfish/1259#-LBDeiQmBHLTOEMHSbuY:-LBDsw_PA_uE92O7BoUG:bl8isdk) ([raw file](https://github.com/batfish/batfish/blob/b66c8caa92ebf886518a017a679c8bc7773560a0/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionStatusAnswerElement.java#L115)):*

What are you thinking?  Something awful has happened if this function gets an object of another type, and even if we check explicitly, I wouldn't know what to do except for throwing up my hands (which the runtime is doing for me anyway). 
</blockquote></details>

Two options I can think of:
- Throw a checked exception so layers above can deal with it (and potentially they have more context to do so)
- Return an ObjectNode containing some representative error message.
However, if neither makes sense here, and we really should just go ¯\_(ツ)_/¯ then leave as is.
I'm marking comment as non-blocking.

Finally, from the SE standpoint I think nthere is another option here, with `TableAnswerElement` being templated, and then `BgpSessionStatusAnswerElement extends TableAnswerElement<BgpSessionInfo> and you get the benefits/checks that come with templated member functions. This would be a completely different PR though.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1259#-:-LBDwMwA7BCoDTCJnu25:bnfp4nl)*

PR #1262, opened 2018-04-29
Make proxy arp a primitive boolean for interfaces

progwriter:
This helps avoid NPEs in `ForwardingAnalysisImpl`.

The way we are handling it right now, this seems to fall in line with defaults. We enable it for IOS and ASA when creating the interface, and it will resolve to `false` for [NXOS](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/sw/5_x/nx-os/unicast/configuration/guide/l3_cli_nxos/l3_ip.html#19966) and [Juniper](https://www.juniper.net/documentation/en_US/junos/topics/task/configuration/interfaces-configuring-restricted-unrestricted-proxy-arp.html)


arifogel:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1262#-:-LBE69OW8kCmh1ZUFvvd:bnfp4nl)*

PR #1267, opened 2018-04-30
pom.xml: enforce that javac in 1.8.X is used.

### LINKED ISSUE #1254: Does not compile with Java 10
Fails to compile with Java 10. Java 9 is untested.

dhalperi:
Fix #1254

### DIFF
--- a/projects/pom.xml
+++ b/projects/pom.xml
@@ -401,6 +401,9 @@
             <configuration>
               <rules>
                 <dependencyConvergence/>
+                <requireJavaVersion>
+                  <version>[1.8.0,1.8.99999)</version>
+                </requireJavaVersion>
                 <requireMavenVersion>
                   <version>3.3.9</version>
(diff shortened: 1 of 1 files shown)


dhalperi:
*[projects/pom.xml, line 405 at r1](https://reviewable.io/reviews/batfish/batfish/1267#-LBN4xjm9AbDWMScrHzg:-LBN5TW-CC9AbpeKGxS0:bp0xdr8) ([raw file](https://github.com/batfish/batfish/blob/ccab336c3e33031fc8243b3134d0344ff84d321c/projects/pom.xml#L405)):*

If you are going to use open interval, why not just `[1.8.0,1.9.0)`?
</blockquote></details>

1.9.0-alpha1 is less than 1.9.0, so this does not work.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1267)*

PR #1272, opened 2018-05-01
Refactor answerAclReachability and avoid encoding extra ACLs

corinaminer:
Fixes first issue described [here](https://docs.google.com/document/d/1CsLM-K9ZJYH1n1gAweSO6bDl4TmIKUdadDnFLnspDLM/edit#).

In the first batch of NoD jobs, `synthesizeAcls()` was encoding all ACLs in each device even though it's called once per ACL, and only needs to encode that ACL. Fixed by passing a set of `aclsToSkip` into `synthesizeAcls()` and then using `setDisabledAcls()` in the `Synthesizer` builder.


arifogel:
Reviewed 1 of 1 files at r1.

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1272#-LBS1Klq2fnVrTMN_KTU:-LBS1Klr7LCHBIhNyNfg:bb9urwe) (no related file):*
I'm having a hard time understanding what ACLs are skipped and when. Can you please add comments explaining the process?
Related: I don't understand why you are using a blacklist rather than a whitelist, i.e. why use `aclsToSkip` over `aclsToProcess`?

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 792 at r1](https://reviewable.io/reviews/batfish/batfish/1272#-LBS1liO-rZuat92_4bq:-LBS1liO-rZuat92_4br:b-jvwtqy) ([raw file](https://github.com/batfish/batfish/blob/541a8fdcd6b819872d9efce2f6ff126ace083d77/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L792)):*
> ```Java
>       }
>       // skip juniper srx inbound filters, as they can't really contain
>       // operator error (todo: check that this is actually be the case)
> ```

TODO comment should be accompanied by link to issue so someone will eventually get to it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

corinaminer:
Review status: 0 of 1 files reviewed at latest revision, 5 unresolved discussions.

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1272#-LBS1Klq2fnVrTMN_KTU:-LBT2g0W3c_BJnCW893q:b-onu05i) (no related file):*

I'm having a hard time understanding what ACLs are skipped and when. Can you please add comments explaining the process?
Related: I don't understand why you are using a blacklist rather than a whitelist, i.e. why use `aclsToSkip` over `aclsToProcess`?
</blockquote></details>

Added some comments, but here's an overview:
- To generate the first set of NoD jobs, it loops through the ACLs in `aclEqSets` and builds a synthesizer for each one. Those synthesizers used to each encode all ACLs on the given ACL's host; now they each only encode the current ACL.
- In r2, thanks to Matt's comment, `generateEarliestMoreGeneralAclLineJobs` also encodes fewer ACLs. In this case it's looping through hostnames, then ACLs within each hostname, so it was building fewer synthesizers (1 per hostname) and encoding all the ACLs for that hostname. Now it's only encoding the subset of those ACLs that have unreachable lines.

Why I'm using a blacklist rather than a whitelist -- great question. The answer is that `SynthesizerInputImpl.java` contains a method `Builder setDisabledAcls()`, but not a method `Builder setEnabledAcls()`. I looked into the class briefly and saw that there is an `_enabledAcls` property, but it's strictly attained with `computeEnabledAcls()`. Seems changeable but I wanted to limit the scope of this PR.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 778 at r1](https://reviewable.io/reviews/batfish/batfish/1272#-LBRvnyq-UEV-ncpoSa1:-LBSEOzt2U1kvaCaXSVE:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/541a8fdcd6b819872d9efce2f6ff126ace083d77/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L778)):*

Can this have a more descriptive name? Perhaps something like `generateUnreachableAclLineJobs`.
</blockquote></details>

Done.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 792 at r1](https://reviewable.io/reviews/batfish/batfish/1272#-LBS1liO-rZuat92_4bq:-LBSF_MB7KatPPlVYATn:b-axrrfs) ([raw file](https://github.com/batfish/batfish/blob/541a8fdcd6b819872d9efce2f6ff126ace083d77/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L792)):*

TODO comment should be accompanied by link to issue so someone will eventually get to it.
</blockquote></details>

Done (link: #1275).

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 823 at r1](https://reviewable.io/reviews/batfish/batfish/1272#-LBRva0F6uaQGKfkN43L:-LBSEO0X0tQ27sS0K00d:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/541a8fdcd6b819872d9efce2f6ff126ace083d77/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L823)):*

Can this have a more descriptive name? Perhaps something like `generateEarliestMoreGeneralAclLineJobs`.
</blockquote></details>

Done.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 838 at r1](https://reviewable.io/reviews/batfish/batfish/1272#-LBRwIIH9_yaq2EKijru:-LBSDgrO4xY3QoRhsUEG:blcmevx) ([raw file](https://github.com/batfish/batfish/blob/541a8fdcd6b819872d9efce2f6ff126ace083d77/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L838)):*

Does this mean ACLs with no unreachable lines will be encoded?
</blockquote></details>

that's a great point. Updated this to avoid encoding ACLs with no unreachable lines. Also added an if statement in the loop below so that it skips those ACLs when looking for unreachable lines.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

arifogel:
Reviewed 1 of 1 files at r2.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 689 at r2](https://reviewable.io/reviews/batfish/batfish/1272#-LBTEKqU3oV-VzgnJNUR:-LBTEKqU3oV-VzgnJNUS:b-vw9opb) ([raw file](https://github.com/batfish/batfish/blob/acbf517ac8b35f12794da0ca744878e705736a3d/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L689)):*
> ```Java
>       // Keep track of which ACLs do/don't have unreachable lines
>       if (!e.getValue()) {
>         aclsWithUnreachableLines.add(aclName);
> ```

Since you are using a set, it looks like you can only have one acl per acl-name - regardless of whether those same-named ACLs are different on different hosts. That is, unless `aclName` is qualified by hostname.
It looks like you are using `String`s and `Set`s all over the place here. Please be certain you are not accidentally losing elements via name collisions.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

corinaminer:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 689 at r2](https://reviewable.io/reviews/batfish/batfish/1272#-LBTEKqU3oV-VzgnJNUR:-LBbGnXK--x7MIW8Bswg:b-g6udbs) ([raw file](https://github.com/batfish/batfish/blob/acbf517ac8b35f12794da0ca744878e705736a3d/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L689)):*

Since you are using a set, it looks like you can only have one acl per acl-name - regardless of whether those same-named ACLs are different on different hosts. That is, unless `aclName` is qualified by hostname.
It looks like you are using `String`s and `Set`s all over the place here. Please be certain you are not accidentally losing elements via name collisions.
</blockquote></details>

This is a good point. I was thinking the distinction between same-named ACLs on different hosts had already been lost at this point (but it hasn't). Pushed another commit to give an idea of what I'm trying now; it works to delineate same-named ACLs, but I want to make sure it runs in a reasonable amount of time in comparison to current master.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

arifogel:
Reviewed 1 of 1 files at r3.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 864 at r3](https://reviewable.io/reviews/batfish/batfish/1272#-LBbU9Pq8yF8ZE2Du-ul:-LBbU9Pq8yF8ZE2Du-um:b-qvuxbi) ([raw file](https://github.com/batfish/batfish/blob/a8583668333e07b52b8e00b9529196ae67300911/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L864)):*
> ```Java
>     for (Entry<String, Map<String, List<AclLine>>> e : unreachableAclLinesMap.entrySet()) {
>       String hostname = e.getKey();
>       System.out.println("Hostname: " + hostname);
> ```

This must be removed before merge. Others as well below.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

corinaminer:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 864 at r3](https://reviewable.io/reviews/batfish/batfish/1272#-LBbU9Pq8yF8ZE2Du-ul:-LBbvz9J3f2tQnWy8yVm:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a8583668333e07b52b8e00b9529196ae67300911/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L864)):*

This must be removed before merge. Others as well below.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272)*

corinaminer:
Put a 100-count loop into answerAclReachability (so each time you run it, it calculates the results 100 times) and added log statements to record the runtime in both this branch and master.  Ran a 2-tailed t-test on the two groups of data and found that this branch runs it faster with a p-value of 0.0043 (master took 270 ± 110 ms and skip-acls-2 took 310 ± 120 ms).

I'm satisfied that I haven't screwed things up in terms of runtime, so as long as memory usage is comparable to before, so I'm no longer blocking merging.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272#-:-LBc0eDeDQdh_O_YXvhd:b-gnhnm4)*

corinaminer:
oops, @anothermattbrown has astutely pointed out that 270 < 310. I had the branches switched in the first comment. Actual results:
- skip-acls-2 took 270 ± 110 ms
- master took 310 ± 120 ms

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272#-:-LBc8N2p6wCFP8utgXjY:b-xhop0s)*

arifogel:
:lgtm:, but I suggest waiting for explicit OK from @anothermattbrown 

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272#-:-LBcq5oA24Ao3g-C2N_q:b-b73mpu)*

anothermattbrown:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1272#-:-LBfyYo82H5CcCPdJbNq:bnfp4nl)*

PR #1279, opened 2018-05-02
CiscoConfiguration: one fewer statement in BGP common export policy

dhalperi:
I know this PR will fail initially b/c the refs are not going to pass.

My question is more educational: the previous design of

```
if(disjunction, return True, fallthrough)
return False
```

seems silly. Is there a design or implementation reason why it it actually is a good idea to structure it as before?


dhalperi:
After discussion, seems better for readability to not inline these things.

dhalperi:
(We can always optimize after the fact)

PR #1285, opened 2018-05-03
Change row type

ratulm:
as you suspected, ObjectNode as the type of row was not a good long-term fit. hiding, mostly not entirely, its internal implementation now.

PR #1291, opened 2018-05-03
Add the IpOwners question

progwriter:
Add a very simple question that allows extracting IpOwners (as computed by `CommonUtil.computeIpOwners`)


ratulm:
@progwriter - we are moving away from this free form output format and looking to produce tabular answers. see the new bgpsessionstatus (especially after #1285) as an example. also, is the uniqueipassignments question producing basically the same information in a different format. my recommendation would then be to combine the two ... feel free to assign that to me  if things are not clear or you don't have time. 

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1291#-:-LBdS0PS2_uuUn7B0G6Q:bgbvhs9)*

progwriter:
@ratulm No problem, happy to combine and "upgrade" the two questions. They are indeed similar, with difference being that UniqueIpAssignments has extra info (interface names) but filters the answer to return duplicates only.

Let's touch base in person so I understand your philosophy behind #1285 before I proceed.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1291#-:-LBdXxbIA7RgWoNU_Iqm:b-h20gvh)*

PR #1321, opened 2018-05-05
CiscoConfiguration: only create BGP_NETWORK6_NETWORKS_FILTER if there are IPv6 networks

dhalperi:
Less clutter in the data model, it's never called anyway, and we definitely shouldn't create it on devices where IPv6 address family isn't even enabled.


arifogel:
:lgtm:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1321#-:-LBrUpac2glacgJ625ug:bnfp4nl)*

PR #1348, opened 2018-05-09
Allow installation of covered static routes in the main RIB

progwriter:
*Context*:
- When activating static routes with next-hop-ip we check that next hop ip is routable.
- If main rib contained a route that fully covered the static route we did not install/activate the static route

*Problem*:
It seems like that is not the desired behavior. After firing up GNS3, confirmed that on IOS all static routes get activated (if covered by other static routes or by a connected route)

*Fix*:
Removed the `containsPrefix` check, modeled the test based on the GNS3 setup.


dhalperi:
Per discussion, this is not yet the right solution. Closing in favor of future work.

PR #1349, opened 2018-05-09
Fix a bgpStatus null pointer

progwriter:
If remote neighbor does not match the node2Regex we should `continue` to avoid evaluating `matchesStatus`


dhalperi:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1349)*

PR #1383, opened 2018-05-14
added javadoc

ratulm:


dhalperi:
Reviewed 1 of 1 files at r1.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 126 at r1](https://reviewable.io/reviews/batfish/batfish/1383#-LCVWHZG29Q0W4eB9b67:-LCVWHZG29Q0W4eB9b68:brv8ap4) ([raw file](https://github.com/batfish/batfish/blob/6e48da95df1bb051d4ad8113ca66a75741c4b528/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L126)):*
> ```Java
>   /**
>    * Converts the {@link Row} object into the actual object that is produced by concrete (child)
>    * TableAnswerElement. This method is overriden by the child classes that are aware of what is
> ```

So is `Object` here a proxy for some type of `TableAnswerElement`? If so, we should consider using types to bound it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1383)*

ratulm:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 126 at r1](https://reviewable.io/reviews/batfish/batfish/1383#-LCVWHZG29Q0W4eB9b67:-LCVWy9o5cqi7jKbPPRE:b-fn8006) ([raw file](https://github.com/batfish/batfish/blob/6e48da95df1bb051d4ad8113ca66a75741c4b528/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L126)):*

So is `Object` here a proxy for some type of `TableAnswerElement`? If so, we should consider using types to bound it.
</blockquote></details>

Object is a proxy for a row of the table, or more precisely, the object (generated by the Answerer) that is "flattened" into a row of the table. There are no constraints on this object at the moment, as long as the Answerer can serialize/deserialize  to Row. So, no natural base type exists. If that makes sense, I can do an empty abstract class or interface (to be constrained later). 

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1383)*

dhalperi:

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java, line 126 at r1](https://reviewable.io/reviews/batfish/batfish/1383#-LCVWHZG29Q0W4eB9b67:-LCVqTgf7w41f4jM-nWS:bncnwrf) ([raw file](https://github.com/batfish/batfish/blob/6e48da95df1bb051d4ad8113ca66a75741c4b528/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/TableAnswerElement.java#L126)):*

Object is a proxy for a row of the table, or more precisely, the object (generated by the Answerer) that is "flattened" into a row of the table. There are no constraints on this object at the moment, as long as the Answerer can serialize/deserialize  to Row. So, no natural base type exists. If that makes sense, I can do an empty abstract class or interface (to be constrained later). 
</blockquote></details>

Summarizing our whiteboard conversation:

It seems like the internal object representation of an individual question is just an implementation detail of a question; instead, questions should expose answers as iterables/lists/etc. of `Row` and then the logic that processes exclusions and assertions and summaries can operate directly on Rows.

For the questions that also wish to expose their internal representation, they should make and document a public API for getting access. This will also make it clear to downstream users that they can rely on a particular API/behavior.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1383)*

PR #1430, opened 2018-05-19
small role parsing bug fix

millstein:
fixed a bug in node-name parsing that was requiring 2+ characters in certain tokens where only 1+ character should be required

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/role/InferRoles.java
@@ -92,7 +92,7 @@ public class InferRoles implements Callable<SortedSet<NodeRoleDimension>> {
         case ALPHA_PLUS_DIGIT_PLUS:
           return plus(ALPHABETIC_REGEX) + plus(DIGIT_REGEX);
         case ALNUM_PLUS:
-          return ALPHABETIC_REGEX + plus(ALPHANUMERIC_REGEX);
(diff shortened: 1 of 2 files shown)


dhalperi:
This seems fine, but two qs:

* I'm confused what "alnum_plus" means -- why isn't it just `\\p{Alnum}+`? Why require first alpha?  (no change in that requirement in this PR tho).
* we should add some unit tests for this code, do you have any / know where the easy place to start is? If not, I could probably take a first pass as a way of trying to grok this code, which I've been meaning to do for a while.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1430#-:-LCprh9p7Unpga_vk_K8:bgdzh72)*

millstein:
alnum_plus indeed used to be just \\p{Alnum}+ but I recently made that change in order to do parsing for node names that don't contain delimiters (non-alphanumeric characters).  for example, we want to parse the name as2border3 into multiple chunks even though it lacks delimiters.  to do that I use numbers as delimiters essentially, which helps me see that this is alnum_plus (as2) alnum_plus (border3).  But if we define alnum_plus as just \\p{Alnum}+ then we will also erroneously capture node names like host2, breaking it into host and 2.  the stricter notion of alnum_plus fixes that problem.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1430#-:-LCqLU5c9-vyUeaC3d3f:b-tm6bbi)*

PR #1433, opened 2018-05-19
Automatic conversion to static methods for all applicable

dhalperi:
Fully automated, no expected issues. We'll see what Travis says about it.


dhalperi:
Don't love it. Not because it's wrong, but because it reveals a few too many crappy APIs in our own code that need to be fixed up before we should do this globally.

PR #1385, opened 2018-05-14
 Flip the bit on NX-OS BGP parser, update refs

dhalperi:
Needs improved testing. So far:

- [x] differences to checked-in refs look correct
- [x] better behavior on several real networks
- [x] single-variant parser fixed a bug in one of our unit tests.
- [ ] ipv6 support, at least in the grammar

What other check boxes should we add?

PR #1436, opened 2018-05-19
Enable the Cisco NX-OS specific parser

dhalperi:
It is not complete, but:
1. the gaps are explicit instead of implicit, and easy to fill in on demand.
2. it's so much more correct that the prior grammar, it's worth it.

Redo of #1385 after adding a bunch of ipv6 grammar and some other new statements.


dhalperi:
Reviewed 2 of 5 files at r1, 3 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1436)*

PR #1447, opened 2018-05-21
Fix for OpenTrace client register logs

haverma:
- Make registering tracing on `ClientBuilder` optional
- Do not register tracing with the `javax.ws.rs.Client` used by coordinators to poll the workers. 
  Because
      - They do not have any useful tracing information
      - Logs keep crowding the client prompt
- Although there will be logs from `WorkMgr.checkTask()`  still as we create a new `Client` for each ping and they look like: 
```
May 21, 2018 12:22:04 PM io.opentracing.contrib.jaxrs2.client.ClientTracingFeature configure
INFO: Registering client OpenTracing, with configuration:io.opentracing.contrib.jaxrs2.client.ClientTracingFeature$Builder@513ba759
```
But they are not as intrusive as they appear when `batfish-client` is performing some task for the user. This will probably get fixed when https://github.com/opentracing-contrib/java-jaxrs/pull/86 is released.

### DIFF
--- a/projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/BfCoordWorkHelper.java
@@ -470,7 +470,8 @@ public class BfCoordWorkHelper {
             _settings.getSslKeystoreFile(),
             _settings.getSslKeystorePassword(),
             _settings.getSslTruststoreFile(),
-            _settings.getSslTruststorePassword())
+            _settings.getSslTruststorePassword(),
+            true)
         .register(MultiPartFeature.class);
   }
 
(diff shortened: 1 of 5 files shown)


dhalperi:
Don't love it, but I see this is a reasonable balance between complexity and crap. The linked jaxrs issue would be very helpful :).

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1447#-:-LD5HM041mEpHVF3Es6g:bpcla1d)*

dhalperi:
(in other words, good job finding a balance...)

---







*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1447#-:-LD5HTQQ8AMk6gpbqUah:b-5r31tc)*

PR #1448, opened 2018-05-21
Location, IpSpace, and Node Specifiers

anothermattbrown:
New abstract and extensible framework for specifying Locations, IpSpaces, and Nodes in questions.

The interfaces of specifiers, factories, and visitors are in batfish-common-protocol. Implementations are currently in batfish.


dhalperi:
Reviewed 12 of 53 files at r1.
Review status: 12 of 53 files reviewed at latest revision, all discussions resolved.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4O6Q-1pF3QjHLfsvM:-LD4O6Q-1pF3QjHLfsvN:b-yjkut5) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java#L6)):*
> ```Java
> 
> /** Identifies the {@link Location} of the link of an interface in the network. */
> public class InterfaceLinkLocation implements Location {
> ```

is this class intended to be extended? if so, make it `abstract`

if not, make it `final` and simplify the code below?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java, line 28 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4OHyPDfq7avEAsCyt:-LD4OHyPDfq7avEAsCyu:bf4nwfh) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java#L28)):*
> ```
>     if (o == null || getClass() != o.getClass()) {
>       return false;
>     }
> ```

assuming final this code can be simplified to `o instanceof InterfaceLinkLocation`. (`null instanceof X` is false, `getClass()` is overkill).

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLocation.java, line 7 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4OXvcDOX3CEDsfETl:-LD4OXvcDOX3CEDsfETm:b-ed92kb) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLocation.java#L7)):*
> ```Java
> 
> /** Identifies the {@link Location} of an interface in the network. */
> public class InterfaceLocation implements Location {
> ```

same comments.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java, line 20 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4SSEzEhOduYYn8l7g:-LD4SSEzEhOduYYn8l7h:bdv96q8) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java#L20)):*
> ```Java
>    *     Location}s, and that no {@link Location} is assigned more than one {@link IpSpace}.
>    */
>   Multimap<IpSpace, Location> resolve(Set<Location> locations, SpecifierContext ctxt);
> ```

Some thoughts, mostly gut thoughts, about the specifier interfaces.

We're putting too much work on Specifier implementors.

* For ease of implementation, should have a `Collection<IpSpace> resolve(Location)` function. This seems like the "inherent" functionality of a resolver. All the rest is downstream optimization that users can opt-into. (Hence having the "bulk resolve" function that allows for optimization, but is not required as the only implementation - it can have default implementation that calls resolve one-at-a-time.)

* My gut is that the return value of bulk resolve should not be a multimap, but rather a `Collection<T>` where T is something like `IpSpace & Collection of locations`. Not `Pair` - we abuse that too much.

    - The return value flips the API -- locations are input keys, no IpSpaces.
    - Why make every implementor pay the cost of uniquifying IpSpaces, optimization across locations, etc. Seems like we'd rather do that in one place that happens for all specifiers.
    - Do we want all code paths to always incur the hashing cost of IpSpace? I'm not sure...

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/LocationVisitor.java, line 12 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4PjUqC6ADT-cnSjGX:-LD4PjUr-FgkdPDs9JjM:bc0sgoo) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/LocationVisitor.java#L12)):*
> ```Java
>   T visitInterfaceLocation(InterfaceLocation interfaceLocation);
> 
>   T visitVrfLocation(VrfLocation vrfLocation);
> ```

This is new.... I thought we just needed links?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/TypedLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4QXrJ9kFPBZ5iCOXU:-LD4QXrJ9kFPBZ5iCOXV:b-vt9i8a) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/TypedLocationSpecifierFactory.java#L6)):*
> ```
> implemeting
> ```

I would not comment on this if we were near the ned of the review, but: minor typo :)

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllNodesNodeSpecifier.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4Qrtu1uBBlSv0NfFH:-LD4Qrtu1uBBlSv0NfFI:bmstipr) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish/src/main/java/org/batfish/specifier/AllNodesNodeSpecifier.java#L6)):*
> ```Java
> 
> /** A {@link NodeSpecifier} specifying all nodes in the network. */
> public class AllNodesNodeSpecifier implements NodeSpecifier {
> ```

For all these classes, consider whether they should be `final`?

(Intuitively, to me this one should be.)

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

dhalperi:
*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/TypedLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4QXrJ9kFPBZ5iCOXU:-LD4VSbCAC8gegXEySxM:b3d54nc) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/TypedLocationSpecifierFactory.java#L6)):*

> ```
> implemeting
> ```

I would not comment on this if we were near the ned of the review, but: minor typo :)
</blockquote></details>

"end->ned" typo not introduced deliberately for comedic value, but let's pretend it was.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

anothermattbrown:
Review status: 12 of 53 files reviewed at latest revision, 7 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java, line 20 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4SSEzEhOduYYn8l7g:-LD7btF-A-qy37hGLtYC:bxpo2g9) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java#L20)):*

Some thoughts, mostly gut thoughts, about the specifier interfaces.

We're putting too much work on Specifier implementors.

* For ease of implementation, should have a `Collection<IpSpace> resolve(Location)` function. This seems like the "inherent" functionality of a resolver. All the rest is downstream optimization that users can opt-into. (Hence having the "bulk resolve" function that allows for optimization, but is not required as the only implementation - it can have default implementation that calls resolve one-at-a-time.)

* My gut is that the return value of bulk resolve should not be a multimap, but rather a `Collection<T>` where T is something like `IpSpace & Collection of locations`. Not `Pair` - we abuse that too much.

    - The return value flips the API -- locations are input keys, no IpSpaces.
    - Why make every implementor pay the cost of uniquifying IpSpaces, optimization across locations, etc. Seems like we'd rather do that in one place that happens for all specifiers.
    - Do we want all code paths to always incur the hashing cost of IpSpace? I'm not sure...
</blockquote></details>

* I am sympathetic to the idea of having a `IpSpace resolve(Location)` method.  It's not clear what `Collection<IpSpace>` should mean though. If it's union, intersection, etc, `IpSpace` is closed under those. It makes sense if the meaning is "run a separate query for each `IpSpace` in the collection", but I'm not sure what the use case would be.
* What we really want is a many-to-one mapping between `Location` and `IpSpace`, where `IpSpace` and `Location` are both unique. `Multimap` seemed the closest fitting collection type, but I agree it's not ideal. Which type is the key seems unimportant to me, since we're unlikely to ever want to do lookups. How about we introduce a class `IpSpaceAssignment` that supports methods `void assign(Location, IpSpace)` and `void assign(Set<Location>, IpSpace)`?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/LocationVisitor.java, line 12 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4PjUqC6ADT-cnSjGX:-LD7enfX-YFQbUoFwCO3:b7y85ac) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/LocationVisitor.java#L12)):*

This is new.... I thought we just needed links?
</blockquote></details>

Yes, I thought at one point that I needed this to preserve current behavior (at least up to ref tests), but I found another way.

I'm happy to remove this. However, it does allow the user to express "any interface on a VRF", while `InterfaceLocation` only supports "every interface on a VRF". This might be useful for perf and/or getting higher information-density answers (one flow per VRF instead of one per interface).

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

anothermattbrown:
Review status: 9 of 54 files reviewed at latest revision, 7 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java, line 20 at r1](https://reviewable.io/reviews/batfish/batfish/1448#-LD4SSEzEhOduYYn8l7g:-LD7slXu2ixjXKVWcUha:b-t8sos) ([raw file](https://github.com/batfish/batfish/blob/96b6f1c8f824622e4a33036840542ab55b110a8a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifier.java#L20)):*

* I am sympathetic to the idea of having a `IpSpace resolve(Location)` method.  It's not clear what `Collection<IpSpace>` should mean though. If it's union, intersection, etc, `IpSpace` is closed under those. It makes sense if the meaning is "run a separate query for each `IpSpace` in the collection", but I'm not sure what the use case would be.
* What we really want is a many-to-one mapping between `Location` and `IpSpace`, where `IpSpace` and `Location` are both unique. `Multimap` seemed the closest fitting collection type, but I agree it's not ideal. Which type is the key seems unimportant to me, since we're unlikely to ever want to do lookups. How about we introduce a class `IpSpaceAssignment` that supports methods `void assign(Location, IpSpace)` and `void assign(Set<Location>, IpSpace)`?
</blockquote></details>

I added `IpSpaceAssignment` as a proposed solution. Internally, it's just list of `(IpSpace, Set<Location>)` pairs. I didn't use a `*Map` because it seems unlikely that we will ever discover syntatically equal `IpSpace`s anyway. What do you think @dhalperi ?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

anothermattbrown:
ready when you are @dhalperi 

---

Review status: 7 of 57 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448#-:-LD9DVds4IacirNf78MD:b-1ew3l6)*

dhalperi:
Reviewed 17 of 24 files at r4.

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllInterfacesLocationSpecifier.java, line 30 at r4](https://reviewable.io/reviews/batfish/batfish/1448#-LDDBH0K-OPXtYWwhzdy:-LDDBH0K-OPXtYWwhzdz:b-qc3jac) ([raw file](https://github.com/batfish/batfish/blob/e065e4d73e2827a92f8a0c6b3618789cf19b5f1e/projects/batfish/src/main/java/org/batfish/specifier/AllInterfacesLocationSpecifier.java#L30)):*
> ```Java
>   }
> 
>   protected Location makeLocation(String node, String iface) {
> ```

private? @Override?

actually, just inline? up to you.

---

*[projects/batfish/src/test/java/org/batfish/specifier/NodeSpecifierTests.java, line 17 at r4](https://reviewable.io/reviews/batfish/batfish/1448#-LDDBFZFBYjmRWZykEAl:-LDDBFZFBYjmRWZykEAm:b-80jztk) ([raw file](https://github.com/batfish/batfish/blob/e065e4d73e2827a92f8a0c6b3618789cf19b5f1e/projects/batfish/src/test/java/org/batfish/specifier/NodeSpecifierTests.java#L17)):*
> ```Java
> import org.junit.Test;
> 
> public class NodeSpecifierTests {
> ```

Test?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

anothermattbrown:
Review status: 53 of 57 files reviewed at latest revision, 3 unresolved discussions.

---

*[projects/batfish/src/main/java/org/batfish/specifier/DescriptionRegexInterfaceLocationSpecifierFactory.java, line 12 at r3](https://reviewable.io/reviews/batfish/batfish/1448#-LDCp9-KDW_jAvJds9Ba:-LDD767k4EGSEC-ZATR2:b-b82zaq) ([raw file](https://github.com/batfish/batfish/blob/82084ffdcd26691835f58e3e84b60df2b3aa0954/projects/batfish/src/main/java/org/batfish/specifier/DescriptionRegexInterfaceLocationSpecifierFactory.java#L12)):*

General warning: `Class` is not really a rich enough type in Java. Will this ever need to support things like `List<Map<String, String>>`?
</blockquote></details>

Yep, that's a limitation of `TypedLocationSpecifierFactory`. In such cases we won't be able to subclass it.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceAssignment.java, line 51 at r3](https://reviewable.io/reviews/batfish/batfish/1448#-LDCmu3K20kEg98z-I3Q:-LDD2-1-1LsyunqXeYkx:baghdxs) ([raw file](https://github.com/batfish/batfish/blob/82084ffdcd26691835f58e3e84b60df2b3aa0954/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceAssignment.java#L51)):*

ditto

here's not as obvious, you could use addAll + sizes, you could use intersection and empty... not sure.
</blockquote></details>

Went with sizes. This means it won't report the duplicate location(s), but that's probably of little value anyway.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

dhalperi:
Reviewed 4 of 6 files at r5.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

anothermattbrown:

---

*[projects/batfish/src/main/java/org/batfish/specifier/NodeRoleRegexLocationSpecifierFactory.java, line 19 at r3](https://reviewable.io/reviews/batfish/batfish/1448#-LDCqmn13BBvaPiK1NTJ:-LDDnK7F8ZJNmyD8QDxg:b9v69a0) ([raw file](https://github.com/batfish/batfish/blob/82084ffdcd26691835f58e3e84b60df2b3aa0954/projects/batfish/src/main/java/org/batfish/specifier/NodeRoleRegexLocationSpecifierFactory.java#L19)):*

open Q: is there a common core we should set up for code from NodeSpecifier?
</blockquote></details>

I propose we leave that for later. It would likely be a straightforward refactoring.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1448)*

PR #1485, opened 2018-05-24
Ensure originType is always set on BGP routes during transformation

progwriter:
BgpRoute.Builder crashes when originType is not set (rightfully so), so ensure we set it when propagating routes.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
@@ -212,7 +212,7 @@ public class BgpRoute extends AbstractRoute {
 
   private final Ip _originatorIp;
 
-  private final OriginType _originType;
+  @Nonnull private final OriginType _originType;
 
   private final RoutingProtocol _protocol;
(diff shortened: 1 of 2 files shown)


progwriter:

---

*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 121 at r1](https://reviewable.io/reviews/batfish/batfish/1485#-LDIH79ICRBuNpwuEmkl:-LDN0ro36LRVJ3jvXcep:b-vzsw3n) ([raw file](https://github.com/batfish/batfish/blob/ffd42497e398654082638ec7d3c44c86242c32d3/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L121)):*

This choice of origin type for non-bgp routes is not common to all routers. If it gets set via this line only, that is a huge red flag. How does this impact manual setting of originType in routing policy? Seems like any time the value is gotten from here that indicates a bug with routing policy,
</blockquote></details>

You're right, the bug is in a different place.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1485)*

PR #1503, opened 2018-05-25
Client: set exit code when, in batch mode, some command fails

dhalperi:
Downstream tools can therefore use exit code for useful things.

Had to delete a broken test -- server is not up for these unit tests.

PR #1510, opened 2018-05-28
don't make exit code dependent on command pass/fail

ratulm:
After #1503, Client exits with non-zero exit code when a test fails. Because of that Travis does not execute the find and diff commands, which are helpful to spot which test failed. This change fixes that.


dhalperi:
Setting up the behavior reversed here was the exact intent of #1503... I'll see what I can do to make Travis better behaved.

dhalperi:

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1510#-LDh1nen1BmfZFpc7WG7:-LDh1nen1BmfZFpc7WG8:b158ak0) (no related file):*
Okay to close this now that #1511 is in?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1510)*

PR #1523, opened 2018-05-30
Revert "Upgrade guava and transitive"

dhalperi:
This reverts commit 049debda33753bf33306fc89ea1d2bcfc8208da5.

New guava doesn't work with findbugs because they switched Nullable annotations and the world is sad. Figuring this out will be more work.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -58,16 +58,15 @@ maven_repository(
         'com.google.auto.service:auto-service:1.0-rc3',
     ],
     force = [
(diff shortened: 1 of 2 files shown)


ratulm:
:lgtm:  The WORKSPACE file is manually maintained?

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1523#-:-LDm8Het2BnjHRziIXBc:b-yojihn)*

dhalperi:
Just like pom.xml, it has to be.

PR #1524, opened 2018-05-30
In travis build, diff example demo out against ref before deleting

anothermattbrown:
The current build script saves the example demo output to `demos/example/commands.ref.testout`, but then deletes it unless batfish returns an error status code. Added an additional check that the output matches the ref. There has been some drift since the ref was generated, which could be a sign of new bugs.


dhalperi:
My understanding from talking to Ratul is that we did not want this because the demo was non-deterministic. Is that no longer the case, @ratulm ?

ratulm:
I don't know if we've hit determinism accidentally but we haven't made an effort toward that. I am inclined to leave the comparison out since most of the questions being called in the demo should be covered elsewhere. Given our current coverage level, I wouldn't mind removing this demo run from Travis altogether.

PR #1534, opened 2018-05-31
provide a safe row building function

ratulm:
To prevent the "type" bug you found today, what do you think of this? Basically, provide a safe build function, which succeeds only if there is type consistency and we'll try to use this function as much as possible going forward.

Another option is to take its core logic and put it in a static function / matcher, and call it from tests.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java
@@ -55,7 +55,7 @@ public class Row implements Comparable<Row> {
       Set<String> columnsInMetadata =
           columnMetadata.stream().map(cm -> cm.getName()).collect(Collectors.toSet());
       Set<String> columnsInData = Row.getColumnNames(_data);
-      if (!columnMetadata.equals(columnsInMetadata)) {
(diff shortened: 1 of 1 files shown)


dhalperi:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java, line 54 at r2](https://reviewable.io/reviews/batfish/batfish/1534#-LEBsZUr-gsK9Pn_Ayg9:-LEBsZUr-gsK9Pn_AygA:b-2ksv4r) ([raw file](https://github.com/batfish/batfish/blob/1de11e69e09b42f09a0d85daeb8bb72400f6fc8f/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java#L54)):*
> ```Java
>      * have the right type.
>      */
>     public Row build(List<ColumnMetadata> columnMetadata) {
> ```

Definitely seems like a step forward.

Not sure how this compares to just `putX(String colName, X colData)` [possibly with `put(String colName, @Nullable Object colData, Schema colSchema)` as a wrapper for `switch colSchema... putX`. Thoughts?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1534)*

PR #1550, opened 2018-06-01
move ReachabilitySettings into the question project

anothermattbrown:
* It has been replaced by ReachabilityParameters in IBatfish.
* No longer unset srcIp in the HeaderSpace, because pathDiff doesn't use
  sources (it originates from VRFs with changed or removed edges). In
  the future, we should add a new question for pathDiff that only
  allows the parameters that are used by it.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/plugin/IBatfish.java
@@ -33,12 +33,12 @@ import org.batfish.datamodel.collections.NamedStructureEquivalenceSets;
 import org.batfish.datamodel.collections.RoutesByVrf;
 import org.batfish.datamodel.pojo.Environment;
 import org.batfish.datamodel.questions.Question;
-import org.batfish.datamodel.questions.ReachabilitySettings;
 import org.batfish.datamodel.questions.smt.HeaderLocationQuestion;
 import org.batfish.datamodel.questions.smt.HeaderQuestion;
 import org.batfish.datamodel.questions.smt.RoleQuestion;
 import org.batfish.grammar.BgpTableFormat;
 import org.batfish.grammar.GrammarSettings;
+import org.batfish.question.ReachabilityParameters;
 import org.batfish.role.NodeRoleDimension;
 import org.batfish.role.NodeRolesData;
 
@@ -127,11 +127,11 @@ public interface IBatfish extends IPluginConsumer {
 
   ParseVendorConfigurationAnswerElement loadParseVendorConfigurationAnswerElement();
 
-  AnswerElement multipath(ReachabilitySettings reachabilitySettings);
+  AnswerElement multipath(ReachabilityParameters reachabilityParameters);
 
   AtomicInteger newBatch(String description, int jobs);
 
-  AnswerElement pathDiff(ReachabilitySettings reachabilitySettings);
+  AnswerElement pathDiff(ReachabilityParameters reachabilityParameters);
 
   void popEnvironment();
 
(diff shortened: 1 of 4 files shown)


dhalperi:
Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2989 at r1](https://reviewable.io/reviews/batfish/batfish/1550#-LDwMhh7Dh5hKw0WcsRW:-LDwMhh7Dh5hKw0WcsRX:bhv5onh) ([raw file](https://github.com/batfish/batfish/blob/89a8b0a98e726add9fdfda5b619905451a5fa3d6/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2989)):*
> ```Java
>     /* TODO pathDiff deserves its own parameter set type.
>      * Sources are ignored, since we always originate from vrfs that have missing or removed
>      * edges. So the user-facing question shouldn't include a source parameter.
> ```

not even as a filter on the changed routes? ("find me reachability differences limited to those differences where sources are like foo")

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/ReachabilitySettings.java, line 354 at r1](https://reviewable.io/reviews/batfish/batfish/1550#-LDwMz4-1vuFhWFlrhXe:-LDwMz4-1vuFhWFlrhXf:b-oqdj5z) ([raw file](https://github.com/batfish/batfish/blob/6acfe6e7ff18c1f84e137bc5d799671fd1a4cfb4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/ReachabilitySettings.java#L354)):*
> ```Java
>     HeaderSpace headerSpace =
>         _headerSpace.toBuilder().setNotSrcIps(nullIpSpace).setSrcIps(nullIpSpace).build();
> ```

I can't link this change with your explanation of it. Isn't this code called for all types, not just pathDiff?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1550)*

anothermattbrown:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2989 at r1](https://reviewable.io/reviews/batfish/batfish/1550#-LDwMhh7Dh5hKw0WcsRW:-LDwOKwfER7ikUcipHvi:b-xkipey) ([raw file](https://github.com/batfish/batfish/blob/89a8b0a98e726add9fdfda5b619905451a5fa3d6/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2989)):*

not even as a filter on the changed routes? ("find me reachability differences limited to those differences where sources are like foo")
</blockquote></details>

It could make sense, but that would be new behavior. currently ingressNode, etc are ignored. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1550)*

anothermattbrown:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/ReachabilitySettings.java, line 354 at r1](https://reviewable.io/reviews/batfish/batfish/1550#-LDwMz4-1vuFhWFlrhXe:-LDwPGDfA8TmzoEs-YM5:btxh0w4) ([raw file](https://github.com/batfish/batfish/blob/6acfe6e7ff18c1f84e137bc5d799671fd1a4cfb4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/ReachabilitySettings.java#L354)):*

I can't link this change with your explanation of it. Isn't this code called for all types, not just pathDiff?
</blockquote></details>

Yes. I was removing the srcIp constraints from headerspace to avoid duplication, and to avoid a potential source of confusion later. However, since we're only using ConstantIpSpaceSpecifier here, it's fine to leave them in. This will probably continue to be true, since we're planning to add a new question with the fancy sane-mode specifiers. 

I was going for the smallest delta that preserves all current behavior.  For pathDiff, we can change it to use the ingress location in some reasonable way, or explicitly check for a ConstantIpSpace specifier. I planned on tackling that in a separate PR, after which we can put this back.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1550)*

dhalperi:

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1550)*

PR #1563, opened 2018-06-02
Specifiers question for querying Location and IpSpace specifiers

anothermattbrown:
* Added some missing specifier factories
* Added method to IBatfish to get a SpecifierContext, so the bulk of
  the question implementation can live in the question class.


dhalperi:
Review status: 2 of 13 files reviewed at latest revision, all discussions resolved.

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LDyrYe94XlodTvrkuMT:-LDyrYe94XlodTvrkuMU:b1y0qud) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java#L6)):*
> ```Java
> 
> @AutoService(LocationSpecifierFactory.class)
> public class AllInterfaceLinksLocationSpecifierFactory implements LocationSpecifierFactory {
> ```

Why is this needed as a separate concept, vs just having `AllInterfaceLinksLocationSpecifier` have a static factory method?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

anothermattbrown:
Review status: 2 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LDyrYe94XlodTvrkuMT:-LE2aKVQ782jpI4HPbO4:buz8f14) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java#L6)):*

Why is this needed as a separate concept, vs just having `AllInterfaceLinksLocationSpecifier` have a static factory method?
</blockquote></details>

Since subtyping doesn't include static methods, we'd have to do something like use reflection to find the static methods for specifiers accessed through the service loader. Seems hacky and non-idiomatic to me. Another option would be to move the (non-static) factory method to the specifier interface and load the specifiers themselves using `AutoService`. I don't like that approach either because it would require each specifier have a zero-input constructor, preventing them from having sensible invariants about their inputs.

In short, this seemed like the cleanest and most idiomatic solution. It does involve some boilerplate, but we're talking about Java after all.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

dhalperi:
Review status: 6 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifierFactory.java, line 1 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DZHm3YPuiYMaM5Ry:-LE3DZHm3YPuiYMaM5Rz:b-bx2out) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifierFactory.java#L1)):*
> ```Java
> package org.batfish.specifier;
> ```

org.batfish.common.specifier

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LDyrYe94XlodTvrkuMT:-LE3ALe885rSiL54BgTh:b-nq6yqa) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java#L6)):*

Since subtyping doesn't include static methods, we'd have to do something like use reflection to find the static methods for specifiers accessed through the service loader. Seems hacky and non-idiomatic to me. Another option would be to move the (non-static) factory method to the specifier interface and load the specifiers themselves using `AutoService`. I don't like that approach either because it would require each specifier have a zero-input constructor, preventing them from having sensible invariants about their inputs.

In short, this seemed like the cleanest and most idiomatic solution. It does involve some boilerplate, but we're talking about Java after all.
</blockquote></details>

Sounds reasonable enough. Can these be inlined in the classes they generate, or is that a bad idea for some reason?

For sanity's sake, take a look at https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/PipelineRunner.java#L42 ?

I think the static factory method is not as crazy as you think if is.

---

*[projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java, line 18 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3D-yN7pMZwzhBCCyZ:-LE3D-yN7pMZwzhBCCy_:bizii9t) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java#L18)):*
> ```Java
>  * specifiers are resolved.
>  */
> public final class SpecifiersQuestion extends Question {
> ```

I can't figure out why this exists, and the refs don't help. Why isn't this just an API call?

---

*[projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java, line 65 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3CZyT-vY8YD9noIvA:-LE3CZyT-vY8YD9noIvB:b-ommupj) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java#L65)):*
> ```Java
> 
>   private static <T> T loadService(
>       Class<T> serviceClass, Function<T, String> nameFunction, String name) {
> ```

Is this the only place we instantiate specifiers? (Seems like this logic belongs somewhere else, likely in the specifiers themselves).

---

*[projects/question/src/main/java/org/batfish/question/specifiers/TestIpSpaceSpecifier.java, line 9 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3Cq1hAGctM2gCYk8K:-LE3Cq1hAGctM2gCYk8L:b-q2o2du) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/TestIpSpaceSpecifier.java#L9)):*
> ```Java
> import org.batfish.specifier.SpecifierContext;
> 
> public class TestIpSpaceSpecifier implements IpSpaceSpecifier {
> ```

inline in test, move to `test` from `main`.

---

*[projects/question/src/test/java/org/batfish/question/specifiers/TestIpSpaceSpecifierFactory.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3Cl2hDRH7LWXtZB5p:-LE3Cl2hDRH7LWXtZB5q:b26qs78) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/test/java/org/batfish/question/specifiers/TestIpSpaceSpecifierFactory.java#L8)):*
> ```Java
> 
> @AutoService(IpSpaceSpecifierFactory.class)
> public class TestIpSpaceSpecifierFactory implements IpSpaceSpecifierFactory {
> ```

definitely inline in test

---

*[tests/basic/commands, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DCfC0FjyWvjpCQ19:-LE3DCfC0FjyWvjpCQ1A:bpcn77x) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/tests/basic/commands#L17)):*
> ```
> test tests/basic/routes.ref get routes
> test tests/basic/ipowners.ref get ipowners
> test tests/basic/specifiers.ref get specifiers locationSpecifierFactory=AllInterfaceLinksLocationSpecifierFactory, ipSpaceSpecifierFactory=InferFromLocationIpSpaceSpecifierFactory
> ```

I don't think it quite makes sense to ask for both specifiers at once. They seem unrelated; should probably be two separate api calls

---

*[tests/basic/commands, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DLBV19zOh70TLo4E:-LE3DLBV19zOh70TLo4F:bwczhr9) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/tests/basic/commands#L17)):*
> ```
> test tests/basic/routes.ref get routes
> test tests/basic/ipowners.ref get ipowners
> test tests/basic/specifiers.ref get specifiers locationSpecifierFactory=AllInterfaceLinksLocationSpecifierFactory, ipSpaceSpecifierFactory=InferFromLocationIpSpaceSpecifierFactory
> ```

unit test, not ref test.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

anothermattbrown:
Review status: 6 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifierFactory.java, line 1 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DZHm3YPuiYMaM5Ry:-LEAf28a59dBhIMdMdI_:b-d7926e) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/IpSpaceSpecifierFactory.java#L1)):*

org.batfish.common.specifier
</blockquote></details>

All other specifier classes live here. If they should all be moved then let's do that as a separate PR.

Why is this necessary? bcp has lots of classes outside of `org.batfish.common` -- should they all be moved?

---

*[projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java, line 6 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LDyrYe94XlodTvrkuMT:-LEAfWPe3DAffZMiQq32:bya80nx) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/batfish/src/main/java/org/batfish/specifier/AllInterfaceLinksLocationSpecifierFactory.java#L6)):*

Sounds reasonable enough. Can these be inlined in the classes they generate, or is that a bad idea for some reason?

For sanity's sake, take a look at https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/PipelineRunner.java#L42 ?

I think the static factory method is not as crazy as you think if is.
</blockquote></details>

Static factory methods are fine for classes known at compile-time, but they don't work well for service classes (specified by a superclass and loaded dynamically) for the reasons I stated previously. 

---

*[projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java, line 18 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3D-yN7pMZwzhBCCyZ:-LEAXWFG1Ie_TDRVGwOG:b-o5uo2n) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java#L18)):*

I can't figure out why this exists, and the refs don't help. Why isn't this just an API call?
</blockquote></details>

I'm confused. You seem to mean you think this either shouldn't exist, or shouldn't be a question. Please make a concrete suggestion if you have one. "An API call" is too vague -- please be more specific (below, you seem to use API to refer to questions, so in that sense this already is one). 

I created this because I saw value in exposing the specifiers API to clients, so users can see how specifiers are resolved for a particular network. This would help understand, debug and refine queries . If you disagree and think there is no value in such a question, please say so straightforwardly.

---

*[projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java, line 65 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3CZyT-vY8YD9noIvA:-LEA_GZmBiLqurFgYNI3:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersQuestion.java#L65)):*

Is this the only place we instantiate specifiers? (Seems like this logic belongs somewhere else, likely in the specifiers themselves).
</blockquote></details>

Done.

---

*[projects/question/src/main/java/org/batfish/question/specifiers/TestIpSpaceSpecifier.java, line 9 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3Cq1hAGctM2gCYk8K:-LEA_Hsm6dAvuOc73onP:bb6un1j) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/main/java/org/batfish/question/specifiers/TestIpSpaceSpecifier.java#L9)):*

inline in test, move to `test` from `main`.
</blockquote></details>

Done

---

*[projects/question/src/test/java/org/batfish/question/specifiers/TestIpSpaceSpecifierFactory.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3Cl2hDRH7LWXtZB5p:-LEAaS4V5MBM0OZX_gZ3:bb6un1j) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/projects/question/src/test/java/org/batfish/question/specifiers/TestIpSpaceSpecifierFactory.java#L8)):*

definitely inline in test
</blockquote></details>

Done

---

*[tests/basic/commands, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DCfC0FjyWvjpCQ19:-LEAb3xQB0O2vPFownI_:b-fc9jh8) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/tests/basic/commands#L17)):*

I don't think it quite makes sense to ask for both specifiers at once. They seem unrelated; should probably be two separate api calls
</blockquote></details>

I don't see the value of two separate calls: this can subsume what I imagine reasonable behavior of two separate API calls to be. A `locationSpecifier` API call would return a list of locations; this will do that if we pass use ConstantIpSpaceSpecifierFactory with input UniverseIpSpace (we can make that the default for those). An `ipSpaceSpecifier` API call would produce the same output as this, but require a list of concrete locations rather than a specifier (we can make a `LocationListLocationSpecifier` to get the same behavior).

---

*[tests/basic/commands, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/1563#-LE3DLBV19zOh70TLo4E:-LEAboS93om8gkJDO8Ck:bvtmo3n) ([raw file](https://github.com/batfish/batfish/blob/406a5a2747039903ca22d4f7d40aa418af062674/tests/basic/commands#L17)):*

unit test, not ref test.
</blockquote></details>

In progress

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

dhalperi:

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1563#-LEBNmrK4wKKZApFxVi8:-LEBNmrK4wKKZApFxVi9:brw18xa) (no related file):*
Per discussion, all good.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

dhalperi:

---

*[projects/batfish-common-protocol/pom.xml, line 302 at r4](https://reviewable.io/reviews/batfish/batfish/1563#-LEG3UXV1-p9hSR9a49V:-LEG3UXW4H9jTs2hUZTN:bmbiaic) ([raw file](https://github.com/batfish/batfish/blob/813020a98dd70eaf71935b60b9e71b312b255c85/projects/batfish-common-protocol/pom.xml#L302)):*
> ```XML
>     <dependency>
>       <groupId>com.google.auto.service</groupId>
>       <artifactId>auto-service</artifactId>
> ```

please alphabetize and list with the main deps, not in the test deps

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

anothermattbrown:

---

*[projects/batfish-common-protocol/pom.xml, line 302 at r4](https://reviewable.io/reviews/batfish/batfish/1563#-LEG3UXV1-p9hSR9a49V:-LEG5pi78COI4eYmXtru:b-83i5os) ([raw file](https://github.com/batfish/batfish/blob/813020a98dd70eaf71935b60b9e71b312b255c85/projects/batfish-common-protocol/pom.xml#L302)):*

please alphabetize and list with the main deps, not in the test deps
</blockquote></details>

gah shouldn't have trusted intellij to do the right thing

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563)*

anothermattbrown:
Working to get coverage above 80%

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563#-:-LEG5zj74QZ5bJDLKA2d:b-v4uf03)*

dhalperi:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563#-:-LEGRBfq8bDXGiTLsAVu:bnfp4nl)*

dhalperi:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1563#-:-LEHJhxn2o7yYvQkrG6c:bnfp4nl)*

PR #1565, opened 2018-06-02
Make traceroute use ForwardingAnalysis for ARP

arifogel:
- traceroute uses ForwardingAnalysis now for ARP decisions
- fix bug where traceroute checked ARP before outgoing ACL


arifogel:
Review status: 14 of 15 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1565#-LE3CIfiDD7dPYjckHzc:-LEAh-yBC6Hl4mmI1oVC:b-8h2mz1) (no related file):*

Nit: Generally I dislike this parameter creep. It looks like the ForwardingAnalysis should be instantiated from within TracerouteEngine. If that means class can no longer be a singleton, that's fine. But constructor of ForwardingAnalysis and signature of `processFlows` has a lot of overlap.
</blockquote></details>

I was trying not to disrupt existing cache mechanism. Keep in mind that `ForwardingAnalysis` is also used by reachability analysis, which may not necessarily involve traceroute.
To be clear, there are three places ForwardingAnalysis is used:
1. Traceroute of intermediate data plane to determine BGP sessions, etc.
2. Traceroute of finalized data plane
3. Reachability analysis for finalized data plane.
Currently cache of ForwardingAnalysis resides within Batfish. Note that data planes are also cached there.
I think it would make sense for Data Plane  to include a cached ForwardingAnalysis via memoization. Then we could pass this ForwardingAnalysis to traceroute automatically as the data plane parameter. Then we don't need to break singleton (although preserving may not be that important) of TraceRouteEngineImpl, since data plane is already parameter everywhere.

Later on we can clean up by adding a traceroute context structure, to take care of existing mess of parameters.

Thoughts?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1565)*

arifogel:
Review status: 12 of 15 files reviewed at latest revision, 2 unresolved discussions.

---

*[projects/batfish/src/test/java/org/batfish/dataplane/TracerouteEngineTest.java, line 272 at r1](https://reviewable.io/reviews/batfish/batfish/1565#-LE3DXd91DKJXbHO-z6G:-LEAmUbFEsBrAWnGEicD:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/2483bfdcfd8ca910eb45329a27e3c2a470bc19a3/projects/batfish/src/test/java/org/batfish/dataplane/TracerouteEngineTest.java#L272)):*

Please add clarifying comments) Why is the flow actually blocked? acl with no lines is an implicit deny all?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1565)*

progwriter:
Review status: 12 of 15 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1565#-LE3CIfiDD7dPYjckHzc:-LEC6eA19dYFBZRdAhs-:bigckxf) (no related file):*

I was trying not to disrupt existing cache mechanism. Keep in mind that `ForwardingAnalysis` is also used by reachability analysis, which may not necessarily involve traceroute.
To be clear, there are three places ForwardingAnalysis is used:
1. Traceroute of intermediate data plane to determine BGP sessions, etc.
2. Traceroute of finalized data plane
3. Reachability analysis for finalized data plane.
Currently cache of ForwardingAnalysis resides within Batfish. Note that data planes are also cached there.
I think it would make sense for Data Plane  to include a cached ForwardingAnalysis via memoization. Then we could pass this ForwardingAnalysis to traceroute automatically as the data plane parameter. Then we don't need to break singleton (although preserving may not be that important) of TraceRouteEngineImpl, since data plane is already parameter everywhere.

Later on we can clean up by adding a traceroute context structure, to take care of existing mess of parameters.

Thoughts?
</blockquote></details>

Ok, I like the idea of the DP having a cached ForwadingAnalysis. Obviously, the work is for a different PR.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1565)*

arifogel:
@anothermattbrown That's just how routers behave. Anyway I just overhauled traceroute to make it more legible and maintainable.

---

Review status: 10 of 16 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1565#-:-LELNSvl5UrSlfKrDWb2:b-ife3si)*

PR #1582, opened 2018-06-04
Multipath-relax should enable ebgp multipath 

### LINKED ISSUE #1556: get routes does not display multipath routes for fattree4 example
As an example, edge-7 should be able to go through either aggregation switch in its pod to reach edge-10 (70.0.10.0/24), but get routes includes only one route via aggregation-4:
`edge-7 vrf:default net:70.0.10.0/24 nhip:10.0.1.0 nhint:dynamic nhnode:aggregation-4 admin:20 cost:0 tag:none prot:bgp`
edge-7 should compute multiple routes, because the `router bgp` configuration on edge-7 contains `bgp bestpath as-path multipath-relax`

progwriter:
Fixes #1556 

`bestpath as-path multipath-relax` actually enables multipath for ebgp without explicit `max-path n` needing to be set.
Fix the extraction, which in turn fixes the bug found in #1556

### DIFF
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/BgpProcessMatchers.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/BgpProcessMatchers.java
@@ -7,8 +7,10 @@ import javax.annotation.Nonnull;
 import org.batfish.datamodel.BgpNeighbor;
 import org.batfish.datamodel.BgpProcess;
 import org.batfish.datamodel.Ip;
+import org.batfish.datamodel.MultipathEquivalentAsPathMatchMode;
 import org.batfish.datamodel.Prefix;
 import org.batfish.datamodel.matchers.BgpProcessMatchersImpl.HasMultipathEbgp;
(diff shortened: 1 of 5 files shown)


arifogel:
:lgtm:

---

---

*[projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java, line 1390 at r1](https://reviewable.io/reviews/batfish/batfish/1582#-LE9xa3f5JepH2u0W568:-LE9xa3f5JepH2u0W569:btj7bfe) ([raw file](https://github.com/batfish/batfish/blob/2baa9c1ed7bc07f2ff6502b69089cefa30735802/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L1390)):*
> ```Java
>         equalTo(MultipathEquivalentAsPathMatchMode.PATH_LENGTH));
> 
>     assertThat(aristaDisabled.getMultipathEbgp(), equalTo(false));
> ```

Prefer to use matchers here so we are not making assertion on a boolean (could also use matchers above). But I'll leave it up to you.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1582#-:-LE9xuAV8gHjE68mq7l-:bnfp4nl)*

PR #1595, opened 2018-06-06
fix an ambiguity of node-name parsing for names that have no delimiters

millstein:
The inferred regex for node names that have no delimiters was too loose, accidentally capturing some node names that have a different format.  That bug was also causing role inference to be non-deterministic.


dhalperi:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1595#-:-LEICihl9UYHuaeRtQlE:bnfp4nl)*

PR #1598, opened 2018-06-06
batfish-common-protocol: establish parser_common to reduce rebuilding parsers

dhalperi:
This PR is a bit more interesting than other Bazel PRs.

Basically, establish an as-small-as-possible set of files in `batfish-common-protocol` that a parser can be compiled against. This way, we don't have to recompile all parsers any time that anything at all in the data model changes.

Trying to do this type of work highlights where encapsulation and separation of functions are useful. It seems strange that we can't build a parser without depending on `AnswerSummary`. Also, this would have been completely impossible had I not broken one dependence on `CommonUtil`, since that common utility class depends on absolutely everything else in the data model FWICT.

### DIFF
--- a/projects/batfish-common-protocol/BUILD
+++ b/projects/batfish-common-protocol/BUILD
@@ -2,6 +2,43 @@ package(default_visibility = ["//visibility:public"])
 
 load("@batfish//skylark:junit.bzl", "junit_tests")
 
+filegroup(
+    name = "parser_files",
+    srcs = [
+        "src/main/java/org/batfish/common/BatfishException.java",
+        "src/main/java/org/batfish/common/DebugBatfishException.java",
+        "src/main/java/org/batfish/common/ParseTreeSentences.java",
+        "src/main/java/org/batfish/common/util/BatfishObjectMapper.java",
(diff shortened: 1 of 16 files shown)


progwriter:
:lgtm:
And yes, `CommonUtil` is a little too... common :)  

---

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java, line 94 at r1](https://reviewable.io/reviews/batfish/batfish/1598#-LELK0Fu82zZ-wigia5W:-LELK0Fu82zZ-wigia5X:b-bnf9sj) ([raw file](https://github.com/batfish/batfish/blob/cb2c994c806d2fff3bb4179c2cf451f175be7dce/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/BatfishCombinedParser.java#L94)):*
> ```Java
>   }
> 
>   public static String escape(String offendingTokenText) {
> ```

nit: Since you're at it, might as well javadoc this. especially since it escapes only a subset of whitespace characters. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1598#-:-LELJkrjBqc5MTwqOu75:bxtjsvb)*

PR #1617, opened 2018-06-08
remove SaneExpr

### LINKED ISSUE #1407: Consider removing `SaneExpr` from various reachability constraints
> I put that in there a long time ago, and I don't remember the exact reason
> It may have been for refinement, so if you disallow one flow, you don't just get another with different fake ICMP code
> I think it was also meant to prevent weird situation where an acl rule only cares about ICMP code, so you get a flow with a specific ICMP code but protocol is not ICMP
> Anyway, the original intent was just not to have simultaneous constraints in a single model on mutually exclusive fields
> Probably I did that because I didn't have confidence  in answers without it, and/or z3 was buggy at the time and I saw weird results
> With our [better] tests now, we may want to just toss it. Probably will improve performance
> We just need to be careful that improper field values don't leak into Flows, causing apparently equivalent Flows not to be such

anothermattbrown:
fixes #1407


arifogel:
:lgtm:

---

Reviewed 19 of 19 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1617#-:-LEVUIkb4EkxaCmj54oj:bnfp4nl)*

PR #1638, opened 2018-06-12
fix negate in Z3ContextJob

anothermattbrown:
We only want to negate the result from NOD, not the headerspace
constraint specified by the user.

The meaning of negate was:
  NOT(AND(Header space constraint, nod result)),
Now its:
  AND(Header space constraint, NOT(nod result)).

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/z3/Z3ContextJob.java
+++ b/projects/batfish/src/main/java/org/batfish/z3/Z3ContextJob.java
@@ -188,25 +188,22 @@ public abstract class Z3ContextJob<R extends BatfishJobResult<?, ?>> extends Bat
             .stream()
             .map(variablesAsConsts::get)
             .collect(Collectors.toList());
-
-    Expr substitutedSmtConstraint =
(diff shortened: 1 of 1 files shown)


dhalperi:
Review status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @arifogel)

---

*[projects/batfish/src/main/java/org/batfish/z3/Z3ContextJob.java, line 207 at r1](https://reviewable.io/reviews/batfish/batfish/1638#-LElKkKj200pSN17qAm7:-LElKkKk0HX8k004zVOB:b-908dex) ([raw file](https://github.com/batfish/batfish/blob/8cc3a57ca5e51e3776f549f66aa40f7cf9228015/projects/batfish/src/main/java/org/batfish/z3/Z3ContextJob.java#L207)):*
> ```Java
> 
>     return context.mkAnd(finalAnswer, substitutedSmtConstraint);
>   }
> ```

Add a test that would have caught this bug.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1638)*

anothermattbrown:
closing, not sure this is the right change to make

PR #1639, opened 2018-06-12
a few random fixes to reduced reachability

anothermattbrown:
- remove blacklistQuery (to be replaced later)
- honor the actions from input parameters
- enable specialization


anothermattbrown:
superseded.

PR #1664, opened 2018-06-14
More Flow fixes.

anothermattbrown:
- Null annotations on all object fields.
- Change builder field types to match corresponding Flow fields.
- Maintain and check the invariant that exactly 1 of _ingressVrf and
  _ingressInterface is non-null.
- Use Object.equals instead of Objects.equals for @Nonnull fields.
- Include _ingressInterface in compareTo
- Remove setter for final field _ingressInterface

The change to compareTo broke a unit test (because more flows are now
considered distinct), so fixed that too.

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/allinone/SpecifiersReachabilityQuestionTest.java
+++ b/projects/allinone/src/test/java/org/batfish/allinone/SpecifiersReachabilityQuestionTest.java
@@ -2,6 +2,7 @@ package org.batfish.allinone;
 
 import static org.batfish.datamodel.matchers.FlowHistoryInfoMatchers.hasFlow;
 import static org.batfish.datamodel.matchers.FlowMatchers.hasDstIp;
+import static org.batfish.datamodel.matchers.FlowMatchers.hasIngressInterface;
 import static org.batfish.datamodel.matchers.FlowMatchers.hasIngressNode;
 import static org.batfish.datamodel.matchers.FlowMatchers.hasSrcIp;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -153,13 +154,39 @@ public class SpecifiersReachabilityQuestionTest {
     AnswerElement answer = new SpecifiersReachabilityAnswerer(question, _batfish).answer();
     assertThat(answer, instanceOf(FlowHistory.class));
     Collection<FlowHistoryInfo> flowHistoryInfos = ((FlowHistory) answer).getTraces().values();
-    assertThat(flowHistoryInfos, hasSize(2));
+    assertThat(flowHistoryInfos, hasSize(4));
     assertThat(
         flowHistoryInfos,
-        hasItem(hasFlow(allOf(hasIngressNode(NODE1), hasSrcIp(new Ip("5.5.5.5"))))));
+        hasItem(
+            hasFlow(
+                allOf(
+                    hasIngressNode(NODE1),
+                    hasIngressInterface(FAST_ETHERNET),
+                    hasSrcIp(new Ip("5.5.5.5"))))));
     assertThat(
         flowHistoryInfos,
-        hasItem(hasFlow(allOf(hasIngressNode(NODE2), hasSrcIp(new Ip("5.5.5.5"))))));
+        hasItem(
+            hasFlow(
+                allOf(
+                    hasIngressNode(NODE1),
+                    hasIngressInterface(LOOPBACK),
+                    hasSrcIp(new Ip("5.5.5.5"))))));
+    assertThat(
+        flowHistoryInfos,
+        hasItem(
+            hasFlow(
+                allOf(
+                    hasIngressNode(NODE2),
+                    hasIngressInterface(FAST_ETHERNET),
+                    hasSrcIp(new Ip("5.5.5.5"))))));
+    assertThat(
+        flowHistoryInfos,
+        hasItem(
+            hasFlow(
+                allOf(
+                    hasIngressNode(NODE2),
+                    hasIngressInterface(LOOPBACK),
+                    hasSrcIp(new Ip("5.5.5.5"))))));
   }
 
   /**
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
@@ -166,6 +166,19 @@ public class CommonUtil {
     return collection == null ? null : collection.isEmpty() ? null : collection;
   }
 
+  /** Compare two nullable comparable objects. null is considered less than non-null. */
+  public static <T extends Comparable<T>> int compareNullable(@Nullable T a, @Nullable T b) {
+    if (a == b) {
+      return 0;
+    } else if (a == null) {
+      return -1;
+    } else if (b == null) {
+      return 1;
+    } else {
+      return a.compareTo(b);
+    }
+  }
+
   private static class TrustAllHostNameVerifier implements HostnameVerifier {
     @Override
     public boolean verify(String hostname, SSLSession session) {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java
@@ -535,7 +535,7 @@ public final class Flow implements Comparable<Flow>, Serializable {
   }
 
   @JsonCreator
-  public static Flow createFlow(
+  private static Flow createFlow(
       @JsonProperty(PROP_INGRESS_NODE) String ingressNode,
       @JsonProperty(PROP_INGRESS_INTERFACE) String ingressInterface,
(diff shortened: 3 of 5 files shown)


arifogel:
Reviewed 5 of 5 files at r1.
Review status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 170 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzVhp31486-s-ILlqE:-LEzVhp31486-s-ILlqF:b-njfj51) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L170)):*
> ```Java
> 
>   /** Compare two nullable comparable objects. null is considered less than non-null. */
>   public static <T extends Comparable<T>> int compareNullable(@Nullable T a, @Nullable T b) {
> ```

1. I don't think you need to handle b being `null`; `a::compareTo` should deal with that.
2. There is already a library way to do this: `Objects.compare(a, b, Comparator.naturalOrder())`

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java, line 481 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzY6q54XfpYwW5x4qf:-LEzY6q54XfpYwW5x4qg:b-i7kuwj) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java#L481)):*
> ```Java
>   @JsonCreator
>   public Flow(
>       @Nonnull @JsonProperty(PROP_INGRESS_NODE) String ingressNode,
> ```

I am wary of using `@Nullable` and `@Nonnull` in a `@JsonCreator`, for which you actually cannot make guarantees on the values with which the function is called.
Instead, I would make a static method the `@JsonCreator`, and have it call this function (where annotations can be enforced). So keep what you have here, but make a same-argument static method that calls this and is the sole `@JsonCreator`.

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/FlowMatchersImpl.java, line 35 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzYwqsALZ2nq9_hmsO:-LEzYwqsALZ2nq9_hmsP:b2x7gr5) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/FlowMatchersImpl.java#L35)):*
> ```Java
>   public static class HasIngressInterface extends FeatureMatcher<Flow, String> {
>     HasIngressInterface(Matcher<? super String> subMatcher) {
>       super(subMatcher, "ingressInterface", "ingressInterface");
> ```

Second argument should be: `"A Flow with ingressInterface:"`

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

anothermattbrown:
Review status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 170 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzVhp31486-s-ILlqE:-LEzgkqH2Nz08B6011cN:b-jkn13s) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L170)):*

1. I don't think you need to handle b being `null`; `a::compareTo` should deal with that.
2. There is already a library way to do this: `Objects.compare(a, b, Comparator.naturalOrder())`
</blockquote></details>

From naturalOrder javadoc: "The returned comparator is serializable and throws `NullPointerException` when comparing `null`", so we need at least the first two branches. Removing the third branch seems risky; it requires the assumption that `T::compareTo` treats `null` the same way here, which is not a safe assumption. For example, `String::compareTo` throws `NPE` on `null` input.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

anothermattbrown:
Review status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @arifogel)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java, line 481 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzY6q54XfpYwW5x4qf:-LEzkz3H4E0tDjzBHJpU:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java#L481)):*

I am wary of using `@Nullable` and `@Nonnull` in a `@JsonCreator`, for which you actually cannot make guarantees on the values with which the function is called.
Instead, I would make a static method the `@JsonCreator`, and have it call this function (where annotations can be enforced). So keep what you have here, but make a same-argument static method that calls this and is the sole `@JsonCreator`.
</blockquote></details>

Done.

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/FlowMatchersImpl.java, line 35 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzYwqsALZ2nq9_hmsO:-LEziwjqDM5pWD59-lbO:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/FlowMatchersImpl.java#L35)):*

Second argument should be: `"A Flow with ingressInterface:"`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

arifogel:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @anothermattbrown)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 170 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzVhp31486-s-ILlqE:-LEzryXgAS-dKhuerb-i:b86zgjw) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L170)):*

From naturalOrder javadoc: "The returned comparator is serializable and throws `NullPointerException` when comparing `null`", so we need at least the first two branches. Removing the third branch seems risky; it requires the assumption that `T::compareTo` treats `null` the same way here, which is not a safe assumption. For example, `String::compareTo` throws `NPE` on `null` input.
</blockquote></details>

Doesn't `Objects.compare` take care of the `null` pieces? Or only in the first position?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java, line 538 at r2](https://reviewable.io/reviews/batfish/batfish/1664#-LEzs7rJABGKzdFBYR6A:-LEzs7rK8vCaRQgW6yUy:b-bf06vf) ([raw file](https://github.com/batfish/batfish/blob/824f53fdfb0d1594c461e23bec60f4b0fd216972/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java#L538)):*
> ```Java
> 
>   @JsonCreator
>   public static Flow createFlow(
> ```

Should be able to make this `private`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

anothermattbrown:
Review status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java, line 170 at r1](https://reviewable.io/reviews/batfish/batfish/1664#-LEzVhp31486-s-ILlqE:-LEzsvXL2MNvghRRU_Ay:bscio5n) ([raw file](https://github.com/batfish/batfish/blob/10c22b94044e4b23da74d026abecfaa62f68ed00/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java#L170)):*

Doesn't `Objects.compare` take care of the `null` pieces? Or only in the first position?
</blockquote></details>

No, it only handles when both objects are `null`, and otherwise delegates to the comparator:
```
    public static <T> int compare(T a, T b, Comparator<? super T> c) {
        return (a == b) ? 0 :  c.compare(a, b);
    }
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

arifogel:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1664)*

PR #1696, opened 2018-06-18
JsonPathQuestion should delegate to its inner question's _differential

anothermattbrown:
added dummy Json property setter for deserialization

### DIFF
--- a/projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java
+++ b/projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java
@@ -1,6 +1,9 @@
 package org.batfish.question.jsonpath;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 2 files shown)


arifogel:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

---

*[projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java, line 456 at r2](https://reviewable.io/reviews/batfish/batfish/1696#-LFKDT-O9u7FaiukCUlg:-LFKDT-P4s2kFMmEqIms:bnj9ocw) ([raw file](https://github.com/batfish/batfish/blob/ab4cd8d18e1d4ffbc788a4706a90e007f482b9c7/projects/question/src/main/java/org/batfish/question/jsonpath/JsonPathQuestionPlugin.java#L456)):*
> ```Java
>     @JsonProperty(BfConsts.PROP_DIFFERENTIAL)
>     @Override
>     public boolean getDifferential() {
> ```

Weren't you going to ignore this? Was that not committed yet?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1696)*

anothermattbrown:
After discussion with @dhalperi, there may be a use case for keeping the differential flags on the outer and inner question, and interpreting each differently. So this change does not appear to the right solution.

PR #1718, opened 2018-06-20
BatfishStorage: support writing a keyed file

dhalperi:


dhalperi:
Review status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @progwriter)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1718#-LFU_34z21o8p74bnHx7:-LFU_34z21o8p74bnHx8:b-8znw0q) (no related file):*
Working still

- [ ] needs tests
- [ ] need to decide if to make storage container-unaware.
- [ ] need to see if I should introduce a StorageKey class or something like that, or if storage should just build it from the String[].

Thoughts?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1718)*

progwriter:
Review status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dhalperi and @progwriter)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1718#-LFU_34z21o8p74bnHx7:-LFUcLDTCwlP1aNxLBAB:bmjw0be) (no related file):*

Working still

- [ ] needs tests
- [ ] need to decide if to make storage container-unaware.
- [ ] need to see if I should introduce a StorageKey class or something like that, or if storage should just build it from the String[].

Thoughts?
</blockquote></details>

If we decided to replace container/snapshot folder names with uuids would StorageKey help?

---

*[projects/batfish/src/main/java/org/batfish/main/BatfishStorage.java, line 189 at r1](https://reviewable.io/reviews/batfish/batfish/1718#-LFUbBqd73TkcUvWnH9C:-LFUbBqd73TkcUvWnH9D:bzelyat) ([raw file](https://github.com/batfish/batfish/blob/5f6f9c1b4653916c7823226b30ffe851779fa956/projects/batfish/src/main/java/org/batfish/main/BatfishStorage.java#L189)):*
> ```Java
> 
>     // Build up the path to the file to be written, ensuring every subpath is actually a subpath of
>     // the prefix. (Catches key == "../foo")
> ```

Almost seems like this should be done with something like security manager, where storage class can only write within global containersDir upon initialization

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1718)*

PR #1721, opened 2018-06-21
Reduce number of ISIS queues

progwriter:
We only care about incoming routes so reduce number of queues kept in memory.
Assuming `node1` is the source/tail of the edge, hence identifies remoteVr


arifogel:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1721)*

PR #1740, opened 2018-06-25
ibdp Node: do not inherit from comparable structure

progwriter:
- Use hostname for ordering
- Still not overriding hashcode or equals: deep versions would be incredibly expensive, and we never node as a map key or in a set.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/Node.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/Node.java
@@ -7,7 +7,7 @@ import javax.annotation.Nonnull;
 import org.batfish.datamodel.Configuration;
 
 /** Dataplane-specific encapsulation of {@link Configuration} */
-public final class Node implements Comparable<Node>, Serializable {
+public final class Node implements Serializable {
 
   private static final long serialVersionUID = 1L;
 
@@ -45,16 +45,5 @@ public final class Node implements Comparable<Node>, Serializable {
   @Nonnull
   SortedMap<String, VirtualRouter> getVirtualRouters() {
     return _virtualRouters;
-  }
-
-  /** Return the hostname of this node. */
-  @Nonnull
-  String getHostName() {
-    return _c.getHostname();
-  }
-
-  @Override
-  public int compareTo(@Nonnull Node o) {
-    return getHostName().compareTo(o.getHostName());
   }
 }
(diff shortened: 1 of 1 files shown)


dhalperi:
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/Node.java, line 58 at r1](https://reviewable.io/reviews/batfish/batfish/1740#-LFsHxKy9tSo4t-DaPFK:-LFsHxKz47hZq4cW9VYx:b9gg46m) ([raw file](https://github.com/batfish/batfish/blob/2839eec616c93277fa695b5290c67bcd00b4ecbc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/Node.java#L58)):*
> ```Java
>   @Override
>   public int compareTo(@Nonnull Node o) {
>     return getHostName().compareTo(o.getHostName());
> ```

This seems wrong... if o == this return true, otherwise if compareTo returns 0 then crash?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1740)*

arifogel:
Review status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @progwriter and @arifogel)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1740#-LFsKyzZ7GeOcgXKcWRS:-LFsKyz_Cq1q6ApMAn9B:bjmpz82) (no related file):*
At this point I am of the opinion that only value types should be `Comparable`. The reason is that there is an expectation of value-ness for `equals` and `hashCode` when something is `Comparable`. If you really need to sort some items that are not value types, you should just provide a `Comparator` at the time.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1740)*

progwriter:
Review status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @progwriter and @arifogel)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1740#-LFsKyzZ7GeOcgXKcWRS:-LFsMWfi4XBrtqAcXCsH:b-shzueq) (no related file):*

At this point I am of the opinion that only value types should be `Comparable`. The reason is that there is an expectation of value-ness for `equals` and `hashCode` when something is `Comparable`. If you really need to sort some items that are not value types, you should just provide a `Comparator` at the time.
</blockquote></details>

Done, removed comparable all together. Turns out we never put node in a sorted collection.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1740)*

arifogel:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1740#-:-LFsNXVE9RcMFrGZT-3W:bnfp4nl)*

PR #1750, opened 2018-06-25
make FlexibleLocationSpecifierFactory tolerant of null inputs.

anothermattbrown:


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/1750#-LFtOKww3h5WZ6AO41Ss:-LFtOKwxA6Z8TG0-5kUi:bae2iq5) (no related file):*
LGTM as-is, but do you want to change something about the specifier interface to move the `null`-handling to a different layer? So that `buildLocationSpecifier` can only be called with `@Nonnull` value, say?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

dhalperi:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

anothermattbrown:
Review status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1750#-LFtOKww3h5WZ6AO41Ss:-LFtOuSB6D4vYceFroZy:b-2gaexq) (no related file):*

LGTM as-is, but do you want to change something about the specifier interface to move the `null`-handling to a different layer? So that `buildLocationSpecifier` can only be called with `@Nonnull` value, say?
</blockquote></details>

My thinking was that letting factories handle null would make it easier to keep defaulting behavior consistent across questions and to keep question implementations simple. For example, all questions could have this logic:

```
public Set<Location> getLocations() {
  return LocationSpecifierFactory.load(
      firstNonNull(_locationSpecifierFactory,DEFAULT_LOCATION_SPECIFIER_FACTORY))
    .buildLocationSpecifier(_locationSpecifierInput);
}
```

Otherwise, each question would have to do something like the following:
```
  if (_locationSpecifierFactory == null && _locationSpecifierInput == null) {
    // return a default specifier or throw an error.
  }
  checkNotNull(_locationSpecifierInput, "Cannot supply a LocationSpecifierFactory without an input");
  return LocationSpecifierFactory.load(
      firstNonNull(_locationSpecifierFactory,DEFAULT_LOCATION_SPECIFIER_FACTORY))
    .buildLocationSpecifier(_locationSpecifierInput);
  }
```

I would like to facilitate a standardized question behavior, but we can achieve that with or without nullable factory inputs. In particular, we can abstract out a method for the second snippet. So the question comes to: do we want all, some, or none of the factories to handle null input? I don't have clear preference. If you'd prefer that factories require `@Nonnull`  input, let's do that.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

dhalperi:
Review status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1750#-LFtOKww3h5WZ6AO41Ss:-LFwz3DY2mphDsZRhmd_:b30ay06) (no related file):*

My thinking was that letting factories handle null would make it easier to keep defaulting behavior consistent across questions and to keep question implementations simple. For example, all questions could have this logic:

```
public Set<Location> getLocations() {
  return LocationSpecifierFactory.load(
      firstNonNull(_locationSpecifierFactory,DEFAULT_LOCATION_SPECIFIER_FACTORY))
    .buildLocationSpecifier(_locationSpecifierInput);
}
```

Otherwise, each question would have to do something like the following:
```
  if (_locationSpecifierFactory == null && _locationSpecifierInput == null) {
    // return a default specifier or throw an error.
  }
  checkNotNull(_locationSpecifierInput, "Cannot supply a LocationSpecifierFactory without an input");
  return LocationSpecifierFactory.load(
      firstNonNull(_locationSpecifierFactory,DEFAULT_LOCATION_SPECIFIER_FACTORY))
    .buildLocationSpecifier(_locationSpecifierInput);
  }
```

I would like to facilitate a standardized question behavior, but we can achieve that with or without nullable factory inputs. In particular, we can abstract out a method for the second snippet. So the question comes to: do we want all, some, or none of the factories to handle null input? I don't have clear preference. If you'd prefer that factories require `@Nonnull`  input, let's do that.
</blockquote></details>

Sorry for being unclear.

I was wondering if an enhancement to factory API could solve this. E.g., add a `@Nonnull getDefaultValue()` function to the factory. Then the upper layer could call this function for missing input, and call build with `@Nonnull input` otherwise.

[Enforcing nullness annotations as above might mean that we're complicating factories that don't actually use the `input` -- not sure of answer, but it might be okay to make the (relatively few) factories that don't need a default create a singleton instead.]

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

dhalperi:
Review status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1750#-LFtOKww3h5WZ6AO41Ss:-LFx-A84AM8X9QULzvON:b-56ezc3) (no related file):*

Sorry for being unclear.

I was wondering if an enhancement to factory API could solve this. E.g., add a `@Nonnull getDefaultValue()` function to the factory. Then the upper layer could call this function for missing input, and call build with `@Nonnull input` otherwise.

[Enforcing nullness annotations as above might mean that we're complicating factories that don't actually use the `input` -- not sure of answer, but it might be okay to make the (relatively few) factories that don't need a default create a singleton instead.]
</blockquote></details>

Actually, I'd probably call it `getDefaultInput`.

This might have the added benefit of making factories think about their defaults...

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

anothermattbrown:
Review status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1750#-LFtOKww3h5WZ6AO41Ss:-LFx1R6M6ltGJ2GfTI9F:bod2n2a) (no related file):*

Actually, I'd probably call it `getDefaultInput`.

This might have the added benefit of making factories think about their defaults...
</blockquote></details>

It's not clear to me why having a `@Nonnull getDefaultInput()` and making do null checking and default resolution is better than just letting the factories do it themselves. Is there a general principle at work here?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

dhalperi:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

dhalperi:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1750)*

PR #1765, opened 2018-06-26
remove Comparable implementation from VerboseBgpEdge

anothermattbrown:


anothermattbrown:
Turns out the Comparable implementation is needed, only this isn't enforced by the static type system. Closing in favor of fixing the implementation rather than removing it.

dhalperi:
Or a customer comparator supplied to the treeset in question?

---







*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1765#-:-LG123QwC79GzGg3OTcF:bk3vz3f)*

PR #1774, opened 2018-06-27
CCPE: minor cleanups

dhalperi:
* Replace lambda by method reference
* Remove unused parameter
* Cleanup building lists
* StringBuilder instead of building string in a loop
* Remove or implement some empty if bodies

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
@@ -1,6 +1,7 @@
 package org.batfish.grammar.cisco;
 
 import static java.util.Comparator.naturalOrder;
+import static java.util.stream.Collectors.toCollection;
 import static org.batfish.datamodel.ConfigurationFormat.ARISTA;
 import static org.batfish.datamodel.ConfigurationFormat.ARUBAOS;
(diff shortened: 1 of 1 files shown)


arifogel:
:lgtm: basically though I am curious about that `assert`

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @progwriter)

---

*[projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java, line 1681 at r1](https://reviewable.io/reviews/batfish/batfish/1774#-LG2h70GAZVkXnjpjt8M:-LG2h70GAZVkXnjpjt8N:b4fnxcf) ([raw file](https://github.com/batfish/batfish/blob/08c9d7b78b3de0eb848f27bbc36dbec801406f03/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L1681)):*
> ```Java
>   public void exitCipt_mode(Cipt_modeContext ctx) {
>     if (ctx.TRANSPORT() != null) {
>       assert true; // do nothing for now
> ```

This is better? Did some tool complain?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1774#-:-LG2hjPfAbszHLVjqXAZ:bx9aokj)*

PR #1786, opened 2018-06-28
Add network configurations helper

progwriter:
*Purpose*:
Have a helper wrapper around `Map<String,Configuration>` that extracts desired objects from the configuration tree based on a key and simplifies a sequential null checks we do in a bunch of places.

First use case (coming soon): pass this around in the dataplane code and remove dependency on `Configuration` from `Node` class.

fyi @arifogel @anothermattbrown

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java
@@ -14,7 +14,7 @@ public final class NetworkConfigurations {
   @Nonnull private final Map<String, Configuration> _configurations;
 
   /** Wrap a configurations map */
-  public NetworkConfigurations(@Nonnull Map<String, Configuration> configurations) {
+  private NetworkConfigurations(@Nonnull Map<String, Configuration> configurations) {
     _configurations = requireNonNull(configurations);
   }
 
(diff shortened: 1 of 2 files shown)


dhalperi:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed, 1 unresolved discussion (waiting on @progwriter)

---

*[a discussion](https://reviewable.io/reviews/batfish/batfish/1786#-LG8pPoB-wYHzoyQoT_5:-LG8pPoB-wYHzoyQoT_6:b-et1wcf) (no related file):*
It's .. fine. Not overwhelmed by it, but fine – I trust you're going somewhere great with this ;)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java, line 43 at r1](https://reviewable.io/reviews/batfish/batfish/1786#-LG8p4k8CVGDHUrjk-9t:-LG8p4k8CVGDHUrjk-9u:b-qvia45) ([raw file](https://github.com/batfish/batfish/blob/87fc5039de0816ac5c51f44dbd5a5f124a4720a6/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java#L43)):*
> ```Java
>   public static NetworkConfigurations of(@Nonnull Map<String, Configuration> configurations) {
>     return new NetworkConfigurations(configurations);
>   }
> ```

why have both public constructor and public static factory method?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1786)*

PR #1788, opened 2018-06-28
IKE phase 1 extraction to changed data-model

haverma:
- and correcting some typos

### DIFF
--- a/projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
+++ b/projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
@@ -127,7 +127,7 @@ public class AllInOne {
         e.printStackTrace();
         System.exit(1);
       }
-      _client.run(new LinkedList<String>());
+      _client.run(new LinkedList<>());
       // The program does not terminate without it if the user misses the
       // quit command
       System.exit(0);
--- a/projects/allinone/src/main/java/org/batfish/allinone/config/Settings.java
+++ b/projects/allinone/src/main/java/org/batfish/allinone/config/Settings.java
@@ -188,7 +188,7 @@ public class Settings extends BaseSettings {
     }
 
     if (getBooleanOptionValue(ARG_VERSION)) {
-      System.out.printf(Version.getCompleteVersionString());
+      System.out.print(Version.getCompleteVersionString());
       System.exit(0);
     }
 
--- a/projects/allinone/src/main/java/org/batfish/allinone/config/Settings.java
+++ b/projects/allinone/src/main/java/org/batfish/allinone/config/Settings.java
@@ -153,7 +153,7 @@ public class Settings extends BaseSettings {
         "mode in which to start batfish",
         Arrays.stream(RunMode.values())
             .filter(v -> v != RunMode.WORKER)
-            .map(v -> v.toString())
+            .map(Object::toString)
             .collect(Collectors.joining("|")));
 
     addOption(ARG_CLIENT_ARGS, "arguments for the client process", "client_args");
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -605,9 +605,7 @@ public class Client extends AbstractClient implements IClient {
     }
     _logger.debug("Uploaded question. Answering now.\n");
     // delete the temporary params file
-    if (questionFile != null) {
-      CommonUtil.deleteIfExists(questionFile);
-    }
+    CommonUtil.deleteIfExists(questionFile);
     // answer the question
     WorkItem wItemAs =
         WorkItemBuilder.getWorkItemAnswerQuestion(
@@ -692,7 +690,7 @@ public class Client extends AbstractClient implements IClient {
         String parameterValue = e.getValue().toString();
         Object parameterObj;
         try {
-          parameterObj = new JSONTokener(parameterValue.toString()).nextValue();
+          parameterObj = new JSONTokener(parameterValue).nextValue();
           questionJson.put(parameterName, parameterObj);
         } catch (JSONException e1) {
           throw new BatfishException(
@@ -2230,9 +2228,8 @@ public class Client extends AbstractClient implements IClient {
       sourceMap
           .get(questionName)
           .forEach(
-              questionContent -> {
-                updateLoadedQuestionsInfo(questionName, questionContent, destinationMap, ae);
-              });
+              questionContent ->
+                  updateLoadedQuestionsInfo(questionName, questionContent, destinationMap, ae));
     }
   }
 
--- a/projects/batfish-client/src/main/java/org/batfish/client/Client.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Client.java
@@ -23,7 +23,6 @@ import io.opentracing.ActiveSpan;
 import io.opentracing.util.GlobalTracer;
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
@@ -539,7 +538,7 @@ public class Client extends AbstractClient implements IClient {
     }
   }
 
-  public Client(String[] args) throws Exception {
+  public Client(String[] args) {
     this(new Settings(args));
   }
 
@@ -756,7 +755,7 @@ public class Client extends AbstractClient implements IClient {
     return true;
   }
 
-  private boolean cat(String[] words) throws IOException, FileNotFoundException {
+  private boolean cat(String[] words) throws IOException {
     if (words.length != 2) {
       _logger.errorf("Invalid arguments: %s\n", Arrays.toString(words));
       printUsage(Command.CAT);
@@ -783,7 +782,7 @@ public class Client extends AbstractClient implements IClient {
     return true;
   }
 
-  private boolean clearScreen(List<String> options, List<String> parameters) throws IOException {
+  private boolean clearScreen(List<String> options, List<String> parameters) {
     if (!isValidArgument(options, parameters, 0, 0, 0, Command.CLEAR_SCREEN)) {
       return false;
     }
@@ -1624,11 +1623,7 @@ public class Client extends AbstractClient implements IClient {
     WorkItem wItemProcessEnv =
         WorkItemBuilder.getWorkItemProcessEnvironment(
             _currContainerName, _currDeltaTestrig, _currDeltaEnv);
-    if (!execute(wItemProcessEnv, outWriter)) {
-      return false;
-    }
-
-    return true;
+    return execute(wItemProcessEnv, outWriter);
   }
 
   private boolean initEnvironment(
@@ -2740,11 +2735,7 @@ public class Client extends AbstractClient implements IClient {
 
     WorkItem wItemParse = WorkItemBuilder.getWorkItemParse(_currContainerName, testrig);
 
-    if (!execute(wItemParse, outWriter)) {
-      return false;
-    }
-
-    return true;
+    return execute(wItemParse, outWriter);
   }
 
   public void run(List<String> initialCommands) {
--- a/projects/batfish-client/src/main/java/org/batfish/client/CommandCompleter.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/CommandCompleter.java
@@ -23,11 +23,7 @@ public class CommandCompleter implements Completer {
     String buffer = parsedLine.word().trim();
 
     if (Strings.isNullOrEmpty(buffer)) {
-      candidates.addAll(
-          _commandStrs
-              .stream()
-              .map(command -> new Candidate(command))
-              .collect(Collectors.toList()));
+      candidates.addAll(_commandStrs.stream().map(Candidate::new).collect(Collectors.toList()));
     } else {
       for (String match : _commandStrs.tailSet(buffer)) {
         if (!match.startsWith(buffer)) {
--- a/projects/batfish-client/src/main/java/org/batfish/client/Main.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/Main.java
@@ -24,6 +24,6 @@ public class Main {
     }
 
     Client client = new Client(settings);
-    client.run(new LinkedList<String>());
+    client.run(new LinkedList<>());
   }
 }
--- a/projects/batfish-client/src/main/java/org/batfish/client/QuestionHelper.java
+++ b/projects/batfish-client/src/main/java/org/batfish/client/QuestionHelper.java
@@ -92,7 +92,7 @@ public class QuestionHelper {
             .readValue(instanceJson.toString(), new TypeReference<InstanceData>() {});
     Map<String, Variable> variables = instanceData.getVariables();
 
-    /**
+    /*
      * We want all variables to be exercised (i.e, assigned values) as part of validation because
      * unexercised parameters and variables are removed as part of template preprocessing in {@link
      * Question#preprocessQuestion(String)} and then we cannot know if the parameter programmed in
--- a/projects/batfish-common-protocol/BUILD
+++ b/projects/batfish-common-protocol/BUILD
@@ -148,7 +148,7 @@ java_library(
         ":RecoveryParserListener.java",
     ],
     deps = [
-        ":common",
+        ":parser_common",
         "@antlr4_runtime//:compile",
     ],
 )
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
@@ -3,7 +3,6 @@ package org.batfish.common;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -29,8 +28,7 @@ public class Warnings implements Serializable {
   public static class Deserializer extends JsonDeserializer<Warnings> {
 
     @Override
-    public Warnings deserialize(JsonParser parser, DeserializationContext ctx)
-        throws IOException, JsonProcessingException {
+    public Warnings deserialize(JsonParser parser, DeserializationContext ctx) throws IOException {
       JsonNode node = parser.getCodec().readTree(parser);
       Warnings warnings = new Warnings();
       if (node.has(PEDANTIC_VAR)) {
@@ -65,7 +63,7 @@ public class Warnings implements Serializable {
 
     @Override
     public void serialize(Warnings value, JsonGenerator jgen, SerializerProvider provider)
-        throws IOException, JsonProcessingException {
+        throws IOException {
       jgen.writeStartObject();
       if (!value._pedanticWarnings.isEmpty()) {
         jgen.writeFieldName(PEDANTIC_VAR);
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CommonUtil.java
@@ -1007,7 +1007,7 @@ public class CommonUtil {
    * DataPlane)} for more details.
    *
    * @param configurations configuration keyed by hostname
-   * @param ipOwners Ip owners (see {@link #computeIpNodeOwners(boolean, Map)}
+   * @param ipOwners Ip owners (see {@link #computeIpNodeOwners(Map, boolean)}
    * @param keepInvalid whether to keep improperly configured neighbors. If performing configuration
    *     checks, you probably want this set to {@code true}, otherwise (e.g., computing dataplane)
    *     you want this to be {@code false}.
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/JsonPathResult.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/JsonPathResult.java
@@ -1,8 +1,8 @@
 package org.batfish.common.util;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ArrayNode;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import org.batfish.common.BatfishException;
@@ -58,7 +58,7 @@ public class JsonPathResult {
   private static List<String> getPrefixParts(JsonNode prefix) {
     String text = prefix.textValue();
     if (text.equals("$")) {
-      return Arrays.asList("$");
+      return Collections.singletonList("$");
     }
     if (text.length() < 2) {
       throw new BatfishException("Unexpected prefix " + text);
@@ -83,6 +83,6 @@ public class JsonPathResult {
 
   /** Tells us if the suffix is null or empty */
   public boolean isNullOrEmptySuffix() {
-    return _suffix == null || (_suffix.isArray() && ((ArrayNode) _suffix).size() == 0);
+    return _suffix == null || (_suffix.isArray() && _suffix.size() == 0);
   }
 }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/JuniperUtils.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/JuniperUtils.java
@@ -66,7 +66,7 @@ public final class JuniperUtils {
     char first = nibble(chars, 1).charAt(0);
     nibble(chars, _characterFamilyReverseIndexMap.get(first));
     char prev = first;
-    String decrypted = "";
+    StringBuilder decrypted = new StringBuilder();
 
     while (chars[0].length() > 0) {
       List<Integer> decode = _codeMatrix.get(decrypted.length() % _codeMatrix.size());
@@ -81,9 +81,9 @@ public final class JuniperUtils {
         gaps.add(g);
       }
       char newChar = gapDecode(gaps, decode);
-      decrypted += newChar;
+      decrypted.append(newChar);
     }
-    return decrypted;
+    return decrypted.toString();
   }
 
   private int gap(char a, char b) {
@@ -138,13 +138,13 @@ public final class JuniperUtils {
 
   private List<List<Integer>> initCodeMatrix() {
     List<List<Integer>> codeMatrix = new ArrayList<>();
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 4, 32}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 16, 32}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 8, 32}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 64}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 32}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 4, 16, 128}));
-    codeMatrix.add(Arrays.asList(new Integer[] {1, 32, 64}));
+    codeMatrix.add(Arrays.asList(1, 4, 32));
+    codeMatrix.add(Arrays.asList(1, 16, 32));
+    codeMatrix.add(Arrays.asList(1, 8, 32));
+    codeMatrix.add(Arrays.asList(1, 64));
+    codeMatrix.add(Arrays.asList(1, 32));
+    codeMatrix.add(Arrays.asList(1, 4, 16, 128));
+    codeMatrix.add(Arrays.asList(1, 32, 64));
     return codeMatrix;
   }
 
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/WorkItemBuilder.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/WorkItemBuilder.java
@@ -131,9 +131,8 @@ public class WorkItemBuilder {
 
     String testrig = workItem.getTestrigName();
     String envName =
-        reqParams.containsKey(BfConsts.ARG_ENVIRONMENT_NAME)
-            ? reqParams.get(BfConsts.ARG_ENVIRONMENT_NAME)
-            : BfConsts.RELPATH_DEFAULT_ENVIRONMENT_NAME;
+        reqParams.getOrDefault(
+            BfConsts.ARG_ENVIRONMENT_NAME, BfConsts.RELPATH_DEFAULT_ENVIRONMENT_NAME);
     String deltaTestrig = reqParams.get(BfConsts.ARG_DELTA_TESTRIG);
     String deltaEnvName = reqParams.get(BfConsts.ARG_DELTA_ENVIRONMENT_NAME);
     if (deltaEnvName != null && deltaTestrig == null) {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute.java
@@ -14,7 +14,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
 @ParametersAreNonnullByDefault
 public abstract class AbstractRoute implements Serializable, Comparable<AbstractRoute> {
 
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute6.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute6.java
@@ -6,7 +6,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import java.io.Serializable;
 import org.batfish.common.BatfishException;
 
-@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
 public abstract class AbstractRoute6 implements Serializable {
 
   protected static final String PROP_ADMINISTRATIVE_COST = "administrativeCost";
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpPeerConfig.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpPeerConfig.java
@@ -5,18 +5,18 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import com.kjetland.jackson.jsonSchema.annotations.JsonSchemaDescription;
+import java.io.Serializable;
 import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import javax.annotation.Nullable;
-import org.batfish.common.util.ComparableStructure;
 import org.batfish.datamodel.NetworkFactory.NetworkFactoryBuilder;
 
 /** Represents a configured BGP peering, at the control plane level */
 @JsonSchemaDescription("A configured e/iBGP peering relationship")
-public final class BgpPeerConfig extends ComparableStructure<Prefix> {
+public final class BgpPeerConfig implements Serializable {
 
   public static class Builder extends NetworkFactoryBuilder<BgpPeerConfig> {
     private Configuration _owner;
@@ -205,8 +205,6 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
 
   private static final String PROP_LOCAL_IP = "localIp";
 
-  private static final String PROP_NAME = "name";
-
   private static final String PROP_OWNER = "owner";
 
   private static final String PROP_REMOTE_AS = "remoteAs";
@@ -277,6 +275,12 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
   /** Whether this session is passive/dynamic (e.g., listens for incoming connections only) */
   private boolean _dynamic;
 
+  /**
+   * The remote peer's IPv4 prefix. Currently nullable because: 1) we sometimes pass null in the
+   * constructor; 2) Jackson deserialization without any enforcement checks.
+   */
+  @Nullable private Prefix _prefix;
+
   /** The autonomous system number that the containing BGP process considers this peer to have. */
   private Long _remoteAs;
 
@@ -310,8 +314,8 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
   }
 
   @JsonCreator
-  public BgpPeerConfig(@JsonProperty(PROP_NAME) Prefix prefix) {
-    super(prefix);
+  public BgpPeerConfig(@Nullable @JsonProperty(PROP_REMOTE_PREFIX) Prefix prefix) {
+    _prefix = prefix;
     _exportPolicySources = new TreeSet<>();
     _generatedRoutes = new LinkedHashSet<>();
     _importPolicySources = new TreeSet<>();
@@ -395,6 +399,37 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
         && this._dynamic == other._dynamic;
   }
 
+  @Override
+  public int hashCode() {
+    // SKIPS: description, generated routes, owner to be consistent with equals
+    return Objects.hash(
+        _additionalPathsReceive,
+        _additionalPathsSelectAll,
+        _additionalPathsSend,
+        _advertiseExternal,
+        _advertiseInactive,
+        _allowLocalAsIn,
+        _allowRemoteAsOut,
+        _authenticationSettings,
+        _clusterId,
+        _defaultMetric,
+        _ebgpMultihop,
+        _enforceFirstAs,
+        _exportPolicy,
+        _exportPolicySources,
+        _group,
+        _importPolicy,
+        _importPolicySources,
+        _localAs,
+        _localIp,
+        _dynamic,
+        _prefix,
+        _remoteAs,
+        _routeReflectorClient,
+        _sendCommunity,
+        _vrf);
+  }
+
   @JsonProperty(PROP_ADDITIONAL_PATHS_RECEIVE)
   public boolean getAdditionalPathsReceive() {
     return _additionalPathsReceive;
@@ -414,8 +449,8 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
   @JsonProperty(PROP_ADDRESS)
   @JsonPropertyDescription("The IPV4 address of the remote peer if not dynamic (passive)")
   public Ip getAddress() {
-    if (_key != null && _key.getPrefixLength() == Prefix.MAX_PREFIX_LENGTH) {
-      return _key.getStartIp();
+    if (_prefix != null && _prefix.getPrefixLength() == Prefix.MAX_PREFIX_LENGTH) {
+      return _prefix.getStartIp();
     } else {
       return null;
     }
@@ -559,7 +594,7 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
       "The remote (destination) IPV4 address of this peering (when not-dynamic), or the "
           + "network of peers allowed to connect on this peering (otherwise)")
   public Prefix getPrefix() {
-    return _key;
+    return _prefix;
   }
 
   @JsonProperty(PROP_REMOTE_AS)
@@ -736,6 +771,6 @@ public final class BgpPeerConfig extends ComparableStructure<Prefix> {
 
   @Override
   public String toString() {
-    return "BgpPeerConfig<Prefix:" + _key + ", AS:" + _remoteAs + ">";
+    return "BgpPeerConfig<Prefix:" + _prefix + ", AS:" + _remoteAs + ">";
   }
 }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.ImmutableSortedSet;
 import com.kjetland.jackson.jsonSchema.annotations.JsonSchemaDescription;
 import java.util.Collections;
@@ -14,6 +15,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.NavigableSet;
+import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -185,7 +187,7 @@ public final class Configuration extends ComparableStructure<String> {
 
   private NavigableMap<String, IkeProposal> _ikeProposals;
 
-  private NavigableMap<String, IkePhase1Key> _ikePhase1keys;
+  private @Nonnull SortedMap<String, IkePhase1Key> _ikePhase1keys;
 
   private NavigableMap<String, IkePhase1Proposal> _ikePhase1Proposals;
 
@@ -273,7 +275,7 @@ public final class Configuration extends ComparableStructure<String> {
     _ikeGateways = new TreeMap<>();
     _ikePolicies = new TreeMap<>();
     _ikeProposals = new TreeMap<>();
-    _ikePhase1keys = new TreeMap<>();
+    _ikePhase1keys = ImmutableSortedMap.of();
     _ikePhase1Policies = new TreeMap<>();
     _ikePhase1Proposals = new TreeMap<>();
     _interfaces = new TreeMap<>();
@@ -440,7 +442,7 @@ public final class Configuration extends ComparableStructure<String> {
 
   @JsonProperty(PROP_IKE_PHASE1_KEYS)
   @JsonPropertyDescription("Dictionary of all IKE phase1 keys for this node.")
-  public NavigableMap<String, IkePhase1Key> getIkePhase1Keys() {
+  public SortedMap<String, IkePhase1Key> getIkePhase1Keys() {
     return _ikePhase1keys;
   }
 
@@ -705,8 +707,9 @@ public final class Configuration extends ComparableStructure<String> {
   }
 
   @JsonProperty(PROP_IKE_PHASE1_KEYS)
-  public void setIkePhase1Keys(NavigableMap<String, IkePhase1Key> ikePhase1Keys) {
-    _ikePhase1keys = ikePhase1Keys;
+  public void setIkePhase1Keys(@Nullable Map<String, IkePhase1Key> ikePhase1Keys) {
+    _ikePhase1keys =
+        ikePhase1Keys == null ? ImmutableSortedMap.of() : ImmutableSortedMap.copyOf(ikePhase1Keys);
   }
 
   @JsonProperty(PROP_IKE_PHASE1_POLICIES)
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
@@ -15,7 +15,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.NavigableSet;
-import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -187,7 +186,7 @@ public final class Configuration extends ComparableStructure<String> {
 
   private NavigableMap<String, IkeProposal> _ikeProposals;
 
-  private @Nonnull SortedMap<String, IkePhase1Key> _ikePhase1keys;
+  private @Nonnull NavigableMap<String, IkePhase1Key> _ikePhase1keys;
 
   private NavigableMap<String, IkePhase1Proposal> _ikePhase1Proposals;
 
@@ -442,7 +441,7 @@ public final class Configuration extends ComparableStructure<String> {
 
   @JsonProperty(PROP_IKE_PHASE1_KEYS)
   @JsonPropertyDescription("Dictionary of all IKE phase1 keys for this node.")
-  public SortedMap<String, IkePhase1Key> getIkePhase1Keys() {
+  public NavigableMap<String, IkePhase1Key> getIkePhase1Keys() {
     return _ikePhase1keys;
   }
 
@@ -707,7 +706,7 @@ public final class Configuration extends ComparableStructure<String> {
   }
 
   @JsonProperty(PROP_IKE_PHASE1_KEYS)
-  public void setIkePhase1Keys(@Nullable Map<String, IkePhase1Key> ikePhase1Keys) {
+  public void setIkePhase1Keys(@Nullable NavigableMap<String, IkePhase1Key> ikePhase1Keys) {
     _ikePhase1keys =
         ikePhase1Keys == null ? ImmutableSortedMap.of() : ImmutableSortedMap.copyOf(ikePhase1Keys);
   }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
@@ -117,6 +117,8 @@ public final class Configuration extends ComparableStructure<String> {
 
   private static final String PROP_IKE_PROPOSALS = "ikeProposals";
 
+  private static final String PROP_IKE_PHASE1_KEYS = "ikePhase1Keys";
+
   private static final String PROP_IKE_PHASE1_POLICIES = "ikePhase1Policies";
 
   private static final String PROP_IKE_PHASE1_PROPOSALS = "ikePhase1Proposals";
@@ -182,6 +184,8 @@ public final class Configuration extends ComparableStructure<String> {
   private NavigableMap<String, IkePolicy> _ikePolicies;
 
   private NavigableMap<String, IkeProposal> _ikeProposals;
+
+  private NavigableMap<String, IkePhase1Key> _ikePhase1keys;
 
   private NavigableMap<String, IkePhase1Proposal> _ikePhase1Proposals;
 
@@ -269,6 +273,7 @@ public final class Configuration extends ComparableStructure<String> {
     _ikeGateways = new TreeMap<>();
     _ikePolicies = new TreeMap<>();
     _ikeProposals = new TreeMap<>();
+    _ikePhase1keys = new TreeMap<>();
     _ikePhase1Policies = new TreeMap<>();
     _ikePhase1Proposals = new TreeMap<>();
     _interfaces = new TreeMap<>();
@@ -431,6 +436,12 @@ public final class Configuration extends ComparableStructure<String> {
   @JsonPropertyDescription("Dictionary of all IKE proposals for this node.")
   public NavigableMap<String, IkeProposal> getIkeProposals() {
     return _ikeProposals;
+  }
+
+  @JsonProperty(PROP_IKE_PHASE1_KEYS)
+  @JsonPropertyDescription("Dictionary of all IKE phase1 keys for this node.")
+  public NavigableMap<String, IkePhase1Key> getIkePhase1Keys() {
+    return _ikePhase1keys;
   }
 
   @JsonProperty(PROP_IKE_PHASE1_POLICIES)
@@ -691,6 +702,11 @@ public final class Configuration extends ComparableStructure<String> {
   @JsonProperty(PROP_IKE_POLICIES)
   public void setIkePolicies(NavigableMap<String, IkePolicy> ikePolicies) {
     _ikePolicies = ikePolicies;
+  }
+
+  @JsonProperty(PROP_IKE_PHASE1_KEYS)
+  public void setIkePhase1Keys(NavigableMap<String, IkePhase1Key> ikePhase1Keys) {
+    _ikePhase1keys = ikePhase1Keys;
   }
 
   @JsonProperty(PROP_IKE_PHASE1_POLICIES)
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java
@@ -117,6 +117,10 @@ public final class Configuration extends ComparableStructure<String> {
 
   private static final String PROP_IKE_PROPOSALS = "ikeProposals";
 
+  private static final String PROP_IKE_PHASE1_POLICIES = "ikePhase1Policies";
+
+  private static final String PROP_IKE_PHASE1_PROPOSALS = "ikePhase1Proposals";
+
   private static final String PROP_IP_ACCESS_LISTS = "ipAccessLists";
 
   private static final String PROP_IP_SPACES = "ipSpaces";
@@ -178,6 +182,10 @@ public final class Configuration extends ComparableStructure<String> {
   private NavigableMap<String, IkePolicy> _ikePolicies;
 
   private NavigableMap<String, IkeProposal> _ikeProposals;
+
+  private NavigableMap<String, IkePhase1Proposal> _ikePhase1Proposals;
+
+  private NavigableMap<String, IkePhase1Policy> _ikePhase1Policies;
 
   private NavigableMap<String, Interface> _interfaces;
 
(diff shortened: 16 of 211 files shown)


dhalperi:
Nice!

---

Reviewed 11 of 14 files at r1, 6 of 6 files at r2.
Review status: all files reviewed, 12 unresolved discussions (waiting on @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 19 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mI7Y8iinAiUczs7X:-LG8mI7Y8iinAiUczs7Y:b-r4t9nn) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L19)):*
> ```Java
>   private static final String PROP_PRE_SHARED_KEY = "preSharedKey";
> 
>   private List<IkePhase1Proposal> _ikePhase1Proposals;
> ```

Add `@Nonnull` annotation here.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 26 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8lrYK7gSo6hyfyhjE:-LG8lrYK7gSo6hyfyhjF:b-6b6mnn) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L26)):*
> ```
> new ArrayList<>();
> ```

As much as possible, data model classes (especially new ones) should be built immutably. Some notes throughout on this change here. You can also ask me or @arifogel more in person.

Make this immutable using `ImmutableList.of()`.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 43 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8m0FTCrdQRyOkkyJ5:-LG8m0FTCrdQRyOkkyJ6:b4a97bn) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L43)):*
> ```Java
>   @JsonProperty(PROP_IKE_PHASE1_PROPOSALS)
>   public void setIkePhase1Proposals(List<IkePhase1Proposal> ikePhase1Proposals) {
>     _ikePhase1Proposals = ikePhase1Proposals;
> ```

because of Jackson, this is actually nullable (we'll serialize empty list as null). So make the argument `@Nullable`, and then the setter should be `_ikePhase1Proposals = input == null ? ImmutableList.of() : ImmutableList.copyOf(input);`.

Other option for handling nullable without adding `@Nullable` annotation is to put a default string in the `JsonProperty`: `@JsonProperty(value = PROP_IKE_PHASE1_PROPOSALS, defaultValue = "[]")`. Not sure how @arifogel feels about that one?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2185 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8gGQ724uI1x56gkv1:-LG8gGQ724uI1x56gkv2:b-r8m4l6) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2185)):*
> ```Java
>         .forEach(
>             (ikePhase1ProposalEntry ->
>                 ikePhase1Policy.getIkePhase1Proposals().add(ikePhase1ProposalEntry.getValue())));
> ```

These are not sorted by names, right? They're really sorted by number?

Actually, now that we have vendor-specific representation for the cisco Ike proposals, can we store them numerically to make this more direct?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2185 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mSr01nmdDY4CuNtL:-LG8mSr01nmdDY4CuNtM:b-wsukue) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2185)):*
> ```Java
>         .forEach(
>             (ikePhase1ProposalEntry ->
>                 ikePhase1Policy.getIkePhase1Proposals().add(ikePhase1ProposalEntry.getValue())));
> ```

rather than `getIkePhase1Proposals().add()`, just build up an ImmutableList here and call `setIkePhase1Proposals` once.

Same in Juniper side.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2190 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8l2AN0lmH5Be7fvAq:-LG8l2AODzX6JzTjwr9y:b-hfo28t) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2190)):*
> ```Java
>   }
> 
>   private String getMatchingPsk(IsakmpProfile isakmpProfile) {
> ```

This seems like an ideal function to unit test in isolation. Make it `static` (passing in the few pieces of local state you need, `_w` and `_keyrings`?). Then you can write unit tests of that static function to ensure that all the cases are handled well.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2199 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8k-bg-Ecvg-Q5zRMh:-LG8k-bg-Ecvg-Q5zRMi:b-53d160) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2199)):*
> ```
>     } else if (isakmpProfile.getKeyring() == null
>         || !_keyrings.containsKey(isakmpProfile.getKeyring())) {
> ```

these are two different error conditions, aren't they?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2218 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8kUUJAgMkCYpeRSWa:-LG8kUUJAgMkCYpeRSWb:b-phyw2n) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2218)):*
> ```Java
>   }
> 
>   private static IkePhase1Proposal toIkePhase1Proposal(IsakmpPolicy isakmpPolicy) {
> ```

move this to `CiscoConversions` utility class.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 45 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8fIe49SYBLhYnDr2s:-LG8fIe49SYBLhYnDr2t:b-cwohae) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L45)):*
> ```
> (_localAddress == null 
> ```

what's this change for?

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1192 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nCAV6H4gFij81-dN:-LG8nCAV6H4gFij81-dO:b-u5v9h9) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1192)):*
> ```Java
>                 ikePhase1Policy.getIkePhase1Proposals().add(ikePhase1Proposal);
>               }
>             });
> ```

similar comments as in cisco -- rather than get().add(), build an immutable list and set it.

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1995 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nMrL6gBKeG9OCwMe:-LG8nMrL6gBKeG9OCwMf:b-zbg199) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1995)):*
> ```Java
>       _c.getIkePolicies().put(name, newPolicy);
>       // storing IKE phase 1 policy
>       IkePhase1Policy ikePhase1Policy = toIkePhase1Policy(oldIkePolicy);
> ```

general style: inline this variable in the next line? It's not used anywhere else and it's a short, single-operation line.

---

*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 1294 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nnCv7iiNFnTJHMv3:-LG8nnCv7iiNFnTJHMv4:bsuxsqr) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L1294)):*
> ```
> *
> ```

typically, one would use a single `*` in block comments - they're not javadoc.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788#-:-LG8o9uG9UMXqbKTcLF7:b-83sjoh)*

haverma:
- Updated the datamodel to use IkePhase1Key and added localInterface, remoteIdentity and selfIdentity as discussed in https://docs.google.com/document/d/1EEpYPXGPaqD44b5v3_3gPFifl9gl3HAWAkihV71qQLw/edit?usp=sharing
- Addressed comments received earlier on this PR
- Using `IpWildcard` as `remoteIdentity` in `IkePhase1Key` and `IkePhase1Policy` as `IpSpace` cannot be used for doing `superset` or `subset` like comparisons currently, although we would want them to be `IpSpace` eventually

---

Review status: 5 of 28 files reviewed, 4 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 19 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mI7Y8iinAiUczs7X:-LGPU7hwBSxxsCH__bRp:bqkt339) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L19)):*

Add `@Nonnull` annotation here.
</blockquote></details>

Adding `@NonNull` here causes IDE to show a warning as in the setter we are declaring the argument as `@Nullable`(because of Jackson serialization, as pointed out in another comment)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 26 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8lrYK7gSo6hyfyhjE:-LGPUY9uEykNVGYalpzF:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L26)):*

> ```
> new ArrayList<>();
> ```

As much as possible, data model classes (especially new ones) should be built immutably. Some notes throughout on this change here. You can also ask me or @arifogel more in person.

Make this immutable using `ImmutableList.of()`.
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 43 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8m0FTCrdQRyOkkyJ5:-LGPUZng6jEUqeEL_J4o:bm89mc7) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L43)):*

because of Jackson, this is actually nullable (we'll serialize empty list as null). So make the argument `@Nullable`, and then the setter should be `_ikePhase1Proposals = input == null ? ImmutableList.of() : ImmutableList.copyOf(input);`.

Other option for handling nullable without adding `@Nullable` annotation is to put a default string in the `JsonProperty`: `@JsonProperty(value = PROP_IKE_PHASE1_PROPOSALS, defaultValue = "[]")`. Not sure how @arifogel feels about that one?
</blockquote></details>

done, but this prevents the declaration of `_ikePhase1Proposals` as a `@Nonnull` attribute.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2185 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8gGQ724uI1x56gkv1:-LGPUr8X2r_hx1YX3Bv_:bqvizh9) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2185)):*

These are not sorted by names, right? They're really sorted by number?

Actually, now that we have vendor-specific representation for the cisco Ike proposals, can we store them numerically to make this more direct?
</blockquote></details>

moved to `CiscoConversions`, these are sorted by numbers of the proposals but we are storing the number as names, so in-effect sorting by names achieves the purpose of sorting by numbers. We are iterating through the vendor independent `IkeProposals` which don't have any number/priority field so I cannot think of a way to make it use the number field more directly. Can discuss more in person

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2185 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mSr01nmdDY4CuNtL:-LGPVlQ1-bkzKOeEgsPP:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2185)):*

rather than `getIkePhase1Proposals().add()`, just build up an ImmutableList here and call `setIkePhase1Proposals` once.

Same in Juniper side.
</blockquote></details>

done

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2190 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8l2AN0lmH5Be7fvAq:-LGPVmBF8UgVfqcHs4iU:bwxrwld) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2190)):*

This seems like an ideal function to unit test in isolation. Make it `static` (passing in the few pieces of local state you need, `_w` and `_keyrings`?). Then you can write unit tests of that static function to ensure that all the cases are handled well.
</blockquote></details>

done, and added tests in `CiscoConfigurationTest` class

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2199 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8k-bg-Ecvg-Q5zRMh:-LGPVrx-4QTXZfRMCKjO:byeovm3) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2199)):*

> ```
>     } else if (isakmpProfile.getKeyring() == null
>         || !_keyrings.containsKey(isakmpProfile.getKeyring())) {
> ```

these are two different error conditions, aren't they?
</blockquote></details>

moved to `CiscoConversion` class, yes they are different error conditions, separated now

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2218 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8kUUJAgMkCYpeRSWa:-LGPW17M5uDoRV6GaxzA:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2218)):*

move this to `CiscoConversions` utility class.
</blockquote></details>

done

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 45 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8fIe49SYBLhYnDr2s:-LG9llnh9t3AhdgPD1sa:b-cs3bzo) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L45)):*

> ```
> (_localAddress == null 
> ```

what's this change for?
</blockquote></details>

There can be a case where `local-address` is not configured on a keyring, in which case keyring matching should only be done based on the IP of the remote peer. (https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/sec_conn_vpnips/configuration/12-4t/sec-sec-for-vpns-w-ipsec-12-4t-book/sec-safenet-suppt.html)

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1192 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nCAV6H4gFij81-dN:-LGPW63p9kzdkzUZVg61:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1192)):*

similar comments as in cisco -- rather than get().add(), build an immutable list and set it.
</blockquote></details>

done

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1995 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nMrL6gBKeG9OCwMe:-LGPW8M05Iqu-OZXI0VB:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1995)):*

general style: inline this variable in the next line? It's not used anywhere else and it's a short, single-operation line.
</blockquote></details>

done

---

*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 1294 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8nnCv7iiNFnTJHMv3:-LGPWBGVCUj6bFiSovaP:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L1294)):*

> ```
> *
> ```

typically, one would use a single `*` in block comments - they're not javadoc.
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788#-:-LGPSem85UM1Fpk0mIEy:bnzarxl)*

dhalperi:
Reviewed 16 of 23 files at r3.
Review status: 21 of 28 files reviewed, 17 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 708 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpU0K41Desm6iRyzE:-LGQpU0K41Desm6iRyzF:b-sdt9z7) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L708)):*
> ```Java
> 
>   @JsonProperty(PROP_IKE_PHASE1_KEYS)
>   public void setIkePhase1Keys(NavigableMap<String, IkePhase1Key> ikePhase1Keys) {
> ```

`@Nullable`, default to empty map.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 10 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQoOXv8vo01V2O9u1v:-LGQoOXv8vo01V2O9u1w:b88x7tx) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L10)):*
> ```
> extends ComparableStructure<String>
> ```

Do not extend ComparableStructure

Drop name from here, if it's stored in the map outside.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 23 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQrInaDfmxqp0ohHqh:-LGQrInaDfmxqp0ohHqi:b-1fexzd) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L23)):*
> ```Java
>   private IkeKeyType _keyType;
> 
>   private String _key;
> ```

rename -- this class is the key, and the identity is more than just this field.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 19 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mI7Y8iinAiUczs7X:-LGQmVba9R1xSUz5Q1v-:b-bsuyju) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L19)):*

Adding `@NonNull` here causes IDE to show a warning as in the setter we are declaring the argument as `@Nullable`(because of Jackson serialization, as pointed out in another comment)
</blockquote></details>

It should be possible for the member variable to be `@Nonnull` and the setter parameter to be `@Nullable` -- the setter just needs to substitute the default value before storing to the member variable. Let's chat in person if this is not clear.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 43 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8m0FTCrdQRyOkkyJ5:-LGQmiHGEwuGrrM-w_bg:b-rn5cmw) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L43)):*

done, but this prevents the declaration of `_ikePhase1Proposals` as a `@Nonnull` attribute.
</blockquote></details>

I don't understand why this prevents the declaration of the member variable as `@Nonnull`. Isn't that handled by the next line, which ensures that the value is never null?

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_crypto.g4, line 479 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQosHM8IQKgAwwl6DV:-LGQosHM8IQKgAwwl6DW:bvdzl3u) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_crypto.g4#L479)):*
> ```ANTLR
> cisprf_self_identity
> :
>    SELF_IDENTITY IP_ADDRESS
> ```

Looks like this is only partly implemented, and not tested.

* missing -- a reference to this clause.
* missing `NEWLINE` at the end?
* missing statement in ref file.
* missing extraction?

---

*[projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java, line 1833 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpLBlEmOPLTb4uQVC:-LGQpLBlEmOPLTb4uQVD:bn9xici) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L1833)):*
> ```
> Ip("255.255.255.255")
> ```

Ip.MAX

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 146 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqE4z38imCeb6ciIq:-LGQqE4z38imCeb6ciIr:b-isqdoo) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L146)):*
> ```Java
>     }
>     return primaryIpToIfaceNames;
>   }
> ```

@anothermattbrown surely this already exists in some form?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 149 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqHO20aZ4aEdB0pa6:-LGQqHO20aZ4aEdB0pa7:bjcyt6m) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L149)):*
> ```Java
> 
>   /**
>    * Resolves the interface names of the addresses used as local addresses of IsaKmpProfiles and
> ```

`{@link` any java class names.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/IsakmpProfile.java, line 20 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqV8L9qI8kFEE0guC:-LGQqV8L9qI8kFEE0guD:bau3o05) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/IsakmpProfile.java#L20)):*
> ```Java
> 
>   // TODO: change to IpWildcard
>   private Prefix _matchIdentity;
> ```

Let's resolve this issue now? What is the problem here?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 19 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqfpl9EEe-h04SES7:-LGQqfpl9EEe-h04SES8:bqhrahn) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L19)):*
> ```Java
> 
>   // TODO: deprecate once old Ike Phase 1 data-model is phased out
>   private Ip _remoteAddress;
> ```

Why not remove now?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 51 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqjhTDwul98JTS79f:-LGQqjhTDwul98JTS79g:b403eq4) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L51)):*
> ```Java
>   }
> 
>   public boolean match(Ip localAddress, Prefix matchIdentity) {
> ```

Document (throughout), all new public functions that aren't setters and getters.

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1183 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqqv44YlSoI0R3sSB:-LGQqqv44YlSoI0R3sSC:b-oei1tl) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1183)):*
> ```
> %s-key
> ```

This is not how Batfish-assigned names are represented. Talk to @arifogel to get guidance on this.

---

*[projects/batfish/src/main/java/org/batfish/representation/vyos/VyosConfiguration.java, line 162 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqxJ31yksqHLy7r27:-LGQqxJ31yksqHLy7r28:b-ug1p0a) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/vyos/VyosConfiguration.java#L162)):*
> ```
> %s-key"
> ```

Same comment.

---

*[projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java, line 1954 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQr3ae4JrKRn7ZpEz1:-LGQr3ae4JrKRn7ZpEz2:b1g5rwr) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L1954)):*
> ```
> hasKey
> ```

Looking at how it's used, this should probably not be called `hasKey` -- `has...Key(hasKey(` doesn't really make sense. If the class is named Key, it shouldn't have a field named Key (unless that field is really the only thing in the class.)

---

*[projects/batfish/src/test/java/org/batfish/representation/cisco/CiscoConfigurationTest.java, line 73 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQrnPn3VzfO7LwR9x0:-LGQrnPn3VzfO7LwR9x1:br1q6pi) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/test/java/org/batfish/representation/cisco/CiscoConfigurationTest.java#L73)):*
> ```Java
> 
>   @Test
>   public void testGetMatchingPskInvalidLocalIface() {
> ```

Move these to `CiscoConversionsTest` -- they're only tests of a utility class in that file, not of `CiscoConfigurationTest`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

haverma:
Review status: 7 of 32 files reviewed, 16 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 708 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpU0K41Desm6iRyzE:-LGRu74t7HNJpT5LHUva:b-3bzhw0) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L708)):*

`@Nullable`, default to empty map.
</blockquote></details>

made `@Nullable`, changed the  parameter from type `NavigableMap` to `Map`, as `NavigableMap` does not take in a `ImmutableSortedMap`

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 190 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGRut101SSThfrlLAmF:-LGRut101SSThfrlLAmG:bqkqs17) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L190)):*
> ```Java
>   private NavigableMap<String, IkeProposal> _ikeProposals;
> 
>   private @Nonnull SortedMap<String, IkePhase1Key> _ikePhase1keys;
> ```

Had to change from `@NavigableMap` to `@SortedMap` to assign `ImmutableSortedMap` to it

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 10 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQoOXv8vo01V2O9u1v:-LGRvHFo6X5e2w3xBYrI:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L10)):*

> ```
> extends ComparableStructure<String>
> ```

Do not extend ComparableStructure

Drop name from here, if it's stored in the map outside.
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 23 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQrInaDfmxqp0ohHqh:-LGRvMF794KWIg1oQZPA:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L23)):*

rename -- this class is the key, and the identity is more than just this field.
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 19 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8mI7Y8iinAiUczs7X:-LGRvORhCDBzvTnzjS7Z:b-2ue0o1) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L19)):*

It should be possible for the member variable to be `@Nonnull` and the setter parameter to be `@Nullable` -- the setter just needs to substitute the default value before storing to the member variable. Let's chat in person if this is not clear.
</blockquote></details>

yes, I think I got confused, IntelliJ does not complain if I handle the null value in the setter

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 43 at r2](https://reviewable.io/reviews/batfish/batfish/1788#-LG8m0FTCrdQRyOkkyJ5:-LGRvaMI7TF6_2pVZxpW:b-1max3g) ([raw file](https://github.com/batfish/batfish/blob/3876dd62fbfb5daa8243f01403ea4d592cc7040b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L43)):*

I don't understand why this prevents the declaration of the member variable as `@Nonnull`. Isn't that handled by the next line, which ensures that the value is never null?
</blockquote></details>

yes, that is true. I was wrong, may be IntelliJ takes some time to resolve the warning after the code is changed

---

*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_crypto.g4, line 479 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQosHM8IQKgAwwl6DV:-LGRvqxHEl5Zal5Txzjj:bj9q709) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_crypto.g4#L479)):*

Looks like this is only partly implemented, and not tested.

* missing -- a reference to this clause.
* missing `NEWLINE` at the end?
* missing statement in ref file.
* missing extraction?
</blockquote></details>

Yes, this was left partly implemented
add
- added a ref in the parent stanza
- added `NEWLINE`
- added a statement in cisco_crypto for self-identity 

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/IsakmpProfile.java, line 20 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqV8L9qI8kFEE0guC:-LGRsude9pxj0APJqoqd:b-en6qmk) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/IsakmpProfile.java#L20)):*

Let's resolve this issue now? What is the problem here?
</blockquote></details>

replaced with `IpWIldcard`, changed some existing code which was using this(`CiscoConfiguration.addIkePoliciesAndGateways`)

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 19 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqfpl9EEe-h04SES7:-LGRtE1uDKqS832wwMsc:b-hfpv0l) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L19)):*

Why not remove now?
</blockquote></details>

removed

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 51 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqjhTDwul98JTS79f:-LGRtIQo6V3xzCBPmJlD:bpwqneq) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L51)):*

Document (throughout), all new public functions that aren't setters and getters.
</blockquote></details>

added javadoc in all public functions affected(not in `package private` functions)

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 1183 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqqv44YlSoI0R3sSB:-LGRtoa268_gSA_RiS3y:b4rcb2i) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L1183)):*
> ```
> String.format("~IKE_PHASE1_KEY_%s~", ikePolicy.getName()),
> ```

Talked with @arifogel, now using the name in the format `String.format("~IKE_PHASE1_KEY_%s~", ikePolicy.getName()),`

---

*[projects/batfish/src/main/java/org/batfish/representation/vyos/VyosConfiguration.java, line 162 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqxJ31yksqHLy7r27:-LGRty46AZQsG1yuyHVm:b-n9tkhd) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/vyos/VyosConfiguration.java#L162)):*

> ```
> %s-key"
> ```

Same comment.
</blockquote></details>

changed

---

*[projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java, line 1954 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQr3ae4JrKRn7ZpEz1:-LGRu-R02Mxh_heOkJcu:b-7c8zsy) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L1954)):*

> ```
> hasKey
> ```

Looking at how it's used, this should probably not be called `hasKey` -- `has...Key(hasKey(` doesn't really make sense. If the class is named Key, it shouldn't have a field named Key (unless that field is really the only thing in the class.)
</blockquote></details>

renamed the key field to `keyValue`

---

*[projects/batfish/src/test/java/org/batfish/representation/cisco/CiscoConfigurationTest.java, line 73 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQrnPn3VzfO7LwR9x0:-LGRu4xd3E2DDJ29-x78:b-7mmp82) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/test/java/org/batfish/representation/cisco/CiscoConfigurationTest.java#L73)):*

Move these to `CiscoConversionsTest` -- they're only tests of a utility class in that file, not of `CiscoConfigurationTest`.
</blockquote></details>

moved

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

haverma:
Review status: 7 of 32 files reviewed, 13 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java, line 1833 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpLBlEmOPLTb4uQVC:-LGRxiPH9LTmU3chjX1s:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L1833)):*

> ```
> Ip("255.255.255.255")
> ```

Ip.MAX
</blockquote></details>

done

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 149 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqHO20aZ4aEdB0pa6:-LGRxkwh-0EKLomNG-ZJ:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L149)):*

`{@link` any java class names.
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

dhalperi:
Reviewed 26 of 26 files at r4.
Review status: all files reviewed, 10 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 708 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpU0K41Desm6iRyzE:-LGS0WuuE4aH_rxCijzm:b-3ixhx6) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L708)):*

made `@Nullable`, changed the  parameter from type `NavigableMap` to `Map`, as `NavigableMap` does not take in a `ImmutableSortedMap`
</blockquote></details>

(Seems wrong, like above.)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 190 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGRut101SSThfrlLAmF:-LGS08GeE0go8UWqk2jv:brvn4t6) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L190)):*
> ImmutableSortedMap

Something seems wrong here. ImmutableSortedMap implements NavigableMap.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 445 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS0Gd71LDxEM49dGMK:-LGS0Gd71LDxEM49dGML:b-hay66i) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L445)):*
> ```Java
>   @JsonProperty(PROP_IKE_PHASE1_KEYS)
>   @JsonPropertyDescription("Dictionary of all IKE phase1 keys for this node.")
>   public SortedMap<String, IkePhase1Key> getIkePhase1Keys() {
> ```

revert?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 51 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS0ajTEwAxGgwi4BsE:-LGS0ajU2Sv9LVsEpFnh:b4owler) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L51)):*
> ```Java
> 
>   @JsonPropertyDescription("Value of the key")
>   @JsonProperty(PROP_KEY)
> ```

rename the prop and the text of the prop?

rename to keyHash?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 3240 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS7y-LEr6f0ZfKV-Yr:-LGS7y-LEr6f0ZfKV-Ys:bxao8gb) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L3240)):*
> ```
> new ImmutableSortedMap.Builder<>(Comparator.naturalOrder())
> ```

`ImmutableSortedMap.naturalOrder()`.

In general, with Guava classes you're not supposed to use `new Builder` unless you're subclassing it. They provide static factory methods for this purpose, like the one above.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 146 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQqE4z38imCeb6ciIq:-LGSAIrqE3bn_7bsuJbf:b-n21c60) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L146)):*

@anothermattbrown surely this already exists in some form?
</blockquote></details>

Aha - it exists in common but not in vendor-specific repr.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 140 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSA9O33B-FnG3xE7Qg:-LGSA9O33B-FnG3xE7Qh:b-70rer7) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L140)):*
> ```Java
>     Map<Ip, String> primaryIpToIfaceNames = new HashMap<>();
>     for (Entry<String, Interface> interfaceNameToInterface : interfaces.entrySet()) {
>       InterfaceAddress primaryAddress = interfaceNameToInterface.getValue().getAddress();
> ```

make sure to include all addresses, not just primary, unless you know only primary can be used for this.

If you are unable to tell, prefer allowing all (or really, using gns3 to find out).

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 152 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSANUo86HQtCJpmKJy:-LGSANUo86HQtCJpmKJz:b6ra57j) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L152)):*
> ```
> static void resolveKeyringIsakmpProfileIfaceNames
> ```

not obvious this should just be one function -- it's just doing two almost-unrelated things in parallel, right?

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 176 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSAXZe6hm-h5WVcjqs:-LGSAXZe6hm-h5WVcjqt:b-9jq6ex) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L176)):*
> ```
> ObjectUtils.firstNonNull
> ```

`import static MoreObjects.firstNonNull` from guava and use that over the apache commons version.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 54 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS4gaM8HS1GmWYW5he:-LGS4gaN5M306XyHDEip:bd5rf4q) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L54)):*
> ```
> _remoteIdentity.supersetOf(matchIdentity)
> ```

this is just intersection, right?

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 2006 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSAloo1IIoJACO5nAs:-LGSAloo1IIoJACO5nAt:bwlziih) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2006)):*
> ```Java
> 
>     ImmutableSortedMap.Builder<String, IkePhase1Key> ikePhase1KeysBuilder =
>         new Builder<>(Comparator.naturalOrder());
> ```

same comment about constructor vs static factory method.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

dhalperi:
Reviewed 13 of 13 files at r5.
Review status: all files reviewed, 8 unresolved discussions (waiting on @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 52 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW80nQ7c5GS5crgPOz:-LGW80nQ7c5GS5crgPP-:b-8jfjdj) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L52)):*
> ```
> Value of the key
> ```
Update?

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW84W9-oKjKt6tMnPu:-LGW84W9-oKjKt6tMnPv:b-rrpbgj) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L73)):*
> ```Java
>         && matchIdentity != null
>         && _remoteIdentity.supersetOf(matchIdentity)
>         && (_localInterface.equals(UNSET_INTERFACE_NAME)
> ```

Since you `.` them here, you should

annotate `_localInterface` and `_remoteIdentity` as `@Nonnull` and make sure the nullability analysis confirms they're never null.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java, line 295 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW8MBq-A4UqPKR-owI:-LGW8MBq-A4UqPKR-owJ:bkj1xsq) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java#L295)):*
> ```
> unset_interface_name
> ```

These are not clear enough, in my opinion. If I'm looking at the data model, and I see that a keyring has `localInterface` = `unset_interface_name`, what does that mean?

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/IkePhase1KeyMatchersImpl.java, line 10 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW8VO22T-16PyPJk6a:-LGW8VO22T-16PyPJk6b:b-77yvz3) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/IkePhase1KeyMatchersImpl.java#L10)):*
> ```
> HasKeyValue 
> ```

Update?

And the caller?

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

haverma:
1. Addressed comments
2. Store numbers as name of `IsakmpPolicy`
3. Change `IkePhase1Proposals` in  `IkePhase1Key` to list of proposal names instead of whole object
4. Change `remoteIdentity` in `IkePhase1Key` and `IkePhase1Policy` to `IpSpace`
5. Addressed latest comments

---

Review status: 12 of 34 files reviewed, 6 unresolved discussions (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 708 at r3](https://reviewable.io/reviews/batfish/batfish/1788#-LGQpU0K41Desm6iRyzE:-LGTXna3-qr7GLPHRlvR:bvth56w) ([raw file](https://github.com/batfish/batfish/blob/892b949a3c9f3e90ce294bd260244e316e87c06d/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L708)):*

(Seems wrong, like above.)
</blockquote></details>

reverted

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 190 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGRut101SSThfrlLAmF:-LGTXDqs5DQxKzQ-z7G8:b-yzy6bd) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L190)):*

> ImmutableSortedMap

Something seems wrong here. ImmutableSortedMap implements NavigableMap.
</blockquote></details>

I had created a `ImmutableMap.Builder` instead of `ImmutableSortedMap.Builder` while trying to use the setter for this attribute, and that was causing the compile warning, which I misinterpreted

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java, line 445 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS0Gd71LDxEM49dGMK:-LGTXjPhD819-5pBtLEu:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L445)):*

revert?
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 51 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS0ajTEwAxGgwi4BsE:-LGTXr9f5baLOj6BXvIa:b-hff6wv) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L51)):*

rename the prop and the text of the prop?

rename to keyHash?
</blockquote></details>

renamed

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 52 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW80nQ7c5GS5crgPOz:-LGX1INW-4HJ69xzD8Rw:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L52)):*

> ```
> Value of the key
> ```
Update?
</blockquote></details>

done

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW84W9-oKjKt6tMnPu:-LGX1JNa6632dh_yYmIE:b-dio4bm) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L73)):*

Since you `.` them here, you should

annotate `_localInterface` and `_remoteIdentity` as `@Nonnull` and make sure the nullability analysis confirms they're never null.
</blockquote></details>

annotated and init with default values in constructor

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java, line 295 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW8MBq-A4UqPKR-owI:-LGX1UWJ73XsIMI1h8rH:bciw8up) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java#L295)):*

> ```
> unset_interface_name
> ```

These are not clear enough, in my opinion. If I'm looking at the data model, and I see that a keyring has `localInterface` = `unset_interface_name`, what does that mean?
</blockquote></details>

trying something more meaningful

---

*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/IkePhase1KeyMatchersImpl.java, line 10 at r5](https://reviewable.io/reviews/batfish/batfish/1788#-LGW8VO22T-16PyPJk6a:-LGX1ap_9L139qiqtKIT:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/daf3ed4647f2054834eb722c0f6c08df7a61d2a4/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/IkePhase1KeyMatchersImpl.java#L10)):*

> ```
> HasKeyValue 
> ```

Update?

And the caller?
</blockquote></details>

done

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 3240 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS7y-LEr6f0ZfKV-Yr:-LGTXuMR7qvWklzu18Au:b-fmmqe2) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L3240)):*

> ```
> new ImmutableSortedMap.Builder<>(Comparator.naturalOrder())
> ```

`ImmutableSortedMap.naturalOrder()`.

In general, with Guava classes you're not supposed to use `new Builder` unless you're subclassing it. They provide static factory methods for this purpose, like the one above.
</blockquote></details>

replaced with factory method

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 140 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSA9O33B-FnG3xE7Qg:-LGTXzj05gT6A8R6rU8i:b1i9piu) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L140)):*

make sure to include all addresses, not just primary, unless you know only primary can be used for this.

If you are unable to tell, prefer allowing all (or really, using gns3 to find out).
</blockquote></details>

using `allAddresses` now

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 152 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSANUo86HQtCJpmKJy:-LGTY6N4EatOudzxIovt:b4bwt6m) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L152)):*

> ```
> static void resolveKeyringIsakmpProfileIfaceNames
> ```

not obvious this should just be one function -- it's just doing two almost-unrelated things in parallel, right?
</blockquote></details>

splitted

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 176 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSAXZe6hm-h5WVcjqs:-LGTYCRVDGEHvOyBJpZe:bchrusv) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L176)):*

> ```
> ObjectUtils.firstNonNull
> ```

`import static MoreObjects.firstNonNull` from guava and use that over the apache commons version.
</blockquote></details>

using correct one

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java, line 54 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGS4gaM8HS1GmWYW5he:-LGX1eKtCwH1O9wFESs0:b-f4suz4) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/cisco/Keyring.java#L54)):*

> ```
> _remoteIdentity.supersetOf(matchIdentity)
> ```

this is just intersection, right?
</blockquote></details>

Updated with intersection in match method of `IkePhase1Key`, this match method is only being used for the old data-model conversion, so will get rid of it eventually

---

*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 2006 at r4](https://reviewable.io/reviews/batfish/batfish/1788#-LGSAloo1IIoJACO5nAs:-LGTYIAQ9gMx_VfVxT_I:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/6be361060f0c524d0096f9afca0291163d3633fe/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2006)):*

same comment about constructor vs static factory method.
</blockquote></details>

done

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788#-:-LGX1zk20gsLviZdMkwC:b-y88jtq)*

dhalperi:
Reviewed 22 of 22 files at r6.
Review status: all files reviewed, 4 unresolved discussions (waiting on @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 94 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX1nFTE-BnyVw0VhJk:-LGX1nFTE-BnyVw0VhJl:bhfiilo) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L94)):*
> ```Java
> 
>   @JsonProperty(PROP_REMOTE_IDENTITY)
>   public void setRemoteIdentity(IpSpace remoteIdentity) {
> ```

Remember that every setter annotated with `@JsonProperty` needs to be annotated `@Nullable`. Please fix and handle.

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 87 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX1zoS5EUXAZDsLyAG:-LGX1zoS5EUXAZDsLyAH:b-6jpun3) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L87)):*
> ```Java
> 
>   @JsonProperty(PROP_REMOTE_IDENTITY)
>   public void setRemoteIdentity(IpSpace remoteIdentity) {
> ```

same comment re setters and nullable.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2176 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX2CfB5J5boYul33oT:-LGX2CfB5J5boYul33oU:b-oyvh2d) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2176)):*
> ```
>     /* Pad priority number with zeros, so string sorting sorts in numerical order too */
>     IkeProposal ikeProposal = new IkeProposal(String.format("%03d", isakmpPolicy.getName()));
> ```

Why pad here? We do not need them to sort in numerical order in the vendor-independent model.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

haverma:
Review status: 30 of 34 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @haverma)

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java, line 94 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX1nFTE-BnyVw0VhJk:-LGXCcKw8twQYjA3eeFj:b-z3pc30) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Key.java#L94)):*

Remember that every setter annotated with `@JsonProperty` needs to be annotated `@Nullable`. Please fix and handle.
</blockquote></details>

done, and handled for remoteIdentity and localInterface

---

*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java, line 87 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX1zoS5EUXAZDsLyAG:-LGXCmaT2IHfL7o_-VCD:b3ax3zf) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/IkePhase1Policy.java#L87)):*

same comment re setters and nullable.
</blockquote></details>

done for all setters, only `ikePhase1Proposals` are `@NonNull` here

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2176 at r6](https://reviewable.io/reviews/batfish/batfish/1788#-LGX2CfB5J5boYul33oT:-LGXCvmuCdGFgilzoEDp:b-hfpv0l) ([raw file](https://github.com/batfish/batfish/blob/126d3cc25ea01f5ac55af64db046577f6be40687/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2176)):*

> ```
>     /* Pad priority number with zeros, so string sorting sorts in numerical order too */
>     IkeProposal ikeProposal = new IkeProposal(String.format("%03d", isakmpPolicy.getName()));
> ```

Why pad here? We do not need them to sort in numerical order in the vendor-independent model.
</blockquote></details>

removed

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

dhalperi:
:lgtm:

---

Reviewed 4 of 4 files at r7.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788#-:-LGXDhxXE6tDwq0C4Q1h:bnfp4nl)*

dhalperi:
Reviewed 2 of 2 files at r8.

---



*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1788)*

PR #1802, opened 2018-07-03
Skipping shutdown or interfaces with no IPs for crypto maps

haverma:


dhalperi:
Review status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @haverma and @dhalperi)

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2040 at r1](https://reviewable.io/reviews/batfish/batfish/1802#-LGSzWdR9HrjeEkP_yFV:-LGSzWdR9HrjeEkP_yFW:boh90al) ([raw file](https://github.com/batfish/batfish/blob/ac3c12f35d980dbe31c36d3c35b3daecb38c2c59/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2040)):*
> ```Java
>                 iface.getName(), cryptoMapName));
>         continue;
>       }
> ```

Since this is in conversion, not in activation, I think we should keep shutdown/inactive interfaces. Removing inactive interfaces happens later. I'd drop this clause and just leave the part below.

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2048 at r1](https://reviewable.io/reviews/batfish/batfish/1802#-LGSzm_U8Q6nLjo8RGzB:-LGSzm_V0JM4676Smc-0:bcgrc8r) ([raw file](https://github.com/batfish/batfish/blob/ac3c12f35d980dbe31c36d3c35b3daecb38c2c59/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2048)):*
> ```Java
>         continue;
>       }
>       Ip bindingInterfaceIp = iface.getAddress().getIp();
> ```

Why was this not caught by null pointer analysis?

Consider (likely in a separate PR, as it probably is somewhat wide-ranging) making `getAddress()` nullable so that we can catch NPEs like this in the future.

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1802)*

haverma:
Review status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @haverma and @dhalperi)

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2040 at r1](https://reviewable.io/reviews/batfish/batfish/1802#-LGSzWdR9HrjeEkP_yFV:-LGT2UyI3-NR0M-o9S_R:b-33m40s) ([raw file](https://github.com/batfish/batfish/blob/ac3c12f35d980dbe31c36d3c35b3daecb38c2c59/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2040)):*

Since this is in conversion, not in activation, I think we should keep shutdown/inactive interfaces. Removing inactive interfaces happens later. I'd drop this clause and just leave the part below.
</blockquote></details>

just excluding interfaces with no IP address now

---

*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 2048 at r1](https://reviewable.io/reviews/batfish/batfish/1802#-LGSzm_U8Q6nLjo8RGzB:-LGT2ev52fL8P-u2Bjs5:bser1j4) ([raw file](https://github.com/batfish/batfish/blob/ac3c12f35d980dbe31c36d3c35b3daecb38c2c59/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L2048)):*

Why was this not caught by null pointer analysis?

Consider (likely in a separate PR, as it probably is somewhat wide-ranging) making `getAddress()` nullable so that we can catch NPEs like this in the future.
</blockquote></details>

Good point , will do that in a later PR

---


*Comments from [Reviewable](https://reviewable.io/reviews/batfish/batfish/1802)*

PR #1817, opened 2018-07-07
Specialize nod jobs 1 query per line

corinaminer:
Putting this PR up for visibility; still pending tests. This change reduces the analysis time for my example long ACL from 18 minutes to 5 minutes. It also reduced analysis time for a network with many short ACLs by about 25% (from 30-35 seconds to 20-25 seconds).

Two main things have been changed, as described by the third "low-hanging fruit" idea [here](https://docs.google.com/document/d/1CsLM-K9ZJYH1n1gAweSO6bDl4TmIKUdadDnFLnspDLM/edit#).

1. `AclReachabilityQuerySynthesizer` is now written to produce one reachability program per one ACL line. 
- each instance of `AclReachabilityQuerySynthesizer` represents only one ACL line, and will generate the program for that line
- previously, each instance represented an entire ACL, and generated a reachability program containing one query for each line of the ACL
- At a high level, this change makes finding unreachable lines faster because each reachability program only depends on the lines before the line it's analyzing, whereas the old setup's reachability program was dependent on all lines in the ACL.

2. The ACLs themselves are specialized to the line under analysis.
- For each line of the ACL, a specialized version of the ACL is produced where any lines that can't overlap with the current line are replaced with `FalseExpr` lines
    - those lines are not removed; then it would be difficult to keep track of line numbers
- The specialized ACL only contains the lines up to and including the current line, since there's no point specializing lines after that
- Any dependent ACLs are also specialized.


corinaminer:
Closing in favor of reimplementing specialization with BDDs. Decided not to pursue this route further because it was getting progressively more complex and compounding opportunities for bugs 🐛 🐞 🐝 🦋 🐜

PR #1845, opened 2018-07-11
Sanitize references in And, Not, and Or expressions

corinaminer:
Fixes an issue where references to ACLs, IpSpaces, and interfaces could go unnoticed during ACL sanitization if they were hidden inside And, Or, or Not match expressions.


ratulm:
The code works for me, but letting others review the code.

PR #1855, opened 2018-07-11
Add delay parameter to interfaces

mkremerbbn:
For EIGRP, routers have a delay parameter which is roughly inversely proportional to bandwidth. Only used on Cisco routers as far as I know. This is step 0 in the EIGRP implementation.


mkremerbbn:
Thanks @progwriter, I will integrate the delay parameter into `EigrpInterfaceSettings` in a subsequent PR.

PR #1863, opened 2018-07-12
Batfish: move z3 to common

dhalperi:
In order to allow questions and other plugins to construct z3 queries.

Notes:

- Interestingly, there were some other things that look like they should have been in common all along, but we did not notice:  `org.batfish.datamodel.visitors.IpSpace*Wildcard`
- Was not yet able to move BDD stuff -- seemed fairly tied into Batfish right now.
- We need to decide what the right question/batfish API is. Right now, BatfishJob is not in common, so questions can't really build up a batch of things to submit.
- I left some things in org.batfish.z3 behind that have to do with the above 2 concepts.


dhalperi:
Needs some revisiting to resolve our outstanding discussions.

PR #1891, opened 2018-07-17
BatfishStorage: no empty batches

dhalperi:
Better user feedback if batches can be completed.


dhalperi:
Users need to be able to distinguish between pending and completed work.

PR #1892, opened 2018-07-17
Juniper: parse and extract static route no-install

progwriter:
Was generating a parse warning.
Parsed and extracted according to [this](https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/install-edit-routing-options.html)


progwriter:
On hold until better tests are written.

PR #1894, opened 2018-07-17
Fix bazel build

progwriter:
Add/remove dependencies as needed after changes to question tests, bdd work, and new topology tests.

*Note*: Not sure why topology tests use empty batches (perhaps as a "log statement"?) so had to suppress a couple of warnings.


dhalperi:
#1891 and #1895 fix some of the issues here, but yes! Thanks!

PR #1897, opened 2018-07-17
Cisco: support for IOS-XR community-set asterisk

dhalperi:
Asterisk means any valid half-element.

Q: Would you prefer a new wildcard type instead of reusing subrange?


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java, line 8421 at r1](https://reviewable.io/reviews/batfish/batfish/1897#-LHe8Ksr95_bD8em8AQG:-LHe9-yrAeDB3zOX3qqW:b-32nyt8) ([raw file](https://github.com/batfish/batfish/blob/14f13b0fc5c441bcca62c17d60d15bba17a0aade/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L8421)):*

I thought arbitrary regexes were allowed here. If so, this is a slight improvement that will have to be redone later. Otherwise, this is fine. Either way not a blocker.
</blockquote></details>

As I read the docs, regexes are only allowed when the user flags them as such: `ios-regex` or `dfa-regex`: https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-1/routing/command/reference/b_routing_cr41crs/b_routing_cr41crs_chapter_01000.html#wp4121051042

PR #1906, opened 2018-07-18
javadoc: quiet

dhalperi:
Don't spit out the names of all the files it generates. Still shows errors and warnings.


dhalperi:
(after #1905)

PR #1917, opened 2018-07-20
Basic EIGRP dataplane

mkremerbbn:
Adds EIGRP route and neighbor computations for both classic and mixed EIGRP modes. No route redistribution, passive interfaces, etc.

Attempted support for mixing classic and named EIGRP processes in a single network, but further testing in simulation/hardware is necessary to verify behavior.


mkremerbbn:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/eigrp/EigrpMetric.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/1917#-LI4ZlDX3v8i5NhbfM3p:-LIE8O5Z2SbvcKWP_LH0:brdutp1) ([raw file](https://github.com/batfish/batfish/blob/719087e6518963d682bb76c52040ffff05908618/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/eigrp/EigrpMetric.java#L47)):*
<details><summary><i>Previously, mkremerbbn (Michael Kremer) wrote…</i></summary><blockquote>

Will file an issue.
</blockquote></details>

After further review of the documentation, the way defaults are handled in this revision is correct. Interface delay is derived from the default bandwidth, not the configured bandwidth. Configured bandwidths and delays take priority over calculations, which is handled.

PR #1921, opened 2018-07-20
Add new `FilterTable` based question: `interfaceMtu`

sfraint:
* Add `FilterTable` to `batfish`
* Add a new question: `interfaceMtu` (native-tableanswer question based on `FilterTable` and `InterfaceProperties`)


ratulm:
One tip for optional optimization of questions like this interfaceMtu one: the interfaceProperties question has a parameter called propertSpec that you can use to extract only certain properties. So, instead of extracting all properties and then projecting two columns, you can simply extract the needed property ("mtu") which will yield the same result as the projection.

PR #1932, opened 2018-07-23
Some beginners work removing ComparableStructure from datamodel

dhalperi:
Most were easy.

Did not add equals, hashCode, or toString to any class, except as needed to make tests pass or compiler warnings succeed.

These two were affected by the latter (and have commits with good messages explaining why).

* Configuration
* AsPathAccessList

Thoughts?


dhalperi:
Clearly I should move configuration to another PR. Hold tight.

PR #1948, opened 2018-07-25
AddressGroup and ReferenceAddressGroupIpSpaceSpecifier fixes

dhalperi:
This started small (fixing a compile-issue identified by bazel) and ended up big. There were *lots* of mistakes in the code here.

Please review in detail (I created separate commits for each bug) to see what I fixed.


dhalperi:
NPE - IntelliJ's null pointer analysis caught it.

The `stream().map(` one was interesting catch by bazel -- since streams are lazy, I think that code may well have never been executing.

PR #1949, opened 2018-07-25
adding README to use the question templates

haverma:


dhalperi:
I'd streamline the doc.

1. Make the "common case" easy and clear, the default assumption in the doc. That is, assume that the  coordinator has the library of questions, and clients should just load questions from the server. Explain the entire article from that perspective first.

2. Then, at the bottom, add a section for clients bringing their own questions.

PR #1951, opened 2018-07-25
Cisco: improve ip sla parsing

dhalperi:


dhalperi:
(Any reviewer will do. Key is to understand what the change is and how we test it ;).

PR #1981, opened 2018-07-30
Introduce filter groups to reference books

ratulm:


dhalperi:
I'll take a look after @sfraint reviews.

PR #2002, opened 2018-08-01
Routing properties template

ratulm:
After #2001


ratulm:
This PR is unblocked now.

PR #2010, opened 2018-08-02
WorkMgr: better error feedback

dhalperi:


dhalperi:
If it's the case that WorkItem must always have both, I went ahead and moved the logic to the WorkItem deserializer. Please take another look?

PR #2012, opened 2018-08-02
Adding the outliers for node properties with new design.

vasu018:
Code changes to accommodate the new design of generalizing the outliers calculation. 
Currently only node properties are handled in this PR.


vasu018:
I think there are dependencies on the node and interface properties and hence will move this code as part of outliers logic. Hence, pulling back the current review for now and will update the code accordingly and reopen it.

PR #2039, opened 2018-08-08
added differentialOnly and nonDifferentialOnly fields to Question

erikljungman:
added these fields so that a policy can contain certain questions that will only be run differentially and certain questions that will only be run non-differentially


dhalperi:
This seems like a big API regression. Let's discuss. @ratulm

arifogel:
Closing as dicussed offline.

PR #2044, opened 2018-08-08
Interface: prevent and verify that vrf is never null

dhalperi:
- Add various annotations to setters and getters, plus requireNonNull
- Make builder default to default VRF
- Get rid of two-object-accounting: never store _vrf instead always deriving
  it from vrfName


dhalperi:
Please criticize, I suspect it's ripe for it.

dhalperi:
This strategy does not work.

PR #2051, opened 2018-08-10
bug fix in Row; easy constructor in Table

ratulm:


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java, line 159 at r1](https://reviewable.io/reviews/batfish/batfish/2051#-LJZi0MfD08YUpVRM5WK:-LJZlxfXAHOdXxpOn-CA:b-bk5pm3) ([raw file](https://github.com/batfish/batfish/blob/caed461cefcaa9b41f337fc7aa8c595713cdd0a3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/Row.java#L159)):*

I am not sure what the standard protection mechanism is. Here is how I was tripping up:

```
RowBuilder rowB = Row.builder().put("k", "v1")
Row r1 = rowB.build();
rowB.put("k", "v2")  // replace the value of k
Row r2 = rowB.build();
```

Now, k was mapped to v2 in both r1 and r2. 
</blockquote></details>

ahh, interesting

I might expect that actually the "most standard" behavior is for the second `put` call to throw -- e.g., see Guava's [ImmutableMap.builder](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html#put(K,%20V)).

But that might not be what you want...

PR #2064, opened 2018-08-13
Support additional types of characters in the metada.

vasu018:
Support additional types of characters in the metada required for space and characters such as ~ which is required to support configurations for named structures.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java
@@ -20,8 +20,12 @@ public final class ColumnMetadata {
   private static final String PROP_NAME = "name";
   private static final String PROP_SCHEMA = "schema";
 
-  // must start with letters or underscore, and only contain {-,\w} ( i.e., [-a-zA-Z_0-9])
+  /**
+   * Must start with alpha numerical letters, underscore or tilde. In addition, consecutive
+   * characters could include [-.:~@].
+   */
   public static final String COLUMN_NAME_PATTERN = "[a-zA-Z0-9_~]+[-\\w\\.:~@]*";
+
   private static final Pattern _COLUMN_NAME_PATTERN = Pattern.compile(COLUMN_NAME_PATTERN);
 
   @Nonnull private String _description;
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java
@@ -21,8 +21,8 @@ public final class ColumnMetadata {
   private static final String PROP_SCHEMA = "schema";
 
   /**
-   * Must start with alpha numerical letters, underscore or tilde. In addition, consecutive
-   * characters could include [-.:~@].
+   * Must start with alpha numerical letters, underscore or tilde. In addition, following characters
+   * could include [-.:~@].
    */
   public static final String COLUMN_NAME_PATTERN = "[a-zA-Z0-9_~]+[-\\w\\.:~@]*";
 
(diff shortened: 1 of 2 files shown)


dhalperi:
Do we have tests that the right characters are allowed in column names, and the right strings are rejected? Please update or add.

vasu018:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java, line 23 at r1](https://reviewable.io/reviews/batfish/batfish/2064#-LJoaqVPC-U3K3ARi4It:-LK2O_wYA_x6dZjoVcuU:betfkwr) ([raw file](https://github.com/batfish/batfish/blob/ed8fde8bcf46ad5bd3fc7a1b725ced712f394188/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java#L23)):*

This comment needs updating. Also, since this is getting complex, please add tests for example names that you are trying to support / reject. 

1/ Have you seen names that begin with a number? 
2/ Does '.' not matching anything? 
3/ Have you seen spaces in names (since you specifically added \\s?
</blockquote></details>

Currently I updated the column name to support following characters. 
The list of characters that are allowed as first characters are  -zA-Z0-9_~

In addition, -  . : ~ @ (five special characters) are allowed in the consecutive set of characters. 
Reason being:
- in general characters . : and - are mostly supported by most of the vendors.
- The @ for fqdn and domain names.
I supported space as i used it for column names. As we do not wanted to support space in the column names and hence removed it.

Please let me know if there are additional characters that are required to be added or removed from the existing list. 

Added some test cases, please see if they serve the purpose?

vasu018:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java, line 23 at r1](https://reviewable.io/reviews/batfish/batfish/2064#-LJoaqVPC-U3K3ARi4It:-LK2kjIV8RuK2qD89D59:b-u12jpx) ([raw file](https://github.com/batfish/batfish/blob/ed8fde8bcf46ad5bd3fc7a1b725ced712f394188/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java#L23)):*

Thanks, Vasu. This all makes sense. Two quick questions:

Regarding your five special characters:
  - what does supported by most vendors mean -- you mean in route map (or other structure names)?
  - under what circumstances do you need to use @ as a column name? 

</blockquote></details>

Ratul, I meant considering all other name structures even. 

Also, some vendor devices for sure accepts @ and space in the names such as zone names, network names and so on. 

Technically we should support other characters even unless until if they have special meaning with batfish or used as separator. Alternatively, we can add it as and when we see the need. For now i think i can stick on with following set of characters.

vasu018:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java, line 23 at r1](https://reviewable.io/reviews/batfish/batfish/2064#-LJoaqVPC-U3K3ARi4It:-LK2xRHPCTKqBNQE_aNM:btojbfq) ([raw file](https://github.com/batfish/batfish/blob/ed8fde8bcf46ad5bd3fc7a1b725ced712f394188/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java#L23)):*

the thing with special characters is that is they will interfere when users want to filter based on column names. but this PR is good for now, and down the line, we may need to do proper escaping in places. 
</blockquote></details>

Great.

vasu018:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java, line 23 at r2](https://reviewable.io/reviews/batfish/batfish/2064#-LK2t_jC1uPxwIrYKW6-:-LK4r45E-ogstXUtlCzz:bbak4mx) ([raw file](https://github.com/batfish/batfish/blob/b1c8e67980f1227387976b69e6147b6438a91151/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/table/ColumnMetadata.java#L23)):*

please update the comment before merging. 
</blockquote></details>

Not sure what update is required in this comment. 

Will make changes the below comment.

PR #2076, opened 2018-08-15
Option to select BGP path ranking criteria

nickgian:
This adds the option to select the order of the rules that are used by BGP to select a best path (and potentially skip some of them completely).


dhalperi:
Sorry @nickgian , looks like we missed this for a day or two.

@anothermattbrown, can you take a look?

nickgian:
I believe so :)

dhalperi:
Great. I kicked off another Travis build and it was green, so here we go!

PR #2086, opened 2018-08-15
Fix question desc2

saparikh:
Changed descriptions for questions in experimental folder.


saparikh:
___
*[questions/experimental/aclReachability.json, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/2086#-LK2WuhX3q4haegJF2GL:-LK3-aqw59VJp3cwRVja:b-fm0eue) ([raw file](https://github.com/batfish/batfish/blob/dad6c707f9c3aeea2e7ba707f289018cb65afbe5/questions/experimental/aclReachability.json#L21)):*

Why change this from `include` to `evaluate` but in other questions use `include`?
</blockquote></details>

I am trying to use `include` for questions that are just restricting information we extract, like propertySpec for nodeProperties. `Evaluate` for parameters that are restricting the thing we are evaluating... so in `aclReachability` we are evaluating ACLs vs extracting information from the data-model.

saparikh:
___
*[questions/experimental/aclReachability.json, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/2086#-LK2WuhX3q4haegJF2GL:-LK30f9B7sNvTGSu-obJ:b-cvk5nv) ([raw file](https://github.com/batfish/batfish/blob/dad6c707f9c3aeea2e7ba707f289018cb65afbe5/questions/experimental/aclReachability.json#L21)):*
<details><summary><i>Previously, saparikh (Samir Parikh) wrote…</i></summary><blockquote>

I am trying to use `include` for questions that are just restricting information we extract, like propertySpec for nodeProperties. `Evaluate` for parameters that are restricting the thing we are evaluating... so in `aclReachability` we are evaluating ACLs vs extracting information from the data-model. 
</blockquote></details>

but I will go back and see if there are other places where using `evaluate` makes sense. If there aren't enough, then will just stick with `include`

saparikh:
___
*[questions/experimental/filterTable.json, line 9 at r1](https://reviewable.io/reviews/batfish/batfish/2086#-LK2XtTfBFXVoPqITSQa:-LK315EwEKcPswUC-dSh:boextye) ([raw file](https://github.com/batfish/batfish/blob/dad6c707f9c3aeea2e7ba707f289018cb65afbe5/questions/experimental/filterTable.json#L9)):*

This longDescription does not seem self-contained like the `aclReachability` longDescription:
> ```
> Return ACLs with unreachable lines, along with all of the unreachable lines within the ACL.
> ```

I'd say we should stick with either self-contained long description, or supplemental description,  depending on how this is used by batfish clients.  Same comments applies to the other added longDescriptions.
</blockquote></details>

Good catch. Made change to make it self-contained

PR #2090, opened 2018-08-16
Parsers: make unrecognized lines show up as parser warnings

dhalperi:
Corollary: remove "urf", the "unrecognized as red flag" option. This
warning happens unconditionally.

- Add UnrecognizedLineToken to store structural information about
  line-based parser recovery and support it in the error strategy.
- Modify unrecognized line reporting in all 3 parsers (Cisco-like,
  Juniper, PAN) to write structured warnings if given that token.
- Delete URF and all testing of it.


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/grammar/UnrecognizedLineToken.java, line 28 at r1](https://reviewable.io/reviews/batfish/batfish/2090#-LK4GlR50uU85E3cqQrv:-LK4JzG24LwEjAsTH5OH:b-1f6zpf) ([raw file](https://github.com/batfish/batfish/blob/41ff6e037d855dbcf3a0f65aa697d2112e4080ca/projects/batfish-common-protocol/src/main/java/org/batfish/grammar/UnrecognizedLineToken.java#L28)):*

`0`s for char position, start, and stop actually would be a single character at the start of the input stream. Since this is not meant to refer to that, could you change at least some of these return values?
</blockquote></details>

Sure -- the old code in BatfishANTLRErrorStrategy was hard-coding `-1`, `-1`, `0`. -- updated to match

PR #2092, opened 2018-08-16
Table answer for undefined references

corinaminer:


dhalperi:
___
*[tests/basic/undefinedReferences.ref, line 1 at r1](https://reviewable.io/reviews/batfish/batfish/2092#-LK3JrUC9vSVt6Elo8k1:-LK3ZFXR-RSM2aAP4PFu:bh1q1w6) ([raw file](https://github.com/batfish/batfish/blob/a581c036f9de669b3f0e0c87bb9f16f37f5c5995/tests/basic/undefinedReferences.ref#L1)):*

those tests do seem the same. how are we deciding what gets tested in parsing and what gets tested in basic?
</blockquote></details>

It's about what's a better test. In this case, we're testing the reference tracking logic and the parsing tests exercise much more of the grammar.

We should not be testing your new question with a ref test - it should be easy to unit test. (See the initialization questions for how to supply a trivia PVCAE with the needed content.)

PR #2093, opened 2018-08-16
DifferentialBDDSourceManager

anothermattbrown:
Adapts BDDSourceManager for use in a differential context.


anothermattbrown:
Closing in favor of something simpler: only consider interfaces active in both snapshots.

PR #2098, opened 2018-08-17
Table answer for unused structures

corinaminer:


saparikh:
___
*[questions/stable/unusedStructures.json, line 6 at r3](https://reviewable.io/reviews/batfish/batfish/2098#-LK8mAW--1WzakfvOL7I:-LK8mAW--1WzakfvOL7J:b-uv28hw) ([raw file](https://github.com/batfish/batfish/blob/444d2f609c16d467b6f44de458cd8f834bd5d0b8/questions/stable/unusedStructures.json#L6)):*
> ```JSON
>     "nodeRegex": "${nodeRegex}",
>     "instance": {
>         "description": "Outputs cases where structures (e.g., ACL, routemaps) are defined but not used.",
> ```

`Return nodes with structures such as ACLs, routemaps, etc... are defined but not used.`

I changed all of the other descriptions to follow this format. Instead of output, we should use return.

dhalperi:
.. that this UI issue is a race condition in the UI code...

PR #2132, opened 2018-08-22
Rename ACL actions to permit and deny

progwriter:
Conversion to "deny" makes it more consistent with flow dispositions and configuration languages. Also, opposite of deny is permit in ACL language.

Thoughts? Will be a breaking change.

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/allinone/AclReachability2Test.java
+++ b/projects/allinone/src/test/java/org/batfish/allinone/AclReachability2Test.java
@@ -262,7 +262,7 @@ public class AclReachability2Test {
                 .put(AclLines2Rows.COL_DIFF_ACTION, false)
                 .put(
                     AclLines2Rows.COL_MESSAGE,
-                    "ACLs { c1: acl } contain an unreachable line:\n  [index 0] IpAccessListLine{action=ACCEPT,"
+                    "ACLs { c1: acl } contain an unreachable line:\n  [index 0] IpAccessListLine{action=PERMIT,"
                         + " matchCondition=PermittedByAcl{aclName=???, defaultAccept=false}}"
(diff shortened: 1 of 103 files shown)


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/LineAction.java, line 7 at r1](https://reviewable.io/reviews/batfish/batfish/2132#-LKYuMEK04mi7khyfFgN:-LKYuMEK04mi7khyfFgO:b-smq6f1) ([raw file](https://github.com/batfish/batfish/blob/a89ad81c965b8e767d1a3cba75eca1945c376ed7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/LineAction.java#L7)):*
> ```Java
>   PERMIT,
>   DENY
> }
> ```

The non-breaking version would be:

```
@JsonCreator
public static LineAction forValue(String value) { ... }
```

and handle PERMIT/DENY.

I would do that at least for the duration of merging and fixing all users.

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/LineAction.java, line 7 at r1](https://reviewable.io/reviews/batfish/batfish/2132#-LKYuMEK04mi7khyfFgN:-LKYv8nF1Sh0p2i1uM6i:b-hf5rlt) ([raw file](https://github.com/batfish/batfish/blob/a89ad81c965b8e767d1a3cba75eca1945c376ed7/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/LineAction.java#L7)):*

The non-breaking version would be:

```
@JsonCreator
public static LineAction forValue(String value) { ... }
```

and handle PERMIT/DENY.

I would do that at least for the duration of merging and fixing all users.
</blockquote></details>

sorry -- and handle ACCEPT/REJECT.

PR #2138, opened 2018-08-23
Separate answer metadata from answers

arifogel:
Answer metadata for table answers is now stored separately in `answer-metadata.json` file along `answer.json`.

Breaking change:
- no longer able to request aggregations via coordinator `getAnalysisAnswersMetrics`
- instead, pass '-aggregations ...' to batfish work item args when analysis is run


dhalperi:
> instead, pass '-aggregations ...' to batfish work item args when analysis is run

Batfish work item args are not (well, SHOULD not be) a user facing concept. Why is this type of change different vs another parameter on execute or just always doing it?

PR #2176, opened 2018-08-28
Interface: delete inboundFilter and add warning in Juniper

dhalperi:
@sfraint is out of town this week.

@ari: wasn't obvious to me exactly which bug to file (and update the TODO) -- do you have an opinion?


dhalperi:
The conflict was with the other PR that removed the same field from InterfaceProperties question as a hack while this PR was in flight. SO that was intended.

The problem with this PR is that we don't yet merge the inbound filter into the zone config. Maybe @sfraint has ideas there, but I don't have cycles to push that through. So for now, this is stagnating. Will close it if it's not useful.

PR #2188, opened 2018-08-29
add network and protocols filter to Routes; rename earlier protocol filter to rib

ratulm:
This is a breaking change.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/2188#-LL56sCY3gl9QIoVnCLH:-LL56sCa2ahGUVCSwGl8:bxubwo0) (no related file):*
Can you use filter for the filtering rather than changing routes? That needs to be the type of path we go down in future -- can't have every question support every filter anyone may ever want.

---

PR #2197, opened 2018-08-29
Traceroute: handle multiple nextHopIps on the same interface

dhalperi:
This is bad design, but we shouldn't crash.

Related changes:

* Since we allow multiple different next hops on the same interface, we now need to
  no longer compare FlowTraceHops identically if they have the same edge. We need to
  implement a correct comparator.
* Refactor equals methods / nullable annotations in other classes I thought the bug was
  in before.
* Add a missing toString.


dhalperi:
1. Need to improve testing
2. Happy to refactor out anything that needs it.

Issue #2143, opened 2018-08-24
Remove question path calculation from Batfish


arifogel:
As part of migration of storage to separate module, calculation of question/answer paths should be done in `BatfishStorage` instead of `Batfish`.

PR #2200, opened 2018-08-30
BatfishStorage refactor and overhaul

arifogel:
- Conduct all worker storage operations through `StorageProvider`
- Move all worker file access to `FileBasedStorage`
- Clean up heinous question path hackery

Follow-on work: make Coordinator use new StorageProvider for storage operations


arifogel:
Fixes #2143

PR #2202, opened 2018-08-30
ReachabilityParametersResolver: drop inactive interfaces during resolution

dhalperi:
Is this a good way to do this? Seemed worse to make every specifier filter out inactive interfaces...


projects/batfish/src/main/java/org/batfish/main/ReachabilityParametersResolver.java
```diff
@@ -153,7 +156,33 @@ IpSpace resolveDestinationIpSpace() throws InvalidReachabilityParametersExceptio
...
+                    iface =
+                        new NodeInterfacePair(
+                            ((InterfaceLocation) l).getNodeName(),
+                            ((InterfaceLocation) l).getInterfaceName());
+                  } else {
+                    assert l instanceof InterfaceLinkLocation;
+                    iface =
+                        new NodeInterfacePair(
+                            ((InterfaceLinkLocation) l).getNodeName(),
+                            ((InterfaceLinkLocation) l).getInterfaceName());
```
dhalperi: Obviously, this should be factored out. a hack for testing and review.

saparikh:
We should only use active interfaces. Otherwise in situations where people have the same IP address on 2 devices, but where only one is active we will do the wrong thing.

PR #2211, opened 2018-08-30
InterfaceProperties: add missing parameter

dhalperi:


questions/experimental/interfaceProperties.json
```diff
@@ -12,6 +13,12 @@
             "interfaces"
         ],
         "variables": {
+            "excludeShutInterfaces": {
+                "description": "Exclude interfaces that are shutdown",
+                "type": "boolean",
+                "optional": false,
+                "value": true
```
dhalperi: @ratulm  false or true here?

PR #2242, opened 2018-09-07
negate AclMatchExprs

anothermattbrown:
Negate an AclLineMatchExpr, applying deMorgan to AndMatchExpr and OrMatchExpr (at most once -- uses NotMatchExpr instead of recursion to negate when applying deMorgan). Everything else is wrapped in a NotMatchExpr (including MatchHeaderSpace, which could also be negated using the flag in HeaderSpace), and double-negation is eliminated.


dhalperi:
(ship when ready -- I don't need to re-review this assuming changes are small.)

PR #2249, opened 2018-09-09
Build sets of AclLineMatchExprs conjuncts/disjuncts

anothermattbrown:
The goal of these classes is to build sets of conjuncts/disjuncts that have no redundant members. 
 Most of the logic exists in the abstract base class `AclLineMatchExprSetBuilder`. It uses BDDs to determine which members are redundant and remove them as they are inserted.


dhalperi:
:lgtm_strong:

PR #2250, opened 2018-09-09
memoize conversions to BDD

anothermattbrown:
Adds two new classes `MemoizedAclLineMatchExprToBDD` and `MemoizedIpSpaceToBDD`, which just memoize the their parent classes' `visit` methods. The rest of the changes are one these two things: make `AclLineMatchExprToBDD`/`IpSpaceToBDD` objects (and transitively, things like `HeaderSpaceToBDD` that depend on them) parameters (so we can pass in memoized or non-memoized instances as desired); use the `visit` method to  make sure memoization occurs.


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/symbolic/bdd/MemoizedAclLineMatchExprToBDD.java, line 20 at r2](https://reviewable.io/reviews/batfish/batfish/2250#-LM8VFU77xLVNcnKZ4np:-LM8iSx03XcYxF9KAyWb:bw521t6) ([raw file](https://github.com/batfish/batfish/blob/34ec81e1d5cac96d4731a7480b05d63f49ce929d/projects/batfish/src/main/java/org/batfish/symbolic/bdd/MemoizedAclLineMatchExprToBDD.java#L20)):*

1. Didn't want to use a cache here, just want to remember everything that has been seen before. This is a short-lived object and should be GC'd reasonably quickly.
2. Identity hash map compares using reference equality.
</blockquote></details>

1. ok 
2. I made a typo, I meant non-reference.

PR #2264, opened 2018-09-12
batfish: delete unused questions

dhalperi:


dhalperi:
@jkhourybbn can you say more about your use case? There's lots of alternatives, e.g., FileParseStatus, for specific things.

These questions were completely untested and did not match current standards (table answers, schemas, etc). Happy to review a PR adding a useful question with proper style and tests, or to help design one that meets your constraints.

jkhourybbn:
use case
  * load a test rig at t=0
  * do some work, find a problem in specific config
  * try to figure out the file corresponding to the config

if I understand correctly, the question needed to be upgraded to meet style and tests, so u deleted it instead :)?

dhalperi:
> if I understand correctly, the question needed to be upgraded to meet style and tests, so u deleted it instead :)?

No, it was out-of-date AND as far as we knew unused. We're deleting with prejudice everything that's no longer used. Tests are one indication of use :).

> try to figure out the file corresponding to the config

ParseStatusQuestion returns a much better answer for this. See https://github.com/batfish/batfish/blob/master/tests/basic/parse-status.ref#L30-L40

PR #2271, opened 2018-09-12
Example network reorganization

haverma:
- moving example and example-with-delta in a new parent dir campus-example containing the common topology image
- adding a readme for both snapshots in the parent dir


saparikh:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/2271#-LMEkgzH7bAMT-CZd8e4:-LMEmCpzAqsyCK-8m9Uw:b-l4tbwc) (no related file):*

I think the name of the primary example should be as clear as possible that this is where you start.

Why add `campus-` to `example`? It seems to me that `example` is a better canonical name.

Instead, I'd probably change the snapshot names -- `live` and `candidate`, for example? @ratulm? @saparikh ?
</blockquote></details>

I would just leave it as `example`. But if we did want to change it so that we can have more descriptive future examples, I would do `example-campus`. Then we could have one that is `example-dc`

---

PR #2272, opened 2018-09-13
Delete auto-generated examples networks

haverma:
deletes
- full-mesh examples
- jelly-fish examples
- ring-examples


ratulm:
Various people have used these networks were useful for testing perf. Would it make sense to move them to their own repo for easy access? Generating them again will take a bit of time for anyone.

PR #2284, opened 2018-09-14
Add VI Model question and tests

corinaminer:
The new question produces all the information produced by the existing Nodes and Neighbors questions, new & improved, in one big JSON.

The keys are: nodes, bgpEdges, eigrpEdges, layer1Edges, layer2Edges, layer3Edges, ospfEdges, and ripEdges. Let me know if it would be better to group all the edges output under one `edges` key.

The only (intentional) difference between the new question's answer and the Nodes/Neighbors answers is that Neighbors separates ebgpEdges and ibgpEdges, whereas the new question combines them into bgpEdges (based on advice from @progwriter ).


progwriter:
Oh, and per you comment, I think it is better to put all edges under one `edges` key and drop the `Edges` suffix from the fields.

PR #2296, opened 2018-09-14
Remove diff column from Table diffing

ratulm:
That feature just seems to get in the way more than it helps.


dhalperi:
Deferring to the team using this stuff more frequently.

PR #2299, opened 2018-09-17
factor IpAccessListToBDD out of BDDAcl

anothermattbrown:
Previously BDDAcl did two things: provide an abstraction of an ACL
represented by a BDD for minesweeper, and a factory to convert ACLs to
BDD (handling references to IP spaces and other ACLs, sources, etc).
This splits them into two classes, which simplifies things and makes it
easier and more efficient to reuse the factory functionality.


yifeiyuan:
:lgtm:

PR #2319, opened 2018-09-19
Rename java traceroute questions

progwriter:
* traceroute -> oldtraceroute
* traceroute2 -> traceroute.
* deprecate old traceroute


jkhourybbn:
the constructor of the new TraceRouteQuestion is not public so I cannot create a TRaceRouteQuestion outside the package. Is this by design? and if so, what is the right way to create a question directly from Java

This does not work
```
TracerouteQuestion question =
        new TracerouteQuestion(false, ingressNode,
                PacketHeaderConstraints.builder().setDstIp(dstIp).build());
```

progwriter:
@jkhourybbn sorry about that, I have a PR that should fix this: #2353

Issue #2318, opened 2018-09-19
Move AclReachability fully into `org.batfish.question.aclreachability` 


dhalperi:
BDD stuff is in common and AclReachability no longer uses anything but BDDs. Some things I believe are doable now:

* move `Batfish#answerAclReachability` into AclReachabilityAnswerer and get rid of the API from IBatfish.
* get rid of AclReachabilityRows indirection for the result table and simplify+inline in AclReachabilityAnswerer
* move AclReachability e2e test from allinone to question module, see if it can be simplified to pure unit testing now that Batfish need not be involved at all.

PR #2323, opened 2018-09-19
AclReachability: move out of batfish and integrate test

dhalperi:
Part 1 of batfish/batfish#2318

* Delete IBatfish#answerAclReachability. Inline the logic in AclReachabilityQuestion.
* Move AclReachabilityTest from allinone to question, since it now does not need
  Batfish to run. This required fixing up IBatfishTestAdapter a little.

Lots still to do, but this is a nice first step.

### DIFF
--- a/projects/allinone/BUILD
+++ b/projects/allinone/BUILD
@@ -47,22 +47,6 @@ junit_tests(
 )
 
 junit_tests(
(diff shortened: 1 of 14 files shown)


yifeiyuan:
:lgtm:

PR #2346, opened 2018-09-22
use default throwing implementation inside IBatfishTestAdapter

arifogel:
- non-throwing implementation is unexpected
- it is the job of extenders to provide functionality


dhalperi:
Tl;dr well designed adapters never throw and only make you override
functions you wanted to. Nothing should throw if possible.

See MouseAdapter for example.

Of course it’s hard with apis that aren’t especially well thought out (like
a bag of functions ;).

PR #2358, opened 2018-09-26
bazel: add javadoc generation

dhalperi:
My hope is this can be a foundation of a script to update the website. The main advantage of using bazel for this: it's not affected in any way by the local state so there's no Maven/IDE screwups. And the commands are simpler.

To try it:

1. `bazel build //projects:javadoc`
2. Look at the produced file: `ls $(bazel info bazel-bin)/projects/javadoc.zip`


dhalperi:
Still plenty of things not done yet to get up to part with the Maven-generated docs. But the basic flow is worth looking at.

PR #2363, opened 2018-09-28
Split Neighbor_Unreachable_Or_Exits_Network disposition into three cases

yifeiyuan:
Split Neighbor_Unreachable_Or_Exits_Network disposition into Neighbor_Unreachable, Delivered_To_Subnet, Exits_Network for traceroute engine


yifeiyuan:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/TracerouteEngineImplContext.java, line 231 at r4](https://reviewable.io/reviews/batfish/batfish/2363#-LOKUjpy0lrPVtfLoL9o:-LOdE-jd866S8AntbgRM:b30fml6) ([raw file](https://github.com/batfish/batfish/blob/5eae381da6028084a1fda2dd50e82f6b339f1428/projects/batfish/src/main/java/org/batfish/dataplane/TracerouteEngineImplContext.java#L231)):*

> NodeNameRegexConnectedHostsIpSpaceSpecifier#HOST_SUBNET_MAX_PREFIX_LENGTH
This is an interesting corner case. Confirming with Samir in the Google doc.
</blockquote></details>

So Samir replies that multiple interface addresses should not overlap in practice. So I think we can still use anymatch in this case.

PR #2412, opened 2018-10-09
Add initial forkSnapshot API call

sfraint:
Add new API call to fork snapshots and deactivate nodes, links, interfaces.

Remove init/upload/parse environment since those are replaced by fork snapshot and will not work with blacklist changes.


sfraint:
Duplicate of https://github.com/batfish/batfish/pull/2413

PR #2440, opened 2018-10-11
add headerspace constraint to reduced reachability question

anothermattbrown:
The headerspace constraint becomes an IpSpaceAssignment (per-source
source IpSpace constraints) and an AclLineMatchExpr (for other header
fields).

Also:
- Pass a constraint on the final headerspace into
  BDDReachabilityAnalysis. This will help direct the search.
- Fix bug where InterfaceLink assignments were clobbering each other
  after they were converted to OriginateVrf StateExprs. Now we merge
  them using OR.
- When a headerspace constraint is provided, check that at least one
  source is consistent with that constraint (since we could have source
  IP constraints in the headerspace constraint and in the source
  IpSpaceAssignment).

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/question/multipath/MultipathConsistencyTest.java
+++ b/projects/allinone/src/test/java/org/batfish/question/multipath/MultipathConsistencyTest.java
@@ -1,20 +1,28 @@
 package org.batfish.question.multipath;
 
+import static org.batfish.datamodel.matchers.FlowMatchers.hasDstIp;
+import static org.batfish.datamodel.matchers.FlowMatchers.hasDstPort;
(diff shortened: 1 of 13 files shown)


progwriter:
___
*[tests/questions/experimental/reducedReachability.ref, line 14 at r3](https://reviewable.io/reviews/batfish/batfish/2440#-LO__2TK3-ulmI999dR7:-LO__2TK3-ulmI999dR8:br6vhd2) ([raw file](https://github.com/batfish/batfish/blob/e018523202460276605d642c73f215a388cf33e7/tests/questions/experimental/reducedReachability.ref#L14)):*
> ```
>       "reachability",
>       "reduced reachability"
>     ]
> ```

why no variables? seems like both actions and headers should show up here

PR #2465, opened 2018-10-16
Update forkSnapshot to allow uploading new/updated files

sfraint:
Update forkSnapshot to accept a base64 encoded zip, containing new/updated files to write to the fork.  These files overwrite any conflicting files copied from the base snapshot.

This zip has the same structure requirements as snapshot upload zips (e.g. config files should be located under `<snapshot_name>/configs/`).


dhalperi:
___
*[projects/coordinator/src/main/java/org/batfish/coordinator/resources/ForkSnapshotBean.java, line 53 at r2](https://reviewable.io/reviews/batfish/batfish/2465#-LOzH5B0CK0EeEvceeXC:-LOzJ8GX0BHL1Kpbp13G:b-9zm48s) ([raw file](https://github.com/batfish/batfish/blob/daa7aa07319980fa4eff46a25000b7f5a56c4b3a/projects/coordinator/src/main/java/org/batfish/coordinator/resources/ForkSnapshotBean.java#L53)):*

Ugh this makes me really nervous but let's defer.
</blockquote></details>

byte[] at the very least.

PR #2484, opened 2018-10-18
 handle non-existent source locations and final nodes in bdd reachability engine

anothermattbrown:
Make the engine robust to non-exist final nodes and source locations.  These can arise in the context of differential reachability (if nodes/locations are added or removed). It's not immediately clear what the desired semantics is in this case, but it should be up to the question. So the engine should be able to tolerate handle non-existent nodes/locations.


ratulm:
In case it helps determine the right/consistent behavior, table diffing had a similar issue: what to do with "keys" that are present in only one table. Using "includeOneTableKeys" flag (which is question-level flag), one can do an "inner join" or a "full outer join." I started with default being the former, but switched to the latter because missing keys things were creating confusion even for me :)

PR #2496, opened 2018-10-19
Cisco parsing fixes

elizalucas:
We added fixes for three different cisco parsing errors:

1. For `router ospf`, `distribute-list` may be applied to a specific interface.
2. For `line`, ignore `speed`.
3. For `router ospf`, allow `priority` to be applied to all interfaces.


dhalperi:
Hey @elizalucas - this is beautiful! There's a few things to do across the last mile, mainly around reference tracking. (we want to make sure we track all references to lists and interfaces, so that the really important baseline questions like `bfq.undefinedReferences` work).

That work is a bit more involved, do you want to take it on (in which case I can better define) or hand it off to Batfish team to finalize?
Thanks!

agember:
@dhalperi It would be great if the Batfish team could take care of it.

PR #2510, opened 2018-10-23
Convert ParseVendorConfigurationAnswerElement filemap to a multimap

sfraint:
More than one input file may contribute to the generation of a single vendor configuration, so track all input files, not just the last one.

For example: [networks/iptables-firewall/](https://github.com/batfish/batfish/tree/master/networks/iptables-firewall) snapshot contains a firewall config (`configs/z2-firewall.cfg`) and corresponding overlay (`hosts/z2-firewall.json`) which combine to produce the vendor model `z2-firewall` file.

For this network, a subset of the `ParseVendorConfigurationAnswerElement` (the input fileMap) changes from:
```
        "z2-firewall" : "configs/z2-firewall.cfg",
```
to:
```
        "z2-firewall" : [
          "configs/z2-firewall.cfg",
          "hosts/z2-firewall.json"
        ],
```

Note: `iptables` files are not yet separately tracked (i.e. they are implicitly linked via the hosts json and not actually added to the fileMap)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseVendorConfigurationAnswerElement.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseVendorConfigurationAnswerElement.java
@@ -1,7 +1,8 @@
 package org.batfish.datamodel.answers;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.TreeMultimap;
 import java.io.Serializable;
 import java.util.SortedMap;
 import java.util.TreeMap;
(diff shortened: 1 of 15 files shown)


dhalperi:
___
*[projects/question/src/main/java/org/batfish/question/UndefinedReferencesQuestionPlugin.java, line 59 at r1](https://reviewable.io/reviews/batfish/batfish/2510#-LPTK8IV9VOEKv_kIq0c:-LPTK8IV9VOEKv_kIq0d:bih6zs3) ([raw file](https://github.com/batfish/batfish/blob/06ee8cbac45ff0a124638b743b21f2b6f9f6eaae/projects/question/src/main/java/org/batfish/question/UndefinedReferencesQuestionPlugin.java#L59)):*
> ```Java
>               .stream()
>               .filter(e -> includeNodes.contains(e.getKey()))
>               .map(Entry::getValue)
> ```

I'm probably missing something, but how does this compile?

Isn't `e.getValue()` a List<String> or something?

PR #2512, opened 2018-10-23
Fixing duplicate routes in traceroute

haverma:
Not showing routes in the `Routing` step which differ in properties which are not a part of the `RouteInfo` class.
Also removing the unused `routesForThisNextHopInterface ` attribute of `TransmissioinContext`

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RouteInfo.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RouteInfo.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.batfish.datamodel.Ip;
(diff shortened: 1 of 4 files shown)


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RoutingStep.java, line 38 at r1](https://reviewable.io/reviews/batfish/batfish/2512#-LPTxgXBBdgeu4LN24th:-LPTxgXBBdgeu4LN24ti:b-5fjfks) ([raw file](https://github.com/batfish/batfish/blob/6a6b7e0113b5b0af327eb612ccfdd0fdc53ddcb9/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RoutingStep.java#L38)):*
> ```Java
>      * be multiple in case of ECMP
>      */
>     @Nonnull private Set<RouteInfo> _routes;
> ```

Is this an absolutely necessary solution, or working around the bug in FIBs you filed?

if the latter, I'd prefer you keep this a list and just check containment when adding routes.

haverma:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RoutingStep.java, line 38 at r1](https://reviewable.io/reviews/batfish/batfish/2512#-LPTxgXBBdgeu4LN24th:-LPX72KH28K8KUsiSmpI:blza67d) ([raw file](https://github.com/batfish/batfish/blob/6a6b7e0113b5b0af327eb612ccfdd0fdc53ddcb9/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/RoutingStep.java#L38)):*

Is this an absolutely necessary solution, or working around the bug in FIBs you filed?

if the latter, I'd prefer you keep this a list and just check containment when adding routes.
</blockquote></details>

reverted to list in https://github.com/batfish/batfish/pull/2515

PR #2519, opened 2018-10-23
Add VI model -> Vendor model file mapping

sfraint:
Add VI model -> Vendor model mapping in `ConvertConfigurationAnswerElement` to track which VI model files derive from which intermediate vendor model files.

Setting up for preserving work (knowing what needs to be redone) during snapshot forking.


dhalperi:
___
*[tests/basic/example-iptables-init.ref, line 239 at r1](https://reviewable.io/reviews/batfish/batfish/2519#-LPYDKCB6auLjJUxR_2O:-LPYDKCB6auLjJUxR_2P:bnea4rc) ([raw file](https://github.com/batfish/batfish/blob/5d6ad26a9b3b5e5e92a6926389887778724cd510/tests/basic/example-iptables-init.ref#L239)):*
> ```
>         "z2-firewall" : "z2-firewall",
>         "z2-host1" : "z2-host1",
>         "z2-host2" : "z2-host2"
> ```

This isn't making sense to me... why are these always the same? These do not seem like filenames.

sfraint:
Closing this until we can nail down more of the design.

PR #2522, opened 2018-10-24
 consider network topology when checking if an interface is full

anothermattbrown:
Previously we only considered the owned IPs when determining whether an interface is full. This leads to incorrect disposition assignment (e.g. `DELIVERED_TO_SUBNET` instead of `NEIGHBOR_UNREACHABLE`) when the IP is owned by an interface in the network but there is no connection to that edge.


anothermattbrown:
appears to be wrong

PR #2527, opened 2018-10-24
Remove NUOEN disposition from set of failure dispositions

progwriter:
Since it's no longer supported by the engine.


progwriter:
___
*[projects/question/src/test/java/org/batfish/question/specifiers/DispositionSpecifierTest.java, line 86 at r1](https://reviewable.io/reviews/batfish/batfish/2527#-LPb8Nt--tj0t6ElZ1Mk:-LPb91V8AOfnMcTuO5S_:b-f1gabc) ([raw file](https://github.com/batfish/batfish/blob/81fe08adb76285dcecd966411c323b9b8fc69847/projects/question/src/test/java/org/batfish/question/specifiers/DispositionSpecifierTest.java#L86)):*

Until this disposition is fully removed, it should be considered SUCCESS right? That's what we assumed in other contexts -- because it could be delivered/exits and we're conservative.

(Or we can just cry)
</blockquote></details>

But this is input only and we don't want folks to actually specify this disposition.

PR #2556, opened 2018-10-29
Add separate folder for optional snapshot files

progwriter:
Thoughts? A centralized place for all optional files?


progwriter:
Too stale.

PR #2572, opened 2018-10-31
Implement configureTemplate; other cleanup in Question

corinaminer:
Had this sitting around, no point letting it rot. Implements `configureTemplate`  and fixes a few unrelated PMD issues in the `Question` class.


corinaminer:
leaving `configureTemplate` behind. just letting clients mess with the template json themselves

PR #2600, opened 2018-11-06
QuestionPlugin: mark createQuestion as deprecated - for internal use only

dhalperi:


dhalperi:
Closing until I have time to improve -- seems like mandatory overriding deprecated functions causes warnings, and that wasn't my intention.

PR #2606, opened 2018-11-07
Customizable BGP admin distance: juniper, global

dhalperi:


dhalperi:
Since we have to support per-peer AD, this is grossly complicated, closing for now.  Need to rethink approach.

PR #2618, opened 2018-11-10
Pack allowed vlans as String, not List<String>

ratulm:
Catching up after 6b3a90845876934928ba6bd18851dc5cd1fcdab9. 

(There are downstream tests that started failing after the commit. That nothing failed in this repo reveals a testing gap.)


dhalperi:
Yes I suspect the interface properties question East tested on the example
network rather than on an interface that actually populates all interface
properties.

PR #2619, opened 2018-11-10
Support nested collections in Schema

ratulm:
Any reason to not allow nested collections in Schema? Simplifies life for questions that embed richer things than single level of nesting, and then they don't have to go to Object schema. What might this break?


dhalperi:
The spec for Schema was no nested collections -- and IIRC this was
important for the early laxness of design decisions. Why the change now?

ratulm:
Running into situations now where having nested collections simplifies
things a lot. Typical situations are for questions that take other other
tables as input for consistency checking or diffing. Their answers combine
cells from one or more input tables, and we need a nested collection to
represent these objects. In one place we were using Schema.Object to get
around this limitation. But Schema.object complicates cleint-side
interpretation of data. Nested collections are the better way to go I think.

dhalperi:
I suspect you really want a row type, not nested collections. Have you
considered speccing that out?

Also for review, can you please provide a link to a Pybatfish Notebook that
shows how nested answers will be displayed?

ratulm:
I am pretty sure that I don't want a row type.

I'll check on pybatfish.

ratulm:
I forgot I had already checked on pybatfish for one example analysis, and I got this:

screen shot 2018-11-12 at 2 17 42 pm

I checked on the code, and it seems to be doing the right thing recursively.

ratulm:
I forgot I had already checked with Pybatfish, and I got this:

screen shot 2018-11-12 at 2 20 42 pm

I just checked on the code itself and it seems to be doing the right thing with recursively.

dhalperi:
`List<List<SelfDescribing>>` is *exactly*  what I was worried about. The inner thing is a row, right? Multiple columns that deserve their own names, etc.

PR #2620, opened 2018-11-11
Bgppropspecifier from list

ratulm:


ratulm:
Included in another PR

PR #2625, opened 2018-11-12
 update column names in table diff

anothermattbrown:
Use "Snapshot" and "Reference" instead of "Base" and "Delta".


ratulm:
Is there general agreement on whether these should be the names? The name `Snapshot` seems to need an adjective to differentiate it from `Reference Snapshot`.

PR #2645, opened 2018-11-15
First pass update for README

saparikh:
Big overhaul

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,22 +7,19 @@
 
 Batfish is a network validation tool that provides correctness guarantees for security, reliability, and compliance by analyzing the configuration of network devices. 
 
-The primary use-case for Batfish is to evaluate planned configuration changes in order to understand the impact of the change. Pre-change ( / pre-deployment / proactive) validation is a critical gap in existing network automation workflows.
+The primary use case for Batfish is to evaluate planned configuration changes in order to understand the impact of the change. Pre-deployment validation is a critical gap in existing network automation workflows. 
+
+While pre-deployment validation is the primary use case, Batfish can also be used to provide post-change validation.
 
 Batfish builds complete models of network behavior from device configurations and finds violations of network policies (built-in, user-defined, and best-practices). By integrating Batfish into their network automation workflow, network engineers can close this gap and ensure that only correct changes are deployed.
 
-While pre-change validation is the primary use-case, Batfish can also be used to provide post-change ( / continuous) validation.
 
-**The Batfish service does NOT require direct access to network devices**, as the core analysis just requires the configuration of network devices. However, additional information from the network can also be fed into Batfish to enhance the analysis. That information includes, but is not limited to:
+**The Batfish service does NOT require direct access to network devices**, as the core analysis just requires the configuration of network devices. Additional information from the network can also be fed into Batfish to enhance the analysis. That information includes, but is not limited to:
 
 * BGP routes received from external peers
-	* This enables route policy validation
 * Topology information represented by LLDP/CDP
-	* This enables Layer 2 analysis - Spanning Tree, VLAN connectivity, etc…
-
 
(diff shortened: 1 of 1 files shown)


saparikh:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/2645#-LRKd909ETQmU8gdLKEJ:-LRNxyik4GAolBJgr_2q:be32yek) (no related file):*

This is a step up from what we have. My only worry would be that its a wall of text. We could use screenshot from notebooks (which gets across how easy it is to use) and/or list of checks into a table. 
</blockquote></details>

Let's get a version out and iterate on it. What we have there is bad and we need to get an update out there.

A lot of READMEs are just walls of text. Agree that it is not ideal, but a ton of pictures is also not good.

---

saparikh:
___
*[README.md, line 16 at r2](https://reviewable.io/reviews/batfish/batfish/2645#-LRK_Crw2GYQlXOuARqp:-LRO11nF1tvJ83ZTr9or:brrqm6b) ([raw file](https://github.com/batfish/batfish/blob/34bc01868819cab6891d38c0e457c45e63a5b4a2/README.md#L16)):*

Its not really a "However" just like configuration other data can be fed in without access to the network. Alternative:

**Batfish does NOT require direct access to network devices** It can perform its analysis "offline" based on configuration pulled from the network by other tools (e.g., RANCID) or based on configurations generated by templates.  The analysis can also be enhanced using other data from the network such as a snapshot of BGP routing tables from external peers. 
</blockquote></details>

made some edits

saparikh:
___
*[README.md, line 19 at r2](https://reviewable.io/reviews/batfish/batfish/2645#-LRKZ4t06DpDS1eY32__:-LRO14cr9l9353O8rbtW:b7ocb6w) ([raw file](https://github.com/batfish/batfish/blob/34bc01868819cab6891d38c0e457c45e63a5b4a2/README.md#L19)):*

This phrasing seems to suggest that we cannot do any route policy validation without external BGP data. One suggestion: "This enables validating routing and forwarding to external destinations"?

My preference would be to remove these bullets altogether -- they don't seem to justify the real estate they are taking up. 
</blockquote></details>

got rid of the sub-bullets. i like keeping the bullets. adding them into the sentence just makes it a long sentence.

saparikh:
___
*[README.md, line 28 at r2](https://reviewable.io/reviews/batfish/batfish/2645#-LRKacwi62HTRUTaCs9J:-LRO1MLgCCKIKH3IngO0:bzdri82) ([raw file](https://github.com/batfish/batfish/blob/34bc01868819cab6891d38c0e457c45e63a5b4a2/README.md#L28)):*

The point about complete model of network behavior can be made here. Also relevant to my point about that para seeming out of place. If you do that, the other sentiment in that para about integration into automation workflows can appear after the "no access to network devices" bit.
</blockquote></details>

check out the updated one and see how that works.

saparikh:
___
*[README.md, line 31 at r2](https://reviewable.io/reviews/batfish/batfish/2645#-LRKbRTOA3SKDLsvTUyW:-LRO1SQT-FuMrDCbOAnW:b-j25vu9) ([raw file](https://github.com/batfish/batfish/blob/34bc01868819cab6891d38c0e457c45e63a5b4a2/README.md#L31)):*

Why lead with compliance?
</blockquote></details>

it is the thing most people are focused on with their templating efforts, etc.. . I don't think the ordering matters that much

saparikh:
___
*[README.md, line 45 at r2](https://reviewable.io/reviews/batfish/batfish/2645#-LRKb5Wd--k-x8EDRzP4:-LRO1a6ADu_VrEayvO2E:b-vq9jmn) ([raw file](https://github.com/batfish/batfish/blob/34bc01868819cab6891d38c0e457c45e63a5b4a2/README.md#L45)):*

This is an odd categorization, as all the checks can be "proactive" change analysis. Just "change validation"? 
</blockquote></details>

I had just change analysis, so will go back to that. Agree proactive is not needed

PR #2694, opened 2018-11-21
add invertSearch parameter to specifiers reachability question

anothermattbrown:
this is to prepare for merging differential reachability and specifiers reachability


ratulm:
This change is independently useful and should proceed. @dhalperi has some concerns about merging the two templates -- lets resolve that offline before merging templates.

Issue #2701, opened 2018-11-23
Add support for network objects for ASA firewalls


ratulm:
Cc: @sfraint , @dhalperi

dhalperi:
@ratulm can you say more? There's a lot of network-object support in ASA!

ratulm:
Apparently network *objects* and network *object-groups* are different: https://community.cisco.com/t5/firewalls/difference-between-object-and-object-group/td-p/2429784

dhalperi:
Good news: this is on the surface an easy fix.

Bad news: the current implementation maps BOTH `network object foo` and network object-group foo` to the named IPspace `foo` in the VI model. Presumably, this is bad. The standard solution would be to name them `foo~object` and `foo~object-group` in the VI model, or something like that. Something like that make sense?

(We could presumably, with more work, only add `~object` etc. if there was a collision. But that might be misleading in other cases, as their names of could change between snapshots if we added/removed the collision.)

ratulm:
Your suggestion of mapping them to different names makes sense to me. 

The thing about smartly handling collisions is fragile for the reasons you mention, and also that within the same snapshot, these things may appear in different orders in different files. We'll lose easy comparison across nodes then.

dhalperi:
> within the same snapshot, these things may appear in different orders in different files.

The not-explicit proposal I had was not file-order dependent (use the type, as above, only during collision).

But it's still not worth it. Will do the name thing as well (probably in 2 PRs).

dhalperi:
Sent a first PR with the basic support. Next steps:

1. convert names for all the groups to include their types, since not doing so leads to collisions.
2. fix other (related) todos in the same CCPE function -- address groups are todo as well.

dhalperi:
Resolved - don't need to solve the name conflict thing, because Cisco doesn't let you have name conflicts. We already warn and throw away the second one.

PR #2706, opened 2018-11-25
Cisco: rename network-object and network object-group in VI model

dhalperi:
For batfish/batfish#2701


dhalperi:
Closing for now.

PR #2726, opened 2018-11-30
helper for getting self describing objects from row

ratulm:


ratulm:
All you had to was to think aloud on Slack :)

PR #2734, opened 2018-12-02
assorted juniper parsing

arifogel:
- aggregate route with policy
- as-path-group
- interface tcp-mss
- BGP confederation and members

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -58,6 +58,7 @@ private void setWildcard() {
 tokens {
    ACK,
    BANG,
+   DYNAMIC_DB,
    FIN,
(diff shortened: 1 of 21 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java, line 2469 at r3](https://reviewable.io/reviews/batfish/batfish/2734#-LShf89q8oN3LT7RitD_:-LShf89q8oN3LT7RitDa:b-9mh4ul) ([raw file](https://github.com/batfish/batfish/blob/cad980cee36323bcb6fccc82f452d7b914e6b331/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2469)):*
> ```Java
>         name,
>         AS_PATH_GROUP_AS_PATH_SELF_REFERENCE,
>         ctx.name.getStart().getLine());
> ```

reference tracking bug for nested structures: can't currently trust the line in the token. this needs to be using `[this#]getLine`

(throughout!)

PR #2750, opened 2018-12-04
Juniper: treat reth and ae the same

dhalperi:


dhalperi:
Sounded like @progwriter had a better handle on this

PR #2751, opened 2018-12-04
Parse Juniper interface bandwidths

corinaminer:
Until now, any configured interface bandwidths were ignored and Batfish determined bandwidths based solely on interface names.


dhalperi:
I like all syntax that is expected to parse in parsing-tests. We catch so
many incidental model changes (and crashing bugs, and. ..) in those.

PR #2752, opened 2018-12-04
Compute channel group members and bandwidth on logical agg interfaces

progwriter:


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2413 at r1](https://reviewable.io/reviews/batfish/batfish/2752#-LStet6uCKGxcXYkvX5Y:-LStgEgJ8TPJYogY2I52:b-dgkuw8) ([raw file](https://github.com/batfish/batfish/blob/7a1e755644374f8577a3267a9e4764a64917d66f/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2413)):*

> ```
>                                 interfaces
>                                     .get(dependency.getInterfaceName())
>                                     .getChannelGroupMembers()
>                                     .stream())
> ```
getting an NPE somewhere in here
</blockquote></details>

Yeah, looking into it now

progwriter:
___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2397 at r1](https://reviewable.io/reviews/batfish/batfish/2752#-LSteTRp5oUywkxAy0-C:-LStgGWf3o08nMNlC4a_:b-u4ffb7) ([raw file](https://github.com/batfish/batfish/blob/7a1e755644374f8577a3267a9e4764a64917d66f/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2397)):*

I'm surprised this is the way you want to do this, rather than logical subinterfaces inheriting properties from aggregated <~physical> interface. Intuition?
</blockquote></details>

Given the current setup, this would allow us to get rid of the hack I put in for not disabling unit interfaces... but now I'm not sure this is the best way to go :P

PR #2761, opened 2018-12-05
Juniper interface: add support for 'c' modifier

dhalperi:
It means cells, which are apparently a unit of 384bps.

Includes a little test simplification using multiple units on the same physical interface.


dhalperi:
[Apparently cells are ATM cells, which had 48-byte payloads (384 bits). Old.]

PR #2762, opened 2018-12-05
return what you can when sending back analysis answers

ratulm:
partially fixes issues that crop up when the code changes the old questions cannot be parsed. the change will let the service return answers for questions that it can still understand instead of failing altogether.


ratulm:
I see. Should I return an error answer?

PR #2769, opened 2018-12-05
RibTreeNode: honor maxPrefixLength

dhalperi:


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/2769#-LT-lwOx0qaGKIrPXz9a:-LT-nwog6Uov_OHaiCGZ:bugtz7i) (no related file):*

I don't understand how this fixes things (I'm probably just being dumb, I'm not saying it doesn't). 
A single node should only contain routes where all routes are of exact same prefix length, so shouldn't the newly added filter be a no-op? Or is the invariant I have in mind gets broken somehow?
</blockquote></details>

Suppose my tree is 0.0.0.0/0 and 1.1.1.1/16

If I search for `1.1.1.1 and <= 10`, I'll visit 0.0.0.0/0 and then 1.1.1.1/16. The old code would realize it was past `10` and then return all routes matching IP in the current node.

This meant it would return 1.1.1.1/16.

---

PR #2802, opened 2018-12-11
Yifei disposition e nvs ii

yifeiyuan:
New disposition computation as in the doc


anothermattbrown:
closing until we're sure we know what the desired behavior is

PR #2810, opened 2018-12-12
Remove vrfOwnedIps from specifier context

progwriter:
Simplifies future work on caching results of ipOwners


dhalperi:
Unused? :shipit:

PR #2815, opened 2018-12-12
Juniper: approximate support for reth interfaces

progwriter:
This is far from full support for reth interfaces (see added TODO), but get's us a little closer to the right behavior for the time being.


progwriter:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/2815#-LTZ_zfQ673d7nkbISmi:-LTZhphE1nnKeQj1BvF0:b-9ginnf) (no related file):*

Please add junit tests s.t.
1. Redundant interface comes up because at least one dependency is up
2. Redundant interface does not come up because all dependencies are down
3. Redundant interface does not come up because it has no dependencies.
</blockquote></details>

This behavior is not implemented, although standby for my next PR 

---

PR #2828, opened 2018-12-14
Change all filter-related steps into a FilterStep

yifeiyuan:
Use a single FilterStep to unify all filters
1. Add FilterStep
2. Split Enter step into Enter step and Filter step
3. Split Exit step into Exit step and Filter step


saparikh:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/FilterStep.java, line 17 at r2](https://reviewable.io/reviews/batfish/batfish/2828#-LTi09nk99bsv94GkLOZ:-LTise8eDDCdJTcbMKOH:b-tl17ir) ([raw file](https://github.com/batfish/batfish/blob/e068f9ff08971c676340c157adb885538cbecb27/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/FilterStep.java#L17)):*

> n which filter can appear
That's a good point. With this PR's change, we can not tell the following two cases: 1. enter -> ingress filter -> exit; 2. enter -> egress filter -> exit.  Is it enough to get the info from the names? cc @dhalperi @saparikh 
</blockquote></details>

No... the names are not expected to provide any insight into where the ACL was applied. We have to keep track of that and distinguish between the cases

PR #2851, opened 2018-12-19
IBDP: add opentracing spans to key inner components

dhalperi:


dhalperi:
@haverma - is there anything simpler/cleaner I can do?

PR #2855, opened 2018-12-19
Search filters hygiene

millstein:
searchFilters should not catch all exceptions from reachFilter -- that was causing errors to be masked.

searchFilters should rename ACLs when they are modified, so they don't conflict with the original versions of those ACLs when generating explanations.


dhalperi:
Ship it

PR #2862, opened 2018-12-20
Don't throw on unmatched InterfaceType

sfraint:
Don't throw on unmatched InterfaceType during conversion to LinkType


sfraint:
Closing this as https://github.com/batfish/batfish/pull/2850 should catch any added interface types.

PR #2893, opened 2019-01-04
[draft] Use parboiled to convert Juniper community regexes to Java regexes

dhalperi:
1. Parboiled seems like a nice at-runtime parser generator we could extend to do autocomplete-like things. I'm playing with it to implement Juniper BGP regexes. Matt helped a lot.

2. I'm using the Juniper tests defined here: https://www.juniper.net/documentation/en_US/junos/topics/usage-guidelines/policy-defining-bgp-communities-and-extended-communities-for-use-in-routing-policy-match-conditions.html#id-10267652

obviously needs
- [ ] checkstyle ignoring
- [ ] comments
- [ ] move tests to unit tests from main
- [ ] negative tests
- [ ] use in grammar
before it could be merged,

but first impressions?


dhalperi:
Current state: works for regular communities but not yet extended communities. Determining whether I should add extended community support or improve Batfish data model for extended communities. TBD.

PR #2894, opened 2019-01-04
Pruning Layer 3 topology based on IPsec peering

haverma:
- using topology from new IPsec datamodel to do the pruning


haverma:
Closing for now as conversion of AWS IPsec VPN configs to the new IPsec datamodel is not yet done. So this is causing `HostTest.java` to fail because interfaces don't get deactivated, the deactivating logic tries to use the new IPsec datamodel which is empty for AWS configurations. Will follow-up with a PR for AWS IPSec VPNs conversions.

PR #2908, opened 2019-01-08
Juniper nat pool address parsing

yifeiyuan:
1. Allow "from ip_prefix to ip_prefix" in Juniper nat pool
2. When use a single ip_prefix to define a nat pool, the first and last ip addresses should not be considered


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java
```diff
@@ -74,16 +74,16 @@ public static Prefix parse(@Nullable String text) {
     return prefix;
   }
 
-  private final Ip _ip;
+  private final Ip _normalizedIp;
```
dhalperi: These changes are a no-go. need to not use prefix for this during conversion.

PR #2922, opened 2019-01-10
PropertySpecifier: switch from a set to a list

dhalperi:
So that the implementor can use order to signify preference, e.g., in property tables.


dhalperi:
cc: @saparikh

PR #2941, opened 2019-01-11
JNPR policy-statement: OR all prefix-lists

corinaminer:
Fixes a bug where routes were required to match every prefix list within a policy statement term instead of any one of them.


progwriter:
We have not (yet), and we shall defer this to a different PR.

PR #2950, opened 2019-01-14
Email exchange for BDD license

ratulm:
Including the email exchange in response from a request from a user.


ratulm:
Glad that I didn't fake it :)

PR #2965, opened 2019-01-15
Yifei screen parsing

yifeiyuan:
1. Add parsing rules to parse all screen options in Juniper SRX
2. Extract supported signature-based screen options into JuniperConfiguration, including:
   icmp large, ip unknown protocols, TCP Winnuke, TCP no flag, TCP SYN FIN, TCP FIN no ACK
3. Convert the above supported screen options into VI model as part of the incoming ACL.


dhalperi:
(Distributing reviews a bit)

PR #2980, opened 2019-01-17
implement juniper interface source-nat interface then clause

anothermattbrown:


dhalperi:
Do you have any guidance for this PR -- docs I should read to understand what's going on?

(Would be useful to include in future PR messages)

PR #2991, opened 2019-01-21
ASA Twice NAT parsing and extraction

mkremerbbn:
Added parsing and VS extraction for twice NAT. This includes
implementation and tests for conversion to Transformations, but
not integrated with the VI model or data plane.

@dhalperi @anothermattbrown @arifogel


mkremerbbn:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/NetworkObject.java, line 9 at r1](https://reviewable.io/reviews/batfish/batfish/2991#-LWmkd2uCVMlx6xoFXc4:-LWwcbMUAs3xU4tWNLWF:bhcfn7n) ([raw file](https://github.com/batfish/batfish/blob/8edf2eb49fd7ba194116c4f6f0847b13a7bb3efc/projects/batfish/src/main/java/org/batfish/representation/cisco/NetworkObject.java#L9)):*
<details><summary><i>Previously, mkremerbbn (Michael Kremer) wrote…</i></summary><blockquote>

Working on it
</blockquote></details>

I replaced NetworkObject with an interface and implemented various NetworkObject types

PR #3008, opened 2019-01-23
Enforce static import for more Guava utilities

dhalperi:


dhalperi:
Rather than commenting in review, seems better to add enforcement so I don't have to ;)

PR #3021, opened 2019-01-25
Remove duplicate definition of roa_route

corinaminer:
Keeping the one that allows for `routing-options aggregate route <prefix>` with no `roa_common` token.

cc: @arifogel -- let me know if there's some reason to ever have duplicate definitions.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3021#-LX5Za589KPFv49SIjwk:-LX5Za589KPFv49SIjwl:b-8nbppg) (no related file):*
Move the right definition to the one that's alphabetically in the right place?

There's no need for duplicate rules, the first one wins IIRC. Or the last one. But whatever.

---

PR #3030, opened 2019-01-26
utility to provide IpSpace for range of IP addreses

arifogel:


arifogel:
Closing in favor of #3036

PR #3031, opened 2019-01-27
update z3 to 2019-01-26-be61a26452735dfe62b14ddb4804ca5df0526dcf

arifogel:
- needed to use `ite` node as guard with doc engine


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3031#-LXKmZmV7Z8yMPsPGrhi:-LXKmZmV7Z8yMPsPGrhj:bfyfhpz) (no related file):*
Travis is failing, too, in the install z3 part. Do you know what's up with that?

---

PR #3055, opened 2019-01-30
Version: add build timestamp

dhalperi:


dhalperi:
Need to think about this if I really want this.

PR #3062, opened 2019-01-31
DisjunctionChain -> FirstMatchChain

corinaminer:
DisjunctionChain took a list of routing policies and returned true for a route if any of them matched and accepted (otherwise evaluated default policy and returned that result). That behavior may be useful somewhere, but the only place it was being used was for chained Juniper BGP import and export policies (e.g. `set protocols bgp import [ POLICY-1 POLICY-2 POLICY-3 ]`). In that context, the correct behavior is to return the result of the first policy that matches the route, whether it accepts or rejects (and if none match, return default policy result as before).

This PR turns DisjunctionChain into FirstMatchChain, which matches those semantics.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Result.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Result.java
@@ -1,6 +1,5 @@
 package org.batfish.datamodel.routing_policy;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.util.Objects;
 import javax.annotation.Nullable;
 import org.batfish.datamodel.routing_policy.expr.BooleanExpr;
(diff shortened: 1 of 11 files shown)


rabeckett:
___
*[projects/batfish/src/main/java/org/batfish/symbolic/smt/TransferSSA.java, line 455 at r2](https://reviewable.io/reviews/batfish/batfish/3062#-LX_IEMo9szW7GTa_iV0:-LX_UEUt-S8JHteS1-VS:bwxqqi1) ([raw file](https://github.com/batfish/batfish/blob/859ea007529490ce835c3ac48b283c46976c2f21/projects/batfish/src/main/java/org/batfish/symbolic/smt/TransferSSA.java#L455)):*

Changed this because `result.addChangedVariables(r)` doesn't look like it modifies `result`, but rather returns a new `TransferResult` that is a modified version of `result`. All other uses of this method follow the `result = result.addChangedVariables(r)` pattern. @anothermattbrown can you confirm this makes sense to change?
</blockquote></details>

I believe what you changed it to is correct.

PR #3096, opened 2019-02-05
Fix AclExplainer to instantiate MemoizedIpAccessListToBDD correctly

anothermattbrown:


dhalperi:
I have comments I'm somehow unable to publish on reviewable.

dhalperi:
MemoizedIpAccessListToBDD:

```
Add some comments with links between constructors to help users pick the wrong one?

E.g., on the above: "`If your ACLs may care which source interface packets come from, see {@link #...}`"
```

PR #3108, opened 2019-02-07
Split generic PrefixTrie logic from RibTree

progwriter:
Needed for two reasons:
1. Allows some versions of RIBs to store generic objects, not just routes
2. Will let us implement proper FIBs with LPM (and therefore allow moving off code from using RIBs directly. RIBs should be internal to DP only)


progwriter:
based on offline discussion with @anothermattbrown, closing to do more cleanup

progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMap.java, line 89 at r3](https://reviewable.io/reviews/batfish/batfish/3108#-LYDMdqV1BobWL8kEQnn:-LYDmCtF3dtJVbzweDeK:bm124hc) ([raw file](https://github.com/batfish/batfish/blob/9c14453bf65fc8230964d034eb06e10de7d56007/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMap.java#L89)):*

yes, but my point is that this could be the wrong node for the intended key. Also, since this is essentially a multimap, it is simpler to use a standard interface.
</blockquote></details>

Yeah, I suppose that's fine, but that means  I'll have to throw `IllegalArgumentExceptions` if callers are inserting into wrong parts of the subtree

progwriter:
Don't wanna deal with this.

PR #3123, opened 2019-02-10
Parboiled ipspace

ratulm:
It is not fully done yet (needs a few more tests, and need to plumb in autocomplete function), but it is ready for look if it all makes sense. 

Overall arch:
  - the parser goes in Parser.java
  - the specifier factory calls the parser with the input and gets back an AST if parsing is successful
  - the AST is interpreted by specifier when resolve is called
  - if parsing fails, potential matches are processed to generate a helpful message

The factory is not plumbed in anywhere yet.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3123#-LYOIJLTDcFXkQXi2AhA:-LYOIJLTDcFXkQXi2AhB:b-j8gq96) (no related file):*
I don't quite get the overall plan here. Can you link me to the design doc?

Naively, I would have expected all of this to not be implemented as a new specifier, but rather to entirely replace all existing SpecifierFactories.

E.g.,

* Hide all use of Parboiled and AST inside package-private logic, no new public classes whatsoever.
* Public API:  String -> Specifier (or tree thereof) 
* Deletion of all specifier factories

Then the only two ways to express specifiers would be

* String that is directly converted into java classes
* json representation of those same java classes

---

ratulm:
The library does not enforce anything about what goes on "value stack." It
has a few helper classes like this one, but we don't have to use them at
all. I'll fix and commit.

PR #3126, opened 2019-02-11
Fix and test loopback, tunnel delay and bandwidth

mkremerbbn:
Fix delay and bandwidth for tunnel interfaces on Cisco and add tests. Add tests for loopback delay and bandwidth.

@dhalperi

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/eigrp/EigrpMetric.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/eigrp/EigrpMetric.java
@@ -11,6 +11,7 @@ import java.util.Objects;
 import java.util.stream.LongStream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNullableByDefault;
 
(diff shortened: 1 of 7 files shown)


mkremerbbn:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/Interface.java, line 49 at r1](https://reviewable.io/reviews/batfish/batfish/3126#-LYSqvqH-fBCCY39uSUT:-LYStaJH3cfdtoXlkR_W:b-7z1szg) ([raw file](https://github.com/batfish/batfish/blob/f855ed4b5047e4e9e5b3491ee95cd358290b2ae7/projects/batfish/src/main/java/org/batfish/representation/cisco/Interface.java#L49)):*

I'm very surprised to see that 100kbits/sec is the default tunnel bandwidth. Can you please add a source if available, and note to which platforms this is applicable?

The same applies to delay below (but with less surprise :P).

Also if you'll indulge me, please use exponent that is multiple of 3 (i.e. `100E3D` in this case).
</blockquote></details>

Added a comment, specified that these values are specific to IOS since I don't know if they apply to any other platform.

PR #3176, opened 2019-02-15
Interface parser

ratulm:
Only parsing. Resolving the specifier using those visitors not done yet. 

Easier to review after #3164 is merged.


ratulm:
Probably not worth reviewing this -- more work for reviewers and for me. Lets just focus on #3181

PR #3177, opened 2019-02-15
Adding network with ISP configurations

haverma:
Adding example network with ISP configuration to peer with AS2 borders.
The network looks like this (this image is also a part of this PR): 
example-network-isp

Bidirectional Traceroute from host1 to 8.8.8.8:
screen shot 2019-02-15 at 11 49 16 pm


networks/example/live-with-isp/configs/as2border1.cfg
```diff
@@ -129,13 +129,16 @@ no ip http server
 no ip http secure-server
 !
 ip access-list extended INSIDE_TO_AS1
- permit ip 2.0.0.0 0.255.255.255 1.0.0.0 0.255.255.255
+ permit ip 2.0.0.0 0.255.255.255 any
  permit ip 10.12.11.2 0.0.0.0 10.12.11.1 0.0.0.0 
+ permit tcp any eq 179 any
```
saparikh: syntax is incorrect... should be `permit tcp any any eq 179` like the one below

saparikh: The line as written, only permits traffic outbound where TCP 179 is the source port. So the remote router must initiate the TCP connection for BGP session to be established.

PR #3179, opened 2019-02-16
collect definitions of legal names in one place

ratulm:
As I am creating more parsers, I found myself creating two sources of truth for valid names (for things like interfaces, vrfs, zones, reference objects, ....). This PR is the start of checking valid name definitions off of one specification.


dhalperi:
Btw, was the names file not a good place for this? https://github.com/batfish/batfish/blob/master/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Names.java

dhalperi:
(Datamodel restrictions of names should be in the datamodel somewhere, not in parboiled-specific code. And parboiled should accept any valid datamodel names, not add arbitrary restrictions that aren't there)

dhalperi:
Btw, haven't looked at the PR that's just my abstract thoughts from reading the discussion :p

PR #3204, opened 2019-02-21
Simplify traceroute route resolution after #3201

progwriter:
* Switch to using FibEntries instead of the really weird logic we had
* Deprecated newly unused Fib methods
* Add a matcher for FibEntry to simplify some tests


progwriter:
#3201 is broken.

PR #3212, opened 2019-02-21
Valid names

ratulm:
@anothermattbrown - code review.

@arifogel, @dan - can you please look at the regexes for Filter, Interface, Node, Vrf, and Zone in Names.java, to see if they capture all names (including generated names) that we are likely to see in practice. (We need to be specific in the grammar about what to expect for such names.) 

Once we are happy with the regexes, I am thinking about enforcing them at object creation time. (The enforcement was already there for reference objects and table columns names, which I ported here.)


ratulm:
@anothermattbrown , @dan - what do you guys think of this strategy for handling arbitrary strings for names? 

the essential idea is that if you don't have a "special" character in the name and do not begin with a digit or /, you can go ahead and use it without quotes, but you must double quote the name otherwise. special characters include spaces and characters that are part of the grammar as operators (,\&()[]). 

the implementation required adding a new Anchor.Type IGNORE for rules to ignore and washing off opening quotes when doing a search for auto completion. 

this strategy applies to interface, filter, vrf, zone, and node names.

ratulm:
On the back of the infra for supporting valid names, I added support for old-style regexes. The main detection logic is: if we see a string containing '*' where we expect to see a (node, interface, filter) name, we deem that to be a deprecated regex and treat it as such.  (Other back-compat changes will come in other PRs.)

Other changes in this PR: I had to modularize the code on ParserUtils.java a bit -- it was getting needlessly hairy. Added new tests.

PR #3216, opened 2019-02-21
Codecov: see if showing file-level diffs is useful

dhalperi:


dhalperi:
Looks like the change works (see earlier parts of the PR), now I've reverted all test deletions.

PR #3225, opened 2019-02-22
Clean up choice of grammar

ratulm:


ratulm:
There are four points of incompatibility that I am aware of:

1/ Using unenclosed regexes like 'as2.*' -- the new grammar expects regexes
like these to be surrounded by '/'. I will fix this by continuing (for now)
to interpret these things as regexes. This interpretation will be limited
in places of fundamental incompatibility, e.g., if the regex had a space or
other funny character. But my planned approach will work for all
first-party and third-party examples that I have seen.

2/ for location '[interfaceExpr]' by itself is no longer supported, one
must do 'nodeExpr[interfaceExpr]'. I can add support for this without much
impact (though I haven't seen examples of such usage).

3/ the meaning of '-' in ip specification has changed from minus to range
to make it consistent what we do for ports. both meanings cannot be
supported at the same time, so I was planning on leaving this as
incompatible.

4/ the set operator for unioning nodes, interfaces, filters has changed
from '+' to ',' (to make it consistent with ports and ips), and the for
differencing has changed from '-' to '\' because '-' is being used for
range and it also can appear in names. i was planning on leaving this as
incompatible too; given the fragility and limitations of set
union/difference in the old grammar, i suspect there wasn't must use for
that. if this seems problematic, i can add support for '+' to also mean
union for now. '-' has fundamental issues as per point above.

ratulm:
I am working on brining the new grammar closer to the old. Will send a PR soon. 

In the meanwhile, I'd argue for not holding up this PR and the related clean-up PR (#3227). Even if we don't transition to the new grammar, these are consolidating the choice of grammar used in one place as opposed to 50 files. It is using identical grammar (and semantics). No ref tests have been changed and the all notebook tests pass. 

(And having that switch available in master will make my life easier as do the other mentioned PRs.)

PR #3239, opened 2019-02-25
OSPF process configuration question

haverma:
screen shot 2019-02-25 at 1 36 29 am

Note: (cc @saparikh ) This question almost looks like a subset of the `OspfProperties` question but I guess we are planning to deprecate that.


saparikh:
Yes, the goal is to deprecate the old OSPF questions once the new ones are available. For the various routing protocols, our questions should follow a standard pattern - the one we established with BGP.

PR #3240, opened 2019-02-25
Grammar for EIGRP metric weights

mkremerbbn:
The default setting for metric weights is supported, but setting the
them is not supported. Added parsing and a 'todo()' for weights which
allows consolidation of other related warnings. For configurations that
do not set actually set 'metric weights', no warning is necessary.

@dhalperi


dhalperi:
Hey @mkremerbbn - sorry for dropping the ball, I've had a busy week between nanog and the next talk. Hoping one of the 3 I've assigned can take a look.

PR #3241, opened 2019-02-25
Handle inline service objects in extended access lists.

mkremerbbn:
ASA extended access lists allow a service specifier in the form "object
NAME". It is possible to use a protocol name or icmp-type (e.g.
icmp-echo) in place of an explicit service object.

@dhalperi


dhalperi:
Hey @mkremerbbn - sorry again for dropping the ball, I've had a busy week between nanog and the next talk. @sfraint - take a look?

PR #3242, opened 2019-02-25
Refactor traceroute to introduce tracing pipelines, delete RIB code

progwriter:
* Begin laying groundwork for flexible pipelines in traceroute
* Make traceroute interact with FIBs only
* Fix unreadable route resolution

This is an expanded version of #3204 
There are no functionality changes, all just refactoring. The goal is to have more flexible stages/pipelines in traceroute that can call into each other by using a hybrid of a visitor/mediator pattern.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Fib.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Fib.java
@@ -11,8 +11,13 @@ public interface Fib extends Serializable {
   @Nonnull
   Map<AbstractRoute, Map<String, Map<Ip, Set<AbstractRoute>>>> getNextHopInterfaces();
 
-  /** Set of interfaces used to forward traffic destined to this IP. */
+  /**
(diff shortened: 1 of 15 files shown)


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/traceroute/TracePipelineMediator.java, line 7 at r2](https://reviewable.io/reviews/batfish/batfish/3242#-LZem3SLBAII9N8qvgCP:-LZfWVw467agGfksQjHF:bwgabpk) ([raw file](https://github.com/batfish/batfish/blob/259d9808f7e7de7180961ba2b9626ba2ce08bac2/projects/batfish/src/main/java/org/batfish/dataplane/traceroute/TracePipelineMediator.java#L7)):*

It's not yet in doc form :/
</blockquote></details>

https://docs.google.com/document/d/1mZS03ZL8iXyu-IJlQwlwStCC2cxa3ZgZfreOK_6GMJA/

progwriter:
I'll break it up into smaller changes

PR #3246, opened 2019-02-25
F5 BIG-IP DNAT Implementation and Tests

arifogel:


projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixIpSpace.java
```diff
@@ -15,9 +15,11 @@
   private static final long serialVersionUID = 1L;
 
   public static IpSpace of(Prefix prefix) {
```
dhalperi: ? why is this in PR?

PR #3254, opened 2019-02-26
Start nixing old specifiers

ratulm:
By old, I mean really old, that is, TYPE:foo, ROLE:bar etc. ones.


ratulm:
This has turned out to be an uglier task that I imagined. Will do it piecemeal.

PR #3285, opened 2019-02-28
migrate a bunch of related property questions

ratulm:
from the old node**s**Specifier to the grammar based specifiers


ratulm:
Why must I push? :)

https://github.com/batfish/batfish/blob/639fa714fc450065a94406d8729775d253b05968/projects/batfish-common-protocol/src/main/java/org/batfish/common/Answerer.java#L31

ratulm:
Yeah, that is probably a good way to go. In the end, others just want the core implementation. Question is a just a handy way to pass parameters (and share defaults in some cases). But we can move piecemeal to that model.

PR #3293, opened 2019-02-28
Remove annotations from IpWildcard

progwriter:
Because CiscoConversions does most bizarre casts and it results in
runtime nulls where there shouldn't be any. The horror.


progwriter:
Closing until we track down root cause of nulls popping up where they shouldn't

PR #3299, opened 2019-03-01
Prefix: speed up hashcode

progwriter:


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java, line 194 at r1](https://reviewable.io/reviews/batfish/batfish/3299#-LZqw5N-BXniWYeJZyhV:-LZqw5N03FffALzPKgtS:b-1ope5) ([raw file](https://github.com/batfish/batfish/blob/375c858e0e09e741e50b38e17ad1cf619a69890c/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java#L194)):*
> ```Java
>   @Override
>   public int hashCode() {
>     if (_hashcode == 0) {
> ```

The standard pattern is a bit different, since accessing volatiles is expensive:

```
int hashCode = _hashCode;
if (hashCode == 0) {
    hashCode = // recompute;
    _hashCode = hashCode;
}
return hashCode;
```

this way you get at most 1 read and at most 1 write

progwriter:
Will re-open once I figure out ref changing issues.

PR #3301, opened 2019-03-01
static function to return simplifed IpSpace from Prefix

arifogel:


dhalperi:
of course ship it first and fix later is fine, as it's blocking other work

PR #3319, opened 2019-03-02
Avoid creating array of generics via varargs function

arifogel:
- technically the tests are now more restrictive
- Fixes several eclipse warnings
- org.hamcrest.Matchers#contains really should be annotated with
  @SafeVarArgs. If there is a way to mix in that annotation, it
  may be a better solution.


arifogel:
https://github.com/hamcrest/JavaHamcrest/issues/83
So my read of the conversation in the above issue is that they accidentally did not add `@SafeVarargs` to the `contains` method in question for Hamcrest 2.
Unfortunately the only way to suppress these warnings in Eclipse is to ignore all things suppressed by `@SuppressWarnings("unchecked")`, which includes obviously unsafe things like:
```
    List<?> list = ImmutableList.of("foo");
    List<Integer> iList = (List<Integer>) list;
    Integer.hashCode(iList.iterator().next());
```
I concede that that the diff presented is not ideal, so I'll just close this for now.

Perhaps this will eventually be fixed by the maintainers.

PR #3324, opened 2019-03-04
Do not limit interfaces to a VRF when computing dispositions

progwriter:
Since VRF leaking + PBR there is no longer an invariant that a flow exits out of the same VRF as it arrived on (or the lookup was done in).


progwriter:
No longer convinced that this is the right thing to do to get desired behavior in reachability.

PR #3326, opened 2019-03-04
BgpRib: fewer needless copies

dhalperi:
~10m->~2m in one use case


dhalperi:
This is just dumb. I'm dumb.

PR #3333, opened 2019-03-05
Cisco/Arista ACLs: support more than one 'neq' port statement

progwriter:
Valid syntax (see [link](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/sec_data_acl/configuration/xe-3s/sec-data-acl-xe-3s-book/sec-create-ip-al-filter.html#GUID-6510AFE4-B7E8-4BBB-92C4-1413C210C77D) for examples)


progwriter:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_common.g4, line 296 at r1](https://reviewable.io/reviews/batfish/batfish/3333#-L_EQHLFBZ5a4L3E-1BM:-L_ESULrAMVDxVvZpVrm:b-v2hm9k) ([raw file](https://github.com/batfish/batfish/blob/5bedec73cc1c900073eecaf65e6286996685f127/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_common.g4#L296)):*

Do you enforce that you don't have both `NEQ` and `EQ` on a line?
Should you? What would happen to `args` if you don't?
</blockquote></details>

I thought top-level ORing wold take care of that enforcement, no?

PR #3349, opened 2019-03-06
Yifei nat ruleset

yifeiyuan:
In NAT, if a packet matches a rule set but no rules in the rule set match the packet, there will be no NAT for the packet (see https://www.inetzero.com/closer-look-at-static-nat/). Previously we applied the second rule set in such case. This PR fixes this bug.


yifeiyuan:
Verified on GNS3: if no rule in a matching rule set matches the packet, rules in low-priority rule-sets will be checked and applied if matched.

PR #3427, opened 2019-03-19
prepend as-path on export only after export policy is applied

arifogel:


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 35 at r1](https://reviewable.io/reviews/batfish/batfish/3427#-LaN91l5Dn7CEQ37HQIo:-LaN91l5Dn7CEQ37HQIp:bzbkmq5) ([raw file](https://github.com/batfish/batfish/blob/89141f049ad88ccb87de11614aad6dc4f3ce0b17/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L35)):*
> ```Java
>   /**
>    * Perform BGP export transformations on a given route when sending an advertisement from {@code
>    * fromNeighbor} to {@code toNeighbor} before export policy as applied.
> ```

is applied?

PR #3436, opened 2019-03-20
Handling SYN-ACK for bidirectional traceroute

haverma:


progwriter:
___
*[projects/question/src/main/java/org/batfish/question/traceroute/TracerouteAnswererHelper.java, line 160 at r1](https://reviewable.io/reviews/batfish/batfish/3436#-LaSA9kS72yb1-fnizvA:-LaSC2ZR4EHj3oSEI4__:b-21h6ix) ([raw file](https://github.com/batfish/batfish/blob/c6cd30e3d7bff9c6fb262cb52959bc2ebab83ec4/projects/question/src/main/java/org/batfish/question/traceroute/TracerouteAnswererHelper.java#L160)):*

we only want to do this if the user didn't specify flags explicitly, right? Maybe @progwriter has the right idea for how to do this...
</blockquote></details>

yep, just add `constraints.getTcpFlags() == null` with a comment explaining what's happening.

PR #3457, opened 2019-03-22
Fix OSPF redistribution in SMT formulation

agember:
OSPF redistribution variables should only be included in the SMT formulation if OSPF EXPORT variables exist for at least one interface. If OSPF does not export on any interfaces, then the redistribution variables for that interface will not appear in any predicate. Consequently, the solver will produce counterexamples with arbitrary values for the redistribution variables.


dhalperi:
Thanks! 1 build warning to fix: https://buildkite.com/batfish/batfish-pre-commit/builds/3897#c366fb28-880e-435c-9168-5949e1f68e59

PR #3462, opened 2019-03-22
Move computeBidirectionalTraces to batfish-common-protocol

haverma:
- since this logic may be needed also outside `BidirectionalTracerouteAnswerer`


haverma:
as per the offline discussion with @dhalperi, we should not move this in batfish-common-protocol as this can be done with minimal code wherever it is needed. Main reason is anything related to dataplane should not go in `batfish-common-protocol`

PR #3471, opened 2019-03-25
Make BGP Topology utils use Bidir traceroute

haverma:


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3471#-LarGeUIBLa2pbd24KUK:-LarGeUIBLa2pbd24KUL:bccwpuk) (no related file):*
Per discussion, simplify more and rename function/vars to be BGP specific. Maybe, "canInitiateBgpSession"?

---

PR #3472, opened 2019-03-26
BDD: deploy new ops everywhere

dhalperi:


dhalperi:
@anothermattbrown I am not at all wedded to any specific change in here, just a PR to see what it would look like. Specifically, I find `a.diff(b)` a readability improvement in many cases.

projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDIpSpaceSpecializer.java
```diff
@@ -62,8 +62,7 @@ public static BDDIpSpaceSpecializer specializeByDstIp(
 
   @Override
   protected Optional<IpSpaceSpecializer> restrictSpecializerToBlacklist(Set<IpWildcard> blacklist) {
-    BDD refinedBDD =
-        blacklist.stream().map(_ipSpaceToBDD::toBDD).map(BDD::not).reduce(_bdd, BDD::and);
+    BDD refinedBDD = blacklist.stream().map(_ipSpaceToBDD::toBDD).reduce(_bdd, BDD::diff);
```
dhalperi: plausibly backwards?

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDIpSpaceSpecializer.java, line 65 at r1](https://reviewable.io/reviews/batfish/batfish/3472#-LarUL8CN4X79FEZdljK-r1-65:-LbP7zgk-3auKwVsEG-K:bf0upus) ([raw file](https://github.com/batfish/batfish/blob/094cf34b2cde3fc9d59126c80da111580e591e26/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDIpSpaceSpecializer.java#L65)):*

Seems right -- `[a,b,c]` will reduce to `_bdd.diff(a).diff(b).diff(c)`. However, `diff` isn't associative, so it may be better to convert this to a loop.
</blockquote></details>

Done. Also added short-circuiting.

dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3472#-LbPD4_3C_mnxiYfq_AH:-LbPD4_3C_mnxiYfq_AI:b-v1spen) (no related file):*
Working on getting numbers

---

PR #3475, opened 2019-03-26
differential reachability: remove IPs of inactive devices from ACCEPT

anothermattbrown:


projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java
```diff
@@ -37,4 +47,27 @@
     return toImmutableMap(
         reverseEdges, Entry::getKey, entry -> ImmutableMap.copyOf(entry.getValue()));
   }
+
+  public static Map<String, Map<String, IpSpace>> computeIpsNotOwnedByInactiveInterfaces(
```
dhalperi: topology utils?

anothermattbrown:
Going another way with this.

PR #3482, opened 2019-03-26
SpecifiersReachability: disallow running in differential mode and refer to better alternatives

dhalperi:


dhalperi:
___
*[projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersReachabilityAnswerer.java, line 37 at r1](https://reviewable.io/reviews/batfish/batfish/3482#-LaxEM0bDRZCRyFvr0ag:-LbPjan3BbnQqVrUDPpa:bbi8jk1) ([raw file](https://github.com/batfish/batfish/blob/f0fbcdc502eec6c28640069a201b9c6ea1647142/projects/question/src/main/java/org/batfish/question/specifiers/SpecifiersReachabilityAnswerer.java#L37)):*

Thinking that it never hurts to point users at multiple questions they might use to analyze data plane differences.

But no strong opinion here.
</blockquote></details>

@ratulm - any pref?

PR #3490, opened 2019-03-27
BDDReachability: use Table instead of nested Maps for transitions

dhalperi:
Implemented the same under the hood.
Fewer copies.
Better interface.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysis.java
+++ b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysis.java
@@ -1,13 +1,14 @@
 package org.batfish.bddreachability;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static org.batfish.bddreachability.BDDReachabilityUtils.computeForwardEdgeMap;
-import static org.batfish.bddreachability.BDDReachabilityUtils.computeReverseEdgeMap;
+import static org.batfish.bddreachability.BDDReachabilityUtils.computeForwardEdgeTable;
 import static org.batfish.common.util.CommonUtil.toImmutableMap;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Table;
+import com.google.common.collect.Tables;
 import io.opentracing.ActiveSpan;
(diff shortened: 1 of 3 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/transition/Zero.java, line 28 at r1](https://reviewable.io/reviews/batfish/batfish/3490#-Lb-tak42jazzzFdOIoI:-Lb-wHkAB9ITQlPrDFxn:b-81taoq) ([raw file](https://github.com/batfish/batfish/blob/f9b68d9a454d2c727d7279f67f798873e34de9ab/projects/batfish/src/main/java/org/batfish/bddreachability/transition/Zero.java#L28)):*

need to either revert this and get rid of the distinct, or implement this for all other overriders.
</blockquote></details>

reverted.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/3490#-Lb-tMvI6FfDalxP0wcF:-Lb03yrl0xwyYhxCheE2:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f9b68d9a454d2c727d7279f67f798873e34de9ab/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L15)):*

Seems like parallel edges should be allowed here, though it looks like maybe they currently aren't (only one is kept?). We can combine parallel edges using the `Or` transition. Then duplicates would have no logically affect (only a performance one).  The `Transitions.or` factory method should remove duplicates and flatten nested `Or`s too.
</blockquote></details>

Done.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/3490#-Lb-tMvI6FfDalxP0wcF:-Lb0J1Aq4TO4PKraqCJA:b-2v20qk) ([raw file](https://github.com/batfish/batfish/blob/f9b68d9a454d2c727d7279f67f798873e34de9ab/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L15)):*

should we update `Transitions.or` in this PR too or separately?
</blockquote></details>

Oh I had read that as you telling me that `Transitions.or` would handle everything for me :p rather than a request to change `Transitions.or`. I think that's a separate PR?

PR #3503, opened 2019-03-29
JFactory: some more BDD optimizations

dhalperi:
I ran several experiments on various networks, mild savings on small work (computing forwarding analysis), seems larger savings on big work -- but the big work is not deterministic so I'm not confident in it.


dhalperi:
I should really separate the style and math. Closing for now.

PR #3511, opened 2019-03-31
BDD: merge ITE and APPLY caches

dhalperi:
We don't use ITE enough for it to merit its own cache. Instead, merge them
together and double the cache size.

Running detectLoops on a large network, this change did not affect results at all
compared to just doubling the cache size but keeping both.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3511#-LbKtHVvD6GUP_BklV4M:-LbKtHVw0FMNlxWsWqYv:b-z5bna6) (no related file):*
Working: After #3505 .

---

PR #3515, opened 2019-04-01
BDD: more deployment of sophisticated operators

dhalperi:
* 12% fewer BDDs created during FA
* 15% fewer BDDs created including detectLoops

Wow, making those `not` BDDs just to `and` them out can be expensive.


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/z3/BDDIpAccessListSpecializer.java, line 159 at r1](https://reviewable.io/reviews/batfish/batfish/3515#-LbPTlzJB5xV86qvDNk1:-LbPW1Wn1BGUtb2wYjxE:b-g6pl6o) ([raw file](https://github.com/batfish/batfish/blob/bcc39b507e56fabe07d2975c8fe2aac34a7ace29/projects/batfish/src/main/java/org/batfish/z3/BDDIpAccessListSpecializer.java#L159)):*

* a implies not b
* equiv to: not a or not b
* equiv to: not (a and b)


* not (a and b) is One
* equiv to: (a and b) is Zero
</blockquote></details>

(And also, the code seems much clearer now. If the AND is empty, FalseExpr.)

PR #3522, opened 2019-04-01
Make hostname in Isp conf case-insensitive

haverma:
Batfish preserves the Interface name case so that is still case sensitive


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/ModelingUtils.java, line 133 at r1](https://reviewable.io/reviews/batfish/batfish/3522#-LbQE6_21m99exMbYp68:-LbQE_U44Z5gUO2fEn0O:bc7ezfh) ([raw file](https://github.com/batfish/batfish/blob/c3b19e701dddd6e19ae076e0d9767f0b17c9c250/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/ModelingUtils.java#L133)):*

here is where we'd warn?
</blockquote></details>

(could be separate PR, of course)

PR #3534, opened 2019-04-02
Add interface for RoutingProcess

progwriter:
Rationale: https://docs.google.com/document/d/1DH36bHoFobFWP5Xv_RFpbvj5lRvQQdYC4duo4UbiH3I/


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/RoutingProcess.java, line 28 at r1](https://reviewable.io/reviews/batfish/batfish/3534#-LbUB3XF-kVKjHg8N2td:-LbUBUkG1apinph9mpug:bkp1dn2) ([raw file](https://github.com/batfish/batfish/blob/fa09af82d113be37160c5c4545824a6441b924c0/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/RoutingProcess.java#L28)):*

I assume we can change this as we go forward.

Wondering if this generic "update topology" is a good API, or whether we'll end up sending topology deltas, etc.
</blockquote></details>

Definitely modifiable. Currently we don't have good notion of topology deltas and the processes would need to be careful not to destroy existing queues, etc. But figured better to have a placeholder rather than not think about this requirement at all.

PR #3564, opened 2019-04-05
Cleanup juniper format detection

dhalperi:
1. Recognize newer RANCID which uses vendor-specific comments.
2. Relax _firstChar=# to mean juniper -- too broad.
3. Update tests to support recognition.


dhalperi:
Not very juniper specific, and most of our test configs won't have it.

We had a surprising number of tests that were just getting away with poor
detection because of the '#' checl.

PR #3611, opened 2019-04-15
OSPF distribute list VI 

haverma:
- done only for distribute-lists using prefix-lists (considering the common use)


progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java, line 1748 at r1](https://reviewable.io/reviews/batfish/batfish/3611#-LcYGGGI1pKNLPLJUu7w:-LcYPsrr8oPtjY-LrVWU:bqta3z4) ([raw file](https://github.com/batfish/batfish/blob/5cf8edca446fd08275785db48e47f6646d1a3a16/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java#L1748)):*

Why is this 
</blockquote></details>

sorry, typo: why is this nonNull when the fields is nullable?

PR #3635, opened 2019-04-17
F5 BIG-IP imish fix bgp peer-group assignment

arifogel:
- allow assignment of address-based peer to peer-group without first specifying remote-as

### DIFF
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/MapMatchers.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/MapMatchers.java
@@ -1,0 +1,57 @@
+package org.batfish.datamodel.matchers;
+
+import static org.hamcrest.Matchers.containsInAnyOrder;
+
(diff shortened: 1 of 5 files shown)


progwriter:
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/MapMatchers.java, line 16 at r2](https://reviewable.io/reviews/batfish/batfish/3635#-LchISYjBmTDxsMSAUWT:-LchJSAeEPBjOqqR6Ldo:bt65la7) ([raw file](https://github.com/batfish/batfish/blob/a09b06407f5c623bc7227ab7e3e3c61b9a1b47bf/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/MapMatchers.java#L16)):*

Why create these matchers rather than asserting on `map.keySet()` or `map.values()`, as we have everywhere else?
</blockquote></details>

Also, http://hamcrest.org/JavaHamcrest/javadoc/2.0.0.0/org/hamcrest/collection/IsMapContaining.html

PR #3650, opened 2019-04-19
Restore l2 topology to the edges question

anothermattbrown:


anothermattbrown:
Closing in favor of a more scalable way of querying the l2 topology

PR #3669, opened 2019-04-20
Intern OSPF routes

progwriter:
This greatly reduces memory pressure when computing dataplane on certain networks


progwriter:
Closing for now, does not provide a substantial benefit over #3672 on available data.

PR #3671, opened 2019-04-20
F5 BIG-IP structured parsing overhaul

arifogel:
- much smarter lexer and parser
- remove unnecessary validation


arifogel:
After #3660

PR #3695, opened 2019-04-24
Include L3 adjacency even if interface is not active

ruchitshrestha:
If an interface is inactive, L3 adjacency is not created. This causes the SMT model to assume neighbor must be external to the network which is an incorrect assumption. The attached configs demonstrate a simple network with no external neighbors and a disabled interface where this improper assumption is made.

[configs.zip](https://github.com/batfish/batfish/files/3113403/configs.zip)

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java
@@ -619,7 +619,7 @@ public final class TopologyUtil {
     configurations.forEach(
         (nodeName, node) -> {
           for (Interface iface : node.getAllInterfaces().values()) {
-            if (iface.isLoopback(node.getConfigurationFormat()) || !iface.getActive()) {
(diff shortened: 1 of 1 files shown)


dhalperi:
I suspect you want rawLayer3Topology

dhalperi:
--- a/projects/batfish/src/main/java/org/batfish/symbolic/Graph.java
+++ b/projects/batfish/src/main/java/org/batfish/symbolic/Graph.java
@@ -179,7 +179,8 @@ public class Graph {
     }
     _routers = _configurations.keySet();

-    Topology topology = _batfish.getEnvironmentTopology();
+    Topology topology =
+
_batfish.getTopologyProvider().getRawLayer3Topology(_batfish.getNetworkSnapshot());

     // Remove the routers we don't want to model
     if (routers != null) {

dhalperi:
Though it will obviously be up to you at that point to figure out how to
properly remove inactive interfaces wherever they're not supposed to be
there in later computations?

dhalperi:
Did using the Raw topology work for you? If so, please send it along.

I think we have consensus that this isn't the right approach right now. Thanks!

PR #3698, opened 2019-04-24
Remove TODO after confirming behavior

corinaminer:
[Confirmed](https://docs.google.com/document/d/1KFCT8ckUYOza4dVFMQ87XutsQWDKT_5UONngfA-eXJw/edit#heading=h.eqvs8hr9omtl) that if the route-map is undefined here, the default route is exported as if no route-map were configured.


dhalperi:
On which vendors? :p

Is there a test?

PR #3708, opened 2019-04-25
move most of z3 package to common

anothermattbrown:
Only BatfishJob stuff needs to be in batfish.

This is a first step in moving the reachability engines to common. This moves most of the NOD engine, as well as some stuff that the BDD engine depends on.

I'd like to rename this package to `org.batfish.reachability` (on the batfish side) and `org.batfish.common.reachability` (on the common side). Should that be here or in a separate PR?


dhalperi:
I'd like to do this differently. Let's chat?

PR #3736, opened 2019-04-30
IpSpaceSpecializer: delete abstract and impls

dhalperi:


dhalperi:
Odds this is likely to be used in future?

PR #3738, opened 2019-04-30
Configuration: move finalization from VS to CCJ

dhalperi:
Extract the simplification/finalization logic from Cisco and move it to ConvertConfigurationJob
so that it applies to all vendors.


dhalperi:
Opinions on the ref diffs? (Which all look correct.)

PR #3745, opened 2019-04-30
Plumb TopologyContext through iBDP

arifogel:


progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java, line 96 at r2](https://reviewable.io/reviews/batfish/batfish/3745#-LdoT7op8z7FR6db_VUR:-LdoT7op8z7FR6db_VUS:bbejpj5) ([raw file](https://github.com/batfish/batfish/blob/08b4cc90fcf7d3777ae07901e7c15809d85d1302/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java#L96)):*
> ```Java
>       boolean checkReachability,
>       @Nullable TracerouteEngine tracerouteEngine,
>       @Nonnull Layer2Topology layer2Topology) {
> ```

This is contradictory with the code below that does a null check. please resolve appropriately.

PR #3766, opened 2019-05-01
Misc simplifications

dhalperi:
* Shortcut expensive comparator for self-comparison.
* `Objects.equals(NotNull, NotNull)` -> `a.equals(b)`
* Box less


dhalperi:
Last one for now, all in separate commits.

PR #3771, opened 2019-05-02
BDDInteger: add range helper and use it

dhalperi:
Adds `bddint.range(a, b)` which is equivalent to `bddint.geq(a).and(bddint.leq(b))`.

Internally, there are two differences:
    1. Short-circuit `a == b`
    2. Save work in the case where `a` and `b` have a common prefix.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDFlowConstraintGenerator.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDFlowConstraintGenerator.java
@@ -90,11 +90,7 @@ public final class BDDFlowConstraintGenerator {
   BDD computeUDPConstraint() {
     BDDInteger dstPort = _bddPacket.getDstPort();
     BDDInteger srcPort = _bddPacket.getSrcPort();
-    BDD bdd1 =
-        dstPort
-            .geq(33434)
-            .and(dstPort.leq(33534))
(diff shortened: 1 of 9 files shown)


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3771#-LdrQWMY5K1MJH-6tCil:-LdrQWMY5K1MJH-6tCim:b-evpb04) (no related file):*
Working: After #3769 

---

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDInteger.java, line 228 at r1](https://reviewable.io/reviews/batfish/batfish/3771#-LdtLvtzEIAUbBrn0YVn:-LdtVA8J9fgEpdMVHww3:bxn4uhi) ([raw file](https://github.com/batfish/batfish/blob/2aafd13ff9555c1a8efa35ca43b274f56947cf91/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDInteger.java#L228)):*

this last loop looks like it could be factored out and shared with `value` -- i.e. `valueN`
</blockquote></details>

I thought about this, but decided it was probably confusing. Note that geqN,leqN mean the `N` LSBs (from `true`), but here we are `and`ing together the `N` MSBs (`from a partial accumulator).

(If you feel strongly, this is sensible - we should make all the three helpers take start idx, end idx, and base BDD. Then we can implement gt, geq, lt, leq, value, range, etc all with simple recursion around the three wrappers.)

PR #3797, opened 2019-05-06
Unnumbered BGP in session questions

corinaminer:
- Adds remote interface column to answers
- Local and remote interface columns are only populated for rows representing BGP unnumbered peers; local and remote IP columns are only populated for rows representing standard BGP peers (active or dynamic)
- New session types `EBGP_UNNUMBERED` and `IBGP_UNNUMBERED` to help tell BGP unnumbered rows apart from misconfigured standard peers


progwriter:
___
*[projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionCompatibilityAnswerer.java, line 90 at r2](https://reviewable.io/reviews/batfish/batfish/3797#-LeDqPAS2lYAvZpEUeEd:-LeEMah36yBJ5WOclpXi:bswha5q) ([raw file](https://github.com/batfish/batfish/blob/16445c2de861100f36d0310200f182673277b73b/projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionCompatibilityAnswerer.java#L90)):*
I was referring to the calls to 
> ```
> BgpTopologyUtils.initBgpTopology
> ```
seemingly all over the place. If you don't init the bgp topology, you don't need the l2 topology.

PR #3799, opened 2019-05-06
Multiple OSPF processes in VI datamodel

corinaminer:
May require some follow-up if there are OSPF questions whose answers don't currently include OSPF process ID. Such questions should still work fine, but it may be ambiguous which OSPF process is associated with a given row of an answer.


tests/parsing-tests/unit-tests-nodes.ref
```diff
@@ -11141,8 +11145,9 @@
             ],
             "mtu" : 1500,
             "ospfArea" : 0,
+            "ospfCost" : 10,
             "ospfDeadInterval" : 0,
-            "ospfEnabled" : false,
+            "ospfEnabled" : true,
```
haverma: Was this change expected ? or the diff is messed up ?

PR #3801, opened 2019-05-06
Bazel: sickbay javadoc for real

dhalperi:
Use the `manual` tag to make this target only build if you invoke it directly. Gets around the
issue until Bazel has better java11 support.


dhalperi:
From https://docs.bazel.build/versions/master/guide.html

> Targets with tags = ["manual"] will not be included in wildcard target patterns ( ..., :*, :all, etc.). You should specify such test targets with explicit target patterns on the command line if you want Bazel to build/test them.

PR #3811, opened 2019-05-07
Use initial layer 2 topology before dataplane generation

corinaminer:


progwriter:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3811#-LeJdFTpAM0dGHCwbECt:-LeJdFTpAM0dGHCwbECu:bhz2w11) (no related file):*
would be great to have a test (even if in followup pr)

---

PR #3817, opened 2019-05-08
Remove unused data plane repair code paths

arifogel:


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 1529 at r1](https://reviewable.io/reviews/batfish/batfish/3817#-LeNvLIVAJ4IyxSWb_eH:-LeNvLIVAJ4IyxSWb_eI:b-ie98u3) ([raw file](https://github.com/batfish/batfish/blob/b5a9dded8b68a43b62f50bdbbfbce0edadffad97/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L1529)):*
> ```Java
>          * repaired, so we still might need to load it from disk.
>          */
>         loadDataPlaneAnswerElement(false);
> ```

the comment and the `false` don't line up here...

PR #3826, opened 2019-05-09
use annotations to generate property descriptors

ratulm:
It is not a complete PR but I am curious if you have any thoughts/objections on this way of doing things: 
1/ Generate the property descriptor map using annotations on methods instead of separate listings
2/ Extend the PropertyDescriptor to include a description, which can then be embedded into the column metadata. 

The main motivation is to be able to attach useful descriptions, and I wanted to avoid doing so in the out of band map as we were doing thus far.


dhalperi:
How would we add properties that are not a simple `getFoo` on the class itself?

ratulm:
The same way as today with property specifier doing something special for
them. The interface outside of property specifiers is not changing; I am
only changing how the property specifier populates its map from property
name to property descriptor (and adding a field to the property
descriptor).

ratulm:
Closing this PR. I'll send a simpler one that does not use annotations but adds descriptions to the property specifier itself.

PR #3837, opened 2019-05-10
don't completely bail when one template fails

ratulm:


dhalperi:
___
*[projects/batfish-client/src/main/java/org/batfish/client/Client.java, line 1952 at r1](https://reviewable.io/reviews/batfish/batfish/3837#-LeU3_mJ5rXTml0Bpzuy:-LeU3_mJ5rXTml0Bpzuz:b-ghmm6w) ([raw file](https://github.com/batfish/batfish/blob/df62008509b04a6f29d853c2b3137ff0eaf206e1/projects/batfish-client/src/main/java/org/batfish/client/Client.java#L1952)):*
> ```Java
>             getQuestionName(questionJSON, jsonQuestionFile.toString()), questionJSON.toString());
>       } catch (Exception e) {
>         if (logger != null) {
> ```

This is a feature. Client is only used for testing, and I want the client to fail on anything fishy like this.

Why this change?

ratulm:
I had one wrong template but all tests failed. From the output, it was hard
to tell which one failed. With this change, only the test with bad template
fails and I could tell which one that was.

dhalperi:
Didn't the python tests of individual files tell you which file was bad?

ratulm:
I didn't run that one then. When I run it now, it does not catch the bug I
had -- repeated key in the JSON. In general too, the Java stuff will catch
different types of errors (e.g., wrong class name). It made sense for only
tests dependent on a template to fail, rather than all of them.

ratulm:
:)

PR #3838, opened 2019-05-10
Remove blacklist getters from IBatfish

arifogel:
- also remove `getEnvironment()` in prep more for blacklist culling


arifogel:
Going a different route

PR #3863, opened 2019-05-14
RibDelta.Builder: add isEmpty

dhalperi:
And use it instead of .build().isEmpty()

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/OspfRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/OspfRoutingProcess.java
@@ -306,7 +306,7 @@ final class OspfRoutingProcess implements RoutingProcess<OspfTopology, OspfRoute
 
   @Override
   public boolean isDirty() {
-    return !_changeset.build().isEmpty()
+    return !_changeset.isEmpty()
         || !_queuedForRedistribution.isEmpty()
         || !_activatedGeneratedRoutes.isEmpty()
         || !_interAreaIncomingRoutes.values().stream().allMatch(Queue::isEmpty)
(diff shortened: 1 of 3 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/RibDelta.java, line 64 at r1](https://reviewable.io/reviews/batfish/batfish/3863#-LeoGInr8DtNw5_c8JAi:-LeoGInr8DtNw5_c8JAj:bavjj7k) ([raw file](https://github.com/batfish/batfish/blob/7d38288d9a7c30124d5c385af975878bb8c8a18b/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibDelta.java#L64)):*
> ```Java
>   /** Check whether this delta is empty (has no outstanding actions) */
>   public boolean isEmpty() {
>     return _actions.isEmpty() || _actions.values().stream().allMatch(List::isEmpty);
> ```

Q: can the RHS be true with the LHS true? For RibDelta?

(for the Builder, I convinced myself it can't be.)

progwriter:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/RibDelta.java, line 64 at r1](https://reviewable.io/reviews/batfish/batfish/3863#-LeoGInr8DtNw5_c8JAi:-LeoGi5S0j-JFDmXj2Ll:b-y2w6q2) ([raw file](https://github.com/batfish/batfish/blob/7d38288d9a7c30124d5c385af975878bb8c8a18b/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibDelta.java#L64)):*

Q: can the RHS be true with the LHS true? For RibDelta?

(for the Builder, I convinced myself it can't be.)
</blockquote></details>

I think the same holds for RibDelta, they are now only constructed by `builder.build()`. (used to be long time ago constructor was public, but we fixed that).

PR #3883, opened 2019-05-15
Cumulus: convert VXLANs

progwriter:
Convert VXLANs to per-VRF VniSettings


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusNcluConfiguration.java, line 552 at r1](https://reviewable.io/reviews/batfish/batfish/3883#-Ley10KRCVDerVrwDpG4:-Lf0fwpJ-uCiMPFO4h6t:b-4jvvji) ([raw file](https://github.com/batfish/batfish/blob/e5b696a11eb390c81b947f3c92da6cf4dc587364/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusNcluConfiguration.java#L552)):*

(can defer to follow-on PR)
</blockquote></details>

Let's defer. Unclear yet if you can combine the two

PR #3890, opened 2019-05-16
demux edges question

ratulm:


progwriter:
___
*[questions/stable/edgesLayer1.json, line 5 at r2](https://reviewable.io/reviews/batfish/batfish/3890#-Lf0YRJFEzQRLAYGna2W:-Lf0uoCe3LorBkwaYoDM:bgge5px) ([raw file](https://github.com/batfish/batfish/blob/ad5f9eb80068c4225308883b82308a21b221f9c2/questions/stable/edgesLayer1.json#L5)):*

was going for something that people use more commonly. what sounds better?
  - layer 1 links
  - layer 1 edges
  - layer 1 adjacencies

the options above are in google order.

</blockquote></details>

In my mind (take with a grain of salt):

* `edges` sounds most natural (and most general). 
* For layer1, `links` makes sense. 
* For l2/l3 `links` does not make sense (far too physical-layer-ish).

PR #3891, opened 2019-05-16
BDD: merge isolated paths in the graph

dhalperi:
For one large network:

Initial size: 868895
Final size: 416566


dhalperi:
Holding off until after we refactor the graph setup and use code.

PR #3900, opened 2019-05-17
Move nullsFirst comparator to right place

corinaminer:
The thing that can be null is `Pair::getSecond`, not `rhs`.


dhalperi:
Can't we just get rid of Pair? There have to be like 5 of them left...

PR #3908, opened 2019-05-20
Edges: move EdgeType internally, proper maintenance of supported values

dhalperi:
EdgeType is only used for Edges question, move it there.
Second, delete the unsupported Layer2 and RIP options.
Then fix the bugs that proper maintenance of EdgeType has identified.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/3908#-LfM34oQCy8M07PkNs7U:-LfM34oR4T1KR4YXjqUI:b-307rue) (no related file):*
Suspect merge race broken now.

---

PR #3929, opened 2019-05-22
Optimizations for Branch transition

anothermattbrown:


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/transition/Transitions.java, line 36 at r2](https://reviewable.io/reviews/batfish/batfish/3929#-LfXLPGE52cQMH8EjKib:-LfXLPGF3-yH4LK3XT8_:b-3j96gg) ([raw file](https://github.com/batfish/batfish/blob/174288762d0768626debc4c149cd324f5d1f1b54/projects/batfish/src/main/java/org/batfish/bddreachability/transition/Transitions.java#L36)):*
> ```
> new Branch
> ```

@anothermattbrown - Actually, should these be recursive calls to this function instead?

PR #3943, opened 2019-05-24
optimize Layer2Topology computation

anothermattbrown:
Don't accumulate edges into a set. Instead create a Layer2Topology
builder that alllows each edge to be GCd after being processed.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Node.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Node.java
@@ -47,6 +47,12 @@ public final class Layer2Node implements Comparable<Layer2Node>, Serializable {
   @Override
   public int compareTo(Layer2Node o) {
     // compareTo is hot for some networks, so inlining the comparator is worth it
+    if (this == o) {
+      return 0;
+    }
+    if (o == null) {
+      return 1;
+    }
     int r = _hostname.compareTo(o._hostname);
     if (r != 0) {
       return r;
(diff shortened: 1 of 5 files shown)


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Node.java, line 55 at r2](https://reviewable.io/reviews/batfish/batfish/3943#-LffF0JND2oWiOwugCbQ:-LffF0JND2oWiOwugCbR:b-7h4ssk) ([raw file](https://github.com/batfish/batfish/blob/ee4c88c1506b96fc3ed14ce7e84f0175147848cf/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Node.java#L55)):*
> ```Java
>     if (o == null) {
>       return 1;
>     }
> ```

@anothermattbrown. @progwriter : compareTo guarantees `@Nonnull`

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Topology.java, line 43 at r2](https://reviewable.io/reviews/batfish/batfish/3943#-LffFL8T9kpk_LNlRUgz:-LffFL8T9kpk_LNlRUh-:brd93rx) ([raw file](https://github.com/batfish/batfish/blob/ee4c88c1506b96fc3ed14ce7e84f0175147848cf/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/Layer2Topology.java#L43)):*
> ```Java
> 
>     private void addNodeIfMissing(Layer2Node node) {
>       if (!_nodes.contains(node)) {
> ```

this would be much faster:

```
if (_nodes.add(node)) {
    _broadcastDomains.addElement(node);
}
```

PR #3950, opened 2019-05-25
PaloAlto: extract tunnel and tunnel units

dhalperi:


dhalperi:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4, line 122 at r2](https://reviewable.io/reviews/batfish/batfish/3950#-LfzGmmk2t89YvxXZtq8:-LfzMp2EBKgo-XdjmVvB:b-5svu9i) ([raw file](https://github.com/batfish/batfish/blob/63892de6d443cbce3c954699111ed52ee6627143/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4#L122)):*

In several of the outstanding PRs, 

```
(
   a
)
```

is used as a clear indicator that there will be `| b | c` ...
</blockquote></details>

in this case -- there are tunnel-specific options that will go here.

PR #4026, opened 2019-06-06
PaloAlto: improvements to non-routing interfaces

dhalperi:
We should not deactivate all interfaces that are not in a router; instead, we should
only deactivate L3 interfaces that are not in a router.

Also rename NULL_VRF to NON_ROUTING, which is more accurate.

Also deactivate interfaces that are not in a zone, since that
is also more accurate.


dhalperi:
FYI, there will be further improvements to this logic, but trying to stage them cleanly.

dhalperi:
This is probably overtaken by #4043 . Will reopen as needed.

Issue #4018, opened 2019-06-05
Cisco switchport mode trunk encapsulation


agember:
On a Cisco Catalyst 6500 running IOS 12.2 it is not necessary to include `switchport mode trunk` to put an interface into trunking mode. Including `switchport trunk encapsulation 8021q` also puts an interface in trunking mode. See
https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst6500/ios/12-2SX/configuration/guide/book/layer2.html#42266 and  https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst6500/ios/12-2SX/configuration/guide/book/layer2.html#35989

However, without `switchport mode trunk` no layer-3 adjacencies between VLAN interfaces are detected.

The following configurations (which are based on real configurations) demonstrate the issue.
[layer1_topology.txt](https://github.com/batfish/batfish/files/3258862/layer1_topology.txt)
[r-core1.txt](https://github.com/batfish/batfish/files/3258864/r-core1.txt)
[r-core2.txt](https://github.com/batfish/batfish/files/3258865/r-core2.txt)

ratulm:
Cc: @haverma

haverma:
thanks @agember for reporting this, I will have a look and send out a fix soon.

PR #4028, opened 2019-06-07
Fix checking of interface mode before setting allowed VLANs

agember:
Further fix to #4018


haverma:
hey Aaron, my fix in https://github.com/batfish/batfish/issues/4018 was incorrect, the newer fix in batfish/batfish#4038 should work fine

dhalperi:
Hey @agember - suspect this is fixed now? Reach out if not. Thanks!

PR #4085, opened 2019-06-14
Changes in visibility and public constructors for RoutesQuestion

haverma:
- making column names public
- restricting the answer return type to TableAnswerElement
- adding a public constructor for `Nodes, VRFs` in `RoutesQuestion`


haverma:
may not be needed

PR #4089, opened 2019-06-14
NextVrf support

arifogel:
- add nextVrf to AbstractRoute
- add FibNextVrf FibAction
- add traceroute support


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/dataplane/traceroute/FlowTracer.java, line 494 at r3](https://reviewable.io/reviews/batfish/batfish/4089#-LhamQz2-7dryHeDClkv:-LhamQz2-7dryHeDClkw:b4pelup) ([raw file](https://github.com/batfish/batfish/blob/6409281976b01331d7b49ebe1d5d853c7743e5a0/projects/batfish/src/main/java/org/batfish/dataplane/traceroute/FlowTracer.java#L494)):*
> ```Java
> 
>   @VisibleForTesting
>   void fibLookup(Ip dstIp, String currentNodeName, Fib fib) {
> ```

potential solution: do a "tail-recursive" call where you pass in a stack of inner next-vr only breadcrumbs

PR #4103, opened 2019-06-19
Fix tunnel Interface types 

haverma:
Changes are based on https://docs.google.com/document/d/12nbZh9u88ae0Ed7RzjAhgJfiqpR1QL4tevSdKwHQnvk/edit?usp=sharing

main points: 
1. Detecting IPSec and GRE tunnels as different interface types 
2. removed VPN interface type
3. two interface types for tunnels: IPSEC_TUNNEL and TUNNEL
4. Not detecting tunnel InterfaceType in Cisco on the basis of iface name, but using `Tunnel mode`
5. removed `TUNNEL_INTERFACE_TYPES ` as it will not be needed (we only need `IPSEC_TUNNELS` now )
6. added detection of GRE tunnels in Juniper


haverma:
closing as this is not the correct and complete solution for fixing interface types

PR #4111, opened 2019-06-20
BDD: add andsat

dhalperi:
`a.andsat(b)` is a shortcut for `!a.and(b).isZero()` that does not
create any BDDs on the way down and can early exit the first it finds any
satisfying assignment.


dhalperi:
Hashes collide

dhalperi:
___
*[projects/bdd/src/test/java/net/sf/javabdd/bdd/BasicTests.java, line 335 at r2](https://reviewable.io/reviews/batfish/batfish/4111#-LiA991Z7joCSM5ydlSY:-LiAXx-K8voki-X_zrYl:bk672fg) ([raw file](https://github.com/batfish/batfish/blob/882cf4d99abebac336dfbac7df2795f50755285b/projects/bdd/src/test/java/net/sf/javabdd/bdd/BasicTests.java#L335)):*

Nope, just bad understanding of order of operations
</blockquote></details>

Yes, but also no impact -- it will just do the same call twice. I guess I should just delete the check :p

Issue #4128, opened 2019-06-21
Dependencies should always exist; child interfaces should always have a parent


dhalperi:
See #4124.

1. We don't add dependencies if the parent interface doesn't exist. This seems bad, as the rest of the pipeline don't know there's a missing dependency.
2. In that issue, the config has `Port-Channel1.1` defined, but not `Port-Channel1`. Presumably this is bad configuration, since it seems PC1 must exist for things to be sensible. Maybe we should implicitly create PC1 and similar in like cases.

PR #4129, opened 2019-06-21
Dependency: add even if parent does not exist

dhalperi:
And add comments to document this, and fix some callers/setters.

This is for batfish/batfish#4128.


dhalperi:
Current state:
* issue is still causing bugs and we'll fix it separately.
* Need vendor-specific work in different extractors to make sure we do the right thing.
* Will split that work into multiple PRs, which are coming.

PR #4154, opened 2019-06-26
EdgesQuestion: add initial parameter, default to false

dhalperi:
Users expect to see the post-dataplane topology by default.

Deliberately changing the behavior of existing questions.

Deliberately only exposing initial in `edges` template.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyProvider.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyProvider.java
@@ -53,18 +53,6 @@ public interface TopologyProvider {
   @Nonnull
   OspfTopology getInitialOspfTopology(@Nonnull NetworkSnapshot networkSnapshot);
 
-  /**
-   * Return the raw layer-3 {@link Topology} for a given {@link NetworkSnapshot}. The layer-3
-   * topology is constructed by inferring layer-3 adjacencies via the layer-3 information in the
-   * configurations, and pruning edges whose vertices are not in the same broadcast domain according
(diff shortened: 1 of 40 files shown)


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/4154#-LiKymr9DRBOQmwwX2PD:-LiL0KJnAcJxqs8nQULG:b-cwruon) (no related file):*

So much to test.

* edges answerer seems kind of gross
* ospf topology was not implemented and I made it up
* ipsec topology is not implemented correctly
</blockquote></details>

Why didn't the variable order check fail?

---

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/topology/TopologyProviderImpl.java, line 254 at r2](https://reviewable.io/reviews/batfish/batfish/4154#-LiL2fVd4Wws2pwJ0iBF:-LiLepCi2m-whv1bsPSp:b-35v0l2) ([raw file](https://github.com/batfish/batfish/blob/fbbea3e0281ee25a3819faa6251e3c08df406dbb/projects/batfish/src/main/java/org/batfish/topology/TopologyProviderImpl.java#L254)):*

getLayer3Topology reaches to `storage` to get post-dp topology from disk (same for bgp/vxlan). While thats not supported for ospf yet, just call `initial...` under the hood?
</blockquote></details>

to clarify, this code is NOT identical to `initialOspfTopology`. The difference is which Layer3Topology getter they call.

dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/4154#-LijWMnuDMixHJdGKKm7:-LijWMnuDMixHJdGKKm8:b3e9w7f) (no related file):*
working - self merge

---

PR #4174, opened 2019-07-02
 Cumulus: better support for address families, extract route-reflector-client

progwriter:
Parsing, and extraction. Holding off on conversion because VI model needs to be tweaked a tiny bit more after #4165


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/4174#-LipHM3_8pOqZY6DerLh:-LipHM3a0Wz37WU6n62v:b-rx4me6) (no related file):*
Deliberate loss of bond1.4094 and associated clag?

---

PR #4185, opened 2019-07-03
Move route reflector setting under address family

progwriter:
Also, plumb through conversion for cumulus route reflector setting.


dhalperi:
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/BgpNeighborMatchersImpl.java, line 55 at r1](https://reviewable.io/reviews/batfish/batfish/4185#-LisoofF9JWDvoiZhlJp:-LispheY2S8FG9ROoAFW:b-katbx2) ([raw file](https://github.com/batfish/batfish/blob/a229cf557d81a765fce491682a12c9ab2486cc69/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers/BgpNeighborMatchersImpl.java#L55)):*

nullable
</blockquote></details>

Why annotate a function that you're overriding and not calling? The `@Nullable` should be inherited from parent.

PR #4188, opened 2019-07-03
Cisco NX-OS port-channel support

arifogel:
- port-channel interfaces
- interface speed and bandwidth computation
- channel-group with and without 'force'
- fix VI port-channel bandwidth computation bug
  - filter contributions from inactive members

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Interface.java
@@ -45,6 +45,8 @@ public final class Interface extends ComparableStructure<String> {
     private boolean _autoState;
     @Nullable private Double _bandwidth;
     private boolean _blacklisted;
+    private @Nullable String _channelGroup;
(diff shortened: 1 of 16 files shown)


dhalperi:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4, line 647 at r1](https://reviewable.io/reviews/batfish/batfish/4188#-LitZElO6BmmmThfW_pz:-LitZElO6BmmmThfW_q-:b590pk2) ([raw file](https://github.com/batfish/batfish/blob/14acdcb611569d8bab58410abf4a858b45abbc74/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4#L647)):*
> ```
>  F_Digit?
> ```
this line looks like it has too many F_Digits: there are 1 req'd positive, 1 required any, and then 8 optional. That means last one of these seems like it makes anything up to 9_999_999_999 valid, which is too big.

PR #4195, opened 2019-07-04
Cisco NX-OS static route extraction and reference tracking

arifogel:
- ipv4 static routes in default and other VRFs

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Batfish supports configurations for a large and growing set of (physical and vir
 * Quanta
 * VyOS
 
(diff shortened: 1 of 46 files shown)


dhalperi:
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route_references, line 15 at r2](https://reviewable.io/reviews/batfish/batfish/4195#-LivURtT7zpu0IBWIkVb:-LivURtU20g-YAzTmry-:b416328) ([raw file](https://github.com/batfish/batfish/blob/64c1d9bfdc3f9b76834c6ae969dc679482bba704/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route_references#L15)):*
> ```
> vrf context vrf1
> 
> ip route 10.0.1.0/24 Ethernet1/1 10.0.0.2 vrf vrf1
> ```

this is not indented. Is it not part of the vrf context?

Issue #4199, opened 2019-07-04
Cumulus: failure to create interfaces


progwriter:
When paring configs linked here:

https://gitlab.com/DylanHamel/batfish-frr-example/tree/bc851d477cf2756ea1b3c8094fbda0e553381953/cumulus_network/nclu/configs

On the Leaf nodes interfaces are not created property. Applying a peer group to a bgp neighbor `peerlink.4096` results in an NPE (during extraction) because neither the interface nor the bgp neighbor exists.

PR #4201, opened 2019-07-04
Cumulus: flesh out more interface grammar

progwriter:
This is for #4199, but doesn't fully fix it (still need to figure out if bond subinerfaces can be referenced before a bond has been defined)
It doesn't flesh out full grammar for interface, but covers:
* interface mtu
* interface ipv6 
* bond mtu
* bond learning off

All based on the console tab completions


progwriter:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_nclu/CumulusNcluLexer.g4, line 1004 at r1](https://reviewable.io/reviews/batfish/batfish/4201#-Lj2Mj8U2A8-BOKM5CPO:-Lj2QTLaBhYh-gv1Z-4P:b9b2kj) ([raw file](https://github.com/batfish/batfish/blob/7f9fe679d241cd920465057dca67785c2f273511/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_nclu/CumulusNcluLexer.g4#L1004)):*

The fragment is not the issue, but `ZERO_LITERAL` above is.

The important thing to understand is that `DEC` will no longer match `0` in the default mode. In order to not break everything, one should make sure that parser rules that handle numbers don't rely on the `DEC` token directly, but instead on some parser `rule` that references `DEC`. Each such rule now must also include `ZERO_LITERAL` as an alternative.

For instance, `uint16` and `uint32` rules in `CumulusNclu_common.g4` need to be updated. However, `vlan_id` and `vni_number` technically do not, since they should never be `0`.

FYI In order to keep number processing comprehensible, prefer to place the `ZERO_LITERAL` token  in the lexer  right above the `DEC` token in the `Complex tokens` section.
</blockquote></details>

Ugh, that seems like not the direction I wanted to go in, thanks for explaining. 
As an alternative, how about I have the `iv6_forward` rule just use a `DEC`, (and optionally add a red flag warning of the DEC matched is anything else but 0 in extraction)?

PR #4231, opened 2019-07-10
Handle cumulus vxlan-anycast-ip in dataplane

progwriter:
Change type3 route propagation to use the anycast IP instead of the
VNI's source address.


progwriter:
This solution now seems suboptimal (it is neither small-impact, nor truly committed to correct behavior). See discussion in this doc: https://docs.google.com/document/d/1xxDAP30FyAiQnU2iygCtyieFjh7yGYuhNYK9CIkoye8/

Closing in favor of a much simpler solution short-term. Will also plan a better long-term solution (see doc linked above)

PR #4234, opened 2019-07-10
mlag id spec

ratulm:
Also sets up general infra for grammar over sets of snapshot/specific names.


dhalperi:
Victor knows more about mlag than I do...

PR #4235, opened 2019-07-10
FRR extraction for neighbors in a VRF

progwriter:
Conversion coming up in a next PR.
This is far from perfect so leaving a trail of comments/todos so we can revisit once we understand more about what can be dumbed in these `printf` lines.


progwriter:
Not the direction I want to go.

PR #4253, opened 2019-07-12
 Support for vendor-specified connected routes

dhalperi:
Some vendors let folks mark route tags, preference, etc. for connected routes.
To support this, we decided to add an optional field to interfaces that, if
present, allows vendors to specify how that interface's connected routes will
be generated from the interface addresses. If absent, the standard default
logic will be used.


dhalperi:
Need to figure out a new or different path for some corner cases.

PR #4274, opened 2019-07-17
ASA: implement network-object range

dhalperi:
We already had the power, no idea why it wasn't plugged in...


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/RangeNetworkObject.java, line 50 at r2](https://reviewable.io/reviews/batfish/batfish/4274#-Lk0ZANaDPMCNEDh5tjH:-Lk1k_EI05exGCxadhF_:be56d0m) ([raw file](https://github.com/batfish/batfish/blob/0d1305e0dce17166a5cdd1a64ae128d95d706c04/projects/batfish/src/main/java/org/batfish/representation/cisco/RangeNetworkObject.java#L50)):*

Asking whether you know for sure that the ASA syntax is inclusive on both ends.
</blockquote></details>

Oh. Yes, definitely.

PR #4280, opened 2019-07-18
Re-add "plain" tunnels in post-DP L3 topology

progwriter:
Keep a topology of non-IPsec tunnels, re-add edges during dataplane
computation.


progwriter:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/4280#-Lk4mAn01d74sIMtsMOv:-Lk4mAn01d74sIMtsMOw:bwr1tng) (no related file):*
Working on figuring out if we're tracerouting to the right IPs (we might be missing some info in VI model)

---

PR #4324, opened 2019-07-23
Converting EIGRP distribute-list to routing policy and store as export policies

haverma:


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 1441 at r2](https://reviewable.io/reviews/batfish/batfish/4324#-LkUoYzy1tfcaiaTyj_w:-LkUoYzy1tfcaiaTyj_x:b-fjuphb) ([raw file](https://github.com/batfish/batfish/blob/a55a43d3ee1bd65303bf9d257a91e9b79b419542/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L1441)):*
> ```Java
>     routingPolicy
>         .getStatements()
>         .add(
> ```

I should have made this explicit before I merged:

This is the right start, but we still need to combine this policy with the redistribution policy (that currently lives at the global process level)

PR #4325, opened 2019-07-23
NX-OS: break out separate address families

dhalperi:
Make parser and VI more correct


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java, line 159 at r2](https://reviewable.io/reviews/batfish/batfish/4325#-LkUxJBi1xpxHZL85MOI:-LkVRLAKBOr33M-Qswe2:b-tnwy58) ([raw file](https://github.com/batfish/batfish/blob/474bbebaf50641724df156fa20bd35be4d0c91af/projects/batfish/src/main/java/org/batfish/job/BatfishJobExecutor.java#L159)):*

Did you mean to commit this? This should remove all parallelism from many jobs.
</blockquote></details>

No, just testing. Done.

PR #4385, opened 2019-07-31
Cumulus: more accurate glob parsing

progwriter:
Attempt to allow more hyphens in globs by removing numbered word.
A glob is now either 
1. a numeric range
2. a word prefix followed by a number
3. a `word`

Full get rid of word segments and unnumbered words.

This means that some of the complexity for "unglobbing" has been moved to the extractor.


progwriter:
After a significant amount of offline deliberation, not convinced that this is an improvement (just pushing the problem somewhere else, into a different corner case).

Leaving #4386, need to figure out how to follow up later.

PR #4396, opened 2019-08-06
AWS: Internet connectivity tests

ratulm:


ratulm:
Need to fix some modeling issues.

PR #4406, opened 2019-08-07
fix description in searchFilters question template

saparikh:

PR #4410, opened 2019-08-07
 Enforce which characters are allowed in question templates

progwriter:
Taking over #4406

### DIFF
--- a/questions/experimental/searchFilters.json
+++ b/questions/experimental/searchFilters.json
@@ -17,7 +17,7 @@
         ],
         "variables": {
             "action": {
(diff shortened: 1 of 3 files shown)


dhalperi:
I don't see an issue affiliated with this change, nor do I see an explanation in this PR. Why?

dhalperi:
(@progwriter)

progwriter:
To aid creation of pybatfish documentation that uses tables (`|` is interpreted as special character in some parts of RST). 
Since descriptions are meant to be simple English and a pipe only existed in one place, fixed it this way.

PR #4416, opened 2019-08-07
Cisco: better interface -> OSPF process mapping

sfraint:
Only associate an interface with an OSPF process if they are explicitly or implicitly (e.g. overlapping OSPF network and interface address) associated with each other in the config.


tests/parsing-tests/unit-tests.ref
```diff
@@ -92842,6 +92842,14 @@
             }
           ]
         },
+        "ios_xr_ospf" : {
+          "Red flags" : [
+            {
+              "tag" : "MISCELLANEOUS",
+              "text" : "Interface: 'Bundle-Ethernet201' contains OSPF settings, but there is no OSPF process"
```
progwriter: Just skimming this file looks like ospf process exists. Is this warning correct?

PR #4485, opened 2019-08-13
OSPF: add hello interval to VI model

sfraint:
* Add OSPF hello interval to VI model
* Also update dead interval to be `null`able (both can be unspecified, e.g. non-OSPF interface)


sfraint:
Closing in favor of https://github.com/batfish/batfish/pull/4617

PR #4521, opened 2019-08-15
IOS: add support for ip_nat_source overload

dhalperi:
See: https://www.cisco.com/c/en/us/support/docs/ip/network-address-translation-nat/26704-nat-faq-00.html#q18


dhalperi:
Closing in favor of work done by @progwriter

PR #4596, opened 2019-08-21
NXOS: parse and ignore some crypto lines

progwriter:
I anticipate more stuff will need to be parsed, so I just went ahead and split into a new file to setup for the future.


dhalperi:
Suggest having @haverma take a look at these

PR #4598, opened 2019-08-21
NX-OS: introduce RoutingProtocolInstance and RedistributionPolicy

dhalperi:
These are two constructs that are common across the parser, and they are useful
for reuse in lots of places (e.g., redistribution applies as-is across other
protocols).

Also fixes the bug where we only allowed one redistribution policy per
protocol, instead of one per protocol instance.


dhalperi:
https://buildkite.com/batfish/batfish-pre-commit/builds/11332 took over and passed, I'm not sure why github wasn't updated.

PR #4614, opened 2019-08-23
 NX-OS: basic EIGRP extraction 

dhalperi:
Just existence of process config in VRFs.


dhalperi:
(I marked the base PR as reviewed already)

PR #4694, opened 2019-09-04
Ospf session compat q

sfraint:
Update `ospfSessionCompatibility` to include unestablished/incompatible sessions, with a new `Session_Status` column.


ratulm:
@sfraint - In case it wasn't on your radar, Parameters.md needs updating too with the new specifier. (Can be a separate PR.)

PR #4818, opened 2019-09-25
Runtime data infrastructure + use in NX-OS

corinaminer:
[Design doc link](https://docs.google.com/document/d/1iAdwuq--6rjMUhkRnJ9lCjTzIActy1Q651q2Ebiggts)

For now runtime data only consists of interface bandwidths and speeds. Can add more properties incrementally.

Integrated into conversion by giving `VendorConfiguration` a field containing the runtime data for its config so that all conversion logic can access it. Added logic to use runtime data in NX-OS conversion as a proof of concept and for ease of testing.

Still to do:
- Add more properties to runtime data as needed
- Add logic to use runtime data to other vendors
- Integrate runtime data to `AwsConfiguration`


dhalperi:
If this is only integrated in conversion, why not just have this as another argument to `toVendorIndependentConfiguration`?

PR #4825, opened 2019-09-26
Ignore ipprotocol.ip

ratulm:
Make `IP` an illegal value for IP protocol specification.


ratulm:
I'll wait for #4826 to be merged and then fix things here.

PR #4826, opened 2019-09-26
IpProtocol: remove IP

dhalperi:


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java, line 7660 at r3](https://reviewable.io/reviews/batfish/batfish/4826#-LpiTl1PE1-CZO9qxeJk:-LpiTl1PE1-CZO9qxeJl:b1frpia) ([raw file](https://github.com/batfish/batfish/blob/beb8bb0a03868b50ba7f256d5a3ff7a12fcd1939/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L7660)):*
> ```Java
>         todo(ctx);
>       } else if (ctx.gre_ipv4 != null) {
>         tunnel.setMode(TunnelMode.GRE_MULTIPOINT);
> ```

tl;dr: I think these are all different vendor variants of the same thing.

PR #4856, opened 2019-10-02
NX-OS: cleanup tcp/udp port_number

dhalperi:


dhalperi:
Am I missing something or were we just silly here?

PR #4857, opened 2019-10-02
add interfaceLinkOwnedIps to specifier context

anothermattbrown:
This is based on forwarding analysis, and will include IPs that are
assigned EXITS_NETWORK or DELIVERED_TO_SUBNET when routed out the
interface.


dhalperi:
Think this is (finally) replaced now :)

PR #4887, opened 2019-10-05
PAN: loopback without units, interfaces with multiple addresses

dhalperi:


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/Interface.java, line 68 at r1](https://reviewable.io/reviews/batfish/batfish/4887#-LqOZvZj5pnD9wFRVtH-:-LqOp4B772lXMxZ36O_U:b-sraem0) ([raw file](https://github.com/batfish/batfish/blob/b303873a65ece3c1b26886879154cfd7fb6a04a8/projects/batfish/src/main/java/org/batfish/representation/palo_alto/Interface.java#L68)):*

not guava :) ?
</blockquote></details>

This is a shim, not a copy.

Issue #4864, opened 2019-10-03
router igmp parsing error


agember:
The following Cisco configuration snippet causes an irrecoverable parser failure:
```
!
router igmp
 vrf default
  interface all router disable
 !
!
end
```

dhalperi:
Hey @agember - we have some existing parsing tests with `router igmp` in it. Can you say more about what is unrecoverably bad? Stack trace? Lost parsing afterwards? Something else?

dhalperi:
And for config snippets, we need information about vendor as well -- what OS is this?

agember:
@dhalperi The snippet is Cisco IOS. The parser fails when it encounters the `interface` line. The context at the time is `[s_interface, cisco_configuration]`. 

Looking at the parsing for `router igmp` (https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoParser.g4#L2185), it seems that there is no parsing support for the `vrf` line. I don't know how the parsing recovery works, but I would hypothesize that the following occurs: the parser tries to recover when it encounters the `vrf` line; recovery returns to the top-level context `cisco_configuration`; the subsequent `interface` line causes the parser to match `s_interface`; the parser does not know what to do with `router disable`, so it fails.

dhalperi:
I still don't understand the actual behavior. If recovery is working, we should ignore that entire `router igmp` block, pop up to next level, and exit with unrecognized lines. That would be expected behavior, is that what's happening?

agember:
I was getting status of `FAILED` instead of `PARTIALLY_UNRECOGNIZED`, but now I cannot reproduce it. I created a pull request to add the snippet to the test suite, to avoid regressions, but I don't think there is any further fixing required.

PR #4907, opened 2019-10-08
Add router igmp with vrf to Cisco IOS parsing tests

agember:
Test case for #4864


dhalperi:
As you can see from #4864 and #4915 , the error was entirely somewhere else. #4915 added regression testing in the form of unit test for this syntax, so I'll close this PR.

PR #4984, opened 2019-10-14
Begin differentiating candidate and established BGP topologies

progwriter:
Towards making `BgpSessionProperties` only used for valid/established sessions.


progwriter:
This conflicts now, will revisit later

PR #5026, opened 2019-10-17
BGP: flesh out next hop propagation to be more comprehensive

progwriter:


progwriter:
Not the right approach to the problem I'm trying to solve

PR #5042, opened 2019-10-19
model ospf nbma in VI

yifeiyuan:


progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfTopologyUtils.java, line 296 at r1](https://reviewable.io/reviews/batfish/batfish/5042#-LrWRf0ZBSZh8-BY2dVv:-LrWT7qU-TnZU-MY9uZI:b-udcqj7) ([raw file](https://github.com/batfish/batfish/blob/326e9cd17a01d8236f5a80ecaf713bfc2f201955/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfTopologyUtils.java#L296)):*

this will break existing session establishment unless conversion for vendors is in place.
</blockquote></details>

I suggest returning true optimistically for now, plumbing in correct logic later.

PR #5087, opened 2019-10-30
EVPN: cleanup conversion code to fix crash

dhalperi:


dhalperi:
tl;dr: only create EVPN neighbor if the neighbor has EVPN configuration

PR #5106, opened 2019-11-01
Cumulus concatentated be robust to leading newlines

arifogel:


dhalperi:
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cumulus_concatenated/testconfigs/concatenation_with_leading_newlines, line 1 at r1](https://reviewable.io/reviews/batfish/batfish/5106#-LscDoZt8fTAvGWYRZEc:-LscDoZt8fTAvGWYRZEd:b647wm3) ([raw file](https://github.com/batfish/batfish/blob/6216fae38d457f987d53278ff4aaca485e4617e3/projects/batfish/src/test/resources/org/batfish/grammar/cumulus_concatenated/testconfigs/concatenation_with_leading_newlines#L1)):*
> ```
> 
> ```

am I correct that if there is a space it will be used as the hostname?

PR #5135, opened 2019-11-07
XR: simplify v4 access-list

dhalperi:
IOS-XR does not distinguish between standard and extended access-lists.


dhalperi:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4, line 65 at r2](https://reviewable.io/reviews/batfish/batfish/5135#-Lt80lm-0LyDD19KDEc3:-Lt8QJAg8dY9G0zcM48S:b-9snzqq) ([raw file](https://github.com/batfish/batfish/blob/3301bd7cb1702d67e388d57ed4e4ead76e66af88/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4#L65)):*

what is this?
</blockquote></details>

placeholder

PR #5158, opened 2019-11-19
Buildkite: dagify to see if its faster

dhalperi:
* Make the extensive static analysis tests depend on templates and formatting, but not on the jar. This gives them a ~3.5m head start.

* Make the ref_tests depend only on the jar, instead of also on formatting and templates. This means they will run even if formatting is bad, as long as the jar builds.


dhalperi:
cc: @anothermattbrown - may be possible to speedup web ui with similar tactics.

PR #5167, opened 2019-11-25
new cumulus interface syntax

yifeiyuan:
add rules to parse the following commands:
```
auto swp1
iface swp1 inet static
  link-autoneg on
  link-speed 1000
```


ratulm:
@yifeiyuan - does this PR close any of the open issues?

PR #5201, opened 2019-12-05
IpPermissions: rewrite and comment

dhalperi:
Plus, add entry name to ACL term.


dhalperi:
Ended up removing the ICMP stuff from this PR. That deserves testing separately.

PR #5205, opened 2019-12-06
Add address families column for BGP peer configuration question

haverma:
The answer to `BgpPeerConfiguration` will contain a set of all Address Families configured on a BGP peer.


haverma:
closing, reason mentioned in my last comment above

PR #5215, opened 2019-12-07
Batfish: move towards eliminating the internal snapshot stack

dhalperi:
Relying on global shared state is dangerous, prevents multithreading, prevents encapsulated code, and/or leads to heisenbugs.

The biggest part of shared state right now is in the pushSnapshot()/pushReferenceSnapshot()/popSnapshot() code. Try to get rid of that in many staged commits.

Side effects:
* many api calls now have a mandatory NetworkSnapshot argument -- so we don't rely on the global one
* removed or deprecated the older variants that peeked at the stack, making that peek explicit.
* deleted the notion of `diffactive` -- was too complex to keep up and only the deprecated Java client used it.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/5215#-LvTwGom8Q1NOG2wO6Bm:-LvTwGom8Q1NOG2wO6Bn:bhnw2iy) (no related file):*
Self-merge only, forgot to hit `draft` in github.

---

PR #5221, opened 2019-12-10
Dataplane: attempt a static logger

progwriter:
Tell me all the ways in which this is a bad idea :)


progwriter:
I've convinced myself this is not worth it, and what's worth it instead is incremental improvements towards using a real logging framework.

PR #5235, opened 2019-12-12
diff reachability: only use start locations present and active in both snapshots

anothermattbrown:


anothermattbrown:
closing because reviewable got confused after the rebase/force push. will reopen after addressing comments

PR #5254, opened 2019-12-17
NX-OS: fix logging server parsing

dhalperi:
The options are all optional

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4
@@ -21,7 +21,8 @@ logging_server
   (
     ls_facility ls_use_vrf?
     | ls_use_vrf ls_facility?
-  ) NEWLINE
+  )?
+  NEWLINE
 ;
 
 logging_host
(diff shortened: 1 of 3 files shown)


dhalperi:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/5254#-LwGWuP39UiZs2AyHM11:-LwG_0mZ2rTClEITs6Sf:blfr1h7) ([raw file](https://github.com/batfish/batfish/blob/59c889f986611287e30bc4a7295daeaf7a40e66d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4#L22)):*

> ```
> ?
> ```
this `?` can be removed now, right? Since we have `?` on the block
</blockquote></details>

AFAICT, the real options are:

```
<nothing>
A
B
AB
BA
```

Can that be done more simply than

`(A B? | B A?)?`

Making it `(A B | B A?)?` [if I interpret your suggestion correctly] eliminates just A, right?

haverma:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/5254#-LwGWuP39UiZs2AyHM11:-LwGxnZ20uUogRAJ9Jde:b-8rzqmx) ([raw file](https://github.com/batfish/batfish/blob/59c889f986611287e30bc4a7295daeaf7a40e66d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_logging.g4#L22)):*

I think the current suggestion is wrong, happy to revise if you come back here with a better one.
</blockquote></details>

Yes, my bad. I had misread and misunderstood the rule before.

PR #5260, opened 2019-12-17
Include match criteria in MatchSessionStepDetail

adrianliaw:
- New data type:
    - `SessionMatchExpr`
        - Contains all the information needed to match a session, but simpler and more specific than `AclLineMatchExpr`
- Changes in `FirewallSessionTraceInfo`
    - Use `SessionMatchExpr` for `_sessionFlows`
    - Keep the getter `getSessionFlows` returning `AclLineMatchExpr` to not break all the usages of `FirewallSessionTraceInfo`
    - Add `getMatchCriteria` that returns `_sessionFlows` as is, which is used in `MatchSessionStepDetail`
- Changes in `FlowTracer`
    - `matchSessionReturnFlow` now returns a `SessionMatchExpr` instead of `AclLineMatchExpr`
- Changes in `MatchSessionStepDetail`
    - Added `matchCriteria` field of type `SessionMatchExpr`
- Tests
    - Created tests for `SessionMatchExpr` on equality and JSON serialisation
    - Tested `matchCriteria` field in `TracerouteEngineImplTest`
    - Fixed tests that broke because of the new data type
        - Replacing `AclLineMatchExprs.TRUE` with an arbitrary dummy `SessionMatchExpr` if the match criteria is meaningless in the test
        - Replacing `AclLineMatchExprs.TRUE` with a `SessionMatchExpr` that matches the tested flow if the match criteria is expected to pass

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/FirewallSessionTraceInfo.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/flow/FirewallSessionTraceInfo.java
@@ -14,14 +14,14 @@ public final class FirewallSessionTraceInfo {
   private final @Nonnull String _hostname;
   private final @Nonnull Set<String> _incomingInterfaces;
   private final @Nonnull SessionAction _action;
-  private final @Nonnull AclLineMatchExpr _sessionFlows;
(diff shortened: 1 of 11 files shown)


adrianliaw:
This PR will be broken into 3 parts with a PR for each.

PR #5310, opened 2019-12-23
PAN support application `any` with service `application-default`

sfraint:
Add support for `any` application, which was previously unhandled in Batfish.

In Palo Alto security rules, `application-default` can be specified as service match, which effectively defers traffic matching to application inspection.  In that case, the special `any` application should match any application defined on the device.

After this PR, Batfish assumes that all traffic will match the `any` application.


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java, line 1018 at r2](https://reviewable.io/reviews/batfish/batfish/5310#-LxSBvkk4_Q7hCTqUQED:-LxbWTcy10KiMeZO73Wa:b-ust673) ([raw file](https://github.com/batfish/batfish/blob/6ada608f02f756506148ca726894fb4bcf52fb68/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1018)):*

> ```
>         if (rule.getAction() == LineAction.PERMIT) {
>           // Since Batfish cannot currently match above L4, we follow Cisco-fragments-like logic:
>           // When permitting an application, optimistically permit all traffic where the L4 rule
>           // matches, assuming it is this application. But when blocking a specific application, do
>           // not block all matching L4 traffic, since we can't know it is this specific application.
>           serviceDisjuncts.addAll(matchApplications(rule, vsys));
>         }
> ```
is this the desired behavior even when `any` is in the set of applications for a deny rule? i.e. no traffic will match a "deny any application" rule w/ application-default specified

I haven't seen any deny rules referencing any application with application-default specified, but it is surely possible
</blockquote></details>

Don't you mean all traffic will match?

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java, line 1044 at r2](https://reviewable.io/reviews/batfish/batfish/5310#-LxbWd7W6KVFZgsx-NUS:-LxbWd7W6KVFZgsx-NUT:b3vizxk) ([raw file](https://github.com/batfish/batfish/blob/6ada608f02f756506148ca726894fb4bcf52fb68/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1044)):*

> ```
>       // Assume all traffic matches some application under the "any" definition
>       if (name.equals(CATCHALL_APPLICATION_NAME)) {
>         return ImmutableList.of(TrueExpr.INSTANCE);
>       }
> ```
</details>

This almost certainly needs to go down below where we check built-ins? Although hopefully no one overrides the `any` built-in?

PR #5316, opened 2019-12-27
Frr npes

ratulm:
Change potential NPEs into warnings. 

I am not sure if this is the right pattern if the user is not looking initIssues (though this PR does not change that. It just translates scary NPEs, which were going into initIssues, into warnings.)


ratulm:
Ok, let's close this one.

PR #5322, opened 2019-12-28
Frr default originate

ratulm:
After #5317


ratulm:
I added a behavior test that mirrors the one for Arista but I cannot get it to pass :) PrefixTracer says filtered_out. How can I debug deeper?

progwriter:
___
*[projects/batfish/src/test/java/org/batfish/grammar/cumulus_frr/CumulusFrrGrammarTest.java, line 407 at r6](https://reviewable.io/reviews/batfish/batfish/5322#-LxPLOch0JahbXR4eZa2:-LxPLOch0JahbXR4eZa3:bmmhjtb) ([raw file](https://github.com/batfish/batfish/blob/14171120d713ab87093686e48829068138f71edf/projects/batfish/src/test/java/org/batfish/grammar/cumulus_frr/CumulusFrrGrammarTest.java#L407)):*
> ```Java
>         Bgpv4Route.builder()
>             .setCommunities(
>                 ImmutableSet.of(StandardCommunity.of(7274718L), StandardCommunity.of(21823932L)))
> ```

I see nothing in the test config that sets these communities. Why do you expect them to be present?

progwriter:
Debug by breakpointing where these routes are processed (https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/VirtualRouter.java#L1780) or thereabouts.

ratulm:
@progwriter - thanks, that helped. the real root cause was that the code i had earlier copied my pattern from wasn't even being invoked.

PR #5327, opened 2019-12-30
Changes for "additive" in "set community" for FRR.

raveranj:
Fixed compile errors to enable "additive" in "set community" command.

Removed comments from RouteMapSetCommunity.java.

Removed commented code.


raveranj:
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/RouteMapSetCommunity.java, line 51 at r2](https://reviewable.io/reviews/batfish/batfish/5327#-LxOaLOFCrvCyl0CbFg7:-LxSL5jHB3zVZ_oQ8y0u:b-260smh) ([raw file](https://github.com/batfish/batfish/blob/f4e2dbb7ad304774ffc4b5dc2141b82196fc0604/projects/batfish/src/main/java/org/batfish/representation/cumulus/RouteMapSetCommunity.java#L51)):*

The behavior you have implemented here does not reflect the behavior of a Cumulus device when a `set community xx:yy additive` command is entered,
Whenever a `set community` line appears in a route-map, it replaces any existing one.

The meaning of `additive` is that the communities mentioned should be added to the communities of a route to be processed, rather than replacing the communities of the route to be processed.
This function should not be altered.
</blockquote></details>

Sync-ed up with Ari offline. Fixing up the behavior. Thanks Ari!

raveranj:
Hey @dhalperi - Should have an updated PR today (latest tomorrow).

PR #5337, opened 2020-01-03
Proof-of-concept log4j logging

progwriter:
* Available only to the batfish module, currently. 
* Does not change existing `BatfishLogger` behavior.
* Using dataplane code as the playground since:
    1. good place to experiment with MDC 
    2. Badly in need of better logging


progwriter:
Too stale

PR #5357, opened 2020-01-07
BgpProcess: store confederations as a space

progwriter:


dhalperi:
Discussion: do the other thing instead, for now? If someone really does this in their configs, we can visit that bridge.

progwriter:
Ok, current thinking is that in normal network designs there can only be a small set (< 10) of confederation peers. No need to change VI at this time.

PR #5360, opened 2020-01-07
remove the check for existense of at least one address  family

ratulm:
Well, nothing bad happened .... the peer shows up in other questions as expected


progwriter:
Just to document offline discussion. The reason for this change was to allow seemingly invalid (from dataplane point of view) peers to be visible/examined through questions like `bgpPeerConfig`

PR #5376, opened 2020-01-08
Arista BGP: flip bit to new bgp parser

progwriter:
* `aristabgp` debug flag is now a noop
* `noaristabgp` debug flag now allows reverting to the old hybrid parser
* Fix up syntax in some of our older tests that use EOS configs

### DIFF
--- a/tests/parsing-tests/unit-tests-undefined.ref
+++ b/tests/parsing-tests/unit-tests-undefined.ref
@@ -44,18 +44,6 @@
     "rows" : [
       {
         "File_Name" : "configs/arista_bgp",
-        "Struct_Type" : "bgp peer-group",
-        "Ref_Name" : "UNDEFINED_PEER_GROUP",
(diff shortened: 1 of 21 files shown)


tests/parsing-tests/unit-tests-undefined.ref
```diff
@@ -42,18 +42,6 @@
       "textDesc" : "'${File_Name}' has undefined '${Struct_Type}' called '${Ref_Name}' being used as '${Context}' at line(s) ${Lines}"
     },
     "rows" : [
-      {
-        "File_Name" : "configs/arista_bgp",
-        "Struct_Type" : "bgp peer-group",
-        "Ref_Name" : "UNDEFINED_PEER_GROUP",
```
dhalperi: Is this a bug?

progwriter: Yes. :( #5378

PR #5394, opened 2020-01-10
Migrate off IpSpaceMetadata onto TraceElement

anothermattbrown:


anothermattbrown:
Trying again a simpler way. Fingers crossed

PR #5399, opened 2020-01-11
FRR: merge loopback parsing with other interfaces

ratulm:
NCLU heritage meant that parsing and post-processing of loopback ("lo") being special-cased. We keep seeing more properties that are shared and the separation does not make sense. No functional change in this PR.


progwriter:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/5399#-LyIbn6w6O3LKQJrIuyy:-LyIbn6w6O3LKQJrIuyz:b-qbul8m) (no related file):*
>  the separation does not make sense

Doesn't it? I don't think `clagd` set of properties can be configured on "regular" interfaces.
Similarly, I don't think loopbacks can ever have L2 settings (switchport, etc.)

That said, you've been looking a lot more at the configs than I have recently, so feel free to tell me why I'm wrong.

---

PR #5433, opened 2020-01-15
Remove AclLineMatchExpr comparability

anothermattbrown:
We used this to sort subexpressions of And/Or. Doesn't add value,
and makes it harder to generate good traces.


dhalperi:
___
*[projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java, line 5720 at r1](https://reviewable.io/reviews/batfish/batfish/5433#-LyfNgDb9BGrvjM9qRoD:-LyfOPnz2fD-84u3b3oL:b-yeqi0y) ([raw file](https://github.com/batfish/batfish/blob/ac153bf1bd6cb85f5a8d916207b8fbb99d9a85ff/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L5720)):*

I don't know how this was passing before
</blockquote></details>

Is this not just a descriptive string that is not tested?

PR #5435, opened 2020-01-16
acl tracer improvements

anothermattbrown:


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/acl/AclTracer.java, line 316 at r1](https://reviewable.io/reviews/batfish/batfish/5435#-Lyg-lghDXfPQHroGnpA:-Lyg2HTxA-4tNEOXhtqe:b-jn5s6n) ([raw file](https://github.com/batfish/batfish/blob/305580a64a4d487783bcfbccd52919189708743e/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/acl/AclTracer.java#L316)):*

I am likely missing something here, or maybe just pre-anticipating future work.

Looking at the test, we are now reporting the not-match ACL and the match (denied) acl. Is this what you want?

I thought we'd be adding a subtrace at the top, getting the result, and then discarding or completing the subtract based on the result.


</blockquote></details>

Discussed offline.

anothermattbrown:
went a different directioin

PR #5441, opened 2020-01-16
Feature/frr Support set metric +/- in Cumulus route-maps.

raveranj:


raveranj:
___
*[projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java, line 367 at r2](https://reviewable.io/reviews/batfish/batfish/5441#-LykOFfW7RjMEVCat9oy:-Lykt3oP9eZGrGW5-Cb4:b9xwf25) ([raw file](https://github.com/batfish/batfish/blob/52ced3e950f03e5e44c10fbec24a642fabcada08/projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java#L367)):*

I'd recommend starting with a number greater than 1. Otherwise you cannot tell if underflow correction is always applied, or just when result is less than 0.
Of course you'll have to update expectations below.
</blockquote></details>

I replaced the underflow check in DecrementMetric.java with Math.max(), which will always trigger the underflow check.

PR #5506, opened 2020-01-31
JunOS: trace firewall filters and security policies separately

dhalperi:
Before, we always marked them as firewall filters/terms in the trace.
To do this better, make a trivial start at separating these in the VS.

* Security policies moved to their own structure
* Security policy terms are always concrete
* Plumb types down into the functions that make the ultimate ACL lines


dhalperi:
@yifeiyuan - take first pass, then @anothermattbrown can review?

PR #5523, opened 2020-02-07
introduce LocationInfo

anothermattbrown:
LocationInfo defines how a location should be used in batfish analyses:
whether a location should be used as an endpoint by default, the source
IPs to use for it when sourcing traffic, and (specifically for
InterfaceLinkLocation) the IPs of external devices (not modeled by
batfish) the link would get an ARP reply from.

This is a first integration step: only the source IPs of a LocationInfo
are used, and the generation of LocationInfo objects is only temporary.
Batfish is creating them for now; in a future PR they will be generated
by VendorConfiguration and/or external input. Also, in a future PR we
will serialize them to disk; for now we recreate each time they are
needed.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/5523#-M-Vz83T48QXT3CslDKI:-M-Vz83T48QXT3CslDKJ:b-n1y6v5) (no related file):*
Want to inspect this before too much farther - haven't caught up on design discussions

---

PR #5555, opened 2020-02-13
Add a test for schemautils

ratulm:


progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Node.java, line 48 at r1](https://reviewable.io/reviews/batfish/batfish/5555#-M-x5SbB11I3U4V0CUSo:-M-x5SbCDX54aI9gCIZj:b-3g1sq7) ([raw file](https://github.com/batfish/batfish/blob/55e370173a7703412ae93b8390306075d7ded61b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/pojo/Node.java#L48)):*
> ```Java
> 
>   // this annotation prevents jackson from converting a string to a node object
>   @JsonIgnore
> ```

Is the annotation still necessary after #5551 ?

Definitely keep the test though ( I was going to send a PR for that myself)

PR #5568, opened 2020-02-18
AWS: refine subnet into public and private

dhalperi:
A public subnet is a subnet that's associated with a route table that
has a route to an Internet gateway.

https://docs.amazonaws.cn/en_us/vpc/latest/userguide/VPC_Scenario2.html


dhalperi:
cc: @adrianliaw @anothermattbrown

PR #5576, opened 2020-02-21
Juniper: better traces for matching source-address,destination-address

dhalperi:
Use the actual text in the config instead of re-string-ifying the wildcard.

Additionally, track the any and any-ipv4 lines.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -3869,10 +3869,11 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener {
     FwFrom from;
     IpWildcard ipWildcard = formIpWildCard(ctx.fftfa_address_mask_prefix());
     if (ipWildcard != null) {
+      String text = getFullText(ctx.fftfa_address_mask_prefix());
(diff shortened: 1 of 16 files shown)


dhalperi:
___
*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 2780 at r1](https://reviewable.io/reviews/batfish/batfish/5576#-M0dxB-QCCQkJ8BV5JTb:-M0dyhhgCp6fPm6DAx0X:b-ews63m) ([raw file](https://github.com/batfish/batfish/blob/39d2506e9af90fd7c4265a51eee7296fb10c09aa/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L2780)):*

> ```
> 1.2.3.4/
> ```
why is this changed?
</blockquote></details>

The configuration line is `set firewall family inet filter FILTER term TERM from source-address 1.2.3.4/255.0.255.0`

Before, we normalized this to `1.0.3.0:0.255.0.255`, which is different because:

* IpWildcard removes bits that don't matter, so the 1.2.3.4 becomes 1.0.3.0
* Batfish wildcards have bits for "don't care", but Juniper apparently has bits for "care", so the `255.0.255.0` in Juniper became `0.255.0.255` in Batfish.

These both show why including the original text is a feature!

PR #5578, opened 2020-02-25
Fix match on community lists in match-clauses of FRR route-maps.

raveranj:


raveranj:
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConversions.java, line 1365 at r2](https://reviewable.io/reviews/batfish/batfish/5578#-M11QqaR3SpzNGAyhe99:-M12E06_0SEgvtIDV86G:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/74fc9ee899d1b692c591358f36821f6ad1c5e998/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConversions.java#L1365)):*

nit: add `@Nonnull` return type annotation for both of these
</blockquote></details>

Done.

raveranj:
___
*[projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java, line 510 at r2](https://reviewable.io/reviews/batfish/batfish/5578#-M11RuXX77stlkOAr-Kt:-M12E4Yj9iKtIfyCBAOT:b5ugun3) ([raw file](https://github.com/batfish/batfish/blob/74fc9ee899d1b692c591358f36821f6ad1c5e998/projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java#L510)):*

Please add more tests for coverage:
- standard community-list with multiple communities on a line
  - match route with all such communities
  - match route with all such communities and more
  - reject route with only some such communities
- standard community-list where route is permitted by some line, but denied because it is matched by an earlier deny line
- expanded community-list with matchable regex involving multiple communities
  - match route with communities matching regex
 - expanded community-list with unmatchable regex (say with backwards sort order, like `_2:2_1:1_`)
  - does not match route with what appear to be those communities
- expanded community-list where route is permitted by some line, but denied because it is matched by an earlier deny line
</blockquote></details>

Done. Added more tests.

raveranj:
___
*[projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java, line 623 at r3](https://reviewable.io/reviews/batfish/batfish/5578#-M12TZJu3t7Yzs_TxSY3:-M12fqdKBe5UIv3zby4b:b-38izd0) ([raw file](https://github.com/batfish/batfish/blob/b4bd2e917dc7b7020ef1d2d5a79845a1e236a501/projects/batfish/src/test/java/org/batfish/grammar/cumulus_concatenated/CumulusConcatenatedGrammarTest.java#L623)):*

There is no such thing as order for the input route. Communities are provided as a set. The order of arguments to `ImmutableSet.of` has no impact.
I meant the regex should be unmatchable, because the rendering of the input route will never match the regex.

The rendering of the input route communities is always sorted. So if the expanded list expects a rendered string that is not sorted, it should never match.
Hence the example I gave: `_2:2_1:1_`. This is unmatchable because if the route contains at least 1:1 and 2:2, the rendering of the route's communities to which the regex is applied will be `...1:1...2:2...`, and never `...2:2...1:1...`.

Ordering should never matter for standard community lists, because we are only checking whether the line's communities are a subset of the route's communities.
</blockquote></details>

Done. Thanks for clarifying. I've replaced this block with a test-case for no-match between the regex in the comm-list and input route.

PR #5588, opened 2020-02-28
AWS: Storing human names for more node types

haverma:
specifically, this adds human-names for the following node types (extracts the tag with key = "Name")

- Subnet
- VPCs
- Internet Gateways
- VPN Gateways
- NAT Gateway
- Transit Gateway

Also moves the class `Tag` to its own file


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/aws/Tag.java, line 33 at r1](https://reviewable.io/reviews/batfish/batfish/5588#-M18RQUe72kYh8vYS2lf:-M19RiLA96jq28JWywo4:b-485oxl) ([raw file](https://github.com/batfish/batfish/blob/40fde03640dfbf988755eb8a95d7e92bc80f0e75/projects/batfish/src/main/java/org/batfish/representation/aws/Tag.java#L33)):*

They didn't need that because they were never being serialized (none of the other AWS constructs have that either). But I added the annotation anyways to get the de-ser test to work.
</blockquote></details>

Why add the clone test at all? All these classes are JSON-read-only, aren't they?

PR #5647, opened 2020-03-23
Write atomicity in FileBasedStorage

arifogel:
- should prevent i/o concurrency issues, e.g. reads of incomplete data
for most files


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java, line 1208 at r2](https://reviewable.io/reviews/batfish/batfish/5647#-M3CIjLq8oAtWntDfU0e:-M3CIjLq8oAtWntDfU0f:b-2e6z0x) ([raw file](https://github.com/batfish/batfish/blob/940eee5d7bfe3b872cb7d069c0ef660707540f77/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java#L1208)):*
> ```Java
>     lock.lock();
>     try {
>       return CommonUtil.readFile(file);
> ```

It looks like reads of non-existent files will now block if they are being written. The preferred behavior is to return immediately with the "file is missing" behavior (null or throw).

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java, line 1204 at r2](https://reviewable.io/reviews/batfish/batfish/5647#-M3CIz4ZDxh63A-caMHO:-M3CIz4ZDxh63A-caMHP:b-6m3ca8) ([raw file](https://github.com/batfish/batfish/blob/940eee5d7bfe3b872cb7d069c0ef660707540f77/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java#L1204)):*
> ```
>  throws BatfishException
> ```
No more uses of this class, ever. It's unnecessary, keep it checked, fix callers if needed.

PR #5651, opened 2020-03-25
use the name in tags as security group name

ratulm:
before this PR, we were using GroupName -- which does not seem to be something that is actively managed by users. e.g., TF puts in a GUID there and AWS puts in 'default' there. using tag name also makes consistent with other resources.


dhalperi:
We used group name to match AWS behavior. Did you want to diverge from Amazon?

ratulm:
This is what I see with TF-based management:

Screen Shot 2020-03-24 at 7 03 12 PM

dhalperi:
If you type in a place where Amazon is auto completing, it'll search and display group name and ignore Name Tag

ratulm:
Hmmm..interesting, inconsistent behavior :) Ok, lets stick to what we have for now.

PR #5654, opened 2020-03-26
syntactic sugar for the internet source location

anothermattbrown:
add `internet` as a shorthand for `@enter(internet[out])` to the location specifier grammar.

no tests yet -- seeking feedback on this approach first


anothermattbrown:
note: if users want to specify "all interface locations on the internet node" (not sure why they would), they should still be able to do that via `internet[.*]`

PR #5659, opened 2020-03-27
BDDFlowConstraintGenerator: change the algorithm

dhalperi:
Previous algorithm: picked flows from BDDs by taking a list of constraints and
keeping the first match. The constraints were either very simple ("ICMP
echo-request") or fairly complex ("well-known source IPs to private IPs using
UDP traceroute"). After finding the first matching constraint, we'd fill in the
unconstrainted bits. For simple constraints, we ended up with a lot of "source
IP 0", "source port 0", etc. For complex constraints, the barrier was an
exponential explosion of simple constraints.

Instead, use a long list of "simple" constraints, which need not be related in
any way except that more preferred ones come earlier. This allows us to home in
on a good representative -- we can put very general nudges like "if possible,
exactly one of src and dst port should be ephemeral" without an exponential
explosion. (The cost instead is potentially many operations to compute a
representative, which has not been fully measured).


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/5659#-M3TFVMa4nIc4tIf76nt:-M3TFVMa4nIc4tIf76nu:b-7n9nqf) (no related file):*
FYI no new tests yet.

All tests that changed were, imo, improvements.

---

PR #5666, opened 2020-03-31
BlockReservedAddressesAtInternet: allow documentation IPs

dhalperi:
This is crucial to keeping examples working. Instead, nudge away from the doc subnets
when picking flows. TBD how we'll resolve permanently, but this is a good middle.


dhalperi:
cc: @anothermattbrown

Issue #5598, opened 2020-03-03
Palo Alto aggregate interface parsing issue


oneryalcin:
Hi Batfish team, 

We have few palo alto devices and I noticed that batfish is able to parse interface IP addresses (hence successfully building `bfq.layer3Edges().answer().frame()` )  when the interface is not an aggregate interface, but it fails to parse when the interfaces is defined as an aggregate interface.
Therefore ip addresses under aggregate-ethernet interfaces doesn't show up in `l3Edges()`  

A very minimal example is given below:

- Parsing `ethernet1/3` is successful
- Parsing  `aggregate-ethernet ae1` is failing

```
devices {
  localhost.localdomain {
    network {
      interface {
        ethernet {
          ethernet1/3 {
            layer3 {
              ip {
                10.0.0.1/29;
              }
            }
          }
          ethernet1/21 {
            aggregate-group ae1;
            link-state auto;
            link-duplex auto;
            link-speed auto;
          }
          ethernet1/22 {
            aggregate-group ae1;
            link-state auto;
            link-duplex auto;
            link-speed auto;
          }
          aggregate-ethernet {
            ae1 {
              layer3 {
                units {
                  ae1.290 {
                    tag 290;
                    ip {
                      10.0.1.1/29;
                    }
                  }
                  ae1.200 {
                    tag 200;
                    ip {
                      10.0.2.1/29;
                    }
                  }
                  ae1.201 {
                    ip {
                      10.0.3.1/29;
                    }
                    tag 201;
                  }
                }
              }
              comment "Aggregate interface to "Core Router";
            }
          }
        }
      }
    }
  }
}
```

dhalperi:
Thanks @oneryalcin - we'll take a look.

dhalperi:
Hi @oneryalcin - Did you edit these configs or are they direct off the box? A few followups:

* In this config, `aggregate-ethernet` is inside of `network > interface > ethernet` ; on my box it's just under `network > interface`. Can you confirm?
* Is the comment at the end edited, or does it actually show up with three `"`?

Thanks.

oneryalcin:
First of all thanks for your quick response @dhalperi, and secondly apologies. You are right, the config was quite long so I just wanted to get the relevant sections. Indeed the `aggreagate-ethernet` is inside of `network > interface`. I can confirm that. 

And the comment at the end edited, so in the original config there are no three double quotes. My bad.

Please see below the excerpt from the original config:
[...]

PR #5676, opened 2020-04-06
PAN: compute interface bandwidths

dhalperi:
PaloAlto ethernet interfaces have bandwidths that are either 1Gb/s or 10Gb/s
based on interface number.

Add setting, tests, and bandwidth inheritance to subinterfaces.

This is for #5598


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java, line 1259 at r1](https://reviewable.io/reviews/batfish/batfish/5676#-M4FPXTyAvE1WXE2xTDz:-M4FWlM04lHUkH_pQX5w:b-32v9sr) ([raw file](https://github.com/batfish/batfish/blob/93698db09e667266fd47863172f58ce980cc058d/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1259)):*

> ```
>     } else if (iface.getParent() != null) {
>       org.batfish.datamodel.Interface parentIface =
>           _c.getAllInterfaces().get(iface.getParent().getName());
>       assert parentIface != null; // because interfaces are processed in sorted order
>       newIface.setBandwidth(parentIface.getBandwidth());
>       // do not set speed, that's a physical property.
> ```

@progwriter this is the part I wish would be VI, unless that's crazy. 
</blockquote></details>

everything to do w/ bandwidth is crazy. it's unclear to me that subinterface bandwidth inheritance is actually vendor-independent. we'd have to run tests on cisco family before I'd commit to that.

PR #5691, opened 2020-04-15
Adding support for interface next-hops to FRR

kylehoferamzn:
Null0, blackhole, reject, and generic interfaces should be properly supported by this.


kylehoferamzn:
_"You mention this handles blackhole,reject. What are the semantics for these as the interface name? Is it the same as null routing?
In either case, it needs to be special cased in conversion, similar to how Null0 is. (+ tests)"_

Same as null routing yes, regrettably.  I think that FRR wants to get rid of Null0 syntax and go for something more explicit, but we need to support all three.  You can see this in FRR's unit tests.

Here's an example of how it actually looks in FRR
!
ip route 1.1.1.1/32 Null0
ip route 2.2.2.2/32 blackhole
ip route 3.3.3.3/32 reject
!

S>* 1.1.1.1/32 [1/0] unreachable (blackhole), weight 1, 00:00:39
S>* 2.2.2.2/32 [1/0] unreachable (blackhole), weight 1, 00:00:26
S>* 3.3.3.3/32 [1/0] unreachable (ICMP unreachable), weight 1, 00:00:04

The one difference I believe is that "reject" will result in an ICMP unreachable being sent.  I think that batfish doesn't really care about this, though, except for maybe some esoteric traceroute behavior.   I've updated the code a bit and provided some tests.  The way it works right now is that all this stuff is canonicalized back to "null_interface" in the VI.

_"this seems unused. did you mean to use it in the ip_route rule?"_

Good call - this was an artifact of previous revisions, I've just removed it.

_"just double checking: both can't be specified (like in cisco for example)?"_

You can, actually!

ip route 1.1.1.1/32 2.2.2.2 Null0 is acceptable syntax.

I didn't really plan on adding support for that here and now, though.  The parser will error out on this, as is - so will the handler - it checks that one of next-hop ip or interface MUST be null.




Additional note: I had a merge issue with the original commit and so I had to redo some stuff - you'll notice a fair bit more testing and some fixes to sv_route.  This commit _should_ be running, bit I haven't had a chance to execute all the tests.

kylehoferamzn:
Gonna move this over to a new PR since I goofed on the branch in my fork.

PR #5698, opened 2020-04-17
Adding next-hop interface support to static routes for FRR

kylehoferamzn:
Previous pull request: https://github.com/batfish/batfish/pull/5691

This adds support for syntax such as:

ip route x.x.x.x/x Null0
ip route x.x.x.x/x eth0
ip route x.x.x.x/x blackhole
ip route x.x.x.x/x Eth0

Does NOT support `ip route x.x.x.x/x y.y.y.y/y Null0` although that is valid syntax, the parser will continue to error out on that as it did previously.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ coverage*
 # Ignore python cache files
 __pycache__
 .pytest_cache/
+
+# Ignore Tokens
+*.tokens
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,3 @@ coverage*
 # Ignore python cache files
 __pycache__
 .pytest_cache/
-
-# Ignore Tokens
-*.tokens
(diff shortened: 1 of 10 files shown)


dhalperi:
___
*[.gitignore, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/5698#-M5NtmdE3XAH5gLH12PT:-M5NyHRRBHjaEws2qZgt:b-j80kbo) ([raw file](https://github.com/batfish/batfish/blob/447cb75274a4d621c98d595694462bc39eff95f1/.gitignore#L47)):*

NACK - this is not checked in on purpose, since users need to know that those files existing will mess everything up.
</blockquote></details>

not ignored on purpose, that is.

kylehoferamzn:
"NACK - this is not checked in on purpose, since users need to know that those files existing will mess everything up."

Fair play - I'll remove.


Re: CanonicalizeInterfaceName

Good eye, I worked for years on a Java project where it was this way.  

Public vs Private - I wasn't sure if this method would get re-used in the future, it seemed somewhat useful.  I'll switch it to private and if someone wants to use it in the future they can open it up/refactor.

PR #5700, opened 2020-04-19
Arista: try to clean up parse tree inside of interface

dhalperi:
Mostly, moving `ip foo` all into one place.


dhalperi:
This saves a surprising amount of parsing time.

PR #5732, opened 2020-04-28
Remove get-object

arifogel:
- remove v1 API call
- remove client support
- remove backend support


dhalperi:
Can you kick off the docker container test build as well, which makes sure this doesn't break old pybatfish?

PR #5741, opened 2020-04-30
Clean up IO operations in FileBasedId{Manager,Resolver}

arifogel:
- delegate to IO methods from FileBasedStorage to ensure atomicity
- remove dependency on CommonUtil


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java, line 1113 at r1](https://reviewable.io/reviews/batfish/batfish/5741#-M6CIR_bDywA3zcVBc1U:-M6H-DR_-F1v75HC0Iyv:b5w3f23) ([raw file](https://github.com/batfish/batfish/blob/19e73c3e787f431eb298ed639282ca0cc79a4244/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java#L1113)):*

So I went with some version of 2.
I don't like the idea of the ID manager/resolver constructing paths, but I at least now the security checks can be centralized.
</blockquote></details>

Why does the ID Manager need to think about files or paths instead of just IDs?

PR #5800, opened 2020-05-12
Add support for "force" with next-hop-self in Cumulus configs.

raveranj:


raveranj:
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_bgp.g4, line 306 at r2](https://reviewable.io/reviews/batfish/batfish/5800#-M7AxiTl89aBJjPrXp-O:-M7I_Xft6Sz8haFLgR2b:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/b56c70d7d7372bf41dbe19d5a64b933349389938/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_bgp.g4#L306)):*

Let's handle both `force` and `all` since they are the exact same command apparently
</blockquote></details>

Done.

raveranj:
___
*[projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java, line 695 at r2](https://reviewable.io/reviews/batfish/batfish/5800#-M7Ay3w-ElF1Yq8iaiq9:-M7I_d891fP_9gqj5Vb8:b-ort0ux) ([raw file](https://github.com/batfish/batfish/blob/b56c70d7d7372bf41dbe19d5a64b933349389938/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java#L695)):*

IIUC, `all` is new (and preferred syntax) while `force` is legacy syntax? 
If that's indeed the case, let's have the setters/getters use `All` as the suffix to match new documentation. Then here can leave comments linking to 
http://docs.frrouting.org/en/latest/bgp.html#clicmd-[no]neighborPEERnext-hop-self[all]
and 
https://github.com/FRRouting/frr/pull/4200 
indicating that they are indeed the same command.

</blockquote></details>

Thats right. Added comments with the links.

raveranj:
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConversions.java, line 778 at r2](https://reviewable.io/reviews/batfish/batfish/5800#-M7B-JfK4mYvB8_HtFwG:-M7I_lDD6_HK5y2xQa9X:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/b56c70d7d7372bf41dbe19d5a64b933349389938/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConversions.java#L778)):*

Leave a TODO comment indicating this won't work for dynamic neighbors.
</blockquote></details>

Done.

PR #5839, opened 2020-06-01
Static route distance support to FRR

kylehoferamzn:
Adding distance support to static routes in FRR.


kylehoferamzn:
Fixed up - good catch!

kylehoferamzn:
Fixed up hashcode - one thing to note is that I didn't worry too much about equality because you cant have duplicate static routes, at least not within a VRF - they replace each other.  Maybe there's some other mechanisms I should add to detect something faulty?  Just a thought.

PR #5842, opened 2020-06-01
BDD: add a method to get a deterministic seeded full satisfying assignment

dhalperi:


dhalperi:
___
*[projects/bdd/src/main/java/net/sf/javabdd/JFactory.java, line 3089 at r1](https://reviewable.io/reviews/batfish/batfish/5842#-M8qCfeL5FCENLASXrHX:-M8qhjD1ByplDdPpmSfs:b-ulxcw1) ([raw file](https://github.com/batfish/batfish/blob/666ca1cf6c42a782e3c1a2b01223a4e744eeaf9c/projects/bdd/src/main/java/net/sf/javabdd/JFactory.java#L3089)):*

It seems fundamentally wrong to push without ever popping. It would be cleaner to have the caller to bdd_makenode push/pop the refs they want to keep. E.g. bdd_makesatnode would push/pop child. This is what I think the binops are doing. They push the left child before building the right to make sure any gbc calls made while building the right child don't free the left child. Then they push the right child before building the new node (because that can call gbc too). Then they pop twice to restore the stack the original state (at the time the call began) before returning. It's easier to understand why its correct/safe if we restore the stack.
</blockquote></details>

I can take an action item to clean up the library, if it's important. This pattern is all over the place.

PR #5848, opened 2020-06-02
Restore input object charset detection

arifogel:
- Centralize and make consistent decoding logic
  - add utility functions to
    - decode stream and append newline if nonempty
    - just decode stream
  - for consistency, also try to add newline for iptables files
  - update most callers of detectCharset to use new API
- Refactor TestrigText to store byte arrays instead of Strings 
  - allows more control over test input
  - update getter/setter names to reflect new API


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/main/StreamDecoder.java, line 28 at r4](https://reviewable.io/reviews/batfish/batfish/5848#-M8vn-rv3xSVB6p006sF:-M8vn-rv3xSVB6p006sG:bf84yzc) ([raw file](https://github.com/batfish/batfish/blob/b1be64e9f31beea9788c5627bf7e8a8a9d91231c/projects/batfish/src/main/java/org/batfish/main/StreamDecoder.java#L28)):*
> ```Java
>   static @Nonnull String decodeStreamAndAppendNewline(@Nonnull InputStream inputStream)
>       throws IOException {
>     byte[] rawBytes = IOUtils.toByteArray(inputStream);
> ```

`inputStream` is not closed, I think?

PR #5857, opened 2020-06-03
Adding support for ClusterList and Originator Id to VI Question

kylehoferamzn:
I realize Im missing tests here, I'll add those in but before I do so I wanted to know if the general approach is what you all want


kylehoferamzn:
I realize Im missing tests here, I'll add those in but before I do so I wanted to know if the general approach is what you all want

PR #5902, opened 2020-06-17
AWS: security groups by eni where appropriate

dhalperi:
In general, security groups are associated with specific network interfaces. For EC2 instances, the "instance security groups" are just the [security groups of the primary interface](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html).

> Security groups are associated with network interfaces. After you launch an instance, you can change the security groups that are associated with the instance, which changes the security groups associated with the primary network interface (eth0).

As far as I can tell, the instance appliances like RDS and EDS do not explicitly create ENIs, so for those we still need to assign the sgs to all interfaces we create dynamically.


dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/5902#-MA2Xt3j2h_MozE3XeRl:-MAYAmHV1p8B8IijU2ty:b5isvq1) (no related file):*

Minimum todo:

* add a test with multiple interfaces with different sgs
* remove some of the code reuse.
</blockquote></details>

redundancy removed and test added

---

PR #5914, opened 2020-06-22
Network statement rms

kylehoferamzn:
This PR does a few things:

1) Makes it so network-statements are originated unconditionally
2) network statements apply their configured attribute maps

There are a few quirks however,

1) Batfish seems to treat the network statement attribute map and the BGP Peers configured route-map as one "big" route-map chain.  This results in prefixes not being advertised when they should.

This PR is also missing some potential tests for attribute-map behavior mostly because I wasn't sure how to mock that part.  However I did test it in a compiled version of Batfish and it works as expected, except for the above quirk.


kylehoferamzn:
"this is a pretty big functionality change.. i'd feel better if we had an e2e test with a mini network, asserting that routes propagate correctly, with and w/o the route being in the main rib"

Absolutely agree thanks for the example!

"I'm guessing that computeIfAbsent semantics are no longer correct. If network statement appears twice, each time with different routemap, the later one should win, correct? that's just a put then."

Yep.  Switched it around.

"I don't understand why this is necessary. Isn't ipv4u AF the default AF?"

So we had a discussion about this, but it was overly complicated to implement without a larger structural change to more code and I think that'd be best handled separately.  The way it works is that the default AF and the ipv4u AF are the same functionally, and FRR  prefers the ipv4u AF - if you add config it'll overwrite whats in the default AF.

I was left with two choices:  Implement a parser that simply stuffed everything into the ipv4u AF and forgot about the default VRF or keep with the existing convention of separating them.  I tried the former but it resulted in this code looking very different than the rest and a lot of conversions whenever the two VRFs needed to interact with each other.  I think we need to refactor _ALL_ default VRF behavior across Batfish to unify it and I didn't want to get into that here - I'd rather follow existing conventions - at leas thats why I implemented it the way I did.


"can routemaps on network statements can block the export of a prefix? if yes, then there is a subtle bug here -- evaluation of attribute policy in dataplane computation does not permit/deny the routes, it merely modifies them."

Indeed it can, good catch.  I'll modify the generation policy to accommodate.

"this code is no longer correct"

Why?  I need to advertise from both VRFs.  Sounds like I just need to adjust the generation policy here with the attached route-map?

kylehoferamzn:
> > I tried the former but it resulted in this code looking very different than the rest and a lot of conversions whenever the two VRFs needed to interact with each other
> 
> What do you mean by two VRFs interacting with each other?
> 
> > I think we need to refactor ALL default VRF behavior across Batfish
> > I don't quite follow. VI representation doesn't care about default VRF as a special construct, it is only vendor-specific.
> 
> > I need to advertise from both VRFs.
> > Multiple VRFs have nothing to do with the way networks are advertised. Each VRF is a self-contained entity. The code is not correct because if you leave it, we will double-advertise the networks, once via your new method and once via the old pipeline.

Good eye, I think that will only happen if we somehow goof up generated config (eg: frr.conf shouldnt have this) but its a good point.  I think I can fix this fairly easily, I'll get that in another commit.

> 
> > _[Reviewable](https://reviewable.io/reviews/batfish/batfish/5914#-:-MB_Hblv3guCDUQKOMuj:b-iurv4a)_ status: all files reviewed, 6 unresolved discussions (waiting on @kylehoferamzn)
> 
> _[projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java, line 419 at r1](https://reviewable.io/reviews/batfish/batfish/5914#-MB_J22T0jhTfGG_rsu1:-MB_J22U9eMSfhfraIVb:b-juroi0) ([raw file](https://github.com/batfish/batfish/blob/27fe539e96b3d8c0e31a2e6a01c00d43150c6fcc/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java#L419)):_
> 
> > ```java
> >                 new BgpNetwork(
> >                     Prefix.parse(ctx.prefix().getText()),
> >                     ctx.route_map_name() != null ? ctx.route_map_name().getText() : null));
> > ```
> 
> one more thing i forgot to mention: we need reference tracking for route maps. here and below

Will take a look at this.

Latest commit includes the test rig - thanks for the pointers there.  There's still that small behavior issue that doesn't make much sense whereby routes that seemingly *should* be advertised (in this case 10.2.2.2/32) are actually NOT advertised.  Need to figure out what's going on there.

kylehoferamzn:
Closing this out, after much investigation we've found this functionality will require enhancements to the way Batfish handles routes coming in and out of Main RIB and BGP Rib.  Will reference the branch for posterity:

https://github.com/kylehoferamzn/batfish/tree/network_statement_rms

PR #5926, opened 2020-06-24
FirewallSessionInterfaceInfo: fix deserialization of empty interfaces

dhalperi:
It may round-trip through json as null instead of empty.


corinaminer:
The `sessionInterfaces` field shouldn't be an empty list in the first place. #5932 will correctly enforce that invariant.

dhalperi:
Replaced by #5932

PR #5929, opened 2020-06-25
Precompute new fields in AwsConfiguration

corinaminer:
Tests in progress to confirm the content of the new fields.


progwriter:
___
*[projects/batfish/src/main/java/org/batfish/representation/aws/Region.java, line 611 at r2](https://reviewable.io/reviews/batfish/batfish/5929#-MAhDPez1A2i9BdN01y8:-MAhEC3XB9-6t2_r5vD8:bxjz8mu) ([raw file](https://github.com/batfish/batfish/blob/d83be775d2d8cb7184d13b4dec3ce81d1f7f62b4/projects/batfish/src/main/java/org/batfish/representation/aws/Region.java#L611)):*

Please no. This was intentionally changed to return an immutable collection that cannot be modified from other parts of the code.
</blockquote></details>

To clarify: public functions like these are used from other places that should not concern themselves with internal representation of VS configuration.

PR #5943, opened 2020-06-27
Don't make the AWS service gateway part of AWS aggregate

ratulm:


dhalperi:
This doesn't seem right. The AWS services are definitely part of AWS's network. Especially, given that someday we're going to regionalize them, they will be even further nested.

What motivates this change?

PR #5949, opened 2020-06-30
BDD reachability analysis: materialize the transposed table for fixpoint

dhalperi:
Large, sparse Tables have poor performance on column-oriented operations:
e.g., get the "column vector" for a given column key is O(#rows) since
it involves one lookup in each row's hashMap.

Since fixpoint does continual row-oriented operations, backwards fixpoint does
continual column-oriented operations (as seen in the forward edge table).
Merely transposing the input table leads to poor performance, instead we should also
materialize the transposed table. This is most important in cases that run many
backwards fixpoints over the same table.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysis.java
+++ b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysis.java
@@ -3,6 +3,7 @@ package org.batfish.bddreachability;
 import static org.batfish.bddreachability.BDDReachabilityUtils.computeForwardEdgeTable;
 import static org.batfish.bddreachability.BDDReachabilityUtils.getIngressLocationBdds;
 
+import com.google.common.base.Suppliers;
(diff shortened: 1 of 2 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java, line 120 at r1](https://reviewable.io/reviews/batfish/batfish/5949#-MB5nF3C5Ime9o1c-y8j:-MB5nF3C5Ime9o1c-y8k:b-sr71en) ([raw file](https://github.com/batfish/batfish/blob/a28292c3d794bee333b4d779c6f93eb9b287d51e/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L120)):*
> ```
> transposeAndMaterialize(forwardEdgeTable)
> ```

Q: should I change this one? My belief finally is that yes, the materialization overhead is a net amortized win if the seed states touch any sensible fraction of the network - and the forward edge table has likely already been optimized.

But happy to hear other views...

PR #5959, opened 2020-07-03
Add RDS and ESD instances to the first subnet for hierarchy purposes

ratulm:


ratulm:
Closing until we figure out some other modeling issues here.

PR #5976, opened 2020-07-09
Reachability: factor out constructFlows helper method

anothermattbrown:


dhalperi:
___
*[projects/symbolic/src/main/java/org/batfish/symbolic/BUILD, line 4 at r1](https://reviewable.io/reviews/batfish/batfish/5976#-MBl9rr4Di2sUDxe12_L:-MBl9rr59fZ_EB7EaX01:bsdzhy0) ([raw file](https://github.com/batfish/batfish/blob/ec47e4b814730a5eb57a209a50273423aa3ccbc7/projects/symbolic/src/main/java/org/batfish/symbolic/BUILD#L4)):*
> ```Starlark
> load("@batfish//skylark:pmd_test.bzl", "pmd_test")
> 
> package(default_visibility = ["//visibility:public"])
> ```

No one should depend on this lib, they should only depend on //projects/symbolic

PR #5982, opened 2020-07-09
AWS NLBs: No src NAT for instance targets

corinaminer:
FYI - the modified test seems like it was intended to do what it's doing now, but the author copy-pasted the `testComputeTargetTransformationStep_ipTarget` test and forgot to switch the target type to `INSTANCE` for `testComputeTargetTransformationStep_instanceTarget`.


corinaminer:
Combining with #5975 to make unit tests pass.

PR #5990, opened 2020-07-14
PAN: DNAT port translation conversion

sfraint:
Add conversion to VI model for Palo Alto dest NAT port translation.


sfraint:
Wrong merge target

PR #5992, opened 2020-07-14
AWS: Apply security groups to instance targets' new interfaces

corinaminer:
Note first commit message is inaccurate; it adds security groups on instance targets' interfaces, not NLBs'.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/aws/Region.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/aws/Region.java
@@ -836,8 +836,8 @@ public final class Region implements Serializable {
 
     // Must happen after ESD and RDS are converted (for security groups to get all referrer IPs) but
     // before subnets are converted (so that target instances' special interfaces can get the
-    // instance's security groups applied)
-    computeSecurityGroups(awsConfiguration, warnings);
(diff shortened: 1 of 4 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/aws/Subnet.java, line 436 at r1](https://reviewable.io/reviews/batfish/batfish/5992#-MC9qxc70Qf6TayzFGSN:-MC9sM3D63x7e-zPJ6Dl:blji7vt) ([raw file](https://github.com/batfish/batfish/blob/97be7cd75121cb0626e334df6379470409660f59/projects/batfish/src/main/java/org/batfish/representation/aws/Subnet.java#L436)):*

> ```
>         // New interface on instance target needs security groups
>         Set<String> instanceSecurityGroups =
>             region.getNetworkInterfaces().values().stream()
>                 .filter(ni -> instanceTarget.getId().equals(ni.getAttachmentInstanceId()))
>                 .flatMap(ni -> ni.getGroups().stream())
>                 .collect(ImmutableSet.toImmutableSet());
> ```

1. This is upside down. Get the primary eni ID from the instance.

2. why stream and then collect eni.groups? Just copy it.
</blockquote></details>


(instance target means primary interface of instance)

PR #6012, opened 2020-07-18
A question for symbolic route analysis

millstein:
First version of a SearchRoutePolicies question, analogous to SearchFilters but for route policies.  Supports searching for route announcements that are accepted/denied by a given route policy, and in the accepted case you can also specify constraints on the output announcement.


progwriter:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/Community.java, line 45 at r1](https://reviewable.io/reviews/batfish/batfish/6012#-MCXloGA88mS9X64VhrZ:-MCcY6OuDmY1zc7je4XK:b-c7e5i2) ([raw file](https://github.com/batfish/batfish/blob/949992d902bc016abf3b93338a93283e20deeb9e/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/Community.java#L45)):*

@progwriter: was not exposing this intentional for some reason? E.g., that we didn't want to pretend this function was a standard encoding since every vendor does something completely different?

e.t., 1.2.3.4:100 is an extended community in some vendors.
</blockquote></details>

correct, this parsing logic is only sound for our own internal representation of communities, which is only used in jackson (de-)serialization, and is wrong in the general case (i.e., when attempting to interpret vendor-specific community strings).

PR #6033, opened 2020-07-28
FRR Set Weight Support

kylehoferamzn:


kylehoferamzn:
"just double checking: This can only be literal integer, or can it be an expression (like metric) where you're allowed to say +5 or -10, etc."

From my experience with the CLI you can only set a literal integer and doesn't allow any increment or expressions.  (And now I know why you guys had IntExpr on Metric!)

"this is too much stuff in a single test"

As discussed, will rename this test and create a new, specific test.

PR #6099, opened 2020-08-17
Fix FBS GC tests on macos

arifogel:
- truncate test values of Instant.now() to seconds
  - setLastModifiedTimemacos on APFS only supports granularity of
seconds
  - fixes macOS bugs where same Instant was used for expungeBeforeTime and
setLastModifiedTime, yet subsequenet file last modified time preceded
expungeBeforeTime

### DIFF
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileBasedStorageTest.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileBasedStorageTest.java
@@ -602,8 +602,9 @@ public final class FileBasedStorageTest {
 
   @Test
   public void testExpungeOldEntriesDirectories() throws IOException {
-    Instant newTime = Instant.now().truncatedTo(ChronoUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


dhalperi:
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileBasedStorageTest.java, line 633 at r2](https://reviewable.io/reviews/batfish/batfish/6099#-MEyd9CD3ukcB70XZ1dE:-MEyd9CD3ukcB70XZ1dF:b-4988cv) ([raw file](https://github.com/batfish/batfish/blob/3355e39c50305e9ff9a0faba35113dbf5c2b535b/projects/batfish-common-protocol/src/test/java/org/batfish/storage/FileBasedStorageTest.java#L633)):*
> ```Java
>   @Test
>   public void testExpungeOldEntriesFiles() throws IOException {
>     Instant newTime = Instant.now().truncatedTo(ChronoUnit.SECONDS);
> ```

actually, this doesn't make any sense to me any more. Why do we have to round here, but not in storage?

PR #6167, opened 2020-09-02
BDD: fix randomFullSatOne entropy

dhalperi:
Checking the lowest bit of a PRNG-generated integer does not work well with the
Linear congruential generatoras (LCGs) that are standard in Java. For the
original implementation, it resulted in only two different random assignments.

Check the middle of the int instead, and add a simple (deterministic) test of
the entropy.

### DIFF
--- a/projects/bdd/BUILD
+++ b/projects/bdd/BUILD
@@ -38,8 +38,9 @@ junit_tests(
     deps = [
         ":bdd",
         ":bdd_testlib",
+        "@maven//:com_google_guava_guava",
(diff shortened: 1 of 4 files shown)


anothermattbrown (commented):
> Reviewed 3 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6167)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #6168, opened 2020-09-02
SearchRoutePolicies: AS-path support

millstein:
Support AS-path filtering in the searchRoutePolicies question.  Also support constraints on the AS paths of the input/output route announcements.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6168)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 233 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFcdIODAlDTRJe_R2P:-MGFcdIODAlDTRJe_R2Q:bvh7df1) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L233)):*
> ```
> It is useful when the goal is to produce concrete example routes from a BDDRoute, for instance.
> ```
The constraints don't always have to hold?
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 241 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFd5vDAbxQ-8pPdJSf:-MGFd5vDAbxQ-8pPdJSg:bv9fszq) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L241)):*
> ```
>     // at most one AS-path regex atomic predicate should be true, since by construction their
>     // regexes are all pairwise disjoint
> ```
in this case, can you just allocate log(N) BDD variables instead of N, for N=number of APs? That's what we do in BDDFiniteDomain. In fact, maybe you can use BDDFiniteDomain? It still has a well-formedness constraint, since there may be extra values.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6168)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 233 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFcdIODAlDTRJe_R2P:-MGG3Hzu4Pq6BWYvs0qS:b-s50wtp) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L233)):*

> ```
> It is useful when the goal is to produce concrete example routes from a BDDRoute, for instance.
> ```
The constraints don't always have to hold?
</blockquote></details>

The constraints do always have to hold.  But I don't think it suffices to enforce them during symbolic route analysis (TransferBDD).  That analysis computes a BDD representing the input routes that are accepted by the route map.  So ANDing the well-formedness constraints in there would ensure that any models of that BDD are feasible routes.  But if what a client is interested in is routes that are denied by the route map, then simply negating that whole thing would be incorrect, because it would not ensure that the well-formedness constraints hold.  So I think it's logically a separate notion.
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 241 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFd5vDAbxQ-8pPdJSf:-MGG4Y_b-bgA1OMJcA6O:b-gkw6wj) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L241)):*

> ```
>     // at most one AS-path regex atomic predicate should be true, since by construction their
>     // regexes are all pairwise disjoint
> ```
in this case, can you just allocate log(N) BDD variables instead of N, for N=number of APs? That's what we do in BDDFiniteDomain. In fact, maybe you can use BDDFiniteDomain? It still has a well-formedness constraint, since there may be extra values.
</blockquote></details>

Could do that, although I'm not sure if it's worth it given that I expect N to be small.  Also I've found in other projects that the smallest encoding is not always the best performing one, because it introduces spurious correlations.  For example, with the "unary" encoding after an AS-path regex match some subset of the choices will have the BDD 1 and the rest the BDD 0, while with the binary encoding there will be more complicated BDDs.  Happy to discuss of course.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6168)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 233 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFcdIODAlDTRJe_R2P:-MGGGUvR2c44l96aIAdo:b83ql16) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L233)):*

The constraints do always have to hold.  But I don't think it suffices to enforce them during symbolic route analysis (TransferBDD).  That analysis computes a BDD representing the input routes that are accepted by the route map.  So ANDing the well-formedness constraints in there would ensure that any models of that BDD are feasible routes.  But if what a client is interested in is routes that are denied by the route map, then simply negating that whole thing would be incorrect, because it would not ensure that the well-formedness constraints hold.  So I think it's logically a separate notion.
</blockquote></details>

makes sense. these points would be worth including in the comment.
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 241 at r1](https://reviewable.io/reviews/batfish/batfish/6168#-MGFd5vDAbxQ-8pPdJSf:-MGGDtJgEn48AlVZcLWW:babu7vg) ([raw file](https://github.com/batfish/batfish/blob/abf2d42fc255322d49f0ec46c20a53856d0ee0da/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L241)):*

Could do that, although I'm not sure if it's worth it given that I expect N to be small.  Also I've found in other projects that the smallest encoding is not always the best performing one, because it introduces spurious correlations.  For example, with the "unary" encoding after an AS-path regex match some subset of the choices will have the BDD 1 and the rest the BDD 0, while with the binary encoding there will be more complicated BDDs.  Happy to discuss of course.
</blockquote></details>

Yeah it'll be interesting to see how many APs we see in practice. The other benefit of the log(N) encoding is that it would eliminate the need for this n^2 constraint, since the AP identifiers would be mutually exclusive. In any case, it would be worth making a note of these trade-offs in the comment.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6168)* status: :shipit: complete! all files reviewed, all discussions resolved

anothermattbrown (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6168#-:-MGGwkYZ3W1OCc_N_xTu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6169, opened 2020-09-03
FRR: Adding support for interface shutdown command

kylehoferamzn:
Behavior tested is that if you issue shutdown in Zebra, the interface is admin down.  If you configure "no shut" it just deletes the command - similar to unconfigured.  For that reason I made it non-null and initialized to false as default.  /etc/network/interfaces state parsing happens prior and also controls (hence I do not have a no-shut force admin up for Zebra)

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4
@@ -659,6 +659,11 @@ SERVICE
   'service'
 ;
 
+SHUTDOWN
+:
+  'shutdown'
+;
+
 SOFT_RECONFIGURATION
 :
   'soft-reconfiguration'
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_interface.g4
@@ -14,7 +14,13 @@ s_interface
     | si_ip
     | si_ipv6
     | si_no
+    | si_shutdown
   )*
+;
+
+si_shutdown
+:
+  SHUTDOWN NEWLINE
 ;
 
 si_ip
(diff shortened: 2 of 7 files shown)


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6169)* status: all files reviewed, 1 unresolved discussion (waiting on @kylehoferamzn)

___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/FrrInterface.java, line 19 at r1](https://reviewable.io/reviews/batfish/batfish/6169#-MGKhAuaA6KO-8iZT5GW:-MGKhAuc7PQIUH0mwNAo:b1wal4n) ([raw file](https://github.com/batfish/batfish/blob/a42a971cbca73e6537eaea3f53d6be00b4a147ea/projects/batfish/src/main/java/org/batfish/representation/cumulus/FrrInterface.java#L19)):*
> ```Java
>   private final @Nonnull String _name;
>   private final @Nullable String _vrfName;
>   private @Nonnull Boolean _shutdown = false;
> ```

Are you absolutely sure the behavior you describe (shutdown = false being the default) is always true? i.e., it doesn't vary based on properties of iface (e.g., physical/l3/switchport/portchannel/etc.) Reason I ask, is because we've seen that w/ nxos for example. :roll_eyes: 

if yes, then ok to make this a primitive `boolean`
if no, then mark `@Nullable` and just do `firstNonNull(iface.getShutdown(), false)` in conversion. Later, we can replace `false` w/ dynamically-computed default value if needed.

kylehoferamzn (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6169)* status: all files reviewed, 1 unresolved discussion (waiting on @nonnull)

___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/FrrInterface.java, line 19 at r1](https://reviewable.io/reviews/batfish/batfish/6169#-MGKhAuaA6KO-8iZT5GW:-MGKjNCK5j6gFmKYvqT5:b-npffjj) ([raw file](https://github.com/batfish/batfish/blob/a42a971cbca73e6537eaea3f53d6be00b4a147ea/projects/batfish/src/main/java/org/batfish/representation/cumulus/FrrInterface.java#L19)):*

Are you absolutely sure the behavior you describe (shutdown = false being the default) is always true? i.e., it doesn't vary based on properties of iface (e.g., physical/l3/switchport/portchannel/etc.) Reason I ask, is because we've seen that w/ nxos for example. :roll_eyes: 

if yes, then ok to make this a primitive `boolean`
if no, then mark `@Nullable` and just do `firstNonNull(iface.getShutdown(), false)` in conversion. Later, we can replace `false` w/ dynamically-computed default value if needed.
</blockquote></details>

Well I'm not 100% sure of anything ;)

But I tested in in FRR 7.2 where "unconfigured" means admin up - for any case where the "shutdown" command isn't seen, Batfish's behavior will continue as-is.  All this code does is shutdown interfaces that are configured as "shutdown" in Zebra.  I tested no shutdown and all it does is remove the "shutdown" command (essentially brings it back to default)

I was gonna make it a primitive but then I can't annotate it with @Nonnull.  Figured it'd be useful for IDE inspection but on second thought, it'll probably catch the null anyway.  I'll go ahead and fix that.

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6169)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6171, opened 2020-09-03
SearchRoutePolicies: flexible regex constraints

millstein:
Introduces a new type for regex constraints on route announcements for SearchRoutePolicies.  These are used for constraints on both communities and AS paths.  Each regex constraint is optionally negated, to indicate that the regex should not be satisfied in the route announcement.  This provides more expressiveness than the previous approach, which could only negate an entire set of regex constraints, and it obviates the need for a separate flag to indicate that negation is desired.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6171)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #6179, opened 2020-09-08
use DagTraceRecorder

anothermattbrown:
Use `DagTraceRecorder` to reuse work and previously-constructed `Hop`s when computing traces. Can be significantly more efficient for networks with high ECMP.

Note: I had previously proposed plumbing through a parameter or using a debug flag to control which `TraceRecorder` is used. I have not done that yet because I'm now unsure it's worth the trouble/mess.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6179)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6213, opened 2020-09-16
PAN: add application trace details for security rules

sfraint:
Add top-level application tracing for Palo Alto security rules.
There is no recursive tracing for applications yet (similar to addresses and services).

----

For example, traces for rules matching an `application` might look something like this:

If the rule specifies an application group and a descendant of that group is matched:
```
Matched security rule RULE2
   Matched source address
      Matched address object addr2
   Matched destination address
      Matched address-group addr_group1
   Matched service application-default
      Matched application-group app_group1
```
or if the rule specifies an application directly, which is matched:
```
Matched security rule RULE2
   Matched source address
      Matched address object addr2
   Matched destination address
      Matched address-group addr_group1
   Matched service application-default
      Matched built-in application bgp
```

Because of how applications are matched (based on service values), they show up under the `Matched service application-default` trace details.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/palo_alto/ApplicationGroup.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/palo_alto/ApplicationGroup.java
@@ -21,7 +21,7 @@ public final class ApplicationGroup implements Serializable {
   }
 
   @VisibleForTesting
-  Set<String> getDescendantObjects(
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6213)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6229, opened 2020-09-23
AWS: ACLs representing security groups have 1 line per rule

corinaminer:
ACLs representing a security group's ingress permissions has 1 line per ingress rule; same idea for egress permissions.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/aws/IpPermissions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/aws/IpPermissions.java
@@ -2,7 +2,8 @@ package org.batfish.representation.aws;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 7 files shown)


progwriter (commented):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6229)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/representation/aws/IpPermissions.java, line 433 at r2](https://reviewable.io/reviews/batfish/batfish/6229#-MHwaw22C5A_Jao4GSnY:-MHwaw22C5A_Jao4GSnZ:b-w3l706) ([raw file](https://github.com/batfish/batfish/blob/6b12109c44b5ed6885839742be990ee5dc614138/projects/batfish/src/main/java/org/batfish/representation/aws/IpPermissions.java#L433)):*
> // TODO Should we set this trace element? If so, to what?

the important thing is to set trace elements on the individual `AclLineMatchExprs` which it looks like we do so I wouldn't worry about this one.

PR #6247, opened 2020-09-29
Parsing: shuffle jobs

dhalperi:
Since Java parallel streams are not self-balancing, it pays to randomize files.
It is common that

  filename -> device role

and it is almost certain that

  device role -> configuration complexity -> parsing work

so not grouping by name is likely to pay off.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6247)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #6249, opened 2020-09-29
Defintion/Reference: Reduce space used

dhalperi:
Changing from `SortedSet<Integer>` to `IntegerSpace` is a dramatic reduction in storage, especially for definition lines. In one large network, halfway through parsing there were about 60M objects in memory, of which 34M were in `DefinedStructureInfo`.

By using `IntegerSpace` instead, we reduce sets to ranges and dramatically cut down on the number of objects maintained. Only internal storage changes - see ViModel diffs. Question answers stay the same.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java
@@ -5,7 +5,9 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.Range;
 import java.io.Serializable;
+import java.util.stream.IntStream;
 import javax.annotation.Nonnull;
 
 public class DefinedStructureInfo implements Serializable {
@@ -38,6 +40,16 @@ public class DefinedStructureInfo implements Serializable {
   }
 
   public void addDefinitionLines(int... lines) {
+    _definitionLines = _definitionLines.toBuilder().including(lines).build();
+  }
+
+  public void addDefinitionLines(IntStream lines) {
+    IntegerSpace.Builder newLines = _definitionLines.toBuilder();
+    lines.forEach(newLines::including);
+    _definitionLines = newLines.build();
+  }
+
+  public void addDefinitionLines(Range<Integer> lines) {
     _definitionLines = _definitionLines.toBuilder().including(lines).build();
   }
 
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java
@@ -6,8 +6,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.io.Serializable;
-import java.util.SortedSet;
-import java.util.TreeSet;
 import javax.annotation.Nonnull;
 
 public class DefinedStructureInfo implements Serializable {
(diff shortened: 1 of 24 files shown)


anothermattbrown (commented):
> Reviewed 14 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6249)* status: 14 of 16 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java, line 33 at r1](https://reviewable.io/reviews/batfish/batfish/6249#-MIQfMPnD7bUT_WpAgWj:-MIQfMPnD7bUT_WpAgWk:bkxcs1e) ([raw file](https://github.com/batfish/batfish/blob/556118fbfd31f0f8337d9daffeb748b2d3f47f6a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java#L33)):*
> ```
> addDefinitionLines
> ```
Since this is called a lot of times, it seems pretty expensive to go back and forth between `IntegerSpace.Builder` and `IntegerSpace`. 

It seems much more efficient to collect into a mutable `TreeRangeSet`, and then convert to `IntegerSpace` in `getDefinitionLines`. Is there a reason not to do that?

anothermattbrown (commented):
> Reviewed 2 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6249)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java, line 33 at r1](https://reviewable.io/reviews/batfish/batfish/6249#-MIQfMPnD7bUT_WpAgWj:-MIQrUWi0aEHviVSz5gX:bv16mjo) ([raw file](https://github.com/batfish/batfish/blob/556118fbfd31f0f8337d9daffeb748b2d3f47f6a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java#L33)):*

> ```
> addDefinitionLines
> ```
Since this is called a lot of times, it seems pretty expensive to go back and forth between `IntegerSpace.Builder` and `IntegerSpace`. 

It seems much more efficient to collect into a mutable `TreeRangeSet`, and then convert to `IntegerSpace` in `getDefinitionLines`. Is there a reason not to do that?


</blockquote></details>

I do not think this is much of a problem in practice - in hierarchical and Cisco-style configs, this is all O(1). [It is relevant for flattened configs, like juniper].

What we really need is a clean builder -> built separation; what you're describing is a fairly ugly (IMO) workaround of a type that has gotten us in trouble in the past. Would make the TreeRangeSet transient so that it serializes and deserializes immutably?

Where this code is in the pipeline it seems hard to untangle into a proper builder -> built pattern. This PR moves this code, by far, out from being the bottleneck in either CPU or memory.

anothermattbrown (commented):
> Reviewed 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6249)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java, line 33 at r1](https://reviewable.io/reviews/batfish/batfish/6249#-MIQfMPnD7bUT_WpAgWj:-MIS3Orq69JIjptSVbA0:bxzbe3a) ([raw file](https://github.com/batfish/batfish/blob/556118fbfd31f0f8337d9daffeb748b2d3f47f6a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DefinedStructureInfo.java#L33)):*

I do not think this is much of a problem in practice - in hierarchical and Cisco-style configs, this is all O(1). [It is relevant for flattened configs, like juniper].

What we really need is a clean builder -> built separation; what you're describing is a fairly ugly (IMO) workaround of a type that has gotten us in trouble in the past. Would make the TreeRangeSet transient so that it serializes and deserializes immutably?

Where this code is in the pipeline it seems hard to untangle into a proper builder -> built pattern. This PR moves this code, by far, out from being the bottleneck in either CPU or memory.
</blockquote></details>

Yeah I agree about the builder pattern, but since we don't use that in this part of the code (and use lots of mutable data structures) I was picturing just keeping it mutable. `getDefinitionLines` could return an immutable view. If that's called a ton of times we could think about caching that (and yes using transient) there.

Anyway, not a blocker

PR #6252, opened 2020-09-30
Dataplane: remove configurations

progwriter:
DP should not be the source of truth for configurations.
This also should unblock some follow-up optimizations where we do not serialize Nodes to disk


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6252)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #6254, opened 2020-09-30
PrefixTracer: fix Neighbor inner class

dhalperi:
* a class that is static does not keep a pointer to its parent, using much less storage
* simplify the equals function


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6254)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6257, opened 2020-09-30
Dataplane: materialize on creation, separate full and partial DPs

progwriter:
Trying to optimize space used by final DP by removing all nodes and prefix tracer info from it.
Also, instead of mucking with suppliers, trying to materialize all data in DP on creation. 
This means, to speed up partial DP construction during DP computation, factor out a partial DP class that only does portion of the work.


anothermattbrown (commented):
> Reviewed 6 of 6 files at r1, 1 of 1 files at r2, 5 of 5 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6257)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6262, opened 2020-10-01
IBDPE: keep partial dataplane entirely in the loop

dhalperi:
Note that the builder has no context that it needs to preserve between
loops. By moving the builder into the loop, the memory it uses can be
freed when the loop ends.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6262)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6266, opened 2020-10-01
Location: subclasses should have different hashes

dhalperi:
This is basically similar to batfish/batfish#3855: hash codes should not be the same even if the two fields are.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java
@@ -3,7 +3,6 @@ package org.batfish.specifier;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
-import java.util.Objects;
 import javax.annotation.Nonnull;
 
 /** Identifies the {@link Location} of the link of an interface in the network. */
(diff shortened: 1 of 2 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6266)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java, line 57 at r1](https://reviewable.io/reviews/batfish/batfish/6266#-MIZyKVT8JOGvDQhRHbM:-MIZyKVT8JOGvDQhRHbN:b-pu3f0) ([raw file](https://github.com/batfish/batfish/blob/9ed00f878d7793d8d55c6f4bb11dba85bda3eb25/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java#L57)):*
> ```Java
>   @Override
>   public int hashCode() {
>     return 31 * 31 * getClass().hashCode() + 31 * _interfaceName.hashCode() + _nodeName.hashCode();
> ```

why not just add `getClass` to the `Objects.hash` call?

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java, line 57 at r1](https://reviewable.io/reviews/batfish/batfish/6266#-MIZyKVT8JOGvDQhRHbM:-MI_G1CD5KqP7PrFHhSS:bfojnmm) ([raw file](https://github.com/batfish/batfish/blob/9ed00f878d7793d8d55c6f4bb11dba85bda3eb25/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/InterfaceLinkLocation.java#L57)):*

why not just add `getClass` to the `Objects.hash` call?
</blockquote></details>

tl;dr: Objects.hash creates an array, which is unnecessary churn for code that may be hot.

PR #6270, opened 2020-10-01
Cisco-style: stop warning about OSPF settings on interface not in an area

dhalperi:
This is apparently really common behavior - show run all on many vendors
will include OSPF defaults. So this is not an indication of a likely
misconfiguration.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6270)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @ratulm)

PR #6278, opened 2020-10-05
IOS: stop explicitly listing reference types for concrete structures

dhalperi:
Listing reference types explicitly is important for structures that can be used
with abstract references, aka, ip access-list which can often refer to one of
half a dozen types. It's not necessary for things like route-map or crypto
constructs that are not referenced in an ambiguous way.

Remove this listing, and hey, find a bug or two on the way.


dhalperi:
@ratulm - would have simplified last few PRs I guess.

ratulm (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6278#-:-MIrP_dbD-FCzXSswhTb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6280, opened 2020-10-06
bazel: turn on two ErrorProne warnings

dhalperi:
And fix them. tl;dr: unused variables should be labeled as unused.

Also fixed at least one bug: NodeRoles was ignoring a parameter.

Trying to kill off PMD.


dhalperi:
Due to upstream effects, this looks like a bad idea.

PR #6282, opened 2020-10-06
ForwardingAnalysisImpl: inline two intermediate collections

dhalperi:
Rather than building two filtered maps that are only used once, filter on-the-fly.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6282)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6284, opened 2020-10-06
Search Route Policies: Remove redundant atomic predicate computation

millstein:
This PR changes the answerer so that it builds the symbolic Graph, which computes atomic predicates for the community and AS-path regexes in a configuration, once per node rather than once per node per route policy.  Since computing atomic predicates is potentially expensive, this may provide a useful performance improvement in some cases.  The PR also declares almost all of the Graph's internal state 'final', to document its read-only nature, which allows it to be safely reused.


anothermattbrown (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6284)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6291, opened 2020-10-08
PAN: Support for rule-type in security rules

ratulm:


dhalperi (commented):
> Reviewed 3 of 6 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 4 unresolved discussions (waiting on @ratulm and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java, line 647 at r2](https://reviewable.io/reviews/batfish/batfish/6291#-MJ9Ne5T5hAt093d0F-N:-MJ9Ne5T5hAt093d0F-O:b-3e08ya) ([raw file](https://github.com/batfish/batfish/blob/c7fa8b06b27c7c7eb303eec5f4e9549ea06c8822/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L647)):*
> ```Java
>             .isEmpty();
>     boolean toZoneInRuleTo =
>         !Sets.intersection(rule.getTo(), ImmutableSet.of(toZoneName, CATCHALL_ZONE_NAME)).isEmpty();
> ```

I think what you have written is probably correct, but I think it is confusing.

1. If either condition is false, return false.

2.
```
switch (firstNonNull(ruletype, UNIVERSAL)
case UNIVERSAL:
  return true;
case INTRAZONE:
  return fromZoneName.equals(toZoneName);
case INTERZONE:
  return !fromZoneName.equals(toZoneName);
default:
  throw missing statmeent or whatever.
```

Yes, PAN UI enforces that INTRAZONE => the from and to sets are equal, but I think we should not rely on that here. What if the text is human-generated?

If we want to rely on that here, we need to document it and enforce it sensibly (elsewhere? drop the rule in this case?)
___
*[projects/batfish/src/test/java/org/batfish/representation/palo_alto/PaloAltoConfigurationTest.java, line 651 at r2](https://reviewable.io/reviews/batfish/batfish/6291#-MJ9OlijEUcD1VAYfxt1:-MJ9OlijEUcD1VAYfxt2:bquyx7j) ([raw file](https://github.com/batfish/batfish/blob/c7fa8b06b27c7c7eb303eec5f4e9549ea06c8822/projects/batfish/src/test/java/org/batfish/representation/palo_alto/PaloAltoConfigurationTest.java#L651)):*
> ```Java
> 
>     rule.setRuleType(RuleType.INTRAZONE);
>     assertTrue(securityRuleApplies("A", "A", rule));
> ```

This condition should not pass: A is not a valid toZone
___
*[projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/rulebase-rule-type, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/6291#-MJ9Oyvm6dfYXT4EFZaT:-MJ9Oyvm6dfYXT4EFZaU:bo4159n) ([raw file](https://github.com/batfish/batfish/blob/c7fa8b06b27c7c7eb303eec5f4e9549ea06c8822/projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/rulebase-rule-type#L6)):*

> ```
> set rulebase security rules RULE2 from z1
> set rulebase security rules RULE2 to z2
> set rulebase security rules RULE2 source any
> set rulebase security rules RULE2 destination any
> set rulebase security rules RULE2 rule-type intrazone
> ```
</details>

this seems to violate the constraints?
___
*[projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/rulebase-rule-type, line 24 at r2](https://reviewable.io/reviews/batfish/batfish/6291#-MJ9P0N32FDZp623oAFU:-MJ9P0N32FDZp623oAFV:b-d6mic7) ([raw file](https://github.com/batfish/batfish/blob/c7fa8b06b27c7c7eb303eec5f4e9549ea06c8822/projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/rulebase-rule-type#L24)):*
> ```
> set rulebase security rules RULE3 destination any
> set rulebase security rules RULE3 rule-type universal
> ```

add a "default" case where the value is null?

dhalperi (commented):
> Reviewed 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJAgjIi0YEgynxEn8GY:bdzba8e) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

> ```
>     _mainConfiguration
>         .getVirtualSystems()
>         .values()
>         .forEach(this::removeInvalidIntrazoneSecurityRules);
> ```
</details>


@sfraint - better place to do this? `Extractor#getConfiguration` seems like an odd choice vs, say, during conversion. Do we do similar things elsewhere?

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJAhf9OCA7YeCE1Edw5:b-35lywf) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

> ```
>     _mainConfiguration
>         .getVirtualSystems()
>         .values()
>         .forEach(this::removeInvalidIntrazoneSecurityRules);
> ```

@sfraint - better place to do this? `Extractor#getConfiguration` seems like an odd choice vs, say, during conversion. Do we do similar things elsewhere?
</blockquote></details>

happy to defer to @sfraint's take. the rationale for doing it here: the parser itself produces a "clean" configuration and is in line with how we treat some other instances of the box rejecting config lines.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJAldPTE8Gu3xcsz07U:b5qhl9i) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*
>we treat some other instances of the box rejecting config lines.
Not sure what you mean? AFAIU, we reject a line in the parser only if it would have been rejected by the device when typed right there. That's not this -- the device would let you type these lines in, at least if the `rule-type` was after the `from`.
Can you point me at an example of what you're thinking of?

The other concern I have is that there's nothing in the API contract that links `getConfiguration` to "parsing is done". No guarantees about when or how many times it's called.

dhalperi (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJApCQ1-5z8NRGF4LXL:bs43ycd) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

>we treat some other instances of the box rejecting config lines.
Not sure what you mean? AFAIU, we reject a line in the parser only if it would have been rejected by the device when typed right there. That's not this -- the device would let you type these lines in, at least if the `rule-type` was after the `from`.
Can you point me at an example of what you're thinking of?

The other concern I have is that there's nothing in the API contract that links `getConfiguration` to "parsing is done". No guarantees about when or how many times it's called.
</blockquote></details>

Some context on rejection via Spencer: 

https://intentionet.slack.com/archives/C01C58KHB7F/p1602199116034100?thread_ts=1602198664.031900&cid=C01C58KHB7F

(We should test if the device behavior is based on whats current when it sees the rule-type command. So, entering `from z1` and then `rule-type intrazone` will fail unless `to z1` has been entered before. If that is the behavior, the cleanup discussion is moot.)

Your other concern is valid. Another place I considered was here: https://github.com/batfish/batfish/blob/0ec5bcd4049ed84382599be3d651213c15f30aa8/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoControlPlaneExtractor.java#L36

I couldn't find other examples of such post-parsing clean ups but doing it during conversion "felt wrong."

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJAsEKj4Kon6t6S8Fcw:bu3g6y7) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

Some context on rejection via Spencer: 

https://intentionet.slack.com/archives/C01C58KHB7F/p1602199116034100?thread_ts=1602198664.031900&cid=C01C58KHB7F

(We should test if the device behavior is based on whats current when it sees the rule-type command. So, entering `from z1` and then `rule-type intrazone` will fail unless `to z1` has been entered before. If that is the behavior, the cleanup discussion is moot.)

Your other concern is valid. Another place I considered was here: https://github.com/batfish/batfish/blob/0ec5bcd4049ed84382599be3d651213c15f30aa8/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoControlPlaneExtractor.java#L36

I couldn't find other examples of such post-parsing clean ups but doing it during conversion "felt wrong." 
</blockquote></details>

We have parsers that call `processParseTree` more than once, too. (Juniper).

--

I'm not quite sure why conversion is the wrong time to do this. It feels like "commit time" errors are best mapped to conversion -- at least, it's definitely not something that is part of walking the parse tree, it's something that comes from post-processing the extracted data from the parse tree.

Batfish-standard is to do this kind of thing during conversion; for example that's where we change undefined references into "match all" in Cisco. (That's also what Spencer suggested in the thread you linked.)

I could see a design with a middle phase between parsing and conversion -- parsing is literally just extracting the text into objects, then something where we "normalize" or "validate" (and/or "concretize"?) the parsed data, and then conversion where we blindly create VI objects from the VS ones.

That middle step would also be where we fill in every single null field with its defaults, for example. I could see why it's strange to propagate `firstNonNull(field, it's default)` all over conversion code.

Right now we don't really have that separation tho.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2848 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAu0nE933K-RQuiUPP:-MJAu0nE933K-RQuiUPQ:b-e0n432) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2848)):*
> ```Java
>   }
> 
>   /** Remove intrazone rules that don't have identical From and To zones */
> ```

FYI you'll have to merge master. It looks like your prior PR(s) conflicted with this one.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJDEffA9PmDjbfJdg30:b-3u6gat) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

We have parsers that call `processParseTree` more than once, too. (Juniper).

--

I'm not quite sure why conversion is the wrong time to do this. It feels like "commit time" errors are best mapped to conversion -- at least, it's definitely not something that is part of walking the parse tree, it's something that comes from post-processing the extracted data from the parse tree.

Batfish-standard is to do this kind of thing during conversion; for example that's where we change undefined references into "match all" in Cisco. (That's also what Spencer suggested in the thread you linked.)

I could see a design with a middle phase between parsing and conversion -- parsing is literally just extracting the text into objects, then something where we "normalize" or "validate" (and/or "concretize"?) the parsed data, and then conversion where we blindly create VI objects from the VS ones.

That middle step would also be where we fill in every single null field with its defaults, for example. I could see why it's strange to propagate `firstNonNull(field, it's default)` all over conversion code.

Right now we don't really have that separation tho.
</blockquote></details>

I'd say we should either:
1a. add this check in `PaloAltoConfigurationBuilder`
1b. (better version of 1a) add this check in a new pre-conversion step like @dhalperi suggested
2. add this check during conversion

Generally, I'd say for things PAN cli won't let you enter 1 seems appropriate, and for things you can enter but might be rejected during commit process 2 seems appropriate.

Since PAN enforces validation for `rule-type` when you enter it on the cli (not at commit time), I'm leaning slightly toward doing this in 1. (also I believe @ratulm confirmed `rule-type` comes after `to` / `from` in config dumps)

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJEsqub5ka_1GYwG_3R:bmx5i9x) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

I'd say we should either:
1a. add this check in `PaloAltoConfigurationBuilder`
1b. (better version of 1a) add this check in a new pre-conversion step like @dhalperi suggested
2. add this check during conversion

Generally, I'd say for things PAN cli won't let you enter 1 seems appropriate, and for things you can enter but might be rejected during commit process 2 seems appropriate.

Since PAN enforces validation for `rule-type` when you enter it on the cli (not at commit time), I'm leaning slightly toward doing this in 1. (also I believe @ratulm confirmed `rule-type` comes after `to` / `from` in config dumps)
</blockquote></details>

For 1, as long as we can do it in an `enter|exit_Rule` -- the part that corresponds to CLI entry time -- then I'm down.

dhalperi (commented):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm and @sfraint)

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJEu8iABdt_rSHw54vx:b-enaxs8) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

For 1, as long as we can do it in an `enter|exit_Rule` -- the part that corresponds to CLI entry time -- then I'm down.
</blockquote></details>

:thumbsup: I was thinking in the `rule-type` exit rule

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJF6UNZ2Sa-YoTj-BIu:be8g4e8) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

:thumbsup: I was thinking in the `rule-type` exit rule
</blockquote></details>

just so we are clear, we need checks here both 1) when rule-type is entered; and 2) after parsing is done. 2 is needed because the rule can be made invalid after the fact by changing its From/To zones. 

i'll add 1 to parsing time and build on dan's other PR to 2.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJF7Hji-YMFfIpJCEf8:b695a3x) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

just so we are clear, we need checks here both 1) when rule-type is entered; and 2) after parsing is done. 2 is needed because the rule can be made invalid after the fact by changing its From/To zones. 

i'll add 1 to parsing time and build on dan's other PR to 2. 
</blockquote></details>

My other PR is a discussion piece.

Instead, you could prevent users from changing zones if rule type has been set to intrazone.

If you want to stick to post-processing, just do it in conversion for now so we can move forward. That's what we do everywhere else in Batfish.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2844 at r3](https://reviewable.io/reviews/batfish/batfish/6291#-MJAgjIi0YEgynxEn8GX:-MJFbFwH8iiK2YySs31R:b-tt8cgw) ([raw file](https://github.com/batfish/batfish/blob/c54eb147af85e96c1994fd2b72e75419d87adbf9/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2844)):*

My other PR is a discussion piece.

Instead, you could prevent users from changing zones if rule type has been set to intrazone.

If you want to stick to post-processing, just do it in conversion for now so we can move forward. That's what we do everywhere else in Batfish.
</blockquote></details>

Did both during parsing and conversion, to get closer to device semantics (which checks during rule-type command and during commit, not during from/to commands). 

If we prevented zone changes for intrazone rules, the following config will behave differently on device and batfish:

```
from = zone A
to = zone A
rule-type = intrazone
from = B
rule-type = interzone
```

on device:  interzone rule from zone B -> A
on batfish: interzone rule from A -> A (because we rejected `from = B`)

dhalperi (commented):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: all files reviewed, 2 unresolved discussions (waiting on @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2485 at r6](https://reviewable.io/reviews/batfish/batfish/6291#-MJOPoJP2MePisfsN58Q:-MJOPoJP2MePisfsN58R:b4ekd4d) ([raw file](https://github.com/batfish/batfish/blob/842b9280811a0343812939ebd00f66c53878e15c/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2485)):*
> ```
> Intrazone security
> ```
Maybe something that indicates we skipped this directive?

Error: cannot set `rule-type intrazone` for a security rule with different ...
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java, line 706 at r6](https://reviewable.io/reviews/batfish/batfish/6291#-MJOQ01t2Yr_pNGZUMiT:-MJOQ01t2Yr_pNGZUMiU:b74cy6r) ([raw file](https://github.com/batfish/batfish/blob/842b9280811a0343812939ebd00f66c53878e15c/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L706)):*
> ```
> Invalid
> ```
Skipping invalid... ?

dhalperi (commented):
> Reviewed 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6291)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #6302, opened 2020-10-10
iosxr - unable to init


sflowers0:
Entire config file fails to be ingested due to the below config:

```
extcommunity-set rt ABC
  12345:678910
end-set
!
```

Parse error:
```
Details
"java.lang.NullPointerException
    at org.batfish.grammar.cisco_xr.CiscoXrControlPlaneExtractor.lambda$enterExtcommunity_set_rt$3(CiscoXrControlPlaneExtractor.java:1859)
    at java.util.HashMap.computeIfAbsent(HashMap.java:1127)
    at org.batfish.grammar.cisco_xr.CiscoXrControlPlaneExtractor.enterExtcommunity_set_rt(CiscoXrControlPlaneExtractor.java:1855)
    at org.batfish.grammar.cisco_xr.CiscoXrParser$Extcommunity_set_rtContext.enterRule(CiscoXrParser.java)
    at org.batfish.grammar.BatfishParseTreeWalker.enterRule(BatfishParseTreeWalker.java:27)
    at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:25)
    at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:28)
    at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:28)
    at org.antlr.v4.runtime.tree.ParseTreeWalker.walk(ParseTreeWalker.java:28)
    at org.batfish.grammar.cisco_xr.CiscoXrControlPlaneExtractor.processParseTree(CiscoXrControlPlaneExtractor.java:7546)
    at org.batfish.job.ParseVendorConfigurationJob.parseFile(ParseVendorConfigurationJob.java:368)
    at org.batfish.job.ParseVendorConfigurationJob.parse(ParseVendorConfigurationJob.java:442)
    at org.batfish.main.Batfish.getOrParse(Batfish.java:2505)
    at org.batfish.main.Batfish.lambda$serializeNetworkConfigs$48(Batfish.java:2580)
    at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
[...]

PR #6319, opened 2020-10-15
XR: fix a crash in ExtcommunitySetRt

dhalperi:
As always, parents should almost never call into their children's context.

Fixes batfish/batfish#6302. Note that the actual code inside that issue is wrong, the second number
has to be between 1 and 65535.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6319)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer and @sfraint)

PR #6329, opened 2020-10-20
Arista: add more commonly-used lexer tokens

dhalperi:
We've learned that because the lexer uses semantic predicates in the `VARIABLE` token,
such lexing decisions are not cached. To improve on this, add back many of the most
common lexer tokens (in `show run all`) to the grammar.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6329)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @progwriter)

PR #6344, opened 2020-10-21
Arista: add another show run all for bgp

dhalperi:
The test checks the overwriting-route-map bug we hit in BGP weight.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6344)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6352, opened 2020-10-22
JunOS: rework SRX cross-zone-policy

dhalperi:
We want to be able to analyze the pure-packet-header
parts of the cross-zone policy. To support this, move the
"matchSourceInterface in from-zone" part of the filter
from inside the zone~A~to~zone~B filter to the parent.

The traces change a little, for the better. Instead of "Matched
cross-zone policy", it's now "Permitted by .." and "Denied by ..".

Additionally, since from-zone junos-host is a real thing, moved the
"permit from junos-host" term to after cross-zone policy evaluation.

PR #6360, opened 2020-10-26
Juniper SRX: create a second copy of security policy for analysis

dhalperi:
The version used in the forwarding pipeline inlines the source interface check
for the from zone. For policy analysis, create another version without that
check.


dhalperi:
Replaces #6352 .

PR #6361, opened 2020-10-26
JunOS: keep the original filter and term names during conversion

dhalperi:
This make sure that tracing and source metadata are correct even when
producing an ACL with a different output name,


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6361)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6377, opened 2020-11-01
Don't use Internet source IPs that are blocked at ISPs

ratulm:
We block some some addresses at ISPs but were allowing them as valid source IPs at the Internet. That led to uninteresting failures when testing reachability from the Internet.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/isp/BlockReservedAddressesAtInternet.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/isp/BlockReservedAddressesAtInternet.java
@@ -45,7 +45,7 @@ public final class BlockReservedAddressesAtInternet implements IspTrafficFilteri
   }
 
   /** Prefix -> description for TraceElement (optional). */
-  private static final Map<String, String> BLOCKED_ADDRESS =
+  static final Map<String, String> BLOCKED_ADDRESS =
       ImmutableMap.<String, String>builder()
           .put("0.0.0.0/8", "RFC 6890")
           .put("10.0.0.0/8", "RFC 1918 private")
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/isp/IspModelingUtils.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/isp/IspModelingUtils.java
@@ -3,6 +3,7 @@ package org.batfish.common.util.isp;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Comparator.naturalOrder;
+import static org.batfish.common.util.isp.BlockReservedAddressesAtInternet.BLOCKED_ADDRESS;
 import static org.batfish.datamodel.BgpPeerConfig.ALL_AS_NUMBERS;
 import static org.batfish.datamodel.Configuration.DEFAULT_VRF_NAME;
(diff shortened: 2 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6377)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6377#-ML6BBaQ0Qnns6xn99-a:-ML6BBaQ0Qnns6xn99-b:b-d1c0o2) (no related file):*
Can you say more about the issue? This doesn't seem right. The ISP blocking is a configurable ISP policy.

---

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6377)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6377#-ML6BBaQ0Qnns6xn99-a:-ML6UxZfDQ6_YYEozuXP:bgwsi4e) (no related file):*

Can you say more about the issue? This doesn't seem right. The ISP blocking is a configurable ISP policy.
</blockquote></details>

I have services in my network that I expect are reachable from the internet. I wanted to write a reachability query to confirm that this was indeed the case. But I kept running into failure cases because certain sources IP were being filtered by the ISPs. I didn't provide any special ISP configuration. 

At first I tried eliminating the source IP space that the Internet null_routes, but that space is a subset of the default ISP-filtering space.  Seems like if we are going to automatically filter some source IP space at the ISP, the Internet shouldn't use that space by default. 

Is there another way to address this problem? 

---

dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/6377#-ML6BBaQ0Qnns6xn99-a:-ML9-wv_AmRhcDI7bUuy:busgc9b) (no related file):*

I have services in my network that I expect are reachable from the internet. I wanted to write a reachability query to confirm that this was indeed the case. But I kept running into failure cases because certain sources IP were being filtered by the ISPs. I didn't provide any special ISP configuration. 

At first I tried eliminating the source IP space that the Internet null_routes, but that space is a subset of the default ISP-filtering space.  Seems like if we are going to automatically filter some source IP space at the ISP, the Internet shouldn't use that space by default. 

Is there another way to address this problem? 
</blockquote></details>

We have seen networks where customers did NOT want ISPs to block private traffic, so they could confirm that their border ACL properly blocks private subnets - they had an outage where the ISP accidentally did send traffic (and routes) for private IP space.

I think the definition of "Internet source IPs" is customer-specific and needs to be in the check.

---

dhalperi:
*[a discussion](https://reviewable.io/reviews/batfish/batfish/6377#-ML6BBaQ0Qnns6xn99-a:-ML90OcgER7zEWZoCuXh:b8hu4nz) (no related file):*

We have seen networks where customers did NOT want ISPs to block private traffic, so they could confirm that their border ACL properly blocks private subnets - they had an outage where the ISP accidentally did send traffic (and routes) for private IP space.

I think the definition of "Internet source IPs" is customer-specific and needs to be in the check.
</blockquote></details>

In the network you're testing, is the first rule of the border ACL equivalent to "block all traffic originating from known spammers"? That is something I've seen in nearly every real network, and it's likely going to cause you the same problem.

---

ratulm:
Will move toward user-level control of source IPs.

PR #6392, opened 2020-11-05
Cisco IOS: improve ospf->bgp redistribution, support matching on route types

progwriter:
This fixes 3 bugs:
1. A spurious warning since `ctx.MATCH()` is a list and always non-null
2. Adds support for matching specific route types in redistribute statement
3. Not specifying match types means redistributing both OSPF and OSPF_IA, not just OSPF (intra-area)

I'm aware this is a subpar solution and we need to make VS datamodel for redistribution more NXOS-like, but wasn't going to overhaul a lot of the VS model to unblock this bug fix.


dhalperi (commented):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6392)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6395, opened 2020-11-05
PAN: parse and ignore application-override

progwriter:
This doesn't affect our analysis (it only affects App-ID/DPI), but can generate so may parse warnings, it drowns any useful info out.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4
@@ -163,6 +163,11 @@ APPLICATION_GROUP
     'application-group'
 ;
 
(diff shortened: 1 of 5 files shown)


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6395)* status: all files reviewed, 1 unresolved discussion (waiting on @progwriter)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6395#-MLTvi5vB4Q9bor71rMd:-MLTvi5w2WcEP8GBC7Lz:biwy4oz) (no related file):*
Like we chatted about offline, functionally probably fine for now, but application-override does interact with security rule behavior.

We should at least warn in the case where Batfish might be ignoring something that affects security-rule-analysis behavior.

---

progwriter:
Closing for now, will revisit with a different solution

PR #6396, opened 2020-11-05
NXOS: parse more EIGRP interface properties

progwriter:
At this time only hello-interval and hold-time and leave placeholder tests for proper extraction and conversion of other properties

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4
@@ -1255,6 +1255,8 @@ EIGRP
       case KEY_CHAIN:
         pushMode(M_TwoWords);
         break;
+      case HELLO_INTERVAL:
+      case HOLD_TIME:
       case MODE:
       case REDISTRIBUTE:
       case ROUTER:
@@ -1715,6 +1717,11 @@ HEX_UINT32
 HMM
 :
   'hmm'
+;
+
+HOLD_TIME
+:
+  'hold-time'
 ;
 
 HOST
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxos_interface.g4
@@ -388,6 +388,8 @@ i_ip
     | i_ip_address
     | i_ip_authentication
     | i_ip_dhcp
+    | i_ip_hello_interval
+    | i_ip_hold_time
     | i_ip_forward
     | i_ip_igmp
     | i_ip_null
@@ -467,6 +469,16 @@ i_ip_dhcp_relay
 i_ip_forward
 :
   FORWARD NEWLINE
+;
+
+i_ip_hello_interval
+:
+  HELLO_INTERVAL EIGRP router_eigrp_process_tag time = uint16 NEWLINE
+;
+
+i_ip_hold_time
+:
+  HOLD_TIME EIGRP router_eigrp_process_tag time = uint16 NEWLINE
 ;
 
 i_ip_igmp
--- a/projects/batfish/src/test/java/org/batfish/grammar/cisco_nxos/CiscoNxosGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/grammar/cisco_nxos/CiscoNxosGrammarTest.java
@@ -7613,4 +7613,18 @@ public final class CiscoNxosGrammarTest {
             .getAllowAsIn(),
         nullValue());
   }
+
+  @Test
+  public void testInterfaceEigrpPropertiesExtraction() {
+    String hostname = "nxos_interface_eigrp";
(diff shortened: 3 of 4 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6396)* status: all files reviewed, 1 unresolved discussion (waiting on @progwriter)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4, line 1258 at r1](https://reviewable.io/reviews/batfish/batfish/6396#-MLPut6OEot9CSRFHWbp:-MLPut6OEot9CSRFHWbq:b-pjrdu8) ([raw file](https://github.com/batfish/batfish/blob/41b251d366c8ed09e3c51eab51541c427cfa2e0d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4#L1258)):*
> ```ANTLR
>         pushMode(M_TwoWords);
>         break;
>       case HELLO_INTERVAL:
> ```

looks like `HELLO_INTERVAL` already existed -- where else is it used and will changing mode there mess things up?

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6396)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4, line 1258 at r1](https://reviewable.io/reviews/batfish/batfish/6396#-MLPut6OEot9CSRFHWbp:-MLQ9Fno2o6Orwv2zddQ:b-lo13ev) ([raw file](https://github.com/batfish/batfish/blob/41b251d366c8ed09e3c51eab51541c427cfa2e0d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4#L1258)):*

looks like `HELLO_INTERVAL` already existed -- where else is it used and will changing mode there mess things up?
</blockquote></details>

other things (like ospf) should not be affected. the main token in question here is `EIGRP`. if the _previous_ token was `HELLO_INTERVAL` we switch modes.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6396)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4, line 1258 at r1](https://reviewable.io/reviews/batfish/batfish/6396#-MLPut6OEot9CSRFHWbp:-MLQQePM4N4pS24fdUGE:b-l79fl7) ([raw file](https://github.com/batfish/batfish/blob/41b251d366c8ed09e3c51eab51541c427cfa2e0d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4#L1258)):*

other things (like ospf) should not be affected. the main token in question here is `EIGRP`. if the _previous_ token was `HELLO_INTERVAL` we switch modes.
</blockquote></details>

Unblocking bc it sounds like you have a good grasp on this, but i still don't quite understand how `pushMode(M_Word)` interacts with the `EIGRP` token and why it doesn't affect other things; does `M_Word` make the parser change behavior if the next token is specifically `EIGRP`?

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6396)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4, line 1258 at r1](https://reviewable.io/reviews/batfish/batfish/6396#-MLPut6OEot9CSRFHWbp:-MLQuFdAE28DLahejnlq:bqdt5zv) ([raw file](https://github.com/batfish/batfish/blob/41b251d366c8ed09e3c51eab51541c427cfa2e0d/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_nxos/CiscoNxosLexer.g4#L1258)):*

Unblocking bc it sounds like you have a good grasp on this, but i still don't quite understand how `pushMode(M_Word)` interacts with the `EIGRP` token and why it doesn't affect other things; does `M_Word` make the parser change behavior if the next token is specifically `EIGRP`?
</blockquote></details>

The reason to push mode `M_Word` is so that the lexer correctly lexes the `WORD` token (which is only available in that mode). We need the `WORD` token because that's how `router_eigrp_process_tag` is defined (it can be an arbitrary name, w/o spaces).

Normally lexer would switch modes based on a _single_ token. However w/ `EIGRP` it's not always possible. For example `router eigrp <process_tag>` should push mode just like `ip hello-interval eigrp <process_tag>`. But _not_ `eigrp router-id ...` (because `router-id` is not a `WORD`).

The solution we have for this is: we look one token back (thus `lastTokenType()`) and special case pushing mode only some of the time. 
This does not affect things like `hello-interval ospf` (as an example) because the primary condition for pushing is still `EIGRP` token.
This _could_ become an issue if there were `hello-interval eigrp <process_tag>` and `hello-interval eigrp <some_keyword>` rules (in which case may need to look even more tokens back), but happily for us, that's not the nxos grammar.

PR #6413, opened 2020-11-11
Ignore Loopback as InterfaceLinkLocation TR sources

ratulm:


dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java, line 285 at r1](https://reviewable.io/reviews/batfish/batfish/6413#-MLsL0WeB369PwJscCdi:-MLsL0Wf4eERa9SsOH3L:be7jwvb) ([raw file](https://github.com/batfish/batfish/blob/5720ac2d6b34270ec44c0bd37ba754ee27e0c06f/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java#L285)):*
> ```
> any non-loopback interface;
> ```

why non-Loopback specifically? What about other virtual interfaces, other interfaces with /32? Maybe this needs to generalize the logic elsewhere, just allowing /31 and /30 also.

Interfaces need to have L3 IP addresses
* have a LLA
* have a ConcreteIpAddress with prefix < 32

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6413)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java, line 285 at r1](https://reviewable.io/reviews/batfish/batfish/6413#-MLsL0WeB369PwJscCdi:-MLsxZ301O09MgVSEdpi:bn9f7j9) ([raw file](https://github.com/batfish/batfish/blob/5720ac2d6b34270ec44c0bd37ba754ee27e0c06f/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java#L285)):*

> ```
> any non-loopback interface;
> ```

why non-Loopback specifically? What about other virtual interfaces, other interfaces with /32? Maybe this needs to generalize the logic elsewhere, just allowing /31 and /30 also.

Interfaces need to have L3 IP addresses
* have a LLA
* have a ConcreteIpAddress with prefix < 32
</blockquote></details>

Are these the right conditions then?
  to-enter: non-loopback && (LLA || concreteAddrPrefix < 32)
  to-start: at least one configured address (LLA or concrete)

Do I need to pay attention to InterfaceType?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6413)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java, line 285 at r1](https://reviewable.io/reviews/batfish/batfish/6413#-MLsL0WeB369PwJscCdi:-MLsyz6lCNNusC7XtqK3:bt43aa1) ([raw file](https://github.com/batfish/batfish/blob/5720ac2d6b34270ec44c0bd37ba754ee27e0c06f/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java#L285)):*
> Are these the right conditions then?

AFAICT that seems like the right generalization of sourceWithIPs default inference.

> Do I need to pay attention to InterfaceType?

We don't do that in the dataplane anywhere else, I wouldn't. (We might, e.g., deactivate an interface based on management, etc.)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6413)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6433, opened 2020-11-17
RoutingPolicy: don't build deep unions

dhalperi:
In very long routing policies, especially when they have lots of
continue statements, we might compile them into fairly complicated
structures with Call statements. Sets.union is a shallow operation that
never coalesces, so it seems plausible that building a deep union will
cause stack overflows on containment checks.

To fix this, materialize the sets as we go. This should not in practice
be very expensive unless we are in a case that would cause a stack
overflow with the prior approach.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6433)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown and @corinaminer)

dhalperi:
The bug reported by @dannypetrov took place on Arista with a route-map with 1146 terms.

PR #6434, opened 2020-11-18
TracerouteEngine: add an interface to pass a TraceRecorder

dhalperi:


dhalperi:
We're going to go a different direction.

PR #6436, opened 2020-11-18
TracerouteEngine: add an interface that exposes the TraceDag

dhalperi:
Not all users want the full list of traces. We may want to do other things, such as looking for any path, or randomly sampling the paths, or looking for a particular hop.

* Convert TraceDag into an interface
* Move it to common, along with tests and some utilities that are of common code.
* Add an interface to expose TraceDag in TracerouteEngine
* As an example, update one caller and its tests.


anothermattbrown (commented):
> Reviewed 7 of 7 files at r1, 5 of 5 files at r2, 6 of 6 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6436)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6447, opened 2020-11-20
IOS: tweak default interface delay inference

progwriter:
Handle the most common portchannel cases. Not a perfect solution (we may need to move this to snapshot post-processing as well) but a good enough stop-gap fix.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
@@ -1909,7 +1909,8 @@ public final class CiscoConfiguration extends VendorConfiguration {
     EigrpMetricValues values =
         EigrpMetricValues.builder()
             .setDelay(
-                firstNonNull(iface.getDelay(), Interface.getDefaultDelay(iface.getName(), _vendor)))
+                firstNonNull(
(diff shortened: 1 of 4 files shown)


corinaminer (commented):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6447)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/Interface.java, line 273 at r2](https://reviewable.io/reviews/batfish/batfish/6447#-MMbL2JC68rE-F2GDcyL:-MMbL2JC68rE-F2GDcyM:b2qaknu) ([raw file](https://github.com/batfish/batfish/blob/b878c1ad4d25213507e7927016924924c0c1057c/projects/batfish/src/main/java/org/batfish/representation/cisco/Interface.java#L273)):*
> ```Java
>      * When bandwidth > 1Gb, this formula is used. The interface may report a different value.
>      * For bandwidths < 1Gb, the delay may be interface type-specific.
>      * See https://tools.ietf.org/html/rfc7868#section-5.6.1.2
> ```

i know this isn't new as of your PR, so no action needed, but... according to rfc this formula is only applicable if EIGRP is using wide metrics

PR #6449, opened 2020-11-20
Remove unused fields from external BGP announcements

ratulm:
The srcNode, srcVrf, and dstVrf fields were never being read. 

Minor other cleanups with setting defaults for unset fields.


arifogel (commented):
> Reviewed 1 of 2 files at r1, 5 of 5 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6449)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6449#-MMb1J3rCkvkgxe4_OUW:-MMb1J3s7fqWXPLhnLdi:b-mwwvi1) (no related file):*
The title of this PR refers to these as "unused fields". However, they are used in `equals`, `hashCode`, and `compareTo`.
Are these advertisements never placed in a `Set` or `SortedSet` where those fields might differ? IIRC they were at some point. If you can make this guarantee, then it should be safe to remove these fields. Evidence is in favor of removal since the getters were not called.

---

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6449)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6449#-MMb1J3rCkvkgxe4_OUW:-MMrRl2C2FRFVCmVmCPz:b4th6hl) (no related file):*

The title of this PR refers to these as "unused fields". However, they are used in `equals`, `hashCode`, and `compareTo`.
Are these advertisements never placed in a `Set` or `SortedSet` where those fields might differ? IIRC they were at some point. If you can make this guarantee, then it should be safe to remove these fields. Evidence is in favor of removal since the getters were not called.
</blockquote></details>

They are being placed in a SortedSet (and I don't understand why). Instead of pulling on that thread, I'll just make these inputs optional as they were before. Will send a different PR. 

---

PR #6454, opened 2020-11-24
EnumSetSpecifier: distinguish between unspecified and empty

dhalperi:
We have documented `properties='/no-such-property/'` as a way to get just
the key of the properties. However, we broke this when we reimplemented
the EnumSetSpecifier. Add a test, then fix it by making EnumSetValues
understand the difference between unspecified and empty. Only
unspecified should default to everything.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/parboiled/ParboiledEnumSetSpecifier.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/parboiled/ParboiledEnumSetSpecifier.java
@@ -10,6 +10,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 import org.batfish.specifier.EnumSetSpecifier;
 import org.parboiled.errors.InvalidInputError;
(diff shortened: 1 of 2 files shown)


ratulm (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6454#-:-MMrsqg16XY61z1BCepN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/questions/NodePropertySpecifierTest.java, line 41 at r1](https://reviewable.io/reviews/batfish/batfish/6454#-MMrtCVrEnurHQimrYgL:-MMrtCVs9aWceUkIHT3X:b-naaro2) ([raw file](https://github.com/batfish/batfish/blob/42850bdc275300d518beea646c42f2277045d661/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/questions/NodePropertySpecifierTest.java#L41)):*
> ```Java
> 
>   @Test
>   public void testRegexNoMatch() {
> ```

Such a test could (also) go in ParboiledEnumSetSpecifierTest.java since the change applies to everything that uses EnumsSetSpecifier.

dhalperi:
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/questions/NodePropertySpecifierTest.java, line 41 at r1](https://reviewable.io/reviews/batfish/batfish/6454#-MMrtCVrEnurHQimrYgL:-MMrupJlC1ZzvT_Mz7Ys:bsw3c0z) ([raw file](https://github.com/batfish/batfish/blob/42850bdc275300d518beea646c42f2277045d661/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/questions/NodePropertySpecifierTest.java#L41)):*

Such a test could (also) go in ParboiledEnumSetSpecifierTest.java since the change applies to everything that uses EnumsSetSpecifier. 
</blockquote></details>

Yeah, I was trying to figure this out. It looks like there's no raw unit tests of the specifier -- the ones you mentioned are actually tests of various other specifiers, like NamedProperty or RoutingProtocol. Given that, seemed cleaner to just put it in with the tested property.

PR #6456, opened 2020-11-24
ASA: extract snmp source-interface and trap servers

### LINKED ISSUE #6426: ASA: SNMP_Trap_Servers not parsed
Syntax options:
` snmp-server host interface { hostname \| ip_address } [ trap \| poll ] [ community community-string ] [ version { 1 \| 2c \| 3 username }] [ udp-port port ]`

Syntax description:
`Specifies the recipient of an SNMP notification. Indicates the interface from which traps are sent. Identifies the name and IP address of the NMS or SNMP manager that can connect to the ASA. The trap keyword limits the NMS to receiving traps only. The poll keyword limits the NMS to sending requests (polling) only. By default, SNMP traps are enabled. By default, the UDP port is 162. The community string is a shared secret key between the ASA and the NMS. The key is a case-sensitive value up to 32 alphanumeric characters. Spaces are not permitted. The default community-string is public. The ASA and ASASM use this key to determine whether the incoming SNMP request is valid. For example, you could designate a site with a community string and then configure the ASA or ASASM and the NMS with the same string. The ASA and ASASM use the specified string and does not respond to requests with an invalid community string. For more information about SNMP hosts, see the “SNMP Hosts” section.Note When SNMP Version 3 hosts are configured on the ASA and ASASM, a user must be associated with that host. To receive traps, after you have added the snmp-server host command, make sure that you configure the user on the NMS with the same credentials as the credentials configured on the ASA and ASASM.`

Example of actuall config from devices:
`
snmp-server host inside X.X.X.X poll version 3 some_user
`

Properties " SNMP_Trap_Servers" and "SNMP_Source_Interface" and not populated.

dhalperi:
Split parser along ASA lines to better support this syntax.

Docs: https://www.cisco.com/c/en/us/td/docs/security/asa/asa90/configuration/guide/asa_90_cli_config/monitor_snmp.html

Fix #6426


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6456)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6465, opened 2020-12-02
Arista: port NX-OS route-map logic

dhalperi:
Prior to this PR, Arista modeled route-maps as nested structures, so that if
there were N terms, those terms would go into a recursive structure that was N
terms deep. For large N (hundreds+) this can cause recursion limits in Java,
e.g., in walking the route-map or in serializing it to disk.

Fix this by adopting (nearly verbatim) the logic from NX-OS.

Validated by existing unit tests + Batfish Enterprise compare view: routes are
the same before and after in many real networks.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/arista/AristaConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/arista/AristaConfiguration.java
@@ -39,6 +39,7 @@ import static org.batfish.representation.arista.eos.AristaRedistributeType.OSPF_
 import static org.batfish.representation.arista.eos.AristaRedistributeType.OSPF_NSSA_EXTERNAL_TYPE_2;
 import static org.batfish.representation.arista.eos.AristaRedistributeType.STATIC;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6465)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/representation/arista/AristaConfiguration.java, line 1903 at r1](https://reviewable.io/reviews/batfish/batfish/6465#-MNVyF8s0D1dEED92ekz:-MNVyF8s0D1dEED92el-:bg36ob9) ([raw file](https://github.com/batfish/batfish/blob/f522a52db3a66dda0539ba91dbbb7f5431042a59/projects/batfish/src/main/java/org/batfish/representation/arista/AristaConfiguration.java#L1903)):*
> ```Java
>       // Continue: on match, change the default.
>       if (action == LineAction.PERMIT) {
>         trueStatements.add(Statements.SetLocalDefaultActionAccept.toStaticStatement());
> ```

codecov reveals we need new unit tests of continue statements

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6465)* status: all files reviewed, 1 unresolved discussion

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6465)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6479, opened 2020-12-04
L3 topology inference: fix a bug with aggregate subinterfaces and LLAs

progwriter:
At some point we introduced ability to connect subinterfaces of aggregate interfaces via link-local addresses.
But the code had a bug where we would connect such interfaces even if they had non-LLA, concrete IP addresses.
This fixes that bug.


arifogel (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6479#-:-MNfC95yBVM6OFUP-dFs:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #4684, opened 2019-09-01
Some NX-OS EIGRP commands are not recognized


Tachashi:
I tested NX-OS EIGRP route analysis on batfish/allinone:latest-159 by using two NX-OSv9000 switches.

The snippet of configs are below:

- nexus1

```
hostname nexus1

feature eigrp
feature interface-vlan

vlan 2222

interface Loopback0
 ip address 10.1.1.1 255.255.255.255

interface ethernet 1/1
 no switchport
 ip address 192.168.100.1/24
 ip router eigrp Test1
 no shutdown

interface ethernet 1/2
 switchport access vlan 2222
 no shutdown

interface Vlan2222
 ip address 192.168.200.1/24
 ip router eigrp Test1
 ip passive-interface eigrp Test1
 no shutdown

ip route 0.0.0.0/0 192.168.200.2

route-map Route1 permit 10
 match interface Vlan2222

router eigrp Test1
 autonomous-system 33
 redistribute static route-map Route1
 default-metric 1000000 10 255 1 1500
```

- nexus2

```
hostname nexus2

feature eigrp
feature interface-vlan

vlan 3333

interface Loopback0
 ip address 10.2.2.2 255.255.255.255
 ip router eigrp Test1

interface Vlan3333
 ip address 192.168.100.2/24
 ip router eigrp Test1
 ip distribute-list eigrp Test1 prefix-list Prefix1 out
 no shutdown

interface ethernet 1/1
 switchport access vlan 3333
 no shutdown

router eigrp Test1
 autonomous-system 33

ip prefix-list Prefix1 seq 10 permit 10.2.2.2/32
```

In Cisco VIRL simulation environment, the outputs of EIGRP neighbor and routing table are below:

```
nexus1# sh ip eigrp neighbor
IP-EIGRP neighbors for process 33 VRF default
H   Address                 Interface       Hold  Uptime  SRTT   RTO  Q  Seq
                                            (sec)         (ms)       Cnt Num
0   192.168.100.2           Eth1/1          14   00:23:54  14   84    0   3
```

```
nexus1# sh ip route
IP Route Table for VRF "default"
'*' denotes best ucast next-hop
'**' denotes best mcast next-hop
'[x/y]' denotes [preference/metric]
'%<string>' in via output denotes VRF <string>

0.0.0.0/0, ubest/mbest: 1/0
    *via 192.168.200.2, [1/0], 00:24:03, static
10.1.1.1/32, ubest/mbest: 2/0, attached
[...]

PR #6485, opened 2020-12-08
NXOS iface conversion: Use EIGRP VRF ASN if process ASN not present

corinaminer:
An EIGRP process configuration can specify its ASN in two ways. It can be set directly in the router EIGRP line:
```
! ASN is 1
router eigrp 1
```
Or it can be set on a per-VRF basis using `autonomous-system` (no VRF specified means default VRF):
```
! ASN is 2 for default VRF
router eigrp EIGRP_PROC
  autonomous-system 2
```
In the latter case, we were failing to propagate the ASN to EIGRP interfaces during conversion, which meant those interfaces didn't have EIGRP in the VI model and thus couldn't establish EIGRP edges.

With this PR, the sample configs in #4684 establish an EIGRP edge.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java
@@ -2020,14 +2020,12 @@ public final class CiscoNxosConfiguration extends VendorConfiguration {
       }
     }
     if (eigrpProcess != null) {
-      // TODO Does EigrpVrfConfiguration ASN take precedence over EigrpProcessConfiguration ASN?
-      Integer asn = eigrpProcess.getAsn();
-      if (asn == null) {
-        asn =
-            Optional.ofNullable(eigrpProcess.getVrf(vrfName))
-                .map(EigrpVrfConfiguration::getAsn)
-                .orElse(null);
-      }
+      // Find process ASN for this interface. If an ASN is explicitly configured for the interface's
+      // VRF, it takes precedence over process ASN (which is just process tag interpreted as ASN).
+      Integer asn =
+          Optional.ofNullable(eigrpProcess.getVrf(vrfName))
+              .map(EigrpVrfConfiguration::getAsn)
+              .orElse(eigrpProcess.getAsn());
       if (asn != null) {
         String importPolicyName = eigrpNeighborImportPolicyName(ifaceName, vrfName, asn);
         String exportPolicyName = eigrpNeighborExportPolicyName(ifaceName, vrfName, asn);
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java
@@ -2019,23 +2019,31 @@ public final class CiscoNxosConfiguration extends VendorConfiguration {
         }
       }
     }
-    if (eigrpProcess != null && eigrpProcess.getAsn() != null) {
-      String importPolicyName =
-          eigrpNeighborImportPolicyName(ifaceName, vrfName, eigrpProcess.getAsn());
-      String exportPolicyName =
-          eigrpNeighborExportPolicyName(ifaceName, vrfName, eigrpProcess.getAsn());
-      generateEigrpPolicy(_c, this, iface.getEigrpInboundDistributeList(), importPolicyName);
-      generateEigrpPolicy(_c, this, iface.getEigrpOutboundDistributeList(), exportPolicyName);
+    if (eigrpProcess != null) {
+      // TODO Does EigrpVrfConfiguration ASN take precedence over EigrpProcessConfiguration ASN?
+      Integer asn = eigrpProcess.getAsn();
+      if (asn == null) {
+        asn =
+            Optional.ofNullable(eigrpProcess.getVrf(vrfName))
+                .map(EigrpVrfConfiguration::getAsn)
+                .orElse(null);
+      }
+      if (asn != null) {
+        String importPolicyName = eigrpNeighborImportPolicyName(ifaceName, vrfName, asn);
+        String exportPolicyName = eigrpNeighborExportPolicyName(ifaceName, vrfName, asn);
+        generateEigrpPolicy(_c, this, iface.getEigrpInboundDistributeList(), importPolicyName);
+        generateEigrpPolicy(_c, this, iface.getEigrpOutboundDistributeList(), exportPolicyName);
 
-      newIface.setEigrp(
-          EigrpInterfaceSettings.builder()
-              .setAsn(eigrpProcess.getAsn().longValue())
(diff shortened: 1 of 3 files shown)


dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java, line 1997 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzs1MR0PjTggs4M6td:-MNzs1MR0PjTggs4M6te:b38hmcs) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java#L1997)):*
> ```
>       // TODO Does EigrpVrfConfiguration ASN take precedence over EigrpProcessConfiguration ASN?
>       Integer asn = eigrpProcess.getAsn();
>       if (asn == null) {
> ```
if this is like BGP, the "process ASN" is just the ASN for the default vrf. Then you cannot enter the `autonomous-system` command in the default vrf, and the process ASN is not used at all in the VRF.

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6485)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @corinaminer, @dhalperi, and @progwriter)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java, line 1997 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzs1MR0PjTggs4M6td:-MNzsxyz5VYz7rjkvdGM:b15kyr) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java#L1997)):*

> ```
>       // TODO Does EigrpVrfConfiguration ASN take precedence over EigrpProcessConfiguration ASN?
>       Integer asn = eigrpProcess.getAsn();
>       if (asn == null) {
> ```
if this is like BGP, the "process ASN" is just the ASN for the default vrf. Then you cannot enter the `autonomous-system` command in the default vrf, and the process ASN is not used at all in the VRF.
</blockquote></details>

You can enter the command if the process tag is not a number: https://www.cisco.com/c/en/us/td/docs/switches/datacenter/sw/4_2/nx-os/unicast/configuration/guide/l3_cli_nxos/l3_eigrp.html#wp1062353

I suspect the command takes precedence 
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_interface_eigrp, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzrxu46LRMLO5RK-AM:-MNzrxu46LRMLO5RK-AN:b-a30fpm) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_interface_eigrp#L17)):*
> ```
> !
> interface Ethernet1/1
>   vrf member VRF
> ```

looking more carefully, do we convert this as belonging to process 1? `router eigrp 1` is defined for default vrf only but the interface is in VRF1. This seems suspect, have you verified this?

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java, line 1997 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzs1MR0PjTggs4M6td:-MNzxYid9idOK-SviUx2:b-77f25m) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java#L1997)):*

You can enter the command if the process tag is not a number: https://www.cisco.com/c/en/us/td/docs/switches/datacenter/sw/4_2/nx-os/unicast/configuration/guide/l3_cli_nxos/l3_eigrp.html#wp1062353

I suspect the command takes precedence 
</blockquote></details>

yeah I was just wrong. Likely, the thing that is inherited across vrfs is the "instance ID". If that is an ASN, I think all vrfs get it by default and (defaulting to instance ID) and can override it locally.

If not,  every vrf needs to override or it will be inactive; should NOT be inherited from default vrf.

(Those are guesses based on BGP experience).

Also suspect that vrf-af are not inherited from default vrf

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6485)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java, line 1997 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzs1MR0PjTggs4M6td:-MO9138DBFi16uu9b9Ws:bggkfcj) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/CiscoNxosConfiguration.java#L1997)):*

yeah I was just wrong. Likely, the thing that is inherited across vrfs is the "instance ID". If that is an ASN, I think all vrfs get it by default and (defaulting to instance ID) and can override it locally.

If not,  every vrf needs to override or it will be inactive; should NOT be inherited from default vrf.

(Those are guesses based on BGP experience).

Also suspect that vrf-af are not inherited from default vrf
</blockquote></details>

Did some lab testing. Confirmed that VRF AS takes precedence if configured, and cleared up some other behavior questions. see details [here](https://docs.google.com/document/d/1J4TQU17vsBQZNRQuU6FDwJxmN7c4dj-x2bW0WFBmDCg)
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_interface_eigrp, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/6485#-MNzrxu46LRMLO5RK-AM:-MO9-mfi3POZ8PDzf6I8:bjx6uld) ([raw file](https://github.com/batfish/batfish/blob/510c8866d0865dfa508da70450713b67f930125d/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_interface_eigrp#L17)):*

looking more carefully, do we convert this as belonging to process 1? `router eigrp 1` is defined for default vrf only but the interface is in VRF1. This seems suspect, have you verified this? 
</blockquote></details>

yes (now). Tested that an interface in non-default vrf can have eigrp if it declares the process in its config:
```
router eigrp 1  
interface Ethernet1/1
  no switchport
  vrf member vrf1
  ip address 10.12.11.1/24
  ip router eigrp 1
  no shutdown
```

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6485)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6492, opened 2020-12-10
L3 topology: keep virtual edges when L1 is present

progwriter:
Some interfaces are purely virtual and will not appear in L1. We still should keep those edges, e.g., Cisco has Vasi interfaces.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6492)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6496, opened 2020-12-11
NXOS EIGRP: Avoid spurious warning

corinaminer:
If an interface is configured to match a given EIGRP process (with `ip router eigrp TAG`) and that same EIGRP process is also configured to include the interface (via a network statement), we were generating a spurious warning that the interface matches multiple EIGRP processes.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6496)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6508, opened 2020-12-14
NXOS EIGRP redist: No policy if no supported protocols

corinaminer:
Do not create an EIGRP redistribution policy if none of the redistributed protocols are supported (doesn't make a practical difference, but the policy is pointless if created because it will reject everything).


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6508)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6512, opened 2020-12-16
EIGRP external routes: Maintain tag across nodes

corinaminer:
See sections 5.4.1.1 and 5.4.1.2 in RFC 7868:
https://tools.ietf.org/html/rfc7868#section-5.4.1.1

Tested on NXOS that these routes keep their tags after export (i.e. still have the same tag on the device to which they were exported):
- Tagged static route redistributed into EIGRP
- BGP route redistributed into EIGRP; its tag is its original BGP ASN
- BGP route that goes through a redistribution policy with `set tag`

Concluded that it is always correct to preserve external EIGRP routes' tags.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6512)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6515, opened 2020-12-21
BDP: factor out wrapper around routing policies for each node

progwriter:
Gets us a slightly cleaner APIs over just maps.
Helps us migrate towards better patterns: individual routing processes should not have direct access to configuration (in the long term)

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -97,7 +97,10 @@ final class BgpRoutingProcess implements RoutingProcess<BgpTopology, BgpRoute<?,
   /** Configuration for this process */
   @Nonnull private final BgpProcess _process;
   /** Parent node configuration */
-  @Nonnull private final Configuration _c;
+  @Deprecated @Nonnull private final Configuration _c;
+
+  @Nonnull private final RoutingPolicies _policies;
+  @Nonnull private final String _hostname;
   /** Name of our VRF */
   @Nonnull private final String _vrfName;
   /** Reference to the parent {@link VirtualRouter} main RIB (read-only). */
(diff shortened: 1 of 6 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6515)* status: all files reviewed, 1 unresolved discussion (waiting on @progwriter)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1300 at r1](https://reviewable.io/reviews/batfish/batfish/6515#-MP60rqyB2RT5WwalkyR:-MP60rqyB2RT5WwalkyS:bsvd414) ([raw file](https://github.com/batfish/batfish/blob/01386e815a6364d8f56e6cafc6571e3e0379671f/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1300)):*
> ```Java
>       return null;
>     }
>     RoutingPolicy exportPolicy = _policies.getOrThrow(exportPolicyName);
> ```

any particular reason to use `getOrThrow` here but `assert exportPolicy != null` above? If you're just leaving behavior as you found it, that's fine

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6515)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1300 at r1](https://reviewable.io/reviews/batfish/batfish/6515#-MP60rqyB2RT5WwalkyR:-MP6PxOb-hBSivMlzI9F:bmu3un2) ([raw file](https://github.com/batfish/batfish/blob/01386e815a6364d8f56e6cafc6571e3e0379671f/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1300)):*

any particular reason to use `getOrThrow` here but `assert exportPolicy != null` above? If you're just leaving behavior as you found it, that's fine
</blockquote></details>

yep, for now just not changing too much. over time we should probably migrate to `getOrThrow` because that results in more meaningful error messages containing node/policy name.

PR #6525, opened 2020-12-23
IOS-XE: bgp address-family import path

dhalperi:
This is for BGP Event-based VPN Import, which affects convergence but not our
modeling.

https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/iproute_bgp/configuration/xe-16-8/irg-xe-16-8-book/irg-event-vpn-import.html

Keeps parser properly in context.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6525)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #6527, opened 2020-12-26
IOS dynamic NAT rule order: sort by ACL name

corinaminer:

PR #6524, opened 2020-12-23
IOS: Support VRF-specific NAT rules

corinaminer:

PR #6528, opened 2020-12-26
IOS NAT rules with interface instead of pool

corinaminer:
We support dynamic NAT rules of this form:
`ip nat inside source list 1 pool pool1`

But it is possible to use an interface in place of a pool. This causes the interface's address to be the sole member of the pool (IOS console automatically adds `overload` to this line to allow for more than one NAT entry):
`ip nat inside source list 1 interface Ethernet1/1`

Still to do: This PR does not support NAT rules of the above form where the interface is inactive or has no address; such rules will be ignored. However, on real IOS devices, these rules still match traffic, and matched traffic is dropped.

After #6524 and #6527 (which impacts test).

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_nat.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_nat.g4
@@ -89,8 +89,11 @@ ipnosm_add_route
 
 ipnis_list
 :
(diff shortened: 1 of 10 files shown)


dhalperi (commented):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6528)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6538, opened 2021-01-05
BDDPacket: enhance and expose sane flow constraint

dhalperi:
* TCP/UDP port 0 are actually invalid, should not be used for picking representative flows
* Expose the constraint so it can be used in other places than just picking flows


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6538)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6545, opened 2021-01-06
Routes: refactor to introduce a generic next hop

progwriter:
Let routes have a generic next hop, and require it always be set to reduce likelihood of nonsense next hop combinations.
Next hop (currently) can be one of: discard, next hop ip, next hop interface (with optional ARP ip). Existing functions to modify next hop kept for backwards compatibility but they are not preferred.

Suggest reviewing commit by commit:
- Core data structures
- Update routes classes and production code
- Fixup test wherever we were making nonsense nexthops, make use of testBuilders for convenience
- Ref updates.

Note: introducing next VRF as a special type of next hop is coming up in a followup PR.

PR #6550, opened 2021-01-07
Make nextVrf a special type of next hop

progwriter:
Add a new type of next hop that indicates that a destination needs to be re-resolved in a different VRF. Now such resolution can be applied to any route, not just to static routes.

Follow up on #6545


dhalperi (commented):
> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6550)* status: all files reviewed, 1 unresolved discussion (waiting on @progwriter)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java, line 168 at r1](https://reviewable.io/reviews/batfish/batfish/6550#-MQPTujg1W7Mb0nut0W3:-MQPTujg1W7Mb0nut0W4:b93api8) ([raw file](https://github.com/batfish/batfish/blob/e3ca16bb25f50cb6c46975bdf12aefff46b25fc4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java#L168)):*
> ```Java
>             @Override
>             public FibAction visitNextHopVrf(NextHopVrf nextHopVrf) {
>               return new FibNextVrf(nextHopVrf.getVrfName());
> ```

Extract single static class instance like others? Looks like it needs no local state

dhalperi:
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java, line 168 at r1](https://reviewable.io/reviews/batfish/batfish/6550#-MQPTujg1W7Mb0nut0W3:-MQPY7X1E5fOOQBZ2voB:b-ujbyxd) ([raw file](https://github.com/batfish/batfish/blob/e3ca16bb25f50cb6c46975bdf12aefff46b25fc4/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java#L168)):*

Extract single static class instance like others? Looks like it needs no local state
</blockquote></details>

no, I guess it does reference `node`.

PR #6551, opened 2021-01-07
BgpRoute: remove discard parameter, use correct next hop type

progwriter:
We had this workaround going since as early as #1834. After #6545 we can implement it just with next hops.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6551)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6561, opened 2021-01-12
Use final L3 topology for OSPF session compatibility

ratulm:
Without this change, any OSPF edges that get established later (e.g., tunnel-based edges) were not being considered.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dhalperi, @progwriter, and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6561#-MQsffr0DzsdD00ccBZ6:-MQsffr0DzsdD00ccBZ7:b-3w5guz) (no related file):*
Hmm, usually the "compatibility" version is Pre-Dataplane and the "status" version is Post-Dataplane. 

The Post-Dataplane version can indeed have edges that don't appear pre-dataplane.

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dhalperi, @progwriter, and @ratulm)

___
*[projects/question/src/main/java/org/batfish/question/ospfsession/OspfSessionCompatibilityAnswerer.java, line 64 at r1](https://reviewable.io/reviews/batfish/batfish/6561#-MQsg0vr6wKOIGlYTnEh:-MQsg0vr6wKOIGlYTnEi:brbc10g) ([raw file](https://github.com/batfish/batfish/blob/c51dcd6a33e75191cdf10a976b4563c49706257d/projects/question/src/main/java/org/batfish/question/ospfsession/OspfSessionCompatibilityAnswerer.java#L64)):*
> ```
> _batfish.getSnapshot()
> ```
definitely don't want this change -- the `snapshot` passed in is the one to use.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dhalperi, @progwriter, and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6561#-MQsffr0DzsdD00ccBZ6:-MQsggaKEvPyrSqv0wib:b-lsaibs) (no related file):*

Hmm, usually the "compatibility" version is Pre-Dataplane and the "status" version is Post-Dataplane. 

The Post-Dataplane version can indeed have edges that don't appear pre-dataplane.
</blockquote></details>

I see. There is no 'status' version for OSPF.  Let's close this PR if that is the invariant we want to maintain. The PR was inspired by user confusion in which there edges in ospfEdges question that didn't appear in ospfSessionCompatibility question. If we don't make this change, we should update the documentation for ospfSessionCompatibility. 

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @progwriter and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6561#-MQsffr0DzsdD00ccBZ6:-MQshfIA-kH9D13Q7eM-:bavfpse) (no related file):*

I see. There is no 'status' version for OSPF.  Let's close this PR if that is the invariant we want to maintain. The PR was inspired by user confusion in which there edges in ospfEdges question that didn't appear in ospfSessionCompatibility question. If we don't make this change, we should update the documentation for ospfSessionCompatibility. 
</blockquote></details>

We could rename this one to status, use the final topology, and then fix it up, too. Compatibility vs Status may make less sense for OSPF than BGP?

---

saparikh:
There is value in having a compatibility check for OSPF. 2 routers could be physically connected, but the MTU, area number, md5 key, network type (p2p, broadcast, nbma), etc... could be wrong. Also, in some scenarios like nbma you configure the remote peer ip explicitly, instead of relying on multicast to discover the remote end.

ratulm:
Circling back here, it looks like BGP compatibility considers all possible pairs (modulo some gaps around NAT of course) and judges that based on config settings. BGP status will then help decide if connections happens in the data plane. 

Taking that view, I think we should leave this question as OspfSessionCompatibility since it considers configuration-level compatibility and simply expand the set of pairs considered. An OspfStatus question will then judge actual sessions (which may still get blocked via L2/L3 ACLs). 

@dhalperi - makes sense? (I will fix the which-snapshot-to-use thing).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @progwriter and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6561#-MS055Ql0ySdX6975CbN:-MS055Ql0ySdX6975CbO:b-ple2vp) (no related file):*
Planned change makes sense.

* mark the question are requiring dataplane
* update templates/tests
* update docs, if needed, in pybatfish

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6561)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @progwriter)

PR #6562, opened 2021-01-13
IOS: extract nat overload, add comments

dhalperi:


corinaminer (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6562#-:-MQxJVX58OR4X33oIfOP:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoIosDynamicNat.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/6562#-MQxJDXWDZy0_TgtnUKP:-MQxJDXWDZy0_TgtnUKQ:bz1bly2) ([raw file](https://github.com/batfish/batfish/blob/f96d7c0deac443bbfc6b703a0644a62be0650363/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoIosDynamicNat.java#L47)):*
> ```Java
>   //
>   // The bug modeling this could catch is if a session can traverse the nat rule but some downstream
>   // filter blocks the port-translated flows, which only shows up under load.
> ```

may be worth mentioning in the comment that overload is only available for inside source dynamic nat.

PR #6563, opened 2021-01-13
PAN: fix up default BGP AF capabilities

progwriter:
AllowRemoteAsOut is true; Warn in the case where we do not support correct enable-sender-side-loop-detection semantics.


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6563)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6572, opened 2021-01-16
Hotfix disable-peer-as-check for nxos

progwriter:
This implements the more correct behavior for nxos as a stop-gap. 
https://www.cisco.com/c/en/us/support/docs/ios-nx-os-software/nx-os-software/214619-configure-disable-peer-as-check-in-bgp-o.html

Looks at last AS in the AS path, instead of the entire AS path.
We need to do more vendor tests, and likely we will need to expand our VI model to support more variations on eBGP loop prevention


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6572)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown and @corinaminer)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6572)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #6581, opened 2021-01-21
IpsecPeerConfig: Require valid local address

corinaminer:
Building the IPsec topology can fail if we create peer configs with invalid (`Ip.AUTO`) local addresses.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6581)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6595, opened 2021-01-26
remove erroneous periods from F_SubnetMask

arifogel:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6595)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6595#-MS-E88X6o43R4DBS90N:-MS-E88Y3JZhBsjno7DM:bvw82o2) (no related file):*
What's the impact? Do we need a new test?

---

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6595)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6595#-MS-E88X6o43R4DBS90N:-MS-FL1pCO-W40weDCyw:bb5h2ow) (no related file):*

What's the impact? Do we need a new test?
</blockquote></details>

Looks like there should be no impact for FRR (except that illegal things like 255.255.A255.0 are no longer allowed), because the only rule that uses it also accepts `IP_ADDRESS`.

For NXOS it seems like there should be an impact, unless tests have been using the two alternatives unaffected by the change.

AFAICT no tests have failed, so I should probably add some.

---

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6595)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @arifogel)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6595#-MS-E88X6o43R4DBS90N:-MS-JxhR3dyX9lQ9K44t:b-nl5nmn) (no related file):*

Looks like there should be no impact for FRR (except that illegal things like 255.255.A255.0 are no longer allowed), because the only rule that uses it also accepts `IP_ADDRESS`.

For NXOS it seems like there should be an impact, unless tests have been using the two alternatives unaffected by the change.

AFAICT no tests have failed, so I should probably add some.
</blockquote></details>

added test. manually confirmed it fails prior to fix from this PR.

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6595)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

PR #6251, opened 2020-09-29
Bdp: refactor BGP to to simplify dataflow and speed up a bit

progwriter:
Several improvements:
- Clear VR -> BgpRoutingProcess pipeline that conforms to the redistribute/pull model (like other processes). Moved nearly all BGP code into BgpRoutingProcess
- Clear(er) endOfRound calls for VR and BGP
- No more re-queuing of entire RIBs on each iteration: keep track of previous round's main RIB delta
- No more processing of external advertisements each iteration, only once
- Got rid of `newSessionEstablishedHook` and inlined logic into
`getOutgoingRoutesPerEdge`
- Fewer occurrences of crazy rib delta merging

PR #6606, opened 2021-01-31
bgp: fix detection of new BGP sessions

dhalperi:
Fix a bug introduced in refactoring from batfish/batfish#6251, where
some routes would not be exchanged on BGP links that come up after
the snapshot starts. Fairly rare in datacenters, because BGP links in
datacenters are usually ebgp singlehop.

When sending routes to new links, we need to use the main RIB from
the previous round rather than the current RIB, because depending on
iteration order of the current round the current RIB may include
routes that were not yet advertised in the previous round. This adds
some memory overhead, but pointers to the routes are much smaller
than the routes themselves, so this should be less than 10%.

Add a test.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpAdvertisement.java
@@ -35,6 +35,37 @@ public class BgpAdvertisement implements Comparable<BgpAdvertisement>, Serializa
     IBGP_RECEIVED,
     IBGP_SENT;
 
+    /** Returns true if this advertisement is from an ebgp peer. */
+    public boolean isEbgp() {
+      switch (this) {
+        case EBGP_ORIGINATED:
+        case EBGP_RECEIVED:
+        case EBGP_SENT:
+          return true;
+        default:
+          return false;
+      }
+    }
+
+    /**
+     * Returns true if this advertisement is from a device's local view of the route, aka, it has
+     * already been processed/filtered by import policy. These routes are installed directly in the
+     * RIB.
+     */
+    public boolean isReceived() {
+      switch (this) {
+        case EBGP_RECEIVED:
+        case IBGP_RECEIVED:
+          return true;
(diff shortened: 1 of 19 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6606)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6606#-MSTdnt0BGPsq8wzmZym:-MSTdnt0BGPsq8wzmZyn:bf214i1) (no related file):*
(haven't looked at the code yet, but I have some questions about the description)

> depending on iteration order of the current round the current RIB may include
> routes that were not yet advertised in the previous round

This sounds like we need to advertise the current RIB, but you say we need to advertise the previous round's main RIB. Can you clarify that for me?

Also, I'm curious what kind of variance there is in the "iteration order".  Is it possible to fix that order, and would doing so allow us to avoid this overhead?

---

anothermattbrown (commented):
> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6606)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6606)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @corinaminer)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6606#-MSVdu47-P2LLN0ZiYJu:-MSVdu47-P2LLN0ZiYJv:b-i4j9oo) (no related file):*
Revised and added invariants per discussion.

---

anothermattbrown (commented):
> Reviewed 2 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6606)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

PR #6610, opened 2021-02-08
cisco: remove disable-peer-as-check from non-NX-OS parsers

dhalperi:
This is leftover from before we split out NX-OS.


arifogel (commented):
> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6610)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6613, opened 2021-02-08
granular allowRemoteAsOut behavior

arifogel:
Support 3 behaviors for allowRemoteAsOut via `AllowRemoteAsOutMode`:
Break `false` down into:
- `AllowRemoteAsOutMode.NEVER`: never send adverts whose as-path contains peer as
- `AllowRemoteAsOutMode.EXCEPT_FIRST` do not send adverts whose as-path's most recent as-set contains peer as
Change `true` to:
- `AllowRemoteAsOutMode.ALWAYS`: do not filter outgoing adverts based on presence of peer as in as-path

PR #6614, opened 2021-02-09
JunOS: fix default BGP loop AS Path check

dhalperi:
It actually only checks the most recent ASN, not the entire AS Path.

Fixup batfish/batfish#6613


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6614)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #6615, opened 2021-02-10
CumulusInterfaces: cleanup and add automatic /32, /128 for addresses

dhalperi:
1. add toIp, toPrefix, and use throughout. Standard stuff so we don't reinvent the wheel.
2. interface address commands all default to /32 (v4) or /128 (v6) if the prefix is not given.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6615)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6622, opened 2021-02-11
PAN: loopback interfaces can have units

dhalperi:
And those units can have IP addresses.

I did confirm that the same interface can have both main loopback and
loopback.123, and both can be active and with L3 config.


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6622)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

Issue #6637, opened 2021-02-17
Parse failures causing later false positives


bcavns01:
We're seeing parse failures for lines related to routing and firewall ACLs.  When we later check for things like unusedStructures, we get results for things that _are_ actually in use, but they are undetected because the structures using them failed to parse:

Examples of lines that failed parsing::
```
as-path-group
set policy-options policy-statement OVERFLOW-LOCAL-IN term import_overflow then local-preference subtract 10
set policy-options policy-statement REJECT-ROUTES term in_overflow_rib_5_3_4_0-24 from condition in-overflow-rib_5_3_4_0-24
set interfaces gr-0/1/10 unit 204 family inet address 10.255.0.9/32 destination 10.255.0.8
set firewall family inet filter NSP term SMALL filter SMALL-PACKETS
```

dhalperi:
Hi @bcavns01 - I'm not quite sure I grok this.

E.g., I'm not sure how 

> set policy-options policy-statement OVERFLOW-LOCAL-IN term import_overflow then local-preference subtract 10

failing to parse (which it did!) causes a reference to be lost. That line doesn't reference anything else.

Other questions:

* What type of device is this? I can't find documentation for `set firewall family inet filter NSP term SMALL filter SMALL-PACKETS`, e.g., here: https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/term-firewall-filter-ex-series.html

* `set policy-options policy-statement REJECT-ROUTES term in_overflow_rib_5_3_4_0-24 from condition in-overflow-rib_5_3_4_0-24` -- we don't model this yet, but also it should not lead to unused structures I think: the only reference here is to the condition `in-overflow-rib_5_3_4_0-24`, but those are not tracked for unused structures.

You can sync with us on Batfish Slack if you want a real-time interaction.

Thanks!

bcavns01:
Hi @dhalperi ,

Thanks for the quick follow-up.  Sorry for the ambiguity in the OP.  I think I got too excited by the free-form format for posting issues to this repo. :)

Not all the examples in the OP were related to unused structures.  They were just examples of failures to parse that would seem to directly impact routing and ACLs.

For the issue of unused structures, as-path-group will probably be one of the larger sources of trouble for us, so I'll try for a clearer example of that:

-------------------------------------------------------------------------

**Related Issue seen:**

(This seems to be the only related issue)

[...]

PR #6649, opened 2021-02-20
JunOS: fix reference tacking for as-path-group as-path

dhalperi:
The right self-reference was there, we just didn't actually ever count them.

This is for #6637


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6649)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #6671, opened 2021-03-01
IOS: Transform originating flows to outside interfaces

corinaminer:
IOS inside NAT rules will be applied to traffic that originates on the NAT device, if it is forwarded out an `outside` interface.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6671)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6684, opened 2021-03-04
Reachability: add Blackhole state

dhalperi:
There are some impossible flows generated in the BDD Reachability engine, e.g.,
flows that are only valid if they exit a specific interface. This is safe (no
wrong results are generated), but it is better to specifically blackhole such
flows rather than having them "disappear "in the middle of the reachability
graph. This makes invariant checking easier.

We can also take minor advantage of this to hopefully speed up some ACL
computations: remove irrelevant flows before feeding them through transitions,
applying ACLs, etc.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/bddreachability/BDDOutgoingOriginalFlowFilterManager.java
+++ b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDOutgoingOriginalFlowFilterManager.java
@@ -5,6 +5,7 @@ import static com.google.common.collect.Maps.immutableEntry;
 import static org.batfish.common.bdd.IpAccessListToBdd.toBdds;
 import static org.batfish.common.util.CollectionUtil.toImmutableMap;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6684)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6684#-MUvEZ9N6nIBCh7mpdvB:-MUvEZ9O72tUACCwc-y5:b-qlytlq) (no related file):*
Working: needs tests.

---

anothermattbrown (commented):
> Reviewed 2 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6684)* status: 2 of 9 files reviewed, 2 unresolved discussions (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 1166 at r1](https://reviewable.io/reviews/batfish/batfish/6684#-MV2zfnYA9_DiOR9sqXB:-MV2zfnYA9_DiOR9sqXC:bujwchk) ([raw file](https://github.com/batfish/batfish/blob/0ebb31999dbf1e27b96be95739d3bf124011afdf/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L1166)):*
> ```
> constraint(flowsForIface1)
> ```
if we put this constraint on the in-edges of `PreOutEdge` (as well as the `PreOutInterface<Disposition>` states) can we avoid adding the black-hole state?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6684)* status: 2 of 9 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 1166 at r1](https://reviewable.io/reviews/batfish/batfish/6684#-MV2zfnYA9_DiOR9sqXB:-MV3nhl7-dyb6NiI0SW3:b1oit3x) ([raw file](https://github.com/batfish/batfish/blob/0ebb31999dbf1e27b96be95739d3bf124011afdf/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L1166)):*

> ```
> constraint(flowsForIface1)
> ```
if we put this constraint on the in-edges of `PreOutEdge` (as well as the `PreOutInterface<Disposition>` states) can we avoid adding the black-hole state?
</blockquote></details>

IIUC we push this constraint a bit earlier in the pipeline... but wouldn't we still need the blackhole state where we moved it to? Those flows would still deadend there, otherwise.

PR #6686, opened 2021-03-04
FortiOS: Parse and extract address objects

corinaminer:


sfraint (commented):
> Reviewed 9 of 10 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: all files reviewed, 9 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4, line 59 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyw0dN9N8t8uobYb5R:-MUyw0dN9N8t8uobYb5S:b5bexaz) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4#L59)):*

> ```
>         // Options for unsupported address types
>         | COUNTRY
>         | FQDN
>         | SDN
>         | SUB_TYPE
> ```
</details>

my gut is, these should not be part of the `null` rule; we should probably have warnings for at least some of them, since they could affect Batfish modeling
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4, line 73 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUywL2qBVrllajwXZ-p:-MUywL2qBVrllajwXZ-q:bblc9c5) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4#L73)):*
> ```
> // 1-15 characters for interface, but 1-35 for zone
> interface_or_zone_name: str;
> ```
probably should put this in `Fortios_common.g4`, since we're going to use it elsewhere (at least in policies)
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 130 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyxRyu3jdCj6tngfTB:-MUyxRyu3jdCj6tngfTC:b-s1aeni) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L130)):*
> ```
>       // It's sort of arbitrary whether we consider this undefined reference to be a zone or an
>       // interface, but safer to call it a zone, because the name may not be a valid interface name.
> ```
fyi we could call it an undefined abstract reference, e.g. to `INTERFACE_OR_ZONE`
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 151 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyyG2E3jPENBO8XTGQ:-MUyyG2E3jPENBO8XTGR:bafg07z) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L151)):*
> ```
> _currentAddress.getTypeSpecificFields()...
> ```
if you your address object is of a specific type, does the config let you set type-specific fields for other types?

if not, we should warn here and for the other fields
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 155 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz1J4r2_SGZL2c3D7p:-MUz1J4r2_SGZL2c3D7q:bvblz7d) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L155)):*

> ```
>   @Override
>   public void exitCfa_set_end_ip(Cfa_set_end_ipContext ctx) {
>     _currentAddress.getTypeSpecificFields().setEndIp(toIp(ctx.ip));
>   }
> ```
</details>

does the cli enforce end ip being higher than start ip? do we need to enforce that as well?
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 274 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyzuKGAkI7Z28ElzOO:-MUyzuKGAkI7Z28ElzOP:b-f6in85) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L274)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>       assert ctx.DYNAMIC() != null
>           // FQDN addresses are based on domain names
>           || ctx.FQDN() != null
>           // Geography-type addresses are based on countries
>           || ctx.GEOGRAPHY() != null
>           // Mac-type addresses are based on mac addresses
>           || ctx.MAC() != null;
>       return Address.Type.UNSUPPORTED;
> ```
</details>

can we just store the real type in the VS model and determine if it is support after the fact?
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 506 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz-cXODEPojecZ6G65:-MUz-cXODEPojecZ6G66:b-v85lnj) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L506)):*
> ```
>   private static final Pattern ZONE_NAME_PATTERN = Pattern.compile("^[^\r\n]{1,35}$"); // todo check
> ```
did you mean to leave this `todo` in?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz-qs96XqVwBpjkHUf:-MUz-qs96XqVwBpjkHUg:bgzgfa) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L22)):*
> ```
>   public static class TypeSpecificFields implements Serializable {
> ```
do all of the applicable type-specific fields have to be set for an address? e.g. does it infer end ip == start ip for iprange?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz065C1MmnHWu8cCFE:-MUz065C1MmnHWu8cCFF:br5xxce) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L71)):*
> ```
>   private boolean _allowRouting; // false by default
> ```
I have been using `nullable` values and saving the default separately for anything with default values so Batfish can differentiate between the config having explicitly set the  value and using default

(at least partly because it sounds like different fortios versions often have different default values)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 139 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyy2HJA5cUo3YKmK6p:-MUyy2HJA5cUo3YKmK6q:b-x4rx2c) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L139)):*
> ```
>   public void setType(Type type) {
>     _type = type;
>   }
> ```
if type is changed, do the type-specific fields get cleared?
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/address, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz0wv-0NRZhopZDS9c:-MUz0wv-0NRZhopZDS9d:bsnlf5g) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/address#L16)):*
> ```
> set associated-interface port1
> ```
mainly curious, what does `associated-interface` mean?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: all files reviewed, 10 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 127 at r2](https://reviewable.io/reviews/batfish/batfish/6686#-MUzHEiN4QSAHssOt_wR:-MUzHEiO6-_nOGnOcSwD:bcg76fp) ([raw file](https://github.com/batfish/batfish/blob/8e267edbbab6a9f104b499e0af7dc87cabdc39d0/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L127)):*
> ```Java
>     Map<String, Address> addresses = vc.getAddresses();
>     assertThat(
>         addresses.keySet(),
> ```

fyi: can use `hasKeys` which is a little more concise

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: 1 of 10 files reviewed, 4 unresolved discussions (waiting on @sfraint)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4, line 59 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyw0dN9N8t8uobYb5R:-MUzLP2M8b6J2j0W13lp:buua2vl) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4#L59)):*

> ```
>         // Options for unsupported address types
>         | COUNTRY
>         | FQDN
>         | SDN
>         | SUB_TYPE
> ```
my gut is, these should not be part of the `null` rule; we should probably have warnings for at least some of them, since they could affect Batfish modeling
</blockquote></details>

These are options that are only available for specific address types that we do not support. As of the new revision (r3), we warn that those types aren't supported. I could see having warnings if these are set for an unexpected address type, but that should be extremely rare and it doesn't seem worth extracting them just for that.

If you'd rather just not parse them so that they always give warnings, that'd be easy.
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4, line 73 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUywL2qBVrllajwXZ-p:-MUzIoVCB_aVQZY7t2Nq:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4#L73)):*

> ```
> // 1-15 characters for interface, but 1-35 for zone
> interface_or_zone_name: str;
> ```
probably should put this in `Fortios_common.g4`, since we're going to use it elsewhere (at least in policies)
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 130 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyxRyu3jdCj6tngfTB:-MUzJK3OCKyUpkRpcd7b:bhy1gde) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L130)):*

> ```
>       // It's sort of arbitrary whether we consider this undefined reference to be a zone or an
>       // interface, but safer to call it a zone, because the name may not be a valid interface name.
> ```
fyi we could call it an undefined abstract reference, e.g. to `INTERFACE_OR_ZONE`
</blockquote></details>

ack, updated comment
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 151 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyyG2E3jPENBO8XTGQ:-MUzSGAZ5Yr5GflsPCFa:b-mkb7p4) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L151)):*

> ```
> _currentAddress.getTypeSpecificFields()...
> ```
if you your address object is of a specific type, does the config let you set type-specific fields for other types?

if not, we should warn here and for the other fields
</blockquote></details>

done. There's a lot of complicated interaction between type-specific fields when you switch types, though. For example if you do:
```
set subnet 1.1.0.0 255.255.0.0
set type iprange
```
Then you will end up with:
```
set type iprange
set start-ip 1.1.1.0
set end-ip 255.255.255.0
```

In most cases that can be handled in conversion. We'll probably have some bugs in some pathological cases where the address type changes multiple times, but doesn't seem worth investigating and supporting those now.
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 155 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz1J4r2_SGZL2c3D7p:-MUziE_A8WCbIUbidmiG:b4l2pzg) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L155)):*

> ```
>   @Override
>   public void exitCfa_set_end_ip(Cfa_set_end_ipContext ctx) {
>     _currentAddress.getTypeSpecificFields().setEndIp(toIp(ctx.ip));
>   }
> ```
does the cli enforce end ip being higher than start ip? do we need to enforce that as well?
</blockquote></details>

It will enforce this on `next`. I think we should handle that in conversion, though. Can't just ignore the problematic end IP; if we have:
```
set start-ip 2.2.2.2
set end-ip 3.3.3.3
set end-ip 1.1.1.1
```
then the VS model should have end IP 1.1.1.1, and the address should be marked invalid and ignored.
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 274 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyzuKGAkI7Z28ElzOO:-MUzjv_aAtEoIlXheIHn:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L274)):*

> ```
>       assert ctx.DYNAMIC() != null
>           // FQDN addresses are based on domain names
>           || ctx.FQDN() != null
>           // Geography-type addresses are based on countries
>           || ctx.GEOGRAPHY() != null
>           // Mac-type addresses are based on mac addresses
>           || ctx.MAC() != null;
>       return Address.Type.UNSUPPORTED;
> ```
can we just store the real type in the VS model and determine if it is support after the fact?
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 506 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz-cXODEPojecZ6G65:-MUzkAZO639zmXzEfIKb:b7mgmto) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L506)):*

> ```
>   private static final Pattern ZONE_NAME_PATTERN = Pattern.compile("^[^\r\n]{1,35}$"); // todo check
> ```
did you mean to leave this `todo` in?
</blockquote></details>

done, checked
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz-qs96XqVwBpjkHUf:-MUzkG3P5JM6mjMGHm4J:b-g27gip) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L22)):*

> ```
>   public static class TypeSpecificFields implements Serializable {
> ```
do all of the applicable type-specific fields have to be set for an address? e.g. does it infer end ip == start ip for iprange?
</blockquote></details>

Some have defaults; added them. for end IP, the default is 0.0.0.0, but that's not a permitted end IP (if you say `next`, it'll fail with "end-ip can not be 0", or "end-ip must be greater than start-ip" if you've set a nonzero start IP) so i didn't bother adding a default.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz065C1MmnHWu8cCFE:-MUzmplbCPACErwdnDOT:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L71)):*

> ```
>   private boolean _allowRouting; // false by default
> ```
I have been using `nullable` values and saving the default separately for anything with default values so Batfish can differentiate between the config having explicitly set the  value and using default

(at least partly because it sounds like different fortios versions often have different default values)
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 139 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyy2HJA5cUo3YKmK6p:-MUzn4ev-d1vCzXwFcO1:b51fml4) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L139)):*

> ```
>   public void setType(Type type) {
>     _type = type;
>   }
> ```
if type is changed, do the type-specific fields get cleared?
</blockquote></details>

No, and we don't want them to, because type-specific information sticks around -- e.g. if you do:
```
set subnet 1.1.1.0 255.255.255.0
set type geography // not compatible with subnet
```
then the address won't show any subnet info, but if you switch back to type `ipmask` (which is the default) then your subnet comes back.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/address, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz0wv-0NRZhopZDS9c:-MUzqAMiDxizGTK0Hujs:b-a7oc9l) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/address#L16)):*

> ```
> set associated-interface port1
> ```
mainly curious, what does `associated-interface` mean?
</blockquote></details>

I'm not sure in practical terms. It's defined as the "network interface associated with address", and there's a forum post implying it has to do with the scope in which the address is visible. not sure how that relates to `fabric-object` ("Enabling fabric-object sets the object as a Security Fabric-wide global object that is synchronized to downstream FortiGates"); configuring `associated-interface` doesn't block you from enabling that.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: 1 of 10 files reviewed, 4 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 155 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz1J4r2_SGZL2c3D7p:-MV-1lQb6V37Cxkgm_fX:b-wtnxyo) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L155)):*

It will enforce this on `next`. I think we should handle that in conversion, though. Can't just ignore the problematic end IP; if we have:
```
set start-ip 2.2.2.2
set end-ip 3.3.3.3
set end-ip 1.1.1.1
```
then the VS model should have end IP 1.1.1.1, and the address should be marked invalid and ignored.
</blockquote></details>

I would suggest either enforcing on exit in extractor via clone like we talked about offline, or just asserting it right here for now and adding a todo.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 155 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUz1J4r2_SGZL2c3D7p:-MV-4qwA7g2v_vSKPN6C:b-prqjog) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L155)):*

I would suggest either enforcing on exit in extractor via clone like we talked about offline, or just asserting it right here for now and adding a todo.
</blockquote></details>

Ok, implemented clone idea that we discussed, with todo for further validity checking.

sfraint (commented):
> Reviewed 11 of 11 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4, line 59 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyw0dN9N8t8uobYb5R:-MV-V2fG4PMn3s6VRRXa:b-3vc96g) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_address.g4#L59)):*

These are options that are only available for specific address types that we do not support. As of the new revision (r3), we warn that those types aren't supported. I could see having warnings if these are set for an unexpected address type, but that should be extremely rare and it doesn't seem worth extracting them just for that.

If you'd rather just not parse them so that they always give warnings, that'd be easy.
</blockquote></details>

as long as we have warnings for things that might impact modeling, I'm happy (also, we should have tests for warnings like that)
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 151 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyyG2E3jPENBO8XTGQ:-MV-N_9uBOp7sMIpwvgq:bxao7a1) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L151)):*

done. There's a lot of complicated interaction between type-specific fields when you switch types, though. For example if you do:
```
set subnet 1.1.0.0 255.255.0.0
set type iprange
```
Then you will end up with:
```
set type iprange
set start-ip 1.1.1.0
set end-ip 255.255.255.0
```

In most cases that can be handled in conversion. We'll probably have some bugs in some pathological cases where the address type changes multiple times, but doesn't seem worth investigating and supporting those now.
</blockquote></details>

yuck, ok
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 182 at r3](https://reviewable.io/reviews/batfish/batfish/6686#-MV-J7vXATDtgiw6FRdR:-MV-J7vXATDtgiw6FRdS:b2quhpj) ([raw file](https://github.com/batfish/batfish/blob/86a3b9a0c94e578a51a9743b2f21660b8ed3399f/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L182)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>     // If edited address is valid, add/update the entry in VS addresses map.
>     // TODO: Better validity checking
>     if (ADDRESS_NAME_PATTERN.matcher(_currentAddress.getName()).matches()) {
>       // TODO Add structure definition for address
>       _c.getAddresses().put(_currentAddress.getName(), _currentAddress);
>     }
> ```
</details>

of course can come later, but we could just keep an `isValid` boolean for the structure that we set to false at any point if the edit should be aborted

can also use that to avoid checking the name pattern in two places
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 139 at r1](https://reviewable.io/reviews/batfish/batfish/6686#-MUyy2HJA5cUo3YKmK6p:-MV-U4fy7_M91RnhZJeP:bah36mw) ([raw file](https://github.com/batfish/batfish/blob/89973f55a9742c1f3973d31cb741e2dc28bf5afd/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L139)):*

No, and we don't want them to, because type-specific information sticks around -- e.g. if you do:
```
set subnet 1.1.1.0 255.255.255.0
set type geography // not compatible with subnet
```
then the address won't show any subnet info, but if you switch back to type `ipmask` (which is the default) then your subnet comes back.
</blockquote></details>

weird...this behavior is different from services, but ok

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6686)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 182 at r3](https://reviewable.io/reviews/batfish/batfish/6686#-MV-J7vXATDtgiw6FRdR:-MV2gzM_Dlgmh5JeNEIY:b-yn8yc2) ([raw file](https://github.com/batfish/batfish/blob/86a3b9a0c94e578a51a9743b2f21660b8ed3399f/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L182)):*

> ```
>     // If edited address is valid, add/update the entry in VS addresses map.
>     // TODO: Better validity checking
>     if (ADDRESS_NAME_PATTERN.matcher(_currentAddress.getName()).matches()) {
>       // TODO Add structure definition for address
>       _c.getAddresses().put(_currentAddress.getName(), _currentAddress);
>     }
> ```
of course can come later, but we could just keep an `isValid` boolean for the structure that we set to false at any point if the edit should be aborted

can also use that to avoid checking the name pattern in two places
</blockquote></details>

yeah, would prefer to push that to another PR -- i foresee some complications

PR #6696, opened 2021-03-05
NO BGP NEIGHBOR

jawyoonis:
- I deleted `sbafi_no` method which was not allowing the `sbafi_no_activate` method to take an effect to no-activate ` no neighbor ipv6 activate` under ipv4 family address


ratulm:
Fixed in https://github.com/batfish/batfish/pull/6699

PR #6700, opened 2021-03-06
Reachability engine: explicitly filter rather than deadend

dhalperi:
In some parts of the reachability graph, we are "leaking" flows that are
internal to a node too far, to a node where these flows have no outlet.

For example, flows that are only valid if sent out interface i1 were making it
to i2-related states, and flows that are only valid in vrf A were making it to
vrf B.

This is safe -- no wrong results are generated, the flows will just die, but
it is better to explicitly exclude such flows rather than having them "disappear"
in the middle of the reachability graph. This makes invariant checking easier.

### DIFF
--- a/projects/allinone/BUILD
+++ b/projects/allinone/BUILD
@@ -128,6 +128,7 @@ junit_tests(
         "//projects/batfish:batfish_testlib",
         "//projects/batfish-common-protocol:common",
         "//projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/matchers",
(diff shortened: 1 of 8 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6700)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @anothermattbrown)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6700)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6703, opened 2021-03-08
FortiOS: Convert addresses

corinaminer:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6703)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 143 at r1](https://reviewable.io/reviews/batfish/batfish/6703#-MVIshH06jS6xDC2JXzr:-MVIshH06jS6xDC2JXzs:bihscah) ([raw file](https://github.com/batfish/batfish/blob/4696f0d362e61096a4603cae372ffc594e84a56a/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L143)):*
> ```Java
>                 getType()));
>         return EmptyIpSpace.INSTANCE;
>       case UNKNOWN: // should never be the effective type
> ```

I really don't see the purpose of this enum value. It's not a helpful message to say "Unrecognized adddress type UNKNOWN." In at least one other place we removed UNKNOWN and used `null` in its place. If the address type has not been set, the message should note that instead of referring to a fake address type.
Further, I don't think it's acceptable to crash on conversion just because an address type was not set.
___
*[projects/batfish/src/test/java/org/batfish/representation/fortios/AddressTest.java, line 25 at r1](https://reviewable.io/reviews/batfish/batfish/6703#-MVItRusELC20bKlZ6W7:-MVItRusELC20bKlZ6W8:bkam1lp) ([raw file](https://github.com/batfish/batfish/blob/4696f0d362e61096a4603cae372ffc594e84a56a/projects/batfish/src/test/java/org/batfish/representation/fortios/AddressTest.java#L25)):*
> ```Java
>     Prefix prefix = Prefix.parse("1.1.1.0/24");
>     address.getTypeSpecificFields().setSubnet(prefix);
>     assertConvertsWithoutWarnings(address, prefix.toIpSpace());
> ```

These are conversion tests, no? You should do BDD equivalence tests instead of just checking they don't complain.
See `CiscoNxosGrammarTest.testObjectGroupIpAddressConversion`

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6703)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java, line 143 at r1](https://reviewable.io/reviews/batfish/batfish/6703#-MVIshH06jS6xDC2JXzr:-MVIupvS0ZWMLZx_M4-T:bllmktq) ([raw file](https://github.com/batfish/batfish/blob/4696f0d362e61096a4603cae372ffc594e84a56a/projects/batfish/src/main/java/org/batfish/representation/fortios/Address.java#L143)):*

I really don't see the purpose of this enum value. It's not a helpful message to say "Unrecognized adddress type UNKNOWN." In at least one other place we removed UNKNOWN and used `null` in its place. If the address type has not been set, the message should note that instead of referring to a fake address type.
Further, I don't think it's acceptable to crash on conversion just because an address type was not set.
</blockquote></details>

Removed `UNKNOWN` case.

As for crashing on conversion, that code should be unreachable because the switch is on `getTypeEffective`, which can't return `UNKNOWN` (or `null`).
___
*[projects/batfish/src/test/java/org/batfish/representation/fortios/AddressTest.java, line 25 at r1](https://reviewable.io/reviews/batfish/batfish/6703#-MVItRusELC20bKlZ6W7:-MVJ-x9F1JgOK4qvcGz4:b2md82l) ([raw file](https://github.com/batfish/batfish/blob/4696f0d362e61096a4603cae372ffc594e84a56a/projects/batfish/src/test/java/org/batfish/representation/fortios/AddressTest.java#L25)):*

These are conversion tests, no? You should do BDD equivalence tests instead of just checking they don't complain.
See `CiscoNxosGrammarTest.testObjectGroupIpAddressConversion`
</blockquote></details>

Done (added to `FortiosGrammarTest` -- i don't think testing BDD equivalence is necessary in both tests)

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6703)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

PR #6704, opened 2021-03-08
Reachability: add Blackhole state

dhalperi:
There are some impossible flows generated in the BDD Reachability engine, e.g.,
flows that are only valid if they exit a specific interface. This is safe (no
wrong results are generated), but it is better to specifically blackhole such
flows rather than having them "disappear "in the middle of the reachability
graph. This makes invariant checking easier.

Not plugged in anywhere yet.


anothermattbrown (commented):
> Reviewed 7 of 7 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6704)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #6710, opened 2021-03-10
Junos: Local-as inheritance not working 


ratulm:
Via Slack: https://networktocode.slack.com/archives/CCE02JK7T/p1615364437011000?thread_ts=1615288821.008400&cid=CCE02JK7T

"issue is that when my bgp group has a specific "local-as" number it's not being inherited to the neighbours defined in this group, but instead relies on the global asn setting under "routing-options". When adding specific "local-as" under the neighbour itself it it being assigned properly."

```
routing-options {
    router-id 111.0.0.1;
    autonomous-system 1111 loops 3;
}
groups {
  replace:
  mesh_bgp{
    protocols {
      bgp {    
        group group_a {
            type external;
            import from-group_a;
            export to-group_a;
            local-as 64611 private no-prepend-global-as;
        neighbor 10.255.16.23 {
                description group_a-to_1;
                peer-as 64616;
              }  
        neighbor 10.255.11.8 {
                description group_a-to_2;
                peer-as 64604;
              }  
        neighbor 10.255.11.2 {
                description group_a-to_3;
                peer-as 64601;
              }
        }
        group group_b {
            type external;
            import from-group_b;
            export to-group_b;
            local-as 64611 private no-prepend-global-as;
        neighbor 10.255.42.23 {
                description group_b-to_1;
                peer-as 64642;
              }  
        neighbor 10.255.34.23 {
                description group_b-to_2;
                peer-as 64634;
              }  
        neighbor 10.255.30.23 {
                description group_b-to_3;
                peer-as 64630;
              }      
        }
      }
    }
  }
}
```

dhalperi:
We'll take a look.

dhalperi:
So the given local-AS will be parsed, but the `private` directive is still not implemented.

PR #6711, opened 2021-03-10
JunOS: parse and warn on no-prepend-global-as

dhalperi:
Add missing warnings for related constructs.

This is for batfish/batfish#6710.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -1751,6 +1751,8 @@ NO_PING_RECORD_ROUTE: 'no-ping-record-route';
 
 NO_PING_TIME_STAMP: 'no-ping-time-stamp';
 
+NO_PREPEND_GLOBAL_AS: 'no-prepend-global-as';
+
 NO_READVERTISE: 'no-readvertise';
 
 NO_REDIRECTS: 'no-redirects';
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_bgp.g4
@@ -449,12 +449,18 @@ bl_common
 :
    bl_alias
    | bl_loops
+   | bl_no_prepend_global_as
    | bl_private
 ;
 
 bl_loops
 :
    LOOPS dec
+;
+
+bl_no_prepend_global_as
+:
+   NO_PREPEND_GLOBAL_AS
 ;
 
 bl_number
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -3839,7 +3839,7 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener {
 
   @Override
   public void exitBl_no_prepend_global_as(Bl_no_prepend_global_asContext ctx) {
-    todo(ctx);
+    _currentBgpGroup.setNoPrependGlobalAs(true);
   }
 
   @Override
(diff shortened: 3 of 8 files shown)


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6711)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java, line 3842 at r1](https://reviewable.io/reviews/batfish/batfish/6711#-MVT4ZfH57VLaGBHjBCM:-MVT4ZfH57VLaGBHjBCN:b-r9rzsy) ([raw file](https://github.com/batfish/batfish/blob/a6c8a48ccffff8f10938e005206fb54efe3464f4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3842)):*
> ```Java
>   @Override
>   public void exitBl_no_prepend_global_as(Bl_no_prepend_global_asContext ctx) {
>     todo(ctx);
> ```

I'll double check in a bit but I thought this is what Batfish does by default. If so, we need to warn when something else happens.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6711)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java, line 3842 at r1](https://reviewable.io/reviews/batfish/batfish/6711#-MVT4ZfH57VLaGBHjBCM:-MVT6kDkD4zPZegZMWaz:b548711) ([raw file](https://github.com/batfish/batfish/blob/a6c8a48ccffff8f10938e005206fb54efe3464f4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3842)):*

I'll double check in a bit but I thought this is what Batfish does by default. If so, we need to warn when something else happens. 
</blockquote></details>

Judging by the FRR warning behavior (https://github.com/batfish/batfish/blob/e2af1bf9d24d1646ce2bc9619dd998a37a1b78fb/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java#L490) which I fixed in consultation with Victor, we implement what in FRR is `no-prepend replace-as`. Per https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighborPEERlocal-asAS-NUMBER[no-prepend][replace-as], this means that (only) local-as will be prepended (not global). 

IIUC, JunOS documentation (https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/local-as-edit-protocols-bgp.html) says that `no-prepend-global-as` means that only local-as is prepended. 

If I am reading this all right -- no guarantee there -- then we should warn on private/alias options.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6711)* status: 3 of 8 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java, line 3842 at r1](https://reviewable.io/reviews/batfish/batfish/6711#-MVT4ZfH57VLaGBHjBCM:-MVTMn-e656rrvcJdN9z:bb73vv8) ([raw file](https://github.com/batfish/batfish/blob/a6c8a48ccffff8f10938e005206fb54efe3464f4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3842)):*

Judging by the FRR warning behavior (https://github.com/batfish/batfish/blob/e2af1bf9d24d1646ce2bc9619dd998a37a1b78fb/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java#L490) which I fixed in consultation with Victor, we implement what in FRR is `no-prepend replace-as`. Per https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighborPEERlocal-asAS-NUMBER[no-prepend][replace-as], this means that (only) local-as will be prepended (not global). 

IIUC, JunOS documentation (https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/local-as-edit-protocols-bgp.html) says that `no-prepend-global-as` means that only local-as is prepended. 

If I am reading this all right -- no guarantee there -- then we should warn on private/alias options. 
</blockquote></details>

TAL.

ratulm (commented):
> Reviewed 3 of 5 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6711)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #6712, opened 2021-03-10
Undefined reference tracking refactor

sfraint:
This PR adds undefined reference tracking directly in `VendorConfiguration`, similar to how we handle defined structure and structure reference tracking already.

This will enable future configuration builders to mark undefined references as they are encountered (in parsing/extraction) instead of having to wait until after an answer element is assigned to the vendor configuration. This helps unblock FortiOS undefined reference tracking.

This has a side-effect of slightly changing how hosts/files with no undefined references are represented. Instead of being missing from the root undefined references map in `ConvertConfigurationAnswerElement`, now they will be present but with an empty sub-map (e.g. see ref file changes).

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java
@@ -49,12 +49,19 @@ public abstract class VendorConfiguration implements Serializable {
   private VendorConfiguration _overlayConfiguration;
 
   // Type description -> Name -> DefinedStructureInfo
+  @Nonnull
   protected final SortedMap<String, SortedMap<String, DefinedStructureInfo>> _structureDefinitions;
 
   // StructureType -> Name -> StructureUsage
+  @Nonnull
   protected final SortedMap<
           StructureType, SortedMap<String, SortedMap<StructureUsage, SortedMultiset<Integer>>>>
       _structureReferences;
+
+  // structType -> structName -> usage -> lines
+  @Nonnull
+  protected final SortedMap<String, SortedMap<String, SortedMap<String, SortedSet<Integer>>>>
+      _undefinedReferences;
 
   private transient boolean _unrecognized;
 
@@ -64,6 +71,7 @@ public abstract class VendorConfiguration implements Serializable {
     _runtimeData = SnapshotRuntimeData.EMPTY_SNAPSHOT_RUNTIME_DATA;
     _structureDefinitions = new TreeMap<>();
     _structureReferences = new TreeMap<>();
+    _undefinedReferences = new TreeMap<>();
   }
 
   public String canonicalizeInterfaceName(String name) {
@@ -270,6 +278,7 @@ public abstract class VendorConfiguration implements Serializable {
                                         name,
                                         usage,
                                         line)))));
+    _answerElement.getUndefinedReferences().put(getFilename(), _undefinedReferences);
   }
 
   public void setFilename(String filename) {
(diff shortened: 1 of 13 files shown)


corinaminer (commented):
> Reviewed 10 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6712)* status: 10 of 13 files reviewed, 1 unresolved discussion (waiting on @arifogel and @sfraint)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java, line 281 at r1](https://reviewable.io/reviews/batfish/batfish/6712#-MVTGKC-1HZ05fxAFfyT:-MVTGKC-1HZ05fxAFfyU:bgpn4bv) ([raw file](https://github.com/batfish/batfish/blob/405ea9192419848af460a2bee3351667b8952ebb/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java#L281)):*
> ```Java
>                                         usage,
>                                         line)))));
>     _answerElement.getUndefinedReferences().put(getFilename(), _undefinedReferences);
> ```

Why not make this line conditional on `_undefinedReferences` not being empty? Then you shouldn't need the ref changes or new matcher.

I suspect the reason is because `undefined()` will continue to modify `_undefinedReferences` during conversion, so you need `_undefinedReferences` to be in the answer element. if so, i'd recommend changing `undefined()` to do:
```
if (answer element is defined) { do the old thing }
else { add reference in _undefinedReferences }
```
Alternatively, to make behavior more obvious, you could make separate methods:
```
undefined(...) { do the old thing }
undefinedInExtraction(...) {
  checkState(_answerElement == null, "can't use this method after extraction");
  add to _undefinedReferences
}
```
And/or you could rename `_undefinedReferences` to `_undefinedReferencesFromExtraction` to clarify that that map will only have a subset of the undefined refs.

In any case it seems avoidable to pad all our CCAEs with empty maps when only FortiOS will use this functionality so far.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6712)* status: 10 of 13 files reviewed, 1 unresolved discussion (waiting on @arifogel and @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java, line 281 at r1](https://reviewable.io/reviews/batfish/batfish/6712#-MVTGKC-1HZ05fxAFfyT:-MVTOWdRBYPkvipDqBbN:b-kh54h9) ([raw file](https://github.com/batfish/batfish/blob/405ea9192419848af460a2bee3351667b8952ebb/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java#L281)):*
> Why not make this line conditional on _undefinedReferences not being empty?

I was going for consistency with the other structures, but I can do that instead

> And/or you could rename _undefinedReferences to _undefinedReferencesFromExtraction to clarify that that map will only have a subset of the undefined refs.

The way it is implemented in this PR, this map will have all the undefined references, not just the ones from extraction

> In any case it seems avoidable to pad all our CCAEs with empty maps when only FortiOS will use this functionality so far.

The only time there will be empty maps is when there are no undefined references - that could occur for any vendor, not just fortiOS

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6712)* status: 10 of 13 files reviewed, 1 unresolved discussion (waiting on @arifogel and @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java, line 281 at r1](https://reviewable.io/reviews/batfish/batfish/6712#-MVTGKC-1HZ05fxAFfyT:-MVTQJTkC01rP4r08Zvc:b-dzksma) ([raw file](https://github.com/batfish/batfish/blob/405ea9192419848af460a2bee3351667b8952ebb/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java#L281)):*
> Why not make this line conditional on _undefinedReferences not being empty?
> 
> I was going for consistency with the other structures, but I can do that instead

Actually, I'm not sure this will work, since you won't know if more undefined references will be added later, after the answer element is set

corinaminer (commented):
:lgtm:

> Reviewed 3 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6712#-:-MVTVy6Q-h8jXVk9-Cki:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java, line 281 at r1](https://reviewable.io/reviews/batfish/batfish/6712#-MVTGKC-1HZ05fxAFfyT:-MVTSvgd8WiA647YKM2q:b-8ohpin) ([raw file](https://github.com/batfish/batfish/blob/405ea9192419848af460a2bee3351667b8952ebb/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java#L281)):*

> Why not make this line conditional on _undefinedReferences not being empty?
> 
> I was going for consistency with the other structures, but I can do that instead

Actually, I'm not sure this will work, since you won't know if more undefined references will be added later, after the answer element is set
</blockquote></details>

Ok, talked offline and decided it's not a big deal for there to be per-file empty maps for undefined refs.

PR #6716, opened 2021-03-11
FortiOS: policy conversion

corinaminer:


arifogel (commented):
> Reviewed 1 of 4 files at r2, 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 4 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 194 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVU9w3900nSGkbhGHvF:-MVU9w3900nSGkbhGHvG:byeapdp) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L194)):*
> ```Java
>     ImmutableList.Builder<AclLineMatchExpr> srcAddrExprs = ImmutableList.builder();
>     ImmutableList.Builder<AclLineMatchExpr> dstAddrExprs = ImmutableList.builder();
>     for (String src : Sets.intersection(_srcAddr, namedIpSpaces.keySet())) {
> ```

Why the intersection?
And does `_srcAddr` contain a VI name or a VS name? What about `namedIpSpaces`?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 203 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUADeg9Uv5SBLFEsbB:-MVUADeg9Uv5SBLFEsbC:bblfnq3) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L203)):*
> ```Java
>           new MatchHeaderSpace(hs, String.format("Match destination address %s", dst)));
>     }
>     matchConjuncts.add(or(srcAddrExprs.build()));
> ```

An empty or expr is false. Is this what you want?
Applies in multiple places.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 206 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAKnpAC97EcrGNKXr:-MVUAKnpAC97EcrGNKXs:b5cr63k) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L206)):*
> ```Java
>     matchConjuncts.add(or(dstAddrExprs.build()));
> 
>     // Match services. TODO confirm services should be disjuncted
> ```

nit: the word is `disjoined`
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVUAqXTCS5tQmuleBSA:b79ilwf) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*
> ```Java
>   }
> 
>   public static String computeViName(@Nullable String name, String number) {
> ```

You never use both arguments. Please just make two functions.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 4 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVUBQFE8VLkhE3I4_6G:b-y31stl) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

You never use both arguments. Please just make two functions.
</blockquote></details>

Also, doc since public

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 194 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVU9w3900nSGkbhGHvF:-MVUBu3KENdWbQVXQB5c:b-uu3y8i) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L194)):*
> Why the intersection?

In case some addresses were not successfully converted to IpSpaces. So far that's not possible, but figured i'd guard against it. Assuming that relevant warnings would be filed when the address fails to convert.

> And does _srcAddr contain a VI name or a VS name? What about namedIpSpaces?

Addresses have unique names, so they're the same in VS and VI. If for some reason we start renaming them for VI, and forget to update this, we'll know because the policy conversion test will fail.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 203 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUADeg9Uv5SBLFEsbB:-MVUDX-s1eP16ujFXy_L:b-omaq1a) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L203)):*

An empty or expr is false. Is this what you want?
Applies in multiple places.
</blockquote></details>

Policies are required to specify at least one each of src address, dst address, and service, so this should only occur if conversion fails for all referenced src or dst addresses or services. So far, addresses and services always convert, but if this happens in the future, such a policy should not match anything. Added warnings.

Up for discussion: should policies that don't match anything be converted at all? I'd lean towards yes, so that users can discover the problem via filter questions as well as warnings.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 206 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAKnpAC97EcrGNKXr:-MVUDaZDArwoA2b05vvp:b-4ljbxi) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L206)):*

nit: the word is `disjoined`
</blockquote></details>

heh, fixed
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVUMCF3ApEDMCiMZH-t:b-agkgci) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

Also, doc since public
</blockquote></details>

Added doc and nonnull annotation.

Two functions doesn't make sense to me. Callers would then have to keep track of the fact that the configured policy name takes precedence but is nullable; and if we ever want to change how VI policy names are computed, we'd have to change that logic everywhere instead of just in this function.

arifogel (commented):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 194 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVU9w3900nSGkbhGHvF:-MVUcFOpA21dPaQvkmg3:b-8iol7z) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L194)):*
> In case some addresses were not successfully converted to IpSpaces. So far that's not possible, but figured i'd guard against it. Assuming that relevant warnings would be filed when the address fails to convert.

Would this ever happen without a warning during extraction? Is this warning redundant?

> Addresses have unique names, so they're the same in VS and VI. If for some reason we start renaming them for VI, and forget to update this, we'll know because the policy conversion test will fail.

OK.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 203 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUADeg9Uv5SBLFEsbB:-MVUcjxn068-kf2ZG4l_:b-3we96y) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L203)):*
> should policies that don't match anything be converted at all?

Yes, unless they could not be created on the CLI.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVUdDxlC86NYXySZdWn:b5fcjdl) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

Added doc and nonnull annotation.

Two functions doesn't make sense to me. Callers would then have to keep track of the fact that the configured policy name takes precedence but is nullable; and if we ever want to change how VI policy names are computed, we'd have to change that logic everywhere instead of just in this function.
</blockquote></details>

I think I misunderstood what this function was for.

Can two policies with different numbers have the same name? If not, this is OK if a little odd IMO.
If they can, you should always include the number in the VI name.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 194 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVU9w3900nSGkbhGHvF:-MVUj8Fc8N_OODUCXg3F:ba6fpcl) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L194)):*

> In case some addresses were not successfully converted to IpSpaces. So far that's not possible, but figured i'd guard against it. Assuming that relevant warnings would be filed when the address fails to convert.

Would this ever happen without a warning during extraction? Is this warning redundant?

> Addresses have unique names, so they're the same in VS and VI. If for some reason we start renaming them for VI, and forget to update this, we'll know because the policy conversion test will fail.

OK.
</blockquote></details>

I'm assuming that warnings explaining why addresses fail to convert will be generated in the address conversion code. The only warning we might file here is that the policy can't match any traffic, which may be a result of one or more addresses failing to convert, but feels separate enough to merit its own warning.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVUi5ClBa31AjC8ceK_:bbkji8b) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

I think I misunderstood what this function was for.

Can two policies with different numbers have the same name? If not, this is OK if a little odd IMO.
If they can, you should always include the number in the VI name.
</blockquote></details>

Policy names are unique, as are policy numbers. Unfortunately, you *can* have policy 10 with no name and policy 20 with name "10", which is why this function can't just be `firstNonNull(name, number)`. (Technically it can still clash if there's a policy named `~UNNAMED~POLICY~10~`, but i'm trusting users to refrain from that.)

I looked at some FortiOS configs and concluded that it's fairly common to set policy names, so opted to implement this in a name-centric way (rather than, say, always using `NUMBER~NAME` or just using the number).

And, unrelated: Added a non-static wrapper around this function because in interface conversion, i already have the policy object and it's annoying to have to call its name and number getters (and it seems like that'll be the case everywhere in non-test code).

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 97 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXNxDDBJLgm31ITDxY:-MVXNxDDBJLgm31ITDxZ:b-oortxb) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L97)):*
> ```
> Must happen after c._ipSpaces is populated
> ```
nit: I'd specifically say something about address objects
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 201 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXPLxTEWd4CMkpdzlW:-MVXPLxTEWd4CMkpdzlX:bt3kon5) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L201)):*
> ```
> return new MatchHeaderSpace(hs, String.format("Match source address %s", addr));
> ```
we should use smarter, object aware `TraceElement`s here and elsewhere, [like we do for PAN and other vendors](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoTraceElementCreators.java#L52-L64)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 253 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXQbVKEYxRmciLCPE0:-MVXQbVKEYxRmciLCPE1:bhkjoio) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L253)):*
> ```
>     ExprAclLine.Builder line =
>         getActionEffective() == Action.ALLOW ? ExprAclLine.accepting() : ExprAclLine.rejecting();
>     line.setMatchCondition(and(matchConjuncts.build()));
> ```
should `line` have a `TraceElement` as well? indicating a `firewall policy` was matched?
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 624 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXR9nX22sQaq9VkXqJ:-MVXR9nX22sQaq9VkXqK:bikr4k5) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L624)):*
> ```
> public void testFirewallPolicyConversion()
> ```
does it make sense to split this out into `FortiosPolicyTest.java`?
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy, line 52 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXROHVCLo4jOJouhyb:-MVXROHVCLo4jOJouhyc:b2vk5ar) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy#L52)):*
> ```
>         set service custom_tcp_12_from_11
> ```
why change the name? if anything should be `12_to_11` no `12_from_11` since this is for (source port) 12 -> (dest port) 11

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy, line 52 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXROHVCLo4jOJouhyb:-MVX_4rs4S2NTJvITc1j:b-yhja08) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy#L52)):*

> ```
>         set service custom_tcp_12_from_11
> ```
why change the name? if anything should be `12_to_11` no `12_from_11` since this is for (source port) 12 -> (dest port) 11
</blockquote></details>

never mind, missed that you changed the ports as well, so I have a different question:

why remove this dest-port overlap / shadowing between rules? before, this policy was nearly fully shadowing the next policy - only using a different source port would allow the following rule to match

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 4 of 6 files reviewed, 6 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 97 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXNxDDBJLgm31ITDxY:-MVXgD9o2d1gMzAXHcoP:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L97)):*

> ```
> Must happen after c._ipSpaces is populated
> ```
nit: I'd specifically say something about address objects
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 201 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXPLxTEWd4CMkpdzlW:-MVXflivBYRyPx4tur9c:b-d931it) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L201)):*

> ```
> return new MatchHeaderSpace(hs, String.format("Match source address %s", addr));
> ```
we should use smarter, object aware `TraceElement`s here and elsewhere, [like we do for PAN and other vendors](https://github.com/batfish/batfish/blob/master/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoTraceElementCreators.java#L52-L64)
</blockquote></details>

hm, looks like we'd need access to the filename to add that... tempted to push to another PR. what do you think?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 253 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXQbVKEYxRmciLCPE0:-MVXgGGe2YeblbaVCeB8:b-kqfcdg) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L253)):*

> ```
>     ExprAclLine.Builder line =
>         getActionEffective() == Action.ALLOW ? ExprAclLine.accepting() : ExprAclLine.rejecting();
>     line.setMatchCondition(and(matchConjuncts.build()));
> ```
should `line` have a `TraceElement` as well? indicating a `firewall policy` was matched?
</blockquote></details>

Since this ACL doesn't cover the source interfaces constraint, i was going to put that type of trace elements in export policy lines that are like "match src interfaces and match policy ACL".
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 624 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXR9nX22sQaq9VkXqJ:-MVXgfuWEZ4fD9lTcVh_:byevsnj) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L624)):*

> ```
> public void testFirewallPolicyConversion()
> ```
does it make sense to split this out into `FortiosPolicyTest.java`?
</blockquote></details>

ehh, maybe? the general principle i've been using is "if it uses a test config, put it in the grammar test". Wouldn't worry about it for now.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy, line 52 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXROHVCLo4jOJouhyb:-MVXhQQ7DhRKV5cFU2IZ:b-f6e76x) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy#L52)):*

never mind, missed that you changed the ports as well, so I have a different question:

why remove this dest-port overlap / shadowing between rules? before, this policy was nearly fully shadowing the next policy - only using a different source port would allow the following rule to match
</blockquote></details>

oh hm... i wasn't paying attention to the order these are applied. I changed it to better test that both services were allowed by the resulting ACL (since `custom_tcp_11` includes all traffic matched by `custom_tcp_12_from_11`).

I'm not clear on what was tested before that's not being tested after this change -- can you elaborate? or we can chat offline

sfraint (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 201 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXPLxTEWd4CMkpdzlW:-MVXuECd5pWKV93gYyKM:b93gvoz) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L201)):*

hm, looks like we'd need access to the filename to add that... tempted to push to another PR. what do you think?
</blockquote></details>

either is fine

also, since you're relying on so much state from `FortiosConfiguration` might want to consider putting the conversion logic in that instead and keeping the VS model class simpler / closer to just what is in the config
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 253 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXQbVKEYxRmciLCPE0:-MVXvINl9MWXOwFJ0lqx:b4c9lyr) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L253)):*

Since this ACL doesn't cover the source interfaces constraint, i was going to put that type of trace elements in export policy lines that are like "match src interfaces and match policy ACL".
</blockquote></details>

:thumbsup: 
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 624 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXR9nX22sQaq9VkXqJ:-MVXvLN5DqHP1YM6ck-s:b-yd3rpl) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L624)):*

ehh, maybe? the general principle i've been using is "if it uses a test config, put it in the grammar test". Wouldn't worry about it for now.
</blockquote></details>

that's not quite the pattern we use for other vendors, e.g. [PaloAltoNatTest](https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoNatTest.java) or [PaloAltoSecurityRuleTest](https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoSecurityRuleTest.java) where we break out (some) more complex conversion logic tests from basic grammar parsing/extraction tests
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy, line 52 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXROHVCLo4jOJouhyb:-MVXw-NW004ltq7oSaNw:bkhsehr) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy#L52)):*

oh hm... i wasn't paying attention to the order these are applied. I changed it to better test that both services were allowed by the resulting ACL (since `custom_tcp_11` includes all traffic matched by `custom_tcp_12_from_11`).

I'm not clear on what was tested before that's not being tested after this change -- can you elaborate? or we can chat offline
</blockquote></details>

it's not a big deal - we're not actually testing this yet

when we need to test this functionality we can just explicitly add shadowing cases to the config that test uses

arifogel (commented):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 5 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVXxyxI6u-M8DZFBj0D:bfh10qe) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

Policy names are unique, as are policy numbers. Unfortunately, you *can* have policy 10 with no name and policy 20 with name "10", which is why this function can't just be `firstNonNull(name, number)`. (Technically it can still clash if there's a policy named `~UNNAMED~POLICY~10~`, but i'm trusting users to refrain from that.)

I looked at some FortiOS configs and concluded that it's fairly common to set policy names, so opted to implement this in a name-centric way (rather than, say, always using `NUMBER~NAME` or just using the number).

And, unrelated: Added a non-static wrapper around this function because in interface conversion, i already have the policy object and it's annoying to have to call its name and number getters (and it seems like that'll be the case everywhere in non-test code).
</blockquote></details>

I think in general we need to stop putting so much non-data-manipulating logic in VS classes. You could just have a static function that computes the VI name given a `Policy` parameter in `FortiosConfiguration`, which is no less concise and follows existing pattern.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 222 at r3](https://reviewable.io/reviews/batfish/batfish/6716#-MVUAqXTCS5tQmuleBS9:-MVXzJKO6lDSbJKTP-Cx:b612ql0) ([raw file](https://github.com/batfish/batfish/blob/792fcb8e62f47e02bb787fbb8612e6f90d85781f/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L222)):*

I think in general we need to stop putting so much non-data-manipulating logic in VS classes. You could just have a static function that computes the VI name given a `Policy` parameter in `FortiosConfiguration`, which is no less concise and follows existing pattern.
</blockquote></details>

ok, moved naming methods into `FortiosConfiguration`. (Still want the two-parameter method because tests should be able to compute VI policy names without accessing the vendor config.)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 201 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXPLxTEWd4CMkpdzlW:-MVXyy8j0u2-qzYqcIig:bdjcgpl) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L201)):*

either is fine

also, since you're relying on so much state from `FortiosConfiguration` might want to consider putting the conversion logic in that instead and keeping the VS model class simpler / closer to just what is in the config
</blockquote></details>

yeah i see what you mean. Moved conversion logic into `FortiosConfiguration` and improved trace elements (since filename is readily available there)
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 624 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXR9nX22sQaq9VkXqJ:-MVXzmALDEfP9XVYE1EZ:b-qjvi0o) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L624)):*

that's not quite the pattern we use for other vendors, e.g. [PaloAltoNatTest](https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoNatTest.java) or [PaloAltoSecurityRuleTest](https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoSecurityRuleTest.java) where we break out (some) more complex conversion logic tests from basic grammar parsing/extraction tests
</blockquote></details>

yeah that's true. can always split some tests out later if we decide this file is getting overrun. I think it's fine to leave here for now
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy, line 52 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXROHVCLo4jOJouhyb:-MVY-ZU22Scq1sR6BXGV:bevkaw9) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/firewall_policy#L52)):*

it's not a big deal - we're not actually testing this yet

when we need to test this functionality we can just explicitly add shadowing cases to the config that test uses
</blockquote></details>

ok. also wanted to clarify that when i said "`custom_tcp_11` includes all traffic matched by `custom_tcp_12_from_11`" i actually meant that `custom_tcp_11` includes all traffic that _would've_ been matched by the _old_ `custom_tcp_11_from_12`.

arifogel (commented):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 211 at r8](https://reviewable.io/reviews/batfish/batfish/6716#-MVY6a9DA38PoMBqFu_3:-MVY6a9ECXdlmQdgCJGn:b-scu6al) ([raw file](https://github.com/batfish/batfish/blob/95e2f0f618ad08f5927166ccd6b18027dbe7b934/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L211)):*
> ```Java
>   public static @Nonnull String computeViPolicyName(@Nullable String name, String number) {
>     // TODO: Might need to generate IpAccessList names per VRF/VDOM
>     return Optional.ofNullable(name).orElseGet(() -> String.format("~UNNAMED~POLICY~%s~", number));
> ```

optional per style preference: `return name != null ? name : String.format(...)`

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 201 at r6](https://reviewable.io/reviews/batfish/batfish/6716#-MVXPLxTEWd4CMkpdzlW:-MVYBTR047M8Dpm5M0fq:b4c9lyr) ([raw file](https://github.com/batfish/batfish/blob/a4648ad7245706326fb3202e01e9428e475fccb6/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L201)):*

yeah i see what you mean. Moved conversion logic into `FortiosConfiguration` and improved trace elements (since filename is readily available there)
</blockquote></details>

:thumbsup:

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6716)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 211 at r8](https://reviewable.io/reviews/batfish/batfish/6716#-MVY6a9DA38PoMBqFu_3:-MVYDKpoA0j0-MeS015J:b-3k24fg) ([raw file](https://github.com/batfish/batfish/blob/95e2f0f618ad08f5927166ccd6b18027dbe7b934/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L211)):*

optional per style preference: `return name != null ? name : String.format(...)`
</blockquote></details>

i don't have a strong preference, and i think the pr is otherwise set

PR #6720, opened 2021-03-12
FortiOS: Policy edit block validation

corinaminer:
Fyi, I didn't find any invalid lines you can enter in the CLI that will pop you out of the policy edit block.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
@@ -419,7 +419,6 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
   @Override
   public void enterCfp_edit(Cfp_editContext ctx) {
     Optional<Long> number = toLong(ctx, ctx.policy_number());
-    _currentPolicyValid = number.isPresent();
     Policy existing = number.map(Object::toString).map(_c.getPolicies()::get).orElse(null);
     if (existing != null) {
       // Make a clone to edit
@@ -427,14 +426,15 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
     } else {
       _currentPolicy = new Policy(toString(ctx.policy_number().str()));
     }
+    _currentPolicy.setValid(number.isPresent());
   }
 
   @Override
   public void exitCfp_edit(Cfp_editContext ctx) {
     // If edited policy is valid, add/update the entry in VS map
-    // TODO: Better validity checking
     String number = _currentPolicy.getNumber();
-    if (_currentPolicyValid) {
+    String invalidReason = _currentPolicy.getInvalidReason();
+    if (invalidReason == null) { // policy is valid
       _c.defineStructure(FortiosStructureType.POLICY, number, ctx);
       _c.referenceStructure(
           FortiosStructureType.POLICY,
@@ -442,6 +442,8 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
           FortiosStructureUsage.POLICY_SELF_REF,
           ctx.start.getLine());
       _c.getPolicies().put(number, _currentPolicy);
+    } else {
+      warn(ctx, String.format("Policy edit block ignored: %s", invalidReason));
     }
     _currentPolicy = null;
   }
@@ -1175,7 +1177,6 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
   private Address _currentAddress;
   private Interface _currentInterface;
   private Policy _currentPolicy;
-  private boolean _currentPolicyValid;
   private Replacemsg _currentReplacemsg;
   private Service _currentService;
   private final @Nonnull FortiosConfiguration _c;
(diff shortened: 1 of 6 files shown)


sfraint (commented):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6720)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 157 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZSs867UL69O8Nx85z:-MVZSs867UL69O8Nx86-:bf4voys) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L157)):*
> ```
>   public void setValid(boolean valid) {
>     _valid = valid;
>   }
> ```
do we ever expect (or want) this to make the policy valid?

i.e. should we just expose a `setInvalid()` method with no params?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 188 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZTEp9914pm-T1KRFp:-MVZTEp9914pm-T1KRFq:bxm39q) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L188)):*
> ```
> Whether this policy has invalid lines that would prevent it from being committed in CLI
> ```
nit: could emphasize the entire policy is prevented from being committed (not just the invalid lines)
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 739 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZTcGT1r0VczBTg6pP:-MVZTcGT1r0VczBTg6pQ:b-347xx4) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L739)):*
> ```
>                 hasComment("Expected policy number in range 0-4294967294, but got '4294967295'"),
>                 hasComment("Expected policy number in range 0-4294967294, but got 'not_a_number'"),
> ```
should we remove these warnings if they're redundant with the block-validity check?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6720)* status: all files reviewed, 2 unresolved discussions (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 157 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZSs867UL69O8Nx85z:-MVZWNtzB6AXJR8imbWj:ben16xv) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L157)):*

> ```
>   public void setValid(boolean valid) {
>     _valid = valid;
>   }
> ```
do we ever expect (or want) this to make the policy valid?

i.e. should we just expose a `setInvalid()` method with no params?
</blockquote></details>

yeah, we do need to set the policy to valid when we make clones (valid is transient so doesn't copy -- and shouldn't have to, since any policy we clone is already in the VS model and therefore must be valid). you can see in `enterCfp_edit` that i'm setting `valid` whether or not the policy already exists.

I initially thought i could just set `_valid = true` in the constructor (and implement `setInvalid()` as you suggest), but that didn't work for clones; apparently the constructor isn't used for cloning.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java, line 188 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZTEp9914pm-T1KRFp:-MVZYKG3-hDCLsvV2YvP:b634ivw) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/main/java/org/batfish/representation/fortios/Policy.java#L188)):*

> ```
> Whether this policy has invalid lines that would prevent it from being committed in CLI
> ```
nit: could emphasize the entire policy is prevented from being committed (not just the invalid lines)
</blockquote></details>

"it" here is meant to refer to the policy... maybe not the clearest comment but not worth rerunning CI
___
*[projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java, line 739 at r2](https://reviewable.io/reviews/batfish/batfish/6720#-MVZTcGT1r0VczBTg6pP:-MVZX2lO4IpRhE-kuhM3:b-2g5hek) ([raw file](https://github.com/batfish/batfish/blob/0b6746ee5d28a4e4b1a487a8856b85219d33200d/projects/batfish/src/test/java/org/batfish/grammar/fortios/FortiosGrammarTest.java#L739)):*

> ```
>                 hasComment("Expected policy number in range 0-4294967294, but got '4294967295'"),
>                 hasComment("Expected policy number in range 0-4294967294, but got 'not_a_number'"),
> ```
should we remove these warnings if they're redundant with the block-validity check?
</blockquote></details>

it's a bit redundant, but i'd opt to keep these because:
- these warnings have more detail about why the name isn't valid
- if we ever do find another reason to set `valid` to false, then the other warning is going to be even more vague
- it'd be bad practice to block `toLong(ctx, ctx.policy_number())` from filing a warning; if policies are referred to from anywhere else, those extraction contexts should be able to depend on `toLong` to file warnings.

PR #6727, opened 2021-03-13
FortiOS: Service edit block validation

corinaminer:
Also minor extraction improvement for services: can't set ICMP code before type.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
@@ -610,20 +610,26 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
   @Override
   public void enterCfsc_edit(Cfsc_editContext ctx) {
     Optional<String> name = toString(ctx, ctx.service_name());
-    if (!name.isPresent()) {
-      _currentService = new Service(toString(ctx.service_name().str()), getUUID()); // dummy
-      return;
+    Service existing = name.map(_c.getServices()::get).orElse(null);
+    if (existing != null) {
+      // Make a clone to edit
+      _currentService = SerializationUtils.clone(existing);
+    } else {
+      _currentService = new Service(toString(ctx.service_name().str()), getUUID());
     }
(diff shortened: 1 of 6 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6727)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Service.java, line 254 at r1](https://reviewable.io/reviews/batfish/batfish/6727#-MVeIEMM3T-oo_1w9rcd:-MVeIEMM3T-oo_1w9rce:b-aqcvir) ([raw file](https://github.com/batfish/batfish/blob/5b32089cf028542ff259abea23fe4ff080e61e04/projects/batfish/src/main/java/org/batfish/representation/fortios/Service.java#L254)):*
> ```Java
> 
>   // Whether this service has invalid lines that would prevent committing the service in CLI
>   private transient boolean _valid;
> ```

I realize this is transient, but this is still polluting the vendor model.

This isn't even a property of a service; it's a property of the current proposed change to a service. You will never have two Service objects for which `_valid` must be remembered.

This kind of state - and the logic to compute it - should be kept in the extractor.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6727)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Service.java, line 254 at r1](https://reviewable.io/reviews/batfish/batfish/6727#-MVeIEMM3T-oo_1w9rcd:-MVrC_OO0HsTuYspGfzt:bmf38qk) ([raw file](https://github.com/batfish/batfish/blob/5b32089cf028542ff259abea23fe4ff080e61e04/projects/batfish/src/main/java/org/batfish/representation/fortios/Service.java#L254)):*

I realize this is transient, but this is still polluting the vendor model.

This isn't even a property of a service; it's a property of the current proposed change to a service. You will never have two Service objects for which `_valid` must be remembered.

This kind of state - and the logic to compute it - should be kept in the extractor.
</blockquote></details>

per discussion, deferring to separate PR

PR #6734, opened 2021-03-16
ios use new community processing api

arifogel:
- copy implementation from nx-os
  - throw out sequence numbers, which are nx-os only
- delete EOS setting{,additive} community list in route-map from IOS vendor model
  - force EOS ref unit test file with this feature to be read as EOS instead of IOS
- remove old wonky conversion tests


dhalperi (commented):
> Reviewed 27 of 27 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6734)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6741, opened 2021-03-16
batfish: remove all SMT and z3-related code and tests

dhalperi:
Preserve TransferBDD and the things it uses.

Delete:
* z3 install script, maven config, bazel config
* smt package everywhere
* abstraction package everywhere, except preserving the InterfacePolicy
  class used by TransferBDD.
* utility classes, matchers, and the like that were smt-related.
* unused classes and code


arifogel (commented):
> Reviewed 306 of 309 files at r1, 6 of 6 files at r2, 10 of 10 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6741)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6759, opened 2021-03-21
Arista: redo parsing for ip access-list stanza

### LINKED ISSUE #6758: ip as-path access-lists on Arista EOS
I think I may have found something worth review by the Batfish dev team:

While doing some route analysis with Batfish, I noticed that in my "live" network, I was able to recieve a bgp route announcement at the ultimate destination, six routers away.  However, the announcement was not getting past the third router of the six, in my Batfish test setup. After much investigation, there was a route map set up on a BGP peer using an as-path access list and here's the key: it had the "any" keyword at the end. According to Arista EOS docs (https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1116711) the "any" keyword denotes the origin type and it seems like it may have been causing an issue for Batfish. Running the Test Route Policy question against the route map returned that the prefix would be denied by the route map (which is why the announcement wasn't making it past the third router) when in fact the as-path regex was correct to permit the advertisement of the prefix. I removed the "any" keyword from the end of the as-path access list in the config file, and ran Batfish again. This time all worked as expected and the routing mirrored that of my live network. It took me some time to t-shoot and debug this one, so passing it along with the hope that it helps.

dhalperi:
* Delete old generic code, add new file
* Fix modes to properly extract words
* Add extraction tests
* Add warnings

Fix #6758

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4
@@ -5777,39 +5777,41 @@ M_AsPath_WS
 
 mode M_AsPathAccessList;
 
-M_AsPathAccessList_DEC
+M_AsPathAccessList_WORD
 :
-   F_Digit+ -> type ( DEC )
-;
-
-M_AsPathAccessList_DENY
-:
-   'deny' -> type ( DENY ) , mode ( M_Description )
+   F_NonWhitespace+ -> type(WORD), mode(M_AsPathAccessList_Action)
 ;
 
 M_AsPathAccessList_NEWLINE
 :
-   F_Newline+ -> type ( NEWLINE ) , mode ( DEFAULT_MODE )
-;
-
-M_AsPathAccessList_PERMIT
-:
-   'permit' -> type ( PERMIT ) , mode ( M_Description )
-;
-
-M_AsPathAccessList_SEQ
-:
-   'seq' -> type ( SEQ )
-;
-
-M_AsPathAccessList_VARIABLE
-:
-   F_Variable_RequiredVarChar F_Variable_VarChar* -> type ( VARIABLE )
+   F_Newline+ -> type(NEWLINE), popMode
 ;
 
 M_AsPathAccessList_WS
 :
-   F_Whitespace+ -> channel ( HIDDEN )
+   F_Whitespace+ -> channel(HIDDEN)
+;
+
+mode M_AsPathAccessList_Action;
+
+M_AsPathAccessList_Action_DENY
+:
+   'deny' -> type(DENY), mode(M_Word)
+;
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6759)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4, line 5778 at r1](https://reviewable.io/reviews/batfish/batfish/6759#-MWLhcjAETdDPCb8cWfO:-MWLhcjAETdDPCb8cWfP:bugnaw0) ([raw file](https://github.com/batfish/batfish/blob/ac3b5ddb9c0ca97203af12783854ae12d3dac1aa/projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4#L5778)):*
> ```
>  M_AsPathAccessList;
> ```
FYI this mode is entered only from `M_AsPath`
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4, line 5799 at r1](https://reviewable.io/reviews/batfish/batfish/6759#-MWLhhJpC2QoSbldrfDx:-MWLhhJpC2QoSbldrfDy:blf00h4) ([raw file](https://github.com/batfish/batfish/blob/ac3b5ddb9c0ca97203af12783854ae12d3dac1aa/projects/batfish/src/main/antlr4/org/batfish/grammar/arista/AristaLexer.g4#L5799)):*
> ```ANTLR
> M_AsPathAccessList_Action_DENY
> :
>    'deny' -> type(DENY), mode(M_Word)
> ```

FYI there is no quoted option for the regexes

arifogel (commented):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6759)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/test/java/org/batfish/grammar/arista/AristaGrammarTest.java, line 869 at r1](https://reviewable.io/reviews/batfish/batfish/6759#-MWQ9Ml_7nKjcuHYGc0D:-MWQ9Ml_7nKjcuHYGc0E:b-kznag) ([raw file](https://github.com/batfish/batfish/blob/ac3b5ddb9c0ca97203af12783854ae12d3dac1aa/projects/batfish/src/test/java/org/batfish/grammar/arista/AristaGrammarTest.java#L869)):*
> ```Java
>     assertThat(list1.getLines(), hasSize(5));
>     {
>       IpAsPathAccessListLine line = list1.getLines().get(0);
> ```

optional: use iterator instead of fixed indices to make test easier to maintain if you insert lines in the config.

PR #6764, opened 2021-03-22
SearchRoutePolicies: Support a special case of matching on community-set regexes

millstein:
Adds support to SearchRoutePolicies for a special case of regex matching on community sets.  When the regex only applies to a single community within the set, then we can treat it as a regex on an individual community and use the existing machinery for community regexes to handle it properly.  There is still no support for more general community-set regexes.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java
@@ -98,17 +98,17 @@ public class CommunitySetMatchExprToBDD
 
   @Override
   public BDD visitCommunitySetMatchRegex(CommunitySetMatchRegex communitySetMatchRegex, Arg arg) {
-    // NOTE: when implementing, update CommunitySetMatchExprVarCollector#visitCommunitySetMatchRegex
-    BDD bdd =
-        communityVarsToBDD(
-            communitySetMatchRegex.accept(
-                new CommunitySetMatchExprVarCollector(), arg.getTransferBDD().getConfiguration()),
-            arg);
-    if (bdd.isZero()) {
+    // Only handles the special case when the regex applies to a single community within the set.
+    // If that is not the case, the collected set of community variables will be empty (see
+    // CommunitySetMatchExprVarCollector#visitCommunitySetMatchRegex) and we throw an exception.
+    Set<CommunityVar> cvars =
+        communitySetMatchRegex.accept(
+            new CommunitySetMatchExprVarCollector(), arg.getTransferBDD().getConfiguration());
+    if (cvars.isEmpty()) {
       throw new UnsupportedOperationException(
           "Currently not supporting arbitrary community set regexes");
     } else {
-      return bdd;
+      return communityVarsToBDD(cvars, arg);
     }
   }
 
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java
@@ -24,6 +24,7 @@ import org.batfish.datamodel.routing_policy.communities.CommunitySetMatchRegex;
 import org.batfish.datamodel.routing_policy.communities.CommunitySetNot;
 import org.batfish.datamodel.routing_policy.communities.HasCommunity;
 import org.batfish.minesweeper.CommunityVar;
+import org.batfish.minesweeper.communities.CommunitySetMatchExprVarCollector;
 
 /**
  * Create a BDD from a {@link
@@ -98,7 +99,17 @@ public class CommunitySetMatchExprToBDD
   @Override
   public BDD visitCommunitySetMatchRegex(CommunitySetMatchRegex communitySetMatchRegex, Arg arg) {
     // NOTE: when implementing, update CommunitySetMatchExprVarCollector#visitCommunitySetMatchRegex
-    throw new UnsupportedOperationException("Currently not supporting community set regexes");
+    BDD bdd =
+        communityVarsToBDD(
+            communitySetMatchRegex.accept(
+                new CommunitySetMatchExprVarCollector(), arg.getTransferBDD().getConfiguration()),
+            arg);
+    if (bdd.isZero()) {
+      throw new UnsupportedOperationException(
+          "Currently not supporting arbitrary community set regexes");
+    } else {
+      return bdd;
+    }
   }
 
   @Override
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java
@@ -67,7 +67,7 @@ public class CommunitySetMatchExprVarCollector
     if (cvar.toAutomaton().isEmpty()) {
       return ImmutableSet.of();
     } else {
-      return ImmutableSet.of(CommunityVar.from(regex));
+      return ImmutableSet.of(cvar);
     }
   }
 
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java
@@ -59,9 +59,16 @@ public class CommunitySetMatchExprVarCollector
   @Override
   public Set<CommunityVar> visitCommunitySetMatchRegex(
       CommunitySetMatchRegex communitySetMatchRegex, Configuration arg) {
-    // This is not supported, but rather than throw we do nothing. If we end up needing to model
-    // this structure, the later code will crash instead.
-    return ImmutableSet.of();
+    // See if this regex can be treated as one that applies to a single community.  If so,
+    // we collect it and treat it as such.  If not, we ignore this regex; later if the symbolic
+    // analysis ever needs to consider this regex then it will fail with an exception.
+    String regex = communitySetMatchRegex.getRegex();
+    CommunityVar cvar = CommunityVar.from(regex);
+    if (cvar.toAutomaton().isEmpty()) {
+      return ImmutableSet.of();
+    } else {
+      return ImmutableSet.of(CommunityVar.from(regex));
+    }
   }
 
   @Override
--- a/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDDTest.java
+++ b/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDDTest.java
@@ -26,8 +26,10 @@ import org.batfish.datamodel.routing_policy.communities.CommunitySetAclLine;
 import org.batfish.datamodel.routing_policy.communities.CommunitySetMatchAll;
 import org.batfish.datamodel.routing_policy.communities.CommunitySetMatchAny;
 import org.batfish.datamodel.routing_policy.communities.CommunitySetMatchExprReference;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetMatchRegex;
 import org.batfish.datamodel.routing_policy.communities.CommunitySetNot;
 import org.batfish.datamodel.routing_policy.communities.HasCommunity;
+import org.batfish.datamodel.routing_policy.communities.TypesFirstAscendingSpaceSeparated;
 import org.batfish.datamodel.routing_policy.expr.LiteralCommunity;
 import org.batfish.minesweeper.CommunityVar;
 import org.batfish.minesweeper.Graph;
@@ -141,6 +143,19 @@ public class CommunitySetMatchExprToBDDTest {
     CommunitySetMatchExprReference csmer = new CommunitySetMatchExprReference(name);
 
     BDD result = _communitySetMatchExprToBDD.visitCommunitySetMatchExprReference(csmer, _arg);
+
+    CommunityVar cvar = CommunityVar.from("^20:");
+
+    assertEquals(cvarToBDD(cvar), result);
+  }
+
+  @Test
+  public void testVisitCommunitySetMatchRegex() {
+    CommunitySetMatchRegex csmr =
+        new CommunitySetMatchRegex(
+            new TypesFirstAscendingSpaceSeparated(ColonSeparatedRendering.instance()), "^20:");
+
+    BDD result = _communitySetMatchExprToBDD.visitCommunitySetMatchRegex(csmr, _arg);
 
     CommunityVar cvar = CommunityVar.from("^20:");
 
--- a/projects/minesweeper/src/test/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollectorTest.java
+++ b/projects/minesweeper/src/test/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollectorTest.java
@@ -131,6 +131,19 @@ public class CommunitySetMatchExprVarCollectorTest {
   }
 
   @Test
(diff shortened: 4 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWQwrGq4gtWGFLa02Q-:bvu8xlh) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>     CommunityVar cvar = CommunityVar.from(regex);
>     if (cvar.toAutomaton().isEmpty()) {
>       return ImmutableSet.of();
>     } else {
>       return ImmutableSet.of(cvar);
>     }
> ```
</details>


This is just ignorance: can you illuminate me as to how this logic works? I looked around in CommunityVar but got lost at the automata bit.

Suppose we had some function`boolean canBeTreatedLikeASingleRegex(String s)`.

I'm guessing that would return the following for Cisco inputs:
* false: `_123$`. (requires community 123 to be last)
* false: `^123` (requires community `123(.*)` to be first)
* true: `123` (and matches any community `\d*123\d*`)
* true: `_123` (and matches any community `123\d*`)
* true: `_123_` (and matches community `123`)
* true: `_123:5_` (and matches community `123:5`)
* true: `123:5_` (and matches community `\d*123:5`)
* false: `_12_ _34_` (requires `12` and `34`)
* false: `_12 34_` (requires `12` and `34`)
* false: `_12 34` (requires `12` and `34\d*`)
...

I can refresh how we compile these into VI regexes.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWR4Vr_6JWeI5iWtO6f:b-wis8f8) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

> ```
>     CommunityVar cvar = CommunityVar.from(regex);
>     if (cvar.toAutomaton().isEmpty()) {
>       return ImmutableSet.of();
>     } else {
>       return ImmutableSet.of(cvar);
>     }
> ```

This is just ignorance: can you illuminate me as to how this logic works? I looked around in CommunityVar but got lost at the automata bit.

Suppose we had some function`boolean canBeTreatedLikeASingleRegex(String s)`.

I'm guessing that would return the following for Cisco inputs:
* false: `_123$`. (requires community 123 to be last)
* false: `^123` (requires community `123(.*)` to be first)
* true: `123` (and matches any community `\d*123\d*`)
* true: `_123` (and matches any community `123\d*`)
* true: `_123_` (and matches community `123`)
* true: `_123:5_` (and matches community `123:5`)
* true: `123:5_` (and matches community `\d*123:5`)
* false: `_12_ _34_` (requires `12` and `34`)
* false: `_12 34_` (requires `12` and `34`)
* false: `_12 34` (requires `12` and `34\d*`)
...

I can refresh how we compile these into VI regexes.
</blockquote></details>

Thanks, I just realized that there is an issue with the current approach and was going to try to fix it, so your examples are right on.  Currently all of them will do what you say above, except that the first two will erroneously say true instead of false.

What happens currently is that we add .* to both ends of the given regex and then intersect (the automaton for) it with (that for) CommunityVar::COMMUNITY_REGEX, which represents all valid community regexes.  That final automaton then represents the regex.

I thought that same approach would work here but realized that it won't because it will erroneously allow regexes that include ^ and $.

So I will need a separate canBeTreatedLikeASingleRegex function, as you said.  My current thinking is that such a function can do something similar to the above but replace COMMUNITY_REGEX with an identical regex except that the leading and trailing ^ and $ are each replaced with a space.  I.e, if the intersection with that regex is non-empty, then it is allowed; otherwise not.  Do you agree?

dhalperi:
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWR7x7f1xCdKgtpKGEU:b-y73tk8) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

Thanks, I just realized that there is an issue with the current approach and was going to try to fix it, so your examples are right on.  Currently all of them will do what you say above, except that the first two will erroneously say true instead of false.

What happens currently is that we add .* to both ends of the given regex and then intersect (the automaton for) it with (that for) CommunityVar::COMMUNITY_REGEX, which represents all valid community regexes.  That final automaton then represents the regex.

I thought that same approach would work here but realized that it won't because it will erroneously allow regexes that include ^ and $.

So I will need a separate canBeTreatedLikeASingleRegex function, as you said.  My current thinking is that such a function can do something similar to the above but replace COMMUNITY_REGEX with an identical regex except that the leading and trailing ^ and $ are each replaced with a space.  I.e, if the intersection with that regex is non-empty, then it is allowed; otherwise not.  Do you agree?
</blockquote></details>

I guess `.` or `?` are also problematic?

Also it will be useful to be informed by the underscore handling. Underscore means asn boundary, so I think it is getting turned into begin, end, or middle of: as-path [`^`, ` `, `$`] and of as-set [`{`, `, `, `}`]. That looks annoying to deal with.

Here's how we compile Cisco expanded regexes into Java regexes. Caveat: I'm not sure this is great code.

```java
static String toJavaRegex(String ciscoRegex) {
    String withoutQuotes;
    if (ciscoRegex.charAt(0) == '"' && ciscoRegex.charAt(ciscoRegex.length() - 1) == '"') {
      withoutQuotes = ciscoRegex.substring(1, ciscoRegex.length() - 1);
    } else {
      withoutQuotes = ciscoRegex;
    }
    String underscoreReplacement = "(,|\\\\{|\\\\}|^|\\$| )";
    String output = withoutQuotes.replaceAll("_", underscoreReplacement);
    return output;
  }
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWR99ySDbKw2cve1Wju:b-jw0a1v) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

I guess `.` or `?` are also problematic?

Also it will be useful to be informed by the underscore handling. Underscore means asn boundary, so I think it is getting turned into begin, end, or middle of: as-path [`^`, ` `, `$`] and of as-set [`{`, `, `, `}`]. That looks annoying to deal with.

Here's how we compile Cisco expanded regexes into Java regexes. Caveat: I'm not sure this is great code.

```java
static String toJavaRegex(String ciscoRegex) {
    String withoutQuotes;
    if (ciscoRegex.charAt(0) == '"' && ciscoRegex.charAt(ciscoRegex.length() - 1) == '"') {
      withoutQuotes = ciscoRegex.substring(1, ciscoRegex.length() - 1);
    } else {
      withoutQuotes = ciscoRegex;
    }
    String underscoreReplacement = "(,|\\\\{|\\\\}|^|\\$| )";
    String output = withoutQuotes.replaceAll("_", underscoreReplacement);
    return output;
  }
```
</blockquote></details>

Right, I dealt with underscore handling already so that seems ok.  

Anything containing . and ? will be rejected.

My proposal was that given a regex r, we produce the intersection of .*r.* with _num:num_ (where the underscores are actually spaces).  If the result is non-empty then this regex is a keeper; otherwise not.

One issue with this approach is if the regex is simply .* -- in that case the intersection is non-empty but I'm not sure if .* on a single community and .* on a community set mean the same thing.  Presumably the latter allows the empty set, while the former is requiring there to be some community.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRB-sy1DaposS5HAYi:b-crax6m) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

Right, I dealt with underscore handling already so that seems ok.  

Anything containing . and ? will be rejected.

My proposal was that given a regex r, we produce the intersection of .*r.* with _num:num_ (where the underscores are actually spaces).  If the result is non-empty then this regex is a keeper; otherwise not.

One issue with this approach is if the regex is simply .* -- in that case the intersection is non-empty but I'm not sure if .* on a single community and .* on a community set mean the same thing.  Presumably the latter allows the empty set, while the former is requiring there to be some community.
</blockquote></details>

Whoops formatting changed my regexes.  I meant the intersection of .\*r.\* with <space>num:num<space>.

dhalperi:
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRE0lG9PnLnLuWlqkE:bt1ll19) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*
I'm trying to reconcile these two comments:
> Anything containing `.` and `?` will be rejected.
Dan: did you mean `or`? If not, why?

> One issue with this approach is if the regex is simply `.*`  ...
Dan: if I read the above, this regex is rejected anyway?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRGWE1CuoKWtqFOUng:b-ze6oks) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

I'm trying to reconcile these two comments:
> Anything containing `.` and `?` will be rejected.
Dan: did you mean `or`? If not, why?

> One issue with this approach is if the regex is simply `.*`  ...
Dan: if I read the above, this regex is rejected anyway?
</blockquote></details>

I realize that this approach is doomed to fail.  I should not be intersecting with what I want, because that ends up ignoring all the stuff that can match the regex that I don't want.  .* is accepted because its intersection with num:num is non-empty, for example, and for the same reason a regex that is an or of a few different things, some of them not supported, can still be accepted.

What I should do instead is impose a strict requirement that the given regex have a particular form that I know is safe, such as [_][num]:[num][_] and so only accept them.  One annoying thing there is that I can't use _ but instead have to use the internal translation, as you showed, but that is vendor-specific.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRHXgT-pmDr6CAFRAy:bstrejp) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

I realize that this approach is doomed to fail.  I should not be intersecting with what I want, because that ends up ignoring all the stuff that can match the regex that I don't want.  .* is accepted because its intersection with num:num is non-empty, for example, and for the same reason a regex that is an or of a few different things, some of them not supported, can still be accepted.

What I should do instead is impose a strict requirement that the given regex have a particular form that I know is safe, such as [_][num]:[num][_] and so only accept them.  One annoying thing there is that I can't use _ but instead have to use the internal translation, as you showed, but that is vendor-specific.
</blockquote></details>

[\_][num]:[num][\_]

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRJN71CfP1E11BLKHQ:bn7qo6m) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

[\_][num]:[num][\_]
</blockquote></details>

Alternately, we could try to do this at the IOS level - only use MatchSetExpr as a last resort. (This is likely more efficient to interpret, too)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6764)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java, line 71 at r1](https://reviewable.io/reviews/batfish/batfish/6764#-MWQwrGq4gtWGFLa02Pz:-MWRQ45MB3fTVr7hUfCe:bgn6068) ([raw file](https://github.com/batfish/batfish/blob/f27749a150352e665a4d9b9da47b563fbfded559/projects/minesweeper/src/main/java/org/batfish/minesweeper/communities/CommunitySetMatchExprVarCollector.java#L71)):*

Alternately, we could try to do this at the IOS level - only use MatchSetExpr as a last resort. (This is likely more efficient to interpret, too)
</blockquote></details>

Good idea.  Looks like I could do that in CiscoConfiguration::toCommunitySetAclLine(ExpandedCommunityListLine) -- if the regex meets a specific format then create a HasCommunity(CommunityMatchRegex) instead of a CommunitySetMatchRegex.

If you agree, I'll go that route.

PR #6768, opened 2021-03-23
FortiOS: service group VS support

sfraint:
Add initial service-group paring and extraction.


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768)* status: 0 of 14 files reviewed, all discussions resolved (waiting on @corinaminer)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6768#-MWRV4aE-NG5EXMpSmH2:-MWRV4aE-NG5EXMpSmH3:b-vgr1zx) (no related file):*
Rename and conversion will be handled in separate PR(s)

---

corinaminer (commented):
> Reviewed 13 of 14 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768)* status: all files reviewed, 4 unresolved discussions (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1116 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWS8QEE3hazRqJ0vUIt:-MWS8QEE3hazRqJ0vUIu:bojqdp4) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1116)):*
> ```Java
>                   "Service group %s cannot be added to %s as it would create a cycle",
>                   name, _currentServiceGroup.getName()));
>           return Optional.empty();
> ```

I'd opt to implement this cycle-checking logic in a separate helper function, and call it directly from the `exit` methods for set/append member. it feels like we're overloading this method, whose purpose was previously just to guarantee all the names are valid, which is consistent with all the other `toX(ctx)` functions.

Separately, i'd avoid putting the recursive `contains` logic into `ServiceGroup`, just to keep the VS model pure. If you factor out a separate helper function, it feels reasonable to put the whole cycle-finding logic in that. (Could put the helper in `FortiosPolicyConversions` or a new conversions file to help keep things organized.)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 199 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSAqNpASHBvr4mu35s:-MWSAqNpASHBvr4mu35t:bss0ugf) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L199)):*
> ```Java
>     }
>     TraceElement te =
>         service instanceof Service ? matchServiceTraceElement((Service) service, _filename) : null;
> ```

What about just making `matchServiceTraceElement` take in a `ServiceGroupMember` instead (and then generate slightly different trace elements depending on the input's type)? I expect we'll want a "Matched service group foo" trace element anyway.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java, line 10 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSBJfTE1c5Jfywn1tL:-MWSBJfTE1c5Jfywn1tM:b1m2t2i) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java#L10)):*
> ```Java
>  * FortiOS datamodel interface representing objects which can be a member of firewall service groups
>  */
> public interface ServiceGroupMember extends FortiosRenameableObject {
> ```

Is there a reason this has to be an interface rather than an abstract class? If it were an abstract class, it could own `_comment` and implement the getter/setter itself.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_group_warnings, line 23 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSFSgj5krj5SzVDTGK:-MWSFSgj5krj5SzVDTGL:b-8fvzqv) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_group_warnings#L23)):*
> ```
>     next
>     edit valid
>         append member "self_ref_not_allowed"
> ```

could you also add `append member "valid"` to make sure it recognizes that cycle?

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768)* status: all files reviewed, 4 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1116 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWS8QEE3hazRqJ0vUIt:-MWV5sBO4CJUcSGt3IFo:b4d4i49) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1116)):*

I'd opt to implement this cycle-checking logic in a separate helper function, and call it directly from the `exit` methods for set/append member. it feels like we're overloading this method, whose purpose was previously just to guarantee all the names are valid, which is consistent with all the other `toX(ctx)` functions.

Separately, i'd avoid putting the recursive `contains` logic into `ServiceGroup`, just to keep the VS model pure. If you factor out a separate helper function, it feels reasonable to put the whole cycle-finding logic in that. (Could put the helper in `FortiosPolicyConversions` or a new conversions file to help keep things organized.)
</blockquote></details>

Yeah, it does feel a bit overloaded.

I can update it, feels wrong to put the logic in a conversion file so I left it in `FortiosConfigurationBuilder` for now
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 199 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSAqNpASHBvr4mu35s:-MWV1X0MEDyHxOwtDSCu:b-3hfaa8) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L199)):*

What about just making `matchServiceTraceElement` take in a `ServiceGroupMember` instead (and then generate slightly different trace elements depending on the input's type)? I expect we'll want a "Matched service group foo" trace element anyway.
</blockquote></details>

Yeah I can do something like that, but I was going to revamp this with the separate conversion PR (since this can't be reached yet)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java, line 10 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSBJfTE1c5Jfywn1tL:-MWV1pkT1aTvIfH7yLBq:b-67s5nv) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java#L10)):*

Is there a reason this has to be an interface rather than an abstract class? If it were an abstract class, it could own `_comment` and implement the getter/setter itself.
</blockquote></details>

Just less flexibility in the future if we want to add other abstract classes and there is no default implementation for `toHeaderSpaces` which makes more sense as an interface function.

But I can convert it.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_group_warnings, line 23 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSFSgj5krj5SzVDTGK:-MWV3eF29FzrnCzQj5zQ:b-7tfmjp) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_group_warnings#L23)):*

could you also add `append member "valid"` to make sure it recognizes that cycle?
</blockquote></details>

added

corinaminer (commented):
> Reviewed 7 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768)* status: all files reviewed, 3 unresolved discussions (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 793 at r3](https://reviewable.io/reviews/batfish/batfish/6768#-MWV_urI8HFvSY1IciTL:-MWV_urJ-qjSKjq3IJae:b-qa27g6) ([raw file](https://github.com/batfish/batfish/blob/c2894ca7f2d4b05cd4352d497f31d1aea109892a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L793)):*
> ```Java
>         .ifPresent(
>             newMembers -> {
>               Set<BatfishUUID> members = _currentServiceGroup.getMemberUUIDs();
> ```

looks like this line could move to where `members.clear()` happens. Doesn't matter much, i'm just paranoid about future editors modifying `members` before the cycle check
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1165 at r3](https://reviewable.io/reviews/batfish/batfish/6768#-MWVb-p41pY6RHTND-SP:-MWVb-p41pY6RHTND-SQ:b-d97fmh) ([raw file](https://github.com/batfish/batfish/blob/c2894ca7f2d4b05cd4352d497f31d1aea109892a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1165)):*
> ```Java
>    */
>   @Nullable
>   private ServiceGroup getParentServiceGroup(
> ```

this and its helper can be static
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java, line 199 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSAqNpASHBvr4mu35s:-MWVYd99CqMdFaYnjFLH:b32w1t) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosConfiguration.java#L199)):*

Yeah I can do something like that, but I was going to revamp this with the separate conversion PR (since this can't be reached yet)
</blockquote></details>

Ok
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java, line 10 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSBJfTE1c5Jfywn1tL:-MWVYw_eBp7V6-rRlq4u:bb0aswf) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java#L10)):*

Just less flexibility in the future if we want to add other abstract classes and there is no default implementation for `toHeaderSpaces` which makes more sense as an interface function.

But I can convert it.
</blockquote></details>

You can leave `toHeaderSpaces` abstract -- i do think that one makes more sense to require each child class to implement separately. (Abstract functions in abstract classes are basically the same as all functions in interfaces)

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 793 at r3](https://reviewable.io/reviews/batfish/batfish/6768#-MWV_urI8HFvSY1IciTL:-MWVkG-_4YanGHcgmNm2:bppomzx) ([raw file](https://github.com/batfish/batfish/blob/c2894ca7f2d4b05cd4352d497f31d1aea109892a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L793)):*

looks like this line could move to where `members.clear()` happens. Doesn't matter much, i'm just paranoid about future editors modifying `members` before the cycle check
</blockquote></details>

sure, updated
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1165 at r3](https://reviewable.io/reviews/batfish/batfish/6768#-MWVb-p41pY6RHTND-SP:-MWVjwFi5WZo6avBOaQu:bbxioaw) ([raw file](https://github.com/batfish/batfish/blob/c2894ca7f2d4b05cd4352d497f31d1aea109892a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1165)):*

this and its helper can be static
</blockquote></details>

updated; the helper is already `static`
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java, line 10 at r2](https://reviewable.io/reviews/batfish/batfish/6768#-MWSBJfTE1c5Jfywn1tL:-MWVk2D-7VN7uythrMlt:b4znczy) ([raw file](https://github.com/batfish/batfish/blob/3fcc627c757accec711673d300afac8933a8c095/projects/batfish/src/main/java/org/batfish/representation/fortios/ServiceGroupMember.java#L10)):*

You can leave `toHeaderSpaces` abstract -- i do think that one makes more sense to require each child class to implement separately. (Abstract functions in abstract classes are basically the same as all functions in interfaces)
</blockquote></details>

updated :thumbsup:

corinaminer (commented):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6768#-:-MWVqH-a4f3PfwQGGQmS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6772, opened 2021-03-23
Arista: rewrite VRF with modes

dhalperi:
* Add M_Vrf, and use it everywhere a vrf name can appear
* Handle both old and new EOS variants
* Audit uses of vrf / vrf-also / use-vrf for Arista (vs generic Cisco)
  support and delete lots of invalid ones.

Deleted include:
* s_callhome entirely
* s_line entirely
* s_router_static entirely
* ospf area block statements


arifogel (commented):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6772)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6809, opened 2021-03-30
FortiOS: Convert bgp

corinaminer:
For now, the conversion process is fairly limited. It:
- Requires BGP router-id to be set (otherwise no BGP will convert)
- Ignores filtering (route-maps) and redistribution
- Will attempt to infer neighbors' update sources if not explicitly configured by finding an interface whose address includes the neighbor ID

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/RoutingProtocol.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/RoutingProtocol.java
@@ -129,6 +129,7 @@ public enum RoutingProtocol {
           case CUMULUS_CONCATENATED:
           case CUMULUS_NCLU:
           case FORCE10:
+          case FORTIOS:
           case FOUNDRY:
           case F5:
           case F5_BIGIP_STRUCTURED:
@@ -257,6 +258,7 @@ public enum RoutingProtocol {
           case CUMULUS_CONCATENATED:
           case CUMULUS_NCLU:
           case FORCE10:
+          case FORTIOS:
           case FOUNDRY:
           case F5:
           case F5_BIGIP_STRUCTURED:
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/FortiosLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/FortiosLexer.g4
@@ -144,6 +144,7 @@ UDP_PORTRANGE: 'udp-portrange';
 UNSELECT: 'unselect';
 UNSET: 'unset';
 UP: 'up';
+UPDATE_SOURCE: 'update-source' -> pushMode(M_Str);
 UTM: 'utm';
 UUID: 'uuid' -> pushMode(M_Str);
 VDOM: 'vdom' -> pushMode(M_Str);
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_bgp.g4
@@ -44,7 +44,7 @@ crbcne
 
 crbcne_set_remote_as: REMOTE_AS bgp_remote_as newline;
 
-crbcne_set_update_source: UPDATE_SOURCE interface_name;
+crbcne_set_update_source: UPDATE_SOURCE interface_name newline;
 
 crbc_redistribute: REDISTRIBUTE bgp_redist_protocol newline crbcr*;
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_bgp.g4
@@ -36,12 +36,17 @@ crbcn_edit: EDIT bgp_neighbor_id newline crbcne* NEXT newline;
 
 crbcne
 :
-    SET crbcne_set_remote_as
+    SET (
+        crbcne_set_remote_as
+        | crbcne_set_update_source
+    )
 ;
 
 crbcne_set_remote_as: REMOTE_AS bgp_remote_as newline;
 
-crbc_redistribute: REDISTRIBUTE protocol = str newline crbcr*;
+crbcne_set_update_source: UPDATE_SOURCE interface_name;
+
+crbc_redistribute: REDISTRIBUTE bgp_redist_protocol newline crbcr*;
 
 crbcr
 :
@@ -58,3 +63,5 @@ bgp_remote_as: str;
 
 // An IP (but not using ip_address rule because EDIT pushes str mode)
 bgp_neighbor_id: str;
+
+bgp_redist_protocol: str;
--- a/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java
@@ -1,6 +1,7 @@
 package org.batfish.grammar.fortios;
 
 import static org.batfish.grammar.fortios.FortiosLexer.UNQUOTED_WORD_CHARS;
+import static org.batfish.representation.fortios.FortiosStructureUsage.BGP_UPDATE_SOURCE_INTERFACE;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
@@ -776,6 +777,12 @@ public final class FortiosConfigurationBuilder extends FortiosParserBaseListener
   @Override
   public void exitCrbcne_set_remote_as(Crbcne_set_remote_asContext ctx) {
     toLong(ctx, ctx.bgp_remote_as()).ifPresent(_currentBgpNeighbor::setRemoteAs);
+  }
+
+  @Override
+  public void exitCrbcne_set_update_source(FortiosParser.Crbcne_set_update_sourceContext ctx) {
+    toInterface(ctx, ctx.interface_name(), BGP_UPDATE_SOURCE_INTERFACE)
+        .ifPresent(_currentBgpNeighbor::setUpdateSource);
   }
 
   @Override
--- a/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpNeighbor.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpNeighbor.java
@@ -19,10 +19,19 @@ public class BgpNeighbor implements Serializable {
     return _remoteAs;
   }
 
+  public @Nullable String getUpdateSource() {
+    return _updateSource;
+  }
+
   public void setRemoteAs(Long remoteAs) {
     _remoteAs = remoteAs;
   }
 
+  public void setUpdateSource(String updateSource) {
+    _updateSource = updateSource;
+  }
+
   private final @Nonnull Ip _ip;
   private @Nullable Long _remoteAs;
+  private @Nullable String _updateSource;
 }
--- a/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java
@@ -1,5 +1,7 @@
 package org.batfish.representation.fortios;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
+
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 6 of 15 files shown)


arifogel (commented):
> Reviewed 5 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings, line 25 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3cz6f8sblHwuUw4N_:-MX3cz6f8sblHwuUw4Na:bo16y07) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings#L25)):*
> ```
>         edit "3.3.3.3"
>             # Inactive update-source
>             set remote-as 4294967295
> ```

Note that max uint32 and max uint16 are really not supposed to be used as ASNs either.
They may not appear in paths on the internet. They _should_ not be used internally, but some vendors may allow them to be used.
I suspect fortigate may proscribe its use. There should definitely at minimum be a warning if such a value is used.

https://tools.ietf.org/html/rfc7300

arifogel (commented):
> Reviewed 3 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3dmf81_KcgiINbIaT:-MX3dmf81_KcgiINbIaU:b-7ffis1) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L30)):*
> ```Java
>       Interface viUpdateSourceIface = c.getAllInterfaces().get(updateSource);
>       if (viUpdateSourceIface == null) {
>         // Conversion issue with the update-source interface. Better to ignore the neighbor than try
> ```

Can you really get here? Is an assertion that `viUpdateASourceIface != null` insufficient?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: 8 of 15 files reviewed, 3 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 42 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3ajvrEU-RKN5Eg_-9:-MX8dtNH8W0kD0CysyXF:b-i60pa0) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L42)):*

`0` is not a valid ASN:
https://tools.ietf.org/html/rfc7607

I assume you cannot spin up a process with this? Probably better to keep it null by default.
</blockquote></details>

interesting. You can explicitly `set as 0`, and `show full-configuration` will include `set as 0` if none has been set. Haven't set up an actual peering that I could test; presumably it would fail with AS 0, based on that config.

Since `set as 0` is valid and is default, i'd prefer to leave the VS model as-is, but modified conversion to skip BGP if the process AS is 0.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3dmf81_KcgiINbIaT:-MX8ghgv3LdzylvMIR2d:bqr9ezc) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L30)):*

Can you really get here? Is an assertion that `viUpdateASourceIface != null` insufficient?
</blockquote></details>

So far, you can't get here, but it might be possible in the future. Do you think it's better to assert the VI interface exists? My thinking was, if this does become possible:
1. end users would rather have the peer be ignored than have the snapshot fail (presumably there will be conversion warnings about the interface to help them diagnose)
2. it's unlikely that our tests would catch this case
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings, line 25 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3cz6f8sblHwuUw4N_:-MX8lbAQEseSaD-isdMk:b-yqpx67) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings#L25)):*

Note that max uint32 and max uint16 are really not supposed to be used as ASNs either.
They may not appear in paths on the internet. They _should_ not be used internally, but some vendors may allow them to be used.
I suspect fortigate may proscribe its use. There should definitely at minimum be a warning if such a value is used.

https://tools.ietf.org/html/rfc7300
</blockquote></details>

ack. [Fortinet docs do say they "should not" be used](https://docs.fortinet.com/document/fortiadc/5.4.1/handbook/569661/bgp), although I don't know if the devices will actually ignore BGP processes with those ASNs configured.

Blocked conversion for processes with those ASNs, but not neighbors with those remote ASNs. Those neighbors won't be able to establish successful peerings anyway since the remote's BGP process won't be converted, and having half-open BGP peerings might be helpful for users to diagnose their problem.

Generally I think this is a rare enough edge case that it's not a big problem if we don't get the behavior right. we can always improve it if someone actually runs into problems with it, but i think that's very unlikely.

arifogel (commented):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 42 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3ajvrEU-RKN5Eg_-9:-MX8peP_2dBHraAoje4V:b-ga4ndj) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L42)):*
> Since set as 0 is valid and is default, i'd prefer to leave the VS model as-is, but modified conversion to skip BGP if the process AS is 0.

Sounds good.

> Haven't set up an actual peering that I could test; presumably it would fail with AS 0, based on that config.

Unlike uint16/32max, 0 is just totally invalid. I don't think you need to bother testing an actual peering with 0.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3dmf81_KcgiINbIaT:-MX8o1ipDYC0L1-0ytcr:be7sdl0) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L30)):*
> Do you think it's better to assert the VI interface exists? 

Sure, that's fine. But that's still an assertion error.

I think in general it is not acceptable to write unreachable error-handling code. Assertions are an exception, as they are really just a sanity check (and help with null analysis).

> end users would rather have the peer be ignored than have the snapshot fail (presumably there will be conversion warnings about the interface to help them diagnose)

The proper way to handle this is to not propagate a neighbor that could never have been configured in the first place on the CLI due to a reference to a non-existent interface. Let me know if I am missing a pathological case here that the CLI would not catch.

> it's unlikely that our tests would catch this case

Catch what case? I defer to my comment above about not writing unreachable code. If you _can't_ test it, don't write it.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings, line 25 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3cz6f8sblHwuUw4N_:-MX8no5uB2_WihDiAj4s:b-32w0iv) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/bgp_conversion_warnings#L25)):*

ack. [Fortinet docs do say they "should not" be used](https://docs.fortinet.com/document/fortiadc/5.4.1/handbook/569661/bgp), although I don't know if the devices will actually ignore BGP processes with those ASNs configured.

Blocked conversion for processes with those ASNs, but not neighbors with those remote ASNs. Those neighbors won't be able to establish successful peerings anyway since the remote's BGP process won't be converted, and having half-open BGP peerings might be helpful for users to diagnose their problem.

Generally I think this is a rare enough edge case that it's not a big problem if we don't get the behavior right. we can always improve it if someone actually runs into problems with it, but i think that's very unlikely.
</blockquote></details>

Sounds reasonable.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 42 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3ajvrEU-RKN5Eg_-9:-MX8q94P5UzuJLnjz0Ad:b-r5jskr) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L42)):*

> Since set as 0 is valid and is default, i'd prefer to leave the VS model as-is, but modified conversion to skip BGP if the process AS is 0.

Sounds good.

> Haven't set up an actual peering that I could test; presumably it would fail with AS 0, based on that config.

Unlike uint16/32max, 0 is just totally invalid. I don't think you need to bother testing an actual peering with 0.
</blockquote></details>

..it's possible it may try to infer local AS under some weird circumstances if you leave it at 0. But you can defer that.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 42 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3ajvrEU-RKN5Eg_-9:-MX8v_xECD8oRJ6ZAmvx:b-hdldis) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L42)):*

..it's possible it may try to infer local AS under some weird circumstances if you leave it at 0. But you can defer that.
</blockquote></details>

ack. deferring
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/6809#-MX3dmf81_KcgiINbIaT:-MX8wBMN4sOdt3iabexe:b5pgax7) ([raw file](https://github.com/batfish/batfish/blob/6afc8ccd4d195b76fbe6e6826d0156045d942e74/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L30)):*

> Do you think it's better to assert the VI interface exists? 

Sure, that's fine. But that's still an assertion error.

I think in general it is not acceptable to write unreachable error-handling code. Assertions are an exception, as they are really just a sanity check (and help with null analysis).

> end users would rather have the peer be ignored than have the snapshot fail (presumably there will be conversion warnings about the interface to help them diagnose)

The proper way to handle this is to not propagate a neighbor that could never have been configured in the first place on the CLI due to a reference to a non-existent interface. Let me know if I am missing a pathological case here that the CLI would not catch.

> it's unlikely that our tests would catch this case

Catch what case? I defer to my comment above about not writing unreachable code. If you _can't_ test it, don't write it.

</blockquote></details>

Oh wait, I realized this is already reachable if the update-source is an interface whose type we don't support. Config will be perfectly valid, but VI interface won't exist. Added test.

arifogel (commented):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6809)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

PR #6812, opened 2021-03-30
TransferBDD: Avoid crashing on common idioms for setting communities

millstein:
Identifies expressions of the form (InputCommunities - E), which is common in the translation of config commands to set communities, and ignores them rather than crashing.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6812)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6842, opened 2021-04-04
IOS-XR: remove u32 alternative for community-set

dhalperi:
> Integer community values must be split in half and expressed as two unsigned
> decimal integers in the range from 0 to 65535, separated by a colon. Single
> 32-bit community values are not allowed.

https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter7.html#con_1115490

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4
@@ -40,7 +40,6 @@ community
    | NO_ADVERTISE
    | NO_EXPORT
    | hi = uint16 COLON lo = uint16
-   | u32 = uint32
 ;
 
 description_line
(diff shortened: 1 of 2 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6842)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6842#-MXSk0ogDvuwUA2XM8uG:-MXSk0ohBxzRRgNRVTsO:b90nfvu) (no related file):*
Note that there is one case where you do not need two colon-separated expressions:
You can use wildcards as:
`*` (all 32 bits)
OR
`*:*` (one `*` for each 16 bits)
However, things like parameters (and everything else that isn't a wildcard) can only stand in for 16 bits at a time.

---

PR #6864, opened 2021-04-07
FortiOS: Anticipate unset lines in route-maps, services

corinaminer:
Does not extract unset, but saves the parser from losing context.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
@@ -20,7 +20,9 @@ crrmec_rule: RULE newline crrmecr_edit* END newline;
 
 crrmecr_edit: EDIT route_map_rule_number newline crrmecre* NEXT newline;
 
-crrmecre: crrmecre_set;
+crrmecre: crrmecre_set | crrme_unset;
+
+crrme_unset: UNSET unimplemented;
 
 crrmecre_set: SET (crrmecre_set_action | crrmecre_set_match_ip_address);
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
@@ -20,7 +20,11 @@ crrmec_rule: RULE newline crrmecr_edit* END newline;
 
 crrmecr_edit: EDIT route_map_rule_number newline crrmecre* NEXT newline;
 
-crrmecre: crrmecre_set | unimplemented_edit_stanza;
+crrmecre
+:
+    crrmecre_set
+    | (UNSET | SELECT | UNSELECT | APPEND | CLEAR) unimplemented
+;
 
 crrmecre_set: SET (crrmecre_set_action | crrmecre_set_match_ip_address);
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_route_map.g4
@@ -20,9 +20,7 @@ crrmec_rule: RULE newline crrmecr_edit* END newline;
 
 crrmecr_edit: EDIT route_map_rule_number newline crrmecre* NEXT newline;
 
-crrmecre: crrmecre_set | crrme_unset;
-
-crrme_unset: UNSET unimplemented;
+crrmecre: crrmecre_set | unimplemented_edit_stanza;
 
 crrmecre_set: SET (crrmecre_set_action | crrmecre_set_match_ip_address);
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4
@@ -18,7 +18,7 @@ cfsc_edit: EDIT service_name newline cfsce* NEXT newline;
 cfsce
 :
     SET cfsc_set_singletons
-    | unimplemented_edit_stanza
+    | (UNSET | SELECT | UNSELECT | APPEND | CLEAR) unimplemented
 ;
 
 cfsc_set_singletons:
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4
@@ -15,7 +15,13 @@ cfsc_rename: RENAME current_name = service_name TO new_name = service_name newli
 
 cfsc_edit: EDIT service_name newline cfsce* NEXT newline;
 
-cfsce: SET cfsc_set_singletons;
+cfsce
+:
+    SET cfsc_set_singletons
+    | cfsc_unset
+;
+
+cfsc_unset: UNSET unimplemented;
 
 cfsc_set_singletons:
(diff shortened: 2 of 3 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/6864#-MXdiPVuC7itxA6gJDGD:-MXdiPVv82RqMp8LFHqp:b-nncl7x) ([raw file](https://github.com/batfish/batfish/blob/a6b4fbed64a1caf5d108be27f4abbb484c019510/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4#L21)):*
> ```ANTLR
> :
>     SET cfsc_set_singletons
>     | cfsc_unset
> ```

Should you be using the same pattern here and on the `SET` line above? Can `cfsc_unset` be inlined as `UNSET implemented`? Does `SET cfsc_set_singletons` suffer from lost parser context when you `set` something unrecognized?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/6864#-MXdiPVuC7itxA6gJDGD:-MXe-UMTD3xzbkUDmGHc:b-wkiv2e) ([raw file](https://github.com/batfish/batfish/blob/a6b4fbed64a1caf5d108be27f4abbb484c019510/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4#L21)):*

Should you be using the same pattern here and on the `SET` line above? Can `cfsc_unset` be inlined as `UNSET implemented`? Does `SET cfsc_set_singletons` suffer from lost parser context when you `set` something unrecognized?
</blockquote></details>

I wanted to keep the `unset` pattern consistent between services and routemaps. Unfortunately the pattern for the `SET` lines is different between those files. The pattern in the route-map file is the one we currently prefer (though I've forgotten why; @sfraint made that decision with dan a while back). I checked and this pattern does handle `set asdfghjkjhgfdsdfghjkjhg` without losing context.

I do think inlining as `UNSET unimplemented` would also work, in both cases. is that preferable?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_custom, line 42 at r2](https://reviewable.io/reviews/batfish/batfish/6864#-MXe1WDK9MyC0HrAcvjs:-MXe1WDK9MyC0HrAcvjt:b-1098l0) ([raw file](https://github.com/batfish/batfish/blob/a207cecc90fd19c6dece78c043742eeedbe0bf7d/projects/batfish/src/test/resources/org/batfish/grammar/fortios/testconfigs/service_custom#L42)):*
> ```
>         set protocol ICMP
>         set icmptype 254
>         set icmpcode 254
> ```

noticed when testing unrecognized `set` lines that this file generated a warning because `icmpcode` can't be set before `icmptype`. Neither of these values actually show up in the extracted vendor config because the service's protocol changes (see next line), but the test's intent was for both to be set when it changed

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/6864#-MXdiPVuC7itxA6gJDGD:-MXhTaib4zXwrJD7TGXw:b-eyjwat) ([raw file](https://github.com/batfish/batfish/blob/a6b4fbed64a1caf5d108be27f4abbb484c019510/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4#L21)):*

I wanted to keep the `unset` pattern consistent between services and routemaps. Unfortunately the pattern for the `SET` lines is different between those files. The pattern in the route-map file is the one we currently prefer (though I've forgotten why; @sfraint made that decision with dan a while back). I checked and this pattern does handle `set asdfghjkjhgfdsdfghjkjhg` without losing context.

I do think inlining as `UNSET unimplemented` would also work, in both cases. is that preferable?
</blockquote></details>

re: pattern revision: the current pattern is inconsistent and a little confusing where sometimes rules like `<prefix>_<keyword>` means the keyword is in the rule and sometimes it means the keyword is handled just before the rule.

e.g.
```
cfp_append
:
    cfp_append_dstaddr
    | cfp_append_dstintf
    | cfp_append_service
    | cfp_append_srcaddr
    | cfp_append_srcintf
;
```
vs
```
cfp_edit: EDIT policy_number newline cfpe* NEXT newline;
```

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel and @corinaminer)

arifogel (commented):
> Reviewed 1 of 1 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4, line 21 at r1](https://reviewable.io/reviews/batfish/batfish/6864#-MXdiPVuC7itxA6gJDGD:-MXi7SLDBkuma06Z6QBb:b477ujo) ([raw file](https://github.com/batfish/batfish/blob/a6b4fbed64a1caf5d108be27f4abbb484c019510/projects/batfish/src/main/antlr4/org/batfish/grammar/fortios/Fortios_service.g4#L21)):*

re: pattern revision: the current pattern is inconsistent and a little confusing where sometimes rules like `<prefix>_<keyword>` means the keyword is in the rule and sometimes it means the keyword is handled just before the rule.

e.g.
```
cfp_append
:
    cfp_append_dstaddr
    | cfp_append_dstintf
    | cfp_append_service
    | cfp_append_srcaddr
    | cfp_append_srcintf
;
```
vs
```
cfp_edit: EDIT policy_number newline cfpe* NEXT newline;
```

</blockquote></details>

Some offline discussion has surfaced these conclusions:
1. Eventually, every edit block rule should contain at least one option for each keyword (SET/UNSET/SELECT/UNSELECT/APPEND/CLEAR) that can be used in it.
2. In a given edit block rule, if there are any implemented options for a given keyword, there should be no unimplemented option for that keyword.
3. In general, two sibling rules should never start with the same token; this triggers adaptive prediction, which is slow and unpredictable. (This would have happened with `SET cfsc_set_singletons | unimplemented_edit_stanza`, since the unimplemented rule includes `SET unimplemented`.)

arifogel (commented):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6864)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6879, opened 2021-04-08
FortiOS: Parse and extract BGP network config

corinaminer:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6879)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel, @corinaminer, and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 981 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnFW4Y23QlI7KbNCTO:-MXnFW4Y23QlI7KbNCTP:b-s33o4u) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L981)):*

> ```
>     long candidate = existing.size() + 1;
>     while (existing.contains(candidate)) {
>       candidate++;
>     }
>     return permitted.contains(candidate) ? Optional.of(candidate) : Optional.empty();
> ```
</details>


Would it be simpler to do something like `permitted.intersect(greaterthan(number of current)).diff(space of existing).lowest value`?  (the API isn't right, but I believe all these primitives exist/work.)
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1003 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnH-xk0_IO6O3_8WOq:-MXnH-xk0_IO6O3_8WOr:b-wrmi6a) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1003)):*
> ```Java
>     return null;
>   }
> ```

Do we have to worry about different numbers with the same prefix?
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnG85d9cN4eCuNaM-j:-MXnG85e6LnO6WZ7ruPt:b1rrlko) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L47)):*
> ```
> String
> ```
If we coerce these to Long, can we just use `Long` here? Otherwise, the code in extraction looks crashy: using `Long#parseLong` without guarding it.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6879)* status: 5 of 8 files reviewed, 2 unresolved discussions (waiting on @arifogel, @dhalperi, and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 981 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnFW4Y23QlI7KbNCTO:-MXnJQ7A9GIZsp9GRzBd:b68bqm2) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L981)):*

> ```
>     long candidate = existing.size() + 1;
>     while (existing.contains(candidate)) {
>       candidate++;
>     }
>     return permitted.contains(candidate) ? Optional.of(candidate) : Optional.empty();
> ```

Would it be simpler to do something like `permitted.intersect(greaterthan(number of current)).diff(space of existing).lowest value`?  (the API isn't right, but I believe all these primitives exist/work.)
</blockquote></details>

It's possible, but not quite that simple, because infinite spaces like `greaterthan(number of current)` aren't supported (also, `least()` throws for empty spaces, which makes things uglier). As a first pass, I came up with this attempt:
```
LongSpace existingSpace = LongSpace.builder().includingAll(existing).build();
LongSpace.Builder candidatesBuilder =
  LongSpace.builder().including(permitted).excluding(existingSpace);
assert !permitted.isEmpty();
long lowestPermitted = permitted.least();
if (lowestPermitted <= existing.size()) {
  candidatesBuilder.excluding(Range.closed(lowestPermitted, (long) existing.size()));
}
LongSpace candidates = candidatesBuilder.build();
return candidates.isEmpty() ? Optional.empty() : Optional.of(candidates.least());
```
Is there a reason this type of approach is better? In practice, it seems very unlikely that the while loop would ever iterate more than a few times.
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 1003 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnH-xk0_IO6O3_8WOq:-MXnSCVOAWRkIjc_Zt3r:b-4i6fey) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L1003)):*

Do we have to worry about different numbers with the same prefix?
</blockquote></details>

yes. done
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnG85d9cN4eCuNaM-j:-MXnSOCp0A4n7wtl2EtW:b-81ijdl) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L47)):*

> ```
> String
> ```
If we coerce these to Long, can we just use `Long` here? Otherwise, the code in extraction looks crashy: using `Long#parseLong` without guarding it.
</blockquote></details>

We've been using strings in this type of scenario because:
- when we enter a BGP network edit block, we need to make `_currentBgpNetwork` nonnull even if the edit block's name is invalid, because inner statements will try to modify it
    - therefore we want to make `BgpNetwork._id` a string so that it can hold whatever name the edit block has
    - could make `BgpNetwork._id` a long and use -1 or something to indicate it's invalid, but then we have to ensure that inner statements' extraction methods never surface `_currentBgpNetwork._id` in warnings
- when we exit a BGP network edit block, we only know the current network's ID from `_currentBgpNetwork._id` (and whether it is valid from `_currentBgpNetworkNameValid`)
    - could `parseLong` at that point if the ID is valid -- is that any less crashy-looking than the current setup?

Tl;dr it seems within reach to make this a `Map<Long, BgpNetwork>`, but i don't see a really clean way to do it (but as you point out, the current code isn't clean either). I'm open to alternatives; what's your vote?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6879)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel, @corinaminer, and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6879#-MXnaEArBBpKVatH26ZT:-MXnaEArBBpKVatH26ZU:byqrtaj) (no related file):*
FYI sorry - was trying to take some reviewer load, not sure i helped ;) But I did learn a lot about our fortios vs structure.

---
___
*[projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java, line 981 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnFW4Y23QlI7KbNCTO:-MXnYKhC04iLc_D13hiL:bhc7ze7) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/grammar/fortios/FortiosConfigurationBuilder.java#L981)):*
>  In practice, it seems very unlikely

... risk aversion. Explicit is better than implicit. No loops is better than loops ;).

But this is looking unwieldy, fine to punt.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnG85d9cN4eCuNaM-j:-MXnZ3Rv2bxQEan803FM:b-37ki3w) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L47)):*
> therefore we want to make BgpNetwork._id a string so that it can hold whatever name the edit block has

It's not obvious to me why we want this connection? That is, are we sure we want to force the VS type String so that we can use Vs.getId() to produce an error message in extractor?

Here's an alternate approach using extractor state instead. Not sure if it's better.

```
_currentName = toString(ctx.name);
Long id = tryParseAsNetworkId(_currentName);
if (id == null) {
_currentBgpNetwork = // get the network, clone, whatever you're already doing, using id
_invalid = false;
} else {
   _currentBgpNetwork = new BgpNetwork(0); // dummy: for extraction, but will be invalid
   _invalid = true;
}
```

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6879)* status: 4 of 8 files reviewed, 1 unresolved discussion (waiting on @arifogel, @dhalperi, and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6879#-MXnaEArBBpKVatH26ZT:-MXnti2d7hRDCA0HV2hY:bts52q2) (no related file):*

FYI sorry - was trying to take some reviewer load, not sure i helped ;) But I did learn a lot about our fortios vs structure.
</blockquote></details>

no worries :)

---
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java, line 47 at r1](https://reviewable.io/reviews/batfish/batfish/6879#-MXnG85d9cN4eCuNaM-j:-MXnyjJz3ZjXnzIXlPxK:bd478eu) ([raw file](https://github.com/batfish/batfish/blob/604ab6035811e579ad704455c71ea3f744819553/projects/batfish/src/main/java/org/batfish/representation/fortios/BgpProcess.java#L47)):*

> therefore we want to make BgpNetwork._id a string so that it can hold whatever name the edit block has

It's not obvious to me why we want this connection? That is, are we sure we want to force the VS type String so that we can use Vs.getId() to produce an error message in extractor?

Here's an alternate approach using extractor state instead. Not sure if it's better.

```
_currentName = toString(ctx.name);
Long id = tryParseAsNetworkId(_currentName);
if (id == null) {
_currentBgpNetwork = // get the network, clone, whatever you're already doing, using id
_invalid = false;
} else {
   _currentBgpNetwork = new BgpNetwork(0); // dummy: for extraction, but will be invalid
   _invalid = true;
}
```
</blockquote></details>

I do think that's better. And moreover, we're not even using the current name anywhere right now. I'm just switching everything over to longs and using 0 as a special value indicating an invalid name. If we really want a warning containing the ID later, we can make a `_currentBgpNetworkName` field.

dhalperi (commented):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6879)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @sfraint)

PR #6900, opened 2021-04-12
extended communitities fix transitive check

arifogel:
- extended community is transitive when second-highest-order type bit is
UNSET, not when it is set
- see https://tools.ietf.org/html/rfc4360 Page 3


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6900)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6902, opened 2021-04-13
Trace: improve performance in sanity checks

dhalperi:
Iterables.getLast is an iteration, but since we have a list we can just get
the index.

Notice this only affects development runs or runs with `-ea` set.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6902)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6906, opened 2021-04-13
FlowTracer: intern Breadcrumbs

dhalperi:
Under ECMP, the same breadcrumb may be generated many times. TraceDAG
compressions makes the redundancy less, but interning Breadcrumbs is still
faster and uses less RAM.

We only intern at the FlowTracer level, so that the reference dies after the
traces are completed.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6906)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6907, opened 2021-04-13
DagTraceRecorder: use sets instead of lists for breadcrumbs

dhalperi:
They are never used in order, and they have containment checked often.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6907)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6911, opened 2021-04-14
checkStyle changes

adiapel:
While running checkstyle with some certain modules names, I found a few changes I would like to make in some files in the project.
the first change is in the file
projects/question/src/main/java/org/batfish/question/routes/RoutesAnswererUtil.java
the first method I changed is populateBgpRouteAttributes. before the change the npathcomplexity was 24,576 (while the max value is 200).
the second method was populateRouteAttributes. before the change the npathxomplexity was 1,024.
my change includes some variable assignments and conditions which reduces the npathcomplexity of the methods.
now, running the same checkstyle with the moudles doesn't return a npathcomplexity error.
I also run "question" tests and had now change.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpPeerConfig.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpPeerConfig.java
@@ -1,6 +1,5 @@
 package org.batfish.datamodel;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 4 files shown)


dhalperi:
This doesn't look like an improvement in readability to me. What are you trying to accomplish?

Are you using Batfish in your network? What brings you to this project?

adiapel:
Hi!
I am a computer science student at the Hebrew University of Jerusalem in Israel. 
I am participating in a research which examines whether removing smells detected by static analyzing will improve code quality in different open-code projects.
for more information about the research:
Follow Your Nose - Which Code Smells are Worth Chasing?
Regarding the specific commit that I made - in my opinion, now the code is more readable. Moreover, there were a few conditions in the original code which were checked a few times on each function call. In this code, these conditions are now checked only once on each function call.

dhalperi:
From our perspective these changes are regressions. The code is much more readable beforehand, and some of the changes even violate our style guidelines (such as static import of Guava helpers).

* See the [Guava style guidance](https://github.com/google/guava/wiki/PreconditionsExplained#preconditions) here: "We strongly recommend importing these statically". We apply this same guidance to other Guava helpers that serve similar functions, like `firstNotNull`.

* Keeping `/* optional */` with each parameter is cleaner and more maintainable than something like `/* the last two are optional */`. If we delete a parameter, change a parameter's contract, add a new parameter, reorder them, or otherwise change the code it will be easy to miss updating the comment.
 
* Keeping important information/context near where it's used is a key principle of readability. That is violated by the above and by the refactoring in RoutesAnswerer.

+ The removal of unnecessary parens in `(flow)` is an improvement, but this is not something it's worth your time to submit, our time to review, or our CI resources to test.

Finally, the most important thing when submitting a change to code is to ensure that your code is consistent with the existing project. Consistency is more important than perfection, and in general these changes are not consistent with the code in our project. You can see this principle and others referenced in many styleguides, such as [Google's](https://google.github.io/styleguide/cppguide.html#Goals): "Optimize for the reader, not the writer" and "Be consistent with the existing code -- Using one style consistently through our codebase lets us focus on other (more important) issues."

Thank you for reaching out, but I do not think you should send more changes like this.

PR #6914, opened 2021-04-16
Don't crash for unsupported IpTables actions

### LINKED ISSUE #6913: Certain IPTables Rules Cause Host To Not Show Up In Snapshot
[2021-04-15_snap_lost_host.tar.gz](https://github.com/batfish/batfish/files/6321406/2021-04-15_snap_lost_host.tar.gz)

Hello!

I encountered an interesting issue when trying to load host/iptables configs into batfish alongside network devices. This has been tested on the latest docker pull of `batfish/allinone`. This is a lab build, so no sensitive data are present. I created a lab of 5 Juniper vQFX switches in a spine/leaf, with an edge vMX router. There are three hosts, _host1_, _host2_, _host3_ that I've added in the `hosts` directory.

When populating iptables files, I just run `iptables-save` on the hosts, based on the documentation [here](https://pybatfish.readthedocs.io/en/latest/notebooks/interacting.html#Packaging-snapshot-data).

What I'm noticing is that certain iptables rules will trigger that particular host to not be loaded when taking a snapshot. **Host1** is the test. I added some NAT rules, and in particular `POSTROUTING` and `DOCKER`. When these are evaluated, host 1 does not get registered in the snapshot. I recreate doing the following:

```
bf_session.host = 'localhost'
bf_set_network('prod')
bf_init_snapshot('snaphost/prod/', name='test', overwrite=True)

bfq.nodeProperties().answer()
```

With the iptables NAT rules un-commented, host1 won't show up. When I comment out these rules, host1 _will_ show up. It seems to happen with any rules in the `filter` table other than `INPUT` as well.

ratulm:
Fixes #6913

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesRule.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/iptables/IptablesRule.java
@@ -51,13 +51,18 @@ public class IptablesRule implements Serializable {
     return _actionType;
   }
 
+  /**
+   * Coverts IpTablesAction to a LineAction for the ACCEPT and DROP cases.
+   *
+   * @throws IllegalArgumentException if the action is not one of those two..
+   */
   public LineAction getIpAccessListLineAction() {
     if (_actionType == IptablesActionType.ACCEPT) {
       return LineAction.PERMIT;
     } else if (_actionType == IptablesActionType.DROP) {
       return LineAction.DENY;
     } else {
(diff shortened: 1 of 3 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6914)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6919, opened 2021-04-16
IOS-XR parse and ignore top-level constructs

arifogel:
- cdp
- ipv4 virtual
- nsr
- tcp
- vty-pool


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6919)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4, line 119 at r1](https://reviewable.io/reviews/batfish/batfish/6919#-MYRXok19ah8aSNLni1m:-MYRXok19ah8aSNLni1n:b-bb0euv) ([raw file](https://github.com/batfish/batfish/blob/606292d7b2190ec8ce0fec13819194f15f421ca3/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4#L119)):*
> ```ANTLR
> s_nsr: NSR PROCESS_FAILURES SWITCHOVER NEWLINE;
> 
> s_null
> ```

In general, I don't think `s_null` is safe. E.g., for something like `cdp` or `tcp` it may well cause us to pop out of inner contexts. (I see you had to patch `if` to avoid this here.)

My rule that has worked in other grammars is to always go two or more tokens deep. E.g., on NX-OS we had trouble until we went `ip pim <all fleshed out> null_rest_of_line`.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6919)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4, line 119 at r1](https://reviewable.io/reviews/batfish/batfish/6919#-MYRXok19ah8aSNLni1m:-MYRYN-93z-t6xUf5dor:b-k2p765) ([raw file](https://github.com/batfish/batfish/blob/606292d7b2190ec8ce0fec13819194f15f421ca3/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4#L119)):*

In general, I don't think `s_null` is safe. E.g., for something like `cdp` or `tcp` it may well cause us to pop out of inner contexts. (I see you had to patch `if` to avoid this here.)

My rule that has worked in other grammars is to always go two or more tokens deep. E.g., on NX-OS we had trouble until we went `ip pim <all fleshed out> null_rest_of_line`.
</blockquote></details>

The problem is that in both interface and top-level, `cdp` by itself is valid.
Additionally in top-level, cdp followed by some other stuff is valid. But since it's valid by itself, it's not clear to me what the benefit is of adding that detail at the top-level grammar.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6919)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4, line 119 at r1](https://reviewable.io/reviews/batfish/batfish/6919#-MYRXok19ah8aSNLni1m:-MYRcppb86SHQiRP6Mj5:b-oi5spe) ([raw file](https://github.com/batfish/batfish/blob/606292d7b2190ec8ce0fec13819194f15f421ca3/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrParser.g4#L119)):*

The problem is that in both interface and top-level, `cdp` by itself is valid.
Additionally in top-level, cdp followed by some other stuff is valid. But since it's valid by itself, it's not clear to me what the benefit is of adding that detail at the top-level grammar.
</blockquote></details>

pulled out rules for most of these

dhalperi (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6919)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6922, opened 2021-04-16
IOS-XR ospf parsing

arifogel:
- parse and warn/ignore various OSPF constructs


projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_ospf.g4
```diff
@@ -17,14 +17,17 @@ ro_area
    (
       area_int = uint_legacy
       | area_ip = IP_ADDRESS
```
dhalperi, line 19: FYI we probably need a rule for this soonish.

dhalperi: (Not in this pr unless you really want to)

dhalperi (approved):

PR #6935, opened 2021-04-20
ParseTreePrettyPrinter: indicate in output which rules are not overridden

dhalperi:
Especially in older grammars, it can be hard to find syntax that has been
added in ANTLR4, but not actually implemented in an extractor. Obvious
candidates are those in `_null` rules, but there may be more.

This is a simple change to annotate in parse trees rules that are not
overridden. When run against a config of interest, manual inspection of such
can reveal gaps.

This is by no means a catch-all: plenty of parser rules exist for grouping, for
recovery, for clarity, etc. But this enables informed manual analysis.


tests/basic/routes-diff.ref
```diff
@@ -132,82 +132,11 @@
       ],
       "textDesc" : "Display diff of RIB routes"
     },
-    "rows" : [
-      {
-        "Node" : {
-          "id" : "node-as1border2",
```
dhalperi, line 138: I don't get this.

arifogel (commented):
> Reviewed 15 of 16 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6935)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6943, opened 2021-04-21
Create neighbor even if local IP is null

ratulm:
Mirror what is happening with other vendors (e.g., Arista). If local-ip/update-source is null, warn and create the VI neighbor anyway. That way, this neighbor will show up BgpPeer set of questions.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
@@ -236,9 +236,8 @@ public final class FortiosBgpConversions {
       if (localIp == null) {
         w.redFlag(
             String.format(
-                "Ignoring BGP neighbor %s: Update-source %s has no address",
-                remoteIp, updateSource.getName()));
-        continue;
+                "BGP neighbor %s in vrf %s: Update-source %s has no address",
+                remoteIp, vrf, updateSource.getName()));
       }
       BgpActivePeerConfig.builder()
           .setLocalIp(localIp)
--- a/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
@@ -231,19 +231,17 @@ public final class FortiosBgpConversions {
     for (Ip remoteIp : neighborIdsInVrf) {
       BgpNeighbor neighbor = bgpProcess.getNeighbors().get(remoteIp);
       @Nullable Interface updateSource = updateSources.get(remoteIp);
-      Ip localIp =
-          updateSource == null || updateSource.getConcreteAddress() == null
-              ? null
-              : updateSource.getConcreteAddress().getIp();
-      if (localIp == null) {
-        String warning =
-            updateSource == null
-                ? String.format(
-                    "BGP neighbor %s: Unable to infer its update source", neighbor.getIp())
-                : String.format(
-                    "BGP neighbor %s in vrf %s: Update-source %s has no address",
-                    remoteIp, vrf, updateSource.getName());
-        w.redFlag(warning);
+      @Nullable Ip localIp = null;
+      if (updateSource == null) {
+        w.redFlag(
+            String.format("BGP neighbor %s: Unable to infer its update source", neighbor.getIp()));
+      } else if (updateSource.getConcreteAddress() == null) {
+        w.redFlag(
+            String.format(
+                "BGP neighbor %s in vrf %s: Update-source %s has no address",
+                remoteIp, vrf, updateSource.getName()));
+      } else {
+        localIp = updateSource.getConcreteAddress().getIp();
       }
       BgpActivePeerConfig.builder()
           .setLocalIp(localIp)
--- a/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java
@@ -1,6 +1,9 @@
 package org.batfish.representation.fortios;
 
 import static org.batfish.datamodel.Names.generatedBgpPeerExportPolicyName;
+import static org.batfish.representation.fortios.FortiosConfiguration.computeVrfName;
+import static org.batfish.representation.fortios.Interface.DEFAULT_VDOM;
+import static org.batfish.representation.fortios.Interface.DEFAULT_VRF;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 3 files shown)


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6943)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 231 at r1](https://reviewable.io/reviews/batfish/batfish/6943#-MYpMJxRCXXvs4kMK4Zb:-MYpMJxRCXXvs4kMK4Zc:b-89thpk) ([raw file](https://github.com/batfish/batfish/blob/69450ae878492972e0fb325e78c2723c62b5dba3/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L231)):*
> ```Java
>     for (Ip remoteIp : neighborIdsInVrf) {
>       BgpNeighbor neighbor = bgpProcess.getNeighbors().get(remoteIp);
>       Interface updateSource = updateSources.get(remoteIp);
> ```

PS: not related to the PR but a curiosity, can `updateSource` be null here (causing NPE on the next line).

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6943)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 231 at r1](https://reviewable.io/reviews/batfish/batfish/6943#-MYpMJxRCXXvs4kMK4Zb:-MYpVB4a8rFrgxdQO443:bxtjdme) ([raw file](https://github.com/batfish/batfish/blob/69450ae878492972e0fb325e78c2723c62b5dba3/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L231)):*

PS: not related to the PR but a curiosity, can `updateSource` be null here (causing NPE on the next line).
</blockquote></details>

No, by design both `neighborIdsInVrf` and `updateSources` only contain info about peers whose update source was successfully identified, so all the IPs in `neighborIdsInVrf` are guaranteed to be keys in `updateSources`. that's guaranteed in the caller, though; ideally i probably should've documented that assumption in this function's javadoc

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6943)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 247 at r2](https://reviewable.io/reviews/batfish/batfish/6943#-MYqd1Sf8-QB2yQwV5EX:-MYqd1Sf8-QB2yQwV5EY:byhb52f) ([raw file](https://github.com/batfish/batfish/blob/47224dffcd73d17669b29d1cbb9b7b28c0b48ec9/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L247)):*
> ```Java
>                     remoteIp, vrf, updateSource.getName());
>         w.redFlag(warning);
>       }
> ```

Nonblocking, but i think this could be expressed more cleanly as:
```
Ip localIp = null;
if (updateSource == null) {
    // file first warning
} else if (updateSource has no address) {
    // file second warning
} else {
    // set localIp
}
```
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java, line 49 at r2](https://reviewable.io/reviews/batfish/batfish/6943#-MYqdphf6x9Fv6QGgCOJ:-MYqdphf6x9Fv6QGgCOK:byu7er) ([raw file](https://github.com/batfish/batfish/blob/47224dffcd73d17669b29d1cbb9b7b28c0b48ec9/projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java#L49)):*
> ```Java
>   public static final boolean DEFAULT_SECONDARY_IP_ENABLED = false;
>   public static final Speed DEFAULT_SPEED = Speed.AUTO;
>   public static final String DEFAULT_VDOM = "root";
> ```

do you know if `root` actually has special meaning on fortinet devices? Interfaces require you to configure a vdom, we've just been using "root" by default.

This might've been why I didn't convert neighbors whose update source couldn't be identified -- I didn't know what vrf's BGP process to put them in.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6943)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java, line 49 at r2](https://reviewable.io/reviews/batfish/batfish/6943#-MYqdphf6x9Fv6QGgCOJ:-MYqfopuEo17H_OZlpNP:b-hh1az4) ([raw file](https://github.com/batfish/batfish/blob/47224dffcd73d17669b29d1cbb9b7b28c0b48ec9/projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java#L49)):*

do you know if `root` actually has special meaning on fortinet devices? Interfaces require you to configure a vdom, we've just been using "root" by default.

This might've been why I didn't convert neighbors whose update source couldn't be identified -- I didn't know what vrf's BGP process to put them in.
</blockquote></details>

I do not. I picked root because it I saw root was being used. 

If root is completely our creation, we could call it default as a more descriptive term.

corinaminer (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6943)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java, line 49 at r2](https://reviewable.io/reviews/batfish/batfish/6943#-MYqdphf6x9Fv6QGgCOJ:-MYql1J-7N6cU0mQd47F:b-65g3yi) ([raw file](https://github.com/batfish/batfish/blob/47224dffcd73d17669b29d1cbb9b7b28c0b48ec9/projects/batfish/src/main/java/org/batfish/representation/fortios/Interface.java#L49)):*

I do not. I picked root because it I saw root was being used. 

If root is completely our creation, we could call it default as a more descriptive term. 
</blockquote></details>

Well, the problem is that `c.getVrfs().get(vrf).setBgpProcess(viProc)` (at the end of `convertBgpProcessForVrf`) will throw if the VRF doesn't exist. Iirc, the only way VRFs are currently created in Fortios conversion is if there's an interface configured in them: in interface conversion, we create the VRF corresponding to the interface's configured VDOM and VRF if it doesn't already exist. That means a device with no configured interfaces will have no VRFs in our VI model, so there's no BGP process to hold any BGP peers.

Does the concept of a "default VDOM" exist? Samir or Chirag would probably know.

If yes, then I think it's reasonable to guarantee a corresponding VRF is created during conversion, and any BGP peers without update sources can go in that VRF's BGP process.

If no, do we want to make up our own concept of a "default VRF" just to have somewhere to put those BGP peers? That seems like it may have some confusing-to-users repercussions; need to run it by Dan at least.

ratulm:
Let's backburner this for now.

PR #6961, opened 2021-04-27
annotate: fix

dhalperi:
Since it has a dependency on Batfish, it should either be in a different project
or just be part of Batfish.

(I had thought it didn't need one, then it could be depended-on by Batfish.)

Also fix the classpath in tools.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6961)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6969, opened 2021-04-28
MemoizedIpSpaceToBDD: use identity equality for IpSpace equality

dhalperi:
Complex IpSpace objects (Acl, IpWildcardSet) have an expensive equals method,
which can cause problems for memoization: we have seen cases where converting
ForwardingAnalysis into BDDs took 10s pre-caching, and then a (cloned via
serialization) copy of the same objects took nearly 15m.

However, avoiding memoizing IpSpace -> BDD conversions altogether also causes
performance problems.

Instead, cache IpSpace objects with identity equality, so that multiple
conversions of the same object are remembered, but different equal objects are
converted redundantly.

Note that most IpSpace types are interned, so that `a != b` should imply
`!a.equals(b)` anyway.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6969)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6972, opened 2021-04-28
IOS-XR fix vrf leaking parsing and vendor model

arifogel:
- only 2 possible policies for leaking in each direction
  - 1 for default vrf
  - 1 for ALL other VRFs, not 1 per non-default vrf


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6972)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer and @sfraint)

PR #6978, opened 2021-05-03
InterfacesByVlanRange: canonicalize on the way in

dhalperi:
This should not have any e2e functional changes, but it sure makes the output
easier to understand. It also makes the intersection operation never produce an
empty intersection, since two closedOpen ranges cannot produce one.


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6978)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #6989, opened 2021-05-08
IOS-XR as-path-set and as-path boolean VS model and extraction

arifogel:
- rewrite and flesh out IOS-XR as-path-set and as-path boolean model
- extract to VS instead of VI structures
- fix some missing cisco->java regex conversion
- fewer crashes for unimplemented functionality
- parsing fixes and test config updates


corinaminer (commented):
> Reviewed 24 of 32 files at r1, 6 of 6 files at r2, 18 of 18 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: 48 of 49 files reviewed, 11 unresolved discussions (waiting on @arifogel and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6989#-M_YQ7z7B4REIKt9y8De:-M_YQ7z7B4REIKt9y8Df:bq2wxi5) (no related file):*
Not fully clear on the ref test changes, may need to go over them with you tomorrow.

---
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_as_path_set.g4, line 55 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_XxKd0AXrcL47UPdEI:-M_XxKd0AXrcL47UPdEJ:b7eoa5l) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_as_path_set.g4#L55)):*
> ```ANTLR
> :
>   lo = as_number
>   | BRACKET_LEFT lo = as_number DOTDOT hi = as_number BRACKET_RIGHT
> ```

so to be clear, the `7.8` format for 32-bit numbers isn't supported yet?
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6995 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y24vo9uAKaoWwzIGD:-M_Y24vo9uAKaoWwzIGE:bljb42t) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6995)):*
> ```Java
> 
>   private @Nonnull Optional<List<Range<Long>>> toRanges(
>       ParserRuleContext messageCtx, As_range_expr_listContext ctx) {
> ```

I don't think `as_range_expr_list` should convert to a list of long ranges here, since it can use variables. It doesn't really matter yet since the AS path constructs that use `toRanges` aren't yet converted to VI anyway, but there should at least be a todo.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 7112 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y3NUFBYIN3h4dJ43b:-M_Y3NUFBYIN3h4dJ43c:bolul0y) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L7112)):*
> ```Java
>       return IntComparator.GE;
>     } else if (ctx.IS() != null) {
>       return IntComparator.EQ;
> ```

there's really no difference between xr's `EQ` and `IS`?
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 8303 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y4EVABJ_0bFB-VjFK:-M_Y4EVABJ_0bFB-VjFL:b2imm9n) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L8303)):*
>TODO: create and use VS InlineAsPathSet

is this really necessary or just a preference for using separate classes for different contexts? (no action needed either way)
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/AsPathSetElemVisitor.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YJImQCd-Oow4AZou0:-M_YJImQCd-Oow4AZou1:b-k05hat) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/AsPathSetElemVisitor.java#L6)):*
> ```Java
>  * A visitor of {@link AsPathSetElem} that takes a single argument and returns a generic argument.
>  */
> public interface AsPathSetElemVisitor<T, U> {
> ```

Doesn't look like any implementations use the argument.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 2148 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YEupH3j5Dr4wX4TCT:-M_YEupH3j5Dr4wX4TCU:bcp9rlx) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L2148)):*
> ```Java
>         DfaRegexAsPathSetElem dfaRegexAsPathSetElem, Void arg) {
>       return new AsPathAccessListLine(
>           LineAction.PERMIT, toJavaRegex(dfaRegexAsPathSetElem.getRegex()));
> ```

Should probably have a todo about figuring out how DFA differs from IOS style regex
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/DfaRegexAsPathSetElem.java, line 27 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YFR8Z1byc9g5IEM7o:-M_YFR8Z1byc9g5IEM7p:b-7jmf1o) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/DfaRegexAsPathSetElem.java#L27)):*
> ```Java
> 
>   @Override
>   public boolean equals(Object o) {
> ```

parameter needs `@Nullable` annotation. Same for `equals` method in many of the other new classes
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIn.java, line 49 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YHXBZ2zu0dP6fqKjF:-M_YHXBZ2zu0dP6fqKjG:buskua1) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIn.java#L49)):*
> ```Java
>       return BooleanExprs.FALSE;
>     }
>     return new MatchAsPath(expr);
> ```

nit: why define `expr` rather than just returning `new MatchAsPath(...)` at each point where `expr` is currently set?
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIsLocal.java, line 32 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YI6Py11pHxL3bV3dA:-M_YI6Py11pHxL3bV3dB:by35iki) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIsLocal.java#L32)):*
> ```Java
>   @Override
>   public boolean equals(Object obj) {
>     return this == obj || obj instanceof RoutePolicyBooleanAsPathIsLocal;
> ```

`obj instanceof RoutePolicyBooleanAsPathIsLocal` should be sufficient; same in `UnimplementedBoolean`
___
*[projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java, line 2386 at r3](https://reviewable.io/reviews/batfish/batfish/6989#-M_YKunICz0Y9Sj8DCaS:-M_YKunICz0Y9Sj8DCaT:bpwbdy5) ([raw file](https://github.com/batfish/batfish/blob/36e0abcd82d4fc37a8385f29ddc3419552258127/projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java#L2386)):*
> ```Java
>       }
>       assertFalse(i.hasNext());
>     }
> ```

No need for the outer scope, and i would remove all the inner scopes as well; they don't contain any new variable definitions and are actually quite dependent on each other (to iterate the iterator). In fact, as far as i'm concerned you could remove `elem` entirely and just have a bunch of `assertThat(i.next(), equalTo(...))`. Or even one big `assertThat(set.getElements(), equalTo(ImmutableList.of(...))`.
___
*[projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java, line 2455 at r3](https://reviewable.io/reviews/batfish/batfish/6989#-M_YMvjMDKY7SpOI_9CY:-M_YMvjN6ZAQ22W4Cre9:bflbpe6) ([raw file](https://github.com/batfish/batfish/blob/36e0abcd82d4fc37a8385f29ddc3419552258127/projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java#L2455)):*
> ```Java
>         assertThat(elem, equalTo(new UniqueLengthAsPathSetElem(EQ, 2, true)));
>       }
>       assertFalse(ei.hasNext());
> ```

similar thought here: no need to scope each element's assert
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-as-path-boolean, line 7 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YPB4EEPUCD9RSlTD9:-M_YPB4EEPUCD9RSlTDA:bqmge0j) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-as-path-boolean#L7)):*
> ```
> route-policy rp1
>  # in inline
>  if as-path in (dfa-regex '^1_', ios-regex '^2_', length eq 1, length ge 2 all, neighbor-is '1 [2..3]', originates-from '4 [5..6]', passes-through '7 [8..9]', unique-length eq 1, unique-length eq 2 all) then
> ```

may want to test `exact` in this inline one, since the other inline one extracts to an `UnimplementedBoolean`

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: 48 of 49 files reviewed, 8 unresolved discussions (waiting on @arifogel, @corinaminer, and @sfraint)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_as_path_set.g4, line 55 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_XxKd0AXrcL47UPdEI:-M_b5XFPBcPKqpJtCrs2:bhq1lm5) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_as_path_set.g4#L55)):*

so to be clear, the `7.8` format for 32-bit numbers isn't supported yet?
</blockquote></details>

Good catch. Implemented.
Added dotted AS to test configs.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6995 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y24vo9uAKaoWwzIGD:-M_b86QD2I_Zix6sRHJc:b-swmov0) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6995)):*

I don't think `as_range_expr_list` should convert to a list of long ranges here, since it can use variables. It doesn't really matter yet since the AS path constructs that use `toRanges` aren't yet converted to VI anyway, but there should at least be a todo.
</blockquote></details>

Certainly it will have to change to support params.
Added todo.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 7112 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y3NUFBYIN3h4dJ43b:-M_b8FJq-vOh4XCz40vH:b-nzqjoq) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L7112)):*

there's really no difference between xr's `EQ` and `IS`?
</blockquote></details>

Nope. Just tested again.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 8303 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_Y4EVABJ_0bFB-VjFK:-M_bA7kX3mN8gsrLj2OT:b1x2f30) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L8303)):*

>TODO: create and use VS InlineAsPathSet

is this really necessary or just a preference for using separate classes for different contexts? (no action needed either way)
</blockquote></details>

really this is about supporting params in the future. updated todo to reflect.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/AsPathSetElemVisitor.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YJImQCd-Oow4AZou0:-M_bCTqjBZWqiBQzmhhU:bss4a2b) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/AsPathSetElemVisitor.java#L6)):*

Doesn't look like any implementations use the argument.
</blockquote></details>

Removed.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/DfaRegexAsPathSetElem.java, line 27 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YFR8Z1byc9g5IEM7o:-M_bDH8h-kjPSJ50AwGv:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/DfaRegexAsPathSetElem.java#L27)):*

parameter needs `@Nullable` annotation. Same for `equals` method in many of the other new classes
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIn.java, line 49 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YHXBZ2zu0dP6fqKjF:-M_bDfdR4PxQjC2zyH7V:b1qxvou) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIn.java#L49)):*

nit: why define `expr` rather than just returning `new MatchAsPath(...)` at each point where `expr` is currently set?
</blockquote></details>

Had an original version where all branches used constructor. Doesn't really matter either way.
Updated.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIsLocal.java, line 32 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YI6Py11pHxL3bV3dA:-M_bEdMV0NqbAHCvasE0:b-7tvbxw) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/RoutePolicyBooleanAsPathIsLocal.java#L32)):*

`obj instanceof RoutePolicyBooleanAsPathIsLocal` should be sufficient; same in `UnimplementedBoolean`
</blockquote></details>

I put a reference equality check here for the same reason we do it everywhere. It's faster than the rest of the checks.
___
*[projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java, line 2386 at r3](https://reviewable.io/reviews/batfish/batfish/6989#-M_YKunICz0Y9Sj8DCaS:-M_bFSlh88pM6oiF5MAc:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/36e0abcd82d4fc37a8385f29ddc3419552258127/projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java#L2386)):*

No need for the outer scope, and i would remove all the inner scopes as well; they don't contain any new variable definitions and are actually quite dependent on each other (to iterate the iterator). In fact, as far as i'm concerned you could remove `elem` entirely and just have a bunch of `assertThat(i.next(), equalTo(...))`. Or even one big `assertThat(set.getElements(), equalTo(ImmutableList.of(...))`.
</blockquote></details>

Done.
___
*[projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java, line 2455 at r3](https://reviewable.io/reviews/batfish/batfish/6989#-M_YMvjMDKY7SpOI_9CY:-M_bGdAI5clZptROKYII:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/36e0abcd82d4fc37a8385f29ddc3419552258127/projects/batfish/src/test/java/org/batfish/grammar/cisco_xr/XrGrammarTest.java#L2455)):*

similar thought here: no need to scope each element's assert
</blockquote></details>

done

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: 25 of 50 files reviewed, 7 unresolved discussions (waiting on @arifogel, @corinaminer, and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 2148 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YEupH3j5Dr4wX4TCT:-M_bHuWV21h0Uc7S1Kzd:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L2148)):*

Should probably have a todo about figuring out how DFA differs from IOS style regex
</blockquote></details>

done
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-as-path-boolean, line 7 at r2](https://reviewable.io/reviews/batfish/batfish/6989#-M_YPB4EEPUCD9RSlTD9:-M_bICzN4X_GpDcTsn5u:b-za4i1k) ([raw file](https://github.com/batfish/batfish/blob/d3f1db3cd5db9c44d68a14ea8950956f8a38ba05/projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-as-path-boolean#L7)):*

may want to test `exact` in this inline one, since the other inline one extracts to an `UnimplementedBoolean`
</blockquote></details>

good catch!
done

corinaminer (commented):
> Reviewed 24 of 24 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: 49 of 50 files reviewed, 4 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4, line 38 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bOiFwBU0Exc82zWuC:-M_bOiFx5L-8yXhKfVsX:b-9xrw81) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4#L38)):*
> ```ANTLR
>   // 1-4294967295
>   uint32
>   | hi = uint16 PERIOD lo = uint16
> ```

did you confirm the dotted format is valid everywhere `as_number` is used? (or, is that at least a reasonable assumption based on your experience with the XR cli?)
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6986 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bR5tk9g1J2Ytul77F:-M_bR5tk9g1J2Ytul77G:bu4j98r) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6986)):*
> ```Java
>       assert ctx.hi != null && ctx.lo != null;
>       long hi = toInteger(ctx.hi);
>       long lo = toInteger(ctx.lo);
> ```

nit: these could be ints
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6988 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bRbXb0HfNsLx_Fm_m:-M_bRbXb0HfNsLx_Fm_n:bwx4ylp) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6988)):*
> ```Java
>       long lo = toInteger(ctx.lo);
>       long val = (hi << 16) | lo;
>       if (val == 0) {
> ```

Does `AS_NUMBER_RANGE` go all the way up to 2^32 - 1? If not, the `if` here needs to actually check if `AS_NUMBER_RANGE` contains `val`. (Probably not a bad idea to do that either way, just to keep the check consistent for dotted and non-dotted formats.)
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6989 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bPTxnBxcKw-R8HW18:-M_bPTxnBxcKw-R8HW19:b-otmam3) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6989)):*
> ```Java
>       long val = (hi << 16) | lo;
>       if (val == 0) {
>         warn(messageCtx, "0 is not a valid AS");
> ```

for consistency's sake, i think this should the warning format from `toLongInSpace`:
`warn(messageCtx, String.format("Expected %s in range %s, but got '%d'", name, space, num));`

In this case you could specialize/tweak as needed, presumably like:
`warn(messageCtx, String.format("Expected AS number in range %s, but got '%d.%d'", AS_NUMBER_RANGE, hi, lo))`

corinaminer (commented):
> Reviewed 1 of 32 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: all files reviewed, 4 unresolved discussions (waiting on @arifogel and @sfraint)

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/6989#-M_YQ7z7B4REIKt9y8De:-M_bYs8a36lYcavCNNpH:b-xw7qog) (no related file):*

Not fully clear on the ref test changes, may need to go over them with you tomorrow.
</blockquote></details>

resolving as discussed offline

---
___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4, line 38 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bOiFwBU0Exc82zWuC:-M_bUj6O119n3R1K6wez:bb77ufg) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_common.g4#L38)):*

did you confirm the dotted format is valid everywhere `as_number` is used? (or, is that at least a reasonable assumption based on your experience with the XR cli?)
</blockquote></details>

Yes.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6986 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bR5tk9g1J2Ytul77F:-M_bVkJ7EUi_CkTmAF9F:bhizbi3) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6986)):*

nit: these could be ints
</blockquote></details>

They would have to be cast to long before the shift since the upper AS limit is UINT32MAX. Easier to just declare as `long`.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6988 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bRbXb0HfNsLx_Fm_m:-M_bXmPu0NVOwXMfLGJ3:bwu2eqn) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6988)):*

Does `AS_NUMBER_RANGE` go all the way up to 2^32 - 1? If not, the `if` here needs to actually check if `AS_NUMBER_RANGE` contains `val`. (Probably not a bad idea to do that either way, just to keep the check consistent for dotted and non-dotted formats.)
</blockquote></details>

Yes.
Done, for legibility.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6989 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bPTxnBxcKw-R8HW18:-M_bYlFN76lmERRvcgvb:bsnj8mr) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6989)):*

for consistency's sake, i think this should the warning format from `toLongInSpace`:
`warn(messageCtx, String.format("Expected %s in range %s, but got '%d'", name, space, num));`

In this case you could specialize/tweak as needed, presumably like:
`warn(messageCtx, String.format("Expected AS number in range %s, but got '%d.%d'", AS_NUMBER_RANGE, hi, lo))`
</blockquote></details>

I did a similar tweak.

corinaminer (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6986 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bR5tk9g1J2Ytul77F:-M_bbHPlEh7UA0kgGuLt:bmdrmg7) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6986)):*

They would have to be cast to long before the shift since the upper AS limit is UINT32MAX. Easier to just declare as `long`.
</blockquote></details>

i meant `hi` and `lo` could be ints (`val` can't), which i believe is true and would not necessitate casting to longs anywhere. But, doesn't really matter.
___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6989 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bPTxnBxcKw-R8HW18:-M_bbmPrBSR2OAb7PZ3o:b-9q9pm0) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6989)):*

I did a similar tweak.
</blockquote></details>

i think you have `val` and `ctx.getText()` swapped (assuming you want this to say "1.1 expressed as a single number is 65537")

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6989 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bPTxnBxcKw-R8HW18:-M_beXaZ5pb5hoVT31FD:b-nubdxl) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6989)):*

i think you have `val` and `ctx.getText()` swapped (assuming you want this to say "1.1 expressed as a single number is 65537")
</blockquote></details>

gah, fixed

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: 49 of 50 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6986 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bR5tk9g1J2Ytul77F:-M_beoSn05HGLkdS_qWS:b-3fzxox) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6986)):*

i meant `hi` and `lo` could be ints (`val` can't), which i believe is true and would not necessitate casting to longs anywhere. But, doesn't really matter.
</blockquote></details>

If you shift `hi << 16` where `hi` is an int, then the final value could be negative after the cast.
You have to cast `hi` to long before the shift. It's just easier to use longs.

corinaminer (commented):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6989)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java, line 6986 at r4](https://reviewable.io/reviews/batfish/batfish/6989#-M_bR5tk9g1J2Ytul77F:-M_brUlZ1hkBfWaKLR3a:b-vzs1s7) ([raw file](https://github.com/batfish/batfish/blob/ece3876fecd09a0266e245d9d4a3c3a8ac4317a1/projects/batfish/src/main/java/org/batfish/grammar/cisco_xr/CiscoXrControlPlaneExtractor.java#L6986)):*

If you shift `hi << 16` where `hi` is an int, then the final value could be negative after the cast.
You have to cast `hi` to long before the shift. It's just easier to use longs.
</blockquote></details>

gotcha.

PR #6990, opened 2021-05-08
Embed source information in RouteFilterList 

ratulm:
Enables linking of namedStrutures and defineStructures (as in IpAccessList)


dhalperi:
Should use the way tracing works, not the legacy (unused) stuff in ip access list. Matt knows how that works.

ratulm:
@anothermattbrown - should I use VendorStructureId (instead of sourceName and sourceType separately)? Other things to be aware of?

Goal is to enable users to link a RouteFilterList namedStructure to its VS definition in config files.

ratulm:
After #6991

anothermattbrown (commented):
> Reviewed 27 of 32 files at r4, 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6990)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #6998, opened 2021-05-17
VI implement new as-path matching API

arifogel:
- implement new API for matching as-path
  - follows structure of new community-matching API
- add reference verifier for as-path structures
- rename old `MatchAsPath` to `LegacyMatchAsPath`
- need to support certain IOS-XR RPL constructs
- supported features
  - match on length
  - match against regex
  - search for sequence of ASes within corresponding list of AS ranges
  - match against deduped version of as-path, where repeated AS entries are collapsed


corinaminer (commented):
> Reviewed 17 of 21 files at r1, 19 of 32 files at r2, 2 of 2 files at r3, 1 of 3 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: 40 of 55 files reviewed, 9 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 178 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wzjVz5Y8pwaOnWgpx:-M_wzjVz5Y8pwaOnWgpy:bue729s) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L178)):*
> ```Java
> 
>   private static void verifyAsPathStructures(Configuration c) {
>     // TODO: crash on undefined/circular refs (conversion is responsible for preventing them)
> ```

not sure why that's a todo -- the verifier already crashes appropriately, right?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_whPKPCHQKQ6DdOcRh:-M_whPKPCHQKQ6DdOcRi:bz5ma6u) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java#L73)):*
> ```Java
>       return Optional.empty();
>     }
>     // TOOD: store information necessary to evaluate dynamic expressions
> ```

By "dynamic expressions", do you mean expressions with variables? if not, what are dynamic expressions/what additional information is needed? please clarify todo
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExpr.java, line 17 at r4](https://reviewable.io/reviews/batfish/batfish/6998#-M_wMxvN2VAD8_4L3AdD:-M_wMxvN2VAD8_4L3AdE:b-wp9uip) ([raw file](https://github.com/batfish/batfish/blob/365a8b0f4d184b738699ba2cc2594afe8d284b16/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExpr.java#L17)):*
> ```Java
> 
>   @Override
>   public abstract int hashCode();
> ```

Does overriding here make it so child classes are forced to implement equals/hashcode?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluator.java, line 34 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wpPDp4NH-d3R47pIa:-M_wpPDp4NH-d3R47pIb:bvvqzor) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluator.java#L34)):*
> ```Java
>     ImmutableList.Builder<AsSet> builder = ImmutableList.builder();
>     for (AsSet asSet : input.getAsSets()) {
>       if (asSet == lastAsSet) {
> ```

shouldn't this use `.equals()`?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java, line 35 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wmKfx0GxxuoZlZk9a:-M_wmKfx0GxxuoZlZk9b:b-8l3mwk) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java#L35)):*
> ```Java
>   private @Nonnull List<AsPathMatchExpr> getDisjunctsSorted() {
>     // ordered for refs
>     return ImmutableList.copyOf(_disjuncts);
> ```

isn't this still nondeterministic, since the iteration order of `ImmutableSet` isn't guaranteed? Would it be easier to just make `_disjuncts` a list? You'd have to adjust `equals` and `hashcode` (also `of`, if you care about eliminating duplicate disjuncts on creation).

otherwise, i think you need some arbitrary but deterministic comparator to order the disjuncts here.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 159 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_ww0Yo-1XXd7ZwS0DV:-M_ww0Yo-1XXd7ZwS0DW:bcuzjqk) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L159)):*
> ```Java
>     public Void visitMatchLegacyAsPath(
>         LegacyMatchAsPath legacyMatchAsPath, AsPathStructuresVerifierContext arg) {
>       return null;
> ```

doesn't this need to look into whether `legacyMatchAsPath` matches a `NamedAsPathSet`? Or are undefined/cyclic `AsPathAccessList`s not in the scope of this verifier?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsSetsMatchingRanges.java, line 18 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wjrCeCBnnJj1LUz0Z:-M_wjrCeCBnnJj1LUz0_:blsfy8a) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsSetsMatchingRanges.java#L18)):*
> ```Java
>  * sequence of {@link org.batfish.datamodel.AsSet}s, each of which contains an AS within a
>  * corresponding sequence of AS ranges. The match may depend on the sequence being anchored at
>  * either, both, or neither of the start/end of the {@link org.batfish.datamodel.AsPath}.
> ```

Please also add docs to the getters. `getRanges` doc should clarify that an `AsSet` matches a range if any value in the set is in the range, and also that the AS path must match the ranges strictly sequentially without skipping any ASes.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/HasAsPathLength.java, line 47 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wlrrvBApqFjD4GgbD:-M_wlrrvBApqFjD4GgbE:bp2lbve) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/HasAsPathLength.java#L47)):*
> ```Java
>   }
> 
>   private static final String PROP_COMPARISON = "comparator";
> ```

nit: comparison, not comparator
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/InputAsPath.java, line 33 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wqRpl5QiIpNMxqpYD:-M_wqRpl5QiIpNMxqpYE:b14btws) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/InputAsPath.java#L33)):*
> ```Java
>   }
> 
>   private static final InputAsPath INSTANCE = new InputAsPath();
> ```

i think you need to explicitly define a constructor to make it private
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/LegacyMatchAsPath.java, line 13 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wVfSr5ooF8vQ478JH:-M_wVfSr5ooF8vQ478JI:bffeys) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/LegacyMatchAsPath.java#L13)):*
> ```Java
> /**
>  * Boolean expression that tests whether an {@link Environment} contains a BGP route with an AS path
>  * that matches a given {@link AsPathSetExpr}.
> ```

perhaps update javadoc with a pointer to `MatchAsPath` and/or explanation as to why this one is legacy

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: 33 of 55 files reviewed, 8 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 178 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wzjVz5Y8pwaOnWgpx:-M_x0H17CP4PVO9ZJQdm:b8d4aya) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L178)):*

not sure why that's a todo -- the verifier already crashes appropriately, right?
</blockquote></details>

Just copied from below. I don't really know what it means. This definitely throws when something bad is found.
Removed comment.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_whPKPCHQKQ6DdOcRh:-M_x0_zY2ALENWOOXbZe:b-fa9sz8) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java#L73)):*

By "dynamic expressions", do you mean expressions with variables? if not, what are dynamic expressions/what additional information is needed? please clarify todo
</blockquote></details>

clarified
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExpr.java, line 17 at r4](https://reviewable.io/reviews/batfish/batfish/6998#-M_wMxvN2VAD8_4L3AdD:-M_x0boYA47iD_KVuW4c:b2nign6) ([raw file](https://github.com/batfish/batfish/blob/365a8b0f4d184b738699ba2cc2594afe8d284b16/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExpr.java#L17)):*

Does overriding here make it so child classes are forced to implement equals/hashcode?
</blockquote></details>

yes
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluator.java, line 34 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wpPDp4NH-d3R47pIa:-M_x0lRb4E4wN7m0yDJo:b-cy2lqi) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluator.java#L34)):*

shouldn't this use `.equals()`?
</blockquote></details>

Yes indeed. The tests accidentally passed because of some caching/interning we do for `AsSet`, but we shouldn't rely on that.
Fixed.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java, line 35 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wmKfx0GxxuoZlZk9a:-M_x1PA5DAumZJ8N0ipm:bju4tne) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java#L35)):*

isn't this still nondeterministic, since the iteration order of `ImmutableSet` isn't guaranteed? Would it be easier to just make `_disjuncts` a list? You'd have to adjust `equals` and `hashcode` (also `of`, if you care about eliminating duplicate disjuncts on creation).

otherwise, i think you need some arbitrary but deterministic comparator to order the disjuncts here.
</blockquote></details>

The iteration order of `ImmutableSet` is guaranteed to be insertion order:
```
  /**
   * A builder for creating {@code ImmutableSet} instances. Example:
   *
   * <pre>{@code
   * static final ImmutableSet<Color> GOOGLE_COLORS =
   *     ImmutableSet.<Color>builder()
   *         .addAll(WEBSAFE_COLORS)
   *         .add(new Color(0, 191, 255))
   *         .build();
   * }</pre>
   *
   * <p>Elements appear in the resulting set in the same order they were first added to the builder.
   *
   * <p>Building does not change the state of the builder, so it is still possible to add more
   * elements and to build again.
   *
   * @since 2.0
   */
  public static class Builder<E> extends ImmutableCollection.Builder<E> {
```
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 159 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_ww0Yo-1XXd7ZwS0DV:-M_x1rmT4rwGF-44r6Qn:b-nwhj6u) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L159)):*

doesn't this need to look into whether `legacyMatchAsPath` matches a `NamedAsPathSet`? Or are undefined/cyclic `AsPathAccessList`s not in the scope of this verifier?
</blockquote></details>

Out of scope. Not touching behavior of legacy stuff.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsSetsMatchingRanges.java, line 18 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wjrCeCBnnJj1LUz0Z:-M_x35huAVfUU273sbC1:by06e9r) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsSetsMatchingRanges.java#L18)):*

Please also add docs to the getters. `getRanges` doc should clarify that an `AsSet` matches a range if any value in the set is in the range, and also that the AS path must match the ranges strictly sequentially without skipping any ASes.
</blockquote></details>

documented
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/HasAsPathLength.java, line 47 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wlrrvBApqFjD4GgbD:-M_x3AEK5MMgISIueYr8:b-7ql1td) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/HasAsPathLength.java#L47)):*

nit: comparison, not comparator
</blockquote></details>

fixed
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/InputAsPath.java, line 33 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wqRpl5QiIpNMxqpYD:-M_x3FqG2CDSV0bEVa3y:b-1j8uql) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/InputAsPath.java#L33)):*

i think you need to explicitly define a constructor to make it private
</blockquote></details>

yup, missed it.
done.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/LegacyMatchAsPath.java, line 13 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wVfSr5ooF8vQ478JH:-M_x3wNnEUU9XVCOLwNH:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/LegacyMatchAsPath.java#L13)):*

perhaps update javadoc with a pointer to `MatchAsPath` and/or explanation as to why this one is legacy
</blockquote></details>

done

corinaminer (commented):
> Reviewed 1 of 21 files at r1, 12 of 32 files at r2, 7 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: 53 of 55 files reviewed, 7 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_whPKPCHQKQ6DdOcRh:-Ma-YSUC5hwQF_NYpe8V:b-mut995) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java#L73)):*

clarified
</blockquote></details>

thanks. Also, just noticed, there's a typo in "todo" -- worth fixing in case someone's searching for todos
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java, line 35 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wmKfx0GxxuoZlZk9a:-Ma-YtW45XvN7-Sy_iBr:b72fphu) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java#L35)):*

The iteration order of `ImmutableSet` is guaranteed to be insertion order:
```
  /**
   * A builder for creating {@code ImmutableSet} instances. Example:
   *
   * <pre>{@code
   * static final ImmutableSet<Color> GOOGLE_COLORS =
   *     ImmutableSet.<Color>builder()
   *         .addAll(WEBSAFE_COLORS)
   *         .add(new Color(0, 191, 255))
   *         .build();
   * }</pre>
   *
   * <p>Elements appear in the resulting set in the same order they were first added to the builder.
   *
   * <p>Building does not change the state of the builder, so it is still possible to add more
   * elements and to build again.
   *
   * @since 2.0
   */
  public static class Builder<E> extends ImmutableCollection.Builder<E> {
```
</blockquote></details>

so the same `AsPathMatchAny` instance will always serialize the same -- that's good, but what about two different (but equal) instances? Any time we're generating refs, the VI model we're serializing will be a completely new set of objects since last time refs were generated. Afaict, the only way we might currently have a guarantee that refs will come out consistent is if the iterable passed to the constructor has a deterministic order, which we shouldn't rely on.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathContextTest.java, line 76 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma-x3cN214IPBHmDo5-:-Ma-x3cN214IPBHmDo50:b-5q6bbw) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathContextTest.java#L76)):*
> ```Java
>                             .setOriginatorIp(Ip.ZERO)
>                             .setOriginType(OriginType.INCOMPLETE)
>                             .setProtocol(RoutingProtocol.BGP)
> ```

you should only need to set AS path and network (other stuff is set to arbitrary values in `testBuilder`)
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluatorTest.java, line 36 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma-xfWh6z6AizGvEEIR:-Ma-xfWh6z6AizGvEEIS:b-d5w5sd) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluatorTest.java#L36)):*
> ```Java
>         DedupedAsPath.of(InputAsPath.instance())
>             .accept(AsPathExprEvaluator.instance(), context(asPath)),
>         equalTo(AsPath.ofSingletonAsSets(1L, 2L, 3L, 2L, 4L)));
> ```

I think it should also test deduping an empty `AsPath`, just to cover the bases.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java, line 93 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma0-p2f0q3i25RQFBoQ:-Ma0-p2f0q3i25RQFBoR:bec8d6d) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java#L93)):*
> ```Java
>                 false,
>                 ImmutableList.of(Range.singleton(2L), Range.singleton(3L), Range.singleton(4L))),
>             AsPath.ofSingletonAsSets(2L, 3L, 4L, 5L)));
> ```

Please add some tests here that do not evaluate to true. In particular, needs test cases where:
- there are more ranges to match than ASNs in the path
- the AS path contains AS sets (at least two such tests so we can see it evaluate to both false and true)
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java, line 168 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma010Xt2d0bB1xrL36s:-Ma010Xt2d0bB1xrL36t:b8lha7k) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java#L168)):*
> ```Java
>                 true,
>                 ImmutableList.of(Range.singleton(2L), Range.singleton(3L), Range.singleton(4L))),
>             AsPath.ofSingletonAsSets(2L, 3L, 4L, 5L)));
> ```

Both of these `assertFalse` tests are going to invoke your short-circuit logic in the evaluator:
```
    if (anchorStart) {
      if (anchorEnd && numAsSets != numRanges) {
        return false;
      }
      return tryMatchRanges(asRanges, asSets.subList(0, numRanges));
    }
```
please add a test with an AS path that will make it through to `tryMatchRanges`, but not match -- e.g. `2, 3, 5`.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/MatchAsPathTest.java, line 54 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma04COx5mMDv7fo-2CQ:-Ma04COx5mMDv7fo-2CR:bj7m85h) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/MatchAsPathTest.java#L54)):*
> ```Java
>                     .setOriginalRoute(new ConnectedRoute(Prefix.ZERO, "i1"))
>                     .build())
>             .getBooleanValue());
> ```

Should also test with a route that does have an AsPath but doesn't match the expr.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: 53 of 55 files reviewed, 3 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java, line 73 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_whPKPCHQKQ6DdOcRh:-Ma0pT2l96thyXTEnKu5:b-7ql1td) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathContext.java#L73)):*

thanks. Also, just noticed, there's a typo in "todo" -- worth fixing in case someone's searching for todos
</blockquote></details>

fixed
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java, line 35 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wmKfx0GxxuoZlZk9a:-Ma0pf4w2pdnFnqkvps5:bqxiqii) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java#L35)):*

so the same `AsPathMatchAny` instance will always serialize the same -- that's good, but what about two different (but equal) instances? Any time we're generating refs, the VI model we're serializing will be a completely new set of objects since last time refs were generated. Afaict, the only way we might currently have a guarantee that refs will come out consistent is if the iterable passed to the constructor has a deterministic order, which we shouldn't rely on.
</blockquote></details>

In practice, this is rarely an issue (this pattern is copied from other stuff). We don't add much in the way of ref-generating unit-test configs.
IMO this can be addressed if it pops up as a problem in the future.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathContextTest.java, line 76 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma-x3cN214IPBHmDo5-:-Ma0qPFD-79SnI6WLOXd:bcbci7v) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathContextTest.java#L76)):*

you should only need to set AS path and network (other stuff is set to arbitrary values in `testBuilder`)
</blockquote></details>

This is copied, but sure.
fixed.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluatorTest.java, line 36 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma-xfWh6z6AizGvEEIR:-Ma0r-6fC_AYF_hfWnHy:b-2ojg2x) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathExprEvaluatorTest.java#L36)):*

I think it should also test deduping an empty `AsPath`, just to cover the bases.
</blockquote></details>

ok, done
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java, line 93 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma0-p2f0q3i25RQFBoQ:-Ma0sRkCAIoa6GFgv2eI:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java#L93)):*

Please add some tests here that do not evaluate to true. In particular, needs test cases where:
- there are more ranges to match than ASNs in the path
- the AS path contains AS sets (at least two such tests so we can see it evaluate to both false and true)
</blockquote></details>

done
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java, line 168 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma010Xt2d0bB1xrL36s:-Ma0smVRAE-VfEeX1GUs:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchExprEvaluatorTest.java#L168)):*

Both of these `assertFalse` tests are going to invoke your short-circuit logic in the evaluator:
```
    if (anchorStart) {
      if (anchorEnd && numAsSets != numRanges) {
        return false;
      }
      return tryMatchRanges(asRanges, asSets.subList(0, numRanges));
    }
```
please add a test with an AS path that will make it through to `tryMatchRanges`, but not match -- e.g. `2, 3, 5`.
</blockquote></details>

done
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/MatchAsPathTest.java, line 54 at r6](https://reviewable.io/reviews/batfish/batfish/6998#-Ma04COx5mMDv7fo-2CQ:-Ma0rGlJ3erg0NWrC1NS:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/306fa3c8048ebc7f87fe44eea50c5d0d85d18bc3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/as_path/MatchAsPathTest.java#L54)):*

Should also test with a route that does have an AsPath but doesn't match the expr.
</blockquote></details>

done

corinaminer (commented):
> Reviewed 2 of 3 files at r4, 5 of 5 files at r7, 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java, line 35 at r5](https://reviewable.io/reviews/batfish/batfish/6998#-M_wmKfx0GxxuoZlZk9a:-Ma0zWR-5Gxr2XSqRHDT:bb7ohn6) ([raw file](https://github.com/batfish/batfish/blob/76735aa73abfd04e2192d4c96f2f72c105983d44/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathMatchAny.java#L35)):*

In practice, this is rarely an issue (this pattern is copied from other stuff). We don't add much in the way of ref-generating unit-test configs.
IMO this can be addressed if it pops up as a problem in the future.
</blockquote></details>

sgtm
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 266 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4bwfI4hktbZ4yyltj:-Ma4bwfI4hktbZ4yyltk:b-cgm6u8) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L266)):*
> ```Java
> 
>     @Override
>     public Void visitPassesThroughAsPath(
> ```

Would now be a good time to delete these useless and unused legacy AS path structures? It's somewhat confusing to have them return null in this visitor. OK to punt to separate PR though.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 322 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4hI1q5WgHFnhetFJ7:-Ma4hI1q5WgHFnhetFJ8:b-ea2i9) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L322)):*
> ```Java
>       resolved.accept(AS_PATH_MATCH_EXPR_VERIFIER, arg);
>       arg._verifiedAsPathMatchExprs.add(name);
>       arg._visitedAsPathMatchExprs.remove(name);
> ```

Might be clearer to call `_visitedAsPathMatchExprs` something like `_currentAsPathReferenceChain` to make it clear that it doesn't also include previously-visited-and-verified exprs.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 513 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4jl2SAscU7zsJOry4:-Ma4jl2SAscU7zsJOry5:bgnxk4w) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L513)):*
> ```Java
> 
>   private static final class RibExprVerifier
>       implements RibExprVisitor<Void, AsPathStructuresVerifierContext> {
> ```

The idea here is that having this class defined will make it obvious that we need to add something when some `RibExpr` eventually includes as-path? If/when that happens, will the as-path be in a new type of `RibExpr`? If not, it seems useless to have this class at all; could just put a todo in `testVisitRibIntersectsPrefixSpace` instead.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: 54 of 55 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 266 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4bwfI4hktbZ4yyltj:-Ma5-AZW8M6hQFNoBwYu:b6d95z5) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L266)):*

Would now be a good time to delete these useless and unused legacy AS path structures? It's somewhat confusing to have them return null in this visitor. OK to punt to separate PR though.
</blockquote></details>

punt, definitely
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 322 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4hI1q5WgHFnhetFJ7:-Ma5-Rk8DgBUr0tqSeKe:bhuft8s) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L322)):*

Might be clearer to call `_visitedAsPathMatchExprs` something like `_currentAsPathReferenceChain` to make it clear that it doesn't also include previously-visited-and-verified exprs.
</blockquote></details>

done, here and for `AsPathExprReference`
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 513 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4jl2SAscU7zsJOry4:-Ma518hb23Jard0D0atp:b-xodyc5) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L513)):*

The idea here is that having this class defined will make it obvious that we need to add something when some `RibExpr` eventually includes as-path? If/when that happens, will the as-path be in a new type of `RibExpr`? If not, it seems useless to have this class at all; could just put a todo in `testVisitRibIntersectsPrefixSpace` instead.
</blockquote></details>

This was a stretch even in the code it was copied from (communities).
Just removed it altogether. It really isn't a reasonable expectation that as-path will figure in here.

corinaminer (commented):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java, line 513 at r8](https://reviewable.io/reviews/batfish/batfish/6998#-Ma4jl2SAscU7zsJOry4:-Ma58bHp4kAx-taZhCk5:b-q297q2) ([raw file](https://github.com/batfish/batfish/blob/3baeeeca92aeeb3e715e344d28c0a61305e44110/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/as_path/AsPathStructuresVerifier.java#L513)):*

This was a stretch even in the code it was copied from (communities).
Just removed it altogether. It really isn't a reasonable expectation that as-path will figure in here.
</blockquote></details>

ok, so by your judgment there's no call for a todo in `testVisitRibIntersectsPrefixSpace` even? (fine by me, just checking.)

arifogel (commented):
> Reviewed 4 of 4 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/6998)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7009, opened 2021-05-20
TransferBDD: Don't throw exceptions during community collection

millstein:
Modifies the visitors that collect up communities so that they never throw exceptions due to unsupported features for symbolic route-policy analysis.  Instead, exceptions are only thrown later if those features are actually required by the analysis.


arifogel (commented):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7009)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7010, opened 2021-05-20
IOS: exclude default-routes as valid steps in resolution process

### LINKED ISSUE #6785: IOS should not recursively resolve static route next hops through default route
For IOS static routes with next hop IPs, the forwarding next hops are resolved recursively. If a forwarding next hop cannot be found, the static route should not be activated. Batfish models this part accurately.

However, IOS has a quirk that static route next hop ips will not recursively resolve through a default route from any protocol. Batfish does not implement this restriction, and will erroneously activate static routes with a recursive next-hop longest prefix match through a default route.

sfraint:
IOS doesn't permit resolving next hop interfaces through default-routes. This PR adds a fixed resolution policy for IOS VRFs that prevents this.

Fixes #6785


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7010)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7016, opened 2021-05-24
IOS XR: handle setting OSPF router network type

sfraint:
Handle setting OSPF network type at OSPF router level. This value is used when no network type is set at the OSPF interface level.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7016)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7018, opened 2021-05-24
FRR: Add support for ospf hello and dead interval timers 

pranavbj-amzn:
added support for OSPF hello and dead interval timers for FRR configs
 the support in vendor independent model for OSPF already exists. This PR provides the support specifically for FRR (vendor specific configs).

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4
@@ -130,6 +130,8 @@ CRITICAL: 'critical';
 
 DATACENTER: 'datacenter';
 
+DEAD_INTERVAL: 'dead-interval';
+
 DEBUGGING: 'debugging';
 
 DEFAULT: 'default';
@@ -193,6 +195,8 @@ FRR: 'frr';
 GE: 'ge';
 
 GOTO: 'goto';
+
+HELLO_INTERVAL: 'hello-interval';
 
 HOSTNAME
 :
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrr_interface.g4
@@ -69,6 +69,8 @@ siip_ospf
     | siipo_message_digest_key
     | siipo_network_p2p
     | siipo_cost
+    | siipo_hello_interval
+    | siipo_dead_interval
   )
   NEWLINE
 ;
@@ -98,7 +100,22 @@ siipo_cost
   COST interface_ospf_cost
 ;
 
+siipo_hello_interval
+:
+  HELLO_INTERVAL interface_ospf_interval
+;
+
+siipo_dead_interval
+:
+  DEAD_INTERVAL interface_ospf_interval
+;
+
 interface_ospf_cost
+:
+  uint16
+;
+
+interface_ospf_interval
 :
   uint16
 ;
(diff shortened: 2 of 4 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7018)* status: all files reviewed, 1 unresolved discussion (waiting on @pranavbj-amzn)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java, line 1084 at r1](https://reviewable.io/reviews/batfish/batfish/7018#-MaVcypE8RcHujQu2mo8:-MaVcypE8RcHujQu2mo9:b-fxvj3d) ([raw file](https://github.com/batfish/batfish/blob/979fcc80bfacc50dc2172b72440d08b0495cfb10/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java#L1084)):*
> ```Java
>   @Override
>   public void exitSiipo_hello_interval(Siipo_hello_intervalContext ctx) {
>     _currentInterface.getOrCreateOspf().setHelloInterval(toInteger(ctx.interface_ospf_interval()));
> ```

Recently we have made a move to be more restrictive with numbers parsed from configs like this.
In particular, we only want to allow valid values.

Take a look at this (slightly altered for didactic purposes) example from this file:
```
  private static final IntegerSpace PREFIX_LENGTH_SPACE =
      IntegerSpace.of(Range.closed(0, Prefix.MAX_PREFIX_LENGTH));
...
  private Optional<Integer> toInteger(ParserRuleContext messageCtx, Ip_prefix_lengthContext ctx) {
    return toIntegerInSpace(messageCtx, ctx, PREFIX_LENGTH_SPACE, "prefix length");
  }
...
  @Override
  public void exitRmmipa_prefix_len(Rmmipa_prefix_lenContext ctx) {
    Optional<Integer> maybeLen = toInteger(ctx, ctx.len);
    maybeLen.ifPresent(
        len ->
            _currentRouteMapEntry.setMatchIpAddressPrefixLen(
                new RouteMapMatchIpAddressPrefixLen(len)));
  }
```

What this does is only set the associated property if the value is within the allowed range. And it will emit a useful warning if it is out of range.

You can copy this pattern by defining an `IntegerSpace` constant as is done above for the valid range of hello-interval values, and copying the `ifPresent` pattern when you set the property.

Tab completion in `vtysh` on Cumulus FRR yields:
```
r2(config-if)# ip ospf hello-interval 
  (1-65535)  Seconds
r2(config-if)# ip ospf dead-interval 
  (1-65535)  Seconds
  minimal    Minimal 1s dead-interval with fast sub-second hellos
```
So these are the valid ranges for these properties.

Now you might be thinking, "Well I used `uint16` in the grammar, so why do I have to do extra validation?" But note that the valid range starts with `1`, not `0`, so `uint16` is actually too permissive.

So you should define add a converter function analogous to `toInteger(ParserRuleContext, Ip_prefix_lengthContext)` above, but for `Interface_ospf_intervalContext`.

Please make similar updates for handling of dead-interval.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7018)* status: all files reviewed, 2 unresolved discussions (waiting on @pranavbj-amzn)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7018#-MaVgBzT2k4n6DH3yH3q:-MaVgBzT2k4n6DH3yH3r:b-i1pclc) (no related file):*
FYI you can fix the formatting failure by running the following in the root of the repository:
`tools/fix_java_format.sh`

---

dhalperi:
Closing this after 2 months of no response. Please reopen if you want to pursue.

PR #7022, opened 2021-05-25
IOS XR: fix loopback OSPF cost calculation

sfraint:
For XR loopback interfaces, the OSPF cost is set to `1` unless explicitly configured. This also applies with set network type / `loopback stub-network enable`.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7022)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #7023, opened 2021-05-26
OSPF: calculate VLAN interface cost

sfraint:
Previously, `Vlan` interfaces were special-cased in VI `OspfProcess` to set OSPF cost to 1. This does not work for some vendors (like NXOS, where the [defaults on the device](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/unicast/configuration/guide/l3_cli_nxos/l3_ospf.html) result in an interface with cost of 40).

After this change, IOS devices (like Catalyst) should still have the correct default `Vlan` interface cost of 1 as the [default interface bandwidth is 1 Gbit](https://www.cisco.com/c/en/us/support/docs/lan-switching/inter-vlan-routing/41860-howto-L3-intervlanrouting.html) and IOS [default reference bandwidth is 100 Mbit](https://www.cisco.com/c/en/us/support/docs/ip/open-shortest-path-first-ospf/9237-9.html#qtwo).


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7023)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java, line 373 at r1](https://reviewable.io/reviews/batfish/batfish/7023#-Mae2gUjEw4fK2v_VXbj:-Mae2gUkDkMRVqwav3-G:b-nvbvuk) ([raw file](https://github.com/batfish/batfish/blob/aae30f035b0a012b8a23a4848112ef784e3b030a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java#L373)):*
> ```Java
>     String interfaceName = i.getName();
>     checkState(
>         i.getBandwidth() != null,
> ```

is it safe to assume VLAN interfaces always have bandwidth set? We weren't assuming this before. (I believe it's safe because we set all interfaces' bandwidths in a post-conversion step, but would be good to confirm)
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/ospf/OspfProcessTest.java, line 45 at r1](https://reviewable.io/reviews/batfish/batfish/7023#-Mae4ERk7EW7gdJzzGQu:-Mae4ERk7EW7gdJzzGQv:b4hcnvb) ([raw file](https://github.com/batfish/batfish/blob/aae30f035b0a012b8a23a4848112ef784e3b030a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/ospf/OspfProcessTest.java#L45)):*
> ```Java
>                 .setBandwidth(1e9)
>                 .setOwner(new Configuration("r1", ConfigurationFormat.CISCO_NX))
>                 .build()),
> ```

Not sure this assertion is necessary, since VLAN interfaces aren't special-cased anymore. Would rather see an NXOS unit test showing that a VLAN interface with minimal OSPF setup ends up with the correct default OSPF cost, as well as the same test in IOS if that doesn't already exist.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7023)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java, line 373 at r1](https://reviewable.io/reviews/batfish/batfish/7023#-Mae2gUjEw4fK2v_VXbj:-MaeD5bW-a3Hh3Z0TQhh:b-94we44) ([raw file](https://github.com/batfish/batfish/blob/aae30f035b0a012b8a23a4848112ef784e3b030a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java#L373)):*

is it safe to assume VLAN interfaces always have bandwidth set? We weren't assuming this before. (I believe it's safe because we set all interfaces' bandwidths in a post-conversion step, but would be good to confirm)
</blockquote></details>

We actually don't set any bandwidth in post-processing (presumably because each vendor has different defaults) but we do always set a bandwidth (at least in Cisco and NXOS and the other vendors I looked at) when building the VI configs.
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/ospf/OspfProcessTest.java, line 45 at r1](https://reviewable.io/reviews/batfish/batfish/7023#-Mae4ERk7EW7gdJzzGQu:-MaeDRK758bpdIjpTPvp:buvsusb) ([raw file](https://github.com/batfish/batfish/blob/aae30f035b0a012b8a23a4848112ef784e3b030a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/ospf/OspfProcessTest.java#L45)):*

Not sure this assertion is necessary, since VLAN interfaces aren't special-cased anymore. Would rather see an NXOS unit test showing that a VLAN interface with minimal OSPF setup ends up with the correct default OSPF cost, as well as the same test in IOS if that doesn't already exist.
</blockquote></details>

I was going for a minimal unit test of this function that failed before and passes now.

We already have (at least some) tests making sure the correct bandwidth and reference bandwidth values make it into the VI model, but I can add something a little more end-to-end-y.

corinaminer (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7023#-:-MaeaK5bAACWYaWEFek_:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java, line 373 at r1](https://reviewable.io/reviews/batfish/batfish/7023#-Mae2gUjEw4fK2v_VXbj:-Mae_Zol61U1GBn7T0gu:b-k0rkim) ([raw file](https://github.com/batfish/batfish/blob/aae30f035b0a012b8a23a4848112ef784e3b030a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java#L373)):*

We actually don't set any bandwidth in post-processing (presumably because each vendor has different defaults) but we do always set a bandwidth (at least in Cisco and NXOS and the other vendors I looked at) when building the VI configs.
</blockquote></details>

Ok. It'll at least be a clear error message if it does pop up.

PR #7029, opened 2021-05-27
BatfishLogger: only format string if level is active

dhalperi:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7029)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/BatfishLogger.java, line 346 at r1](https://reviewable.io/reviews/batfish/batfish/7029#-Maimg_JCCDOcpI-CVwH:-Maimg_K4rDMIHhSumLS:b-a93tex) ([raw file](https://github.com/batfish/batfish/blob/fb5f4beeadf64537a3d0be98d274c92d0fd55b34/projects/batfish-common-protocol/src/main/java/org/batfish/common/BatfishLogger.java#L346)):*
> ```Java
> 
>   private synchronized void writef(int level, @Nonnull String format, @Nonnull Object... args) {
>     if (!isActive(level)) {
> ```

Why are there separate checks in each `<level>f ` function if you do the check here?
Is the `write` call below even reachable given the call sites of this function?

Are the `LogManager` log levels orthogonal to the `BatfishLogger` log levels?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7029)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/BatfishLogger.java, line 346 at r1](https://reviewable.io/reviews/batfish/batfish/7029#-Maimg_JCCDOcpI-CVwH:-Mairf7-5hGYmTirIziM:b-fx7wxy) ([raw file](https://github.com/batfish/batfish/blob/fb5f4beeadf64537a3d0be98d274c92d0fd55b34/projects/batfish-common-protocol/src/main/java/org/batfish/common/BatfishLogger.java#L346)):*

Why are there separate checks in each `<level>f ` function if you do the check here?
Is the `write` call below even reachable given the call sites of this function?

Are the `LogManager` log levels orthogonal to the `BatfishLogger` log levels?
</blockquote></details>

Yes, the double logging and orthogonal. We might have to format the string for EITHER level, they are not constrained to be the same.

PR #7031, opened 2021-05-29
Fix Arista: neighbor default-originate route-map

corinaminer:
In `neighbor 1.2.3.4 default-originate route-map FOO` on Arista devices, `FOO` is used as an attribute policy for the generated default route, meaning:
> its set clauses are used to modify attributes of the exported default route, but its match clauses are not used to conditionally advertise the route. The default route is always advertised to the specified neighbor. ([source](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1116958))

In Batfish, this route-map is correctly set as the `attributePolicy` in the neighbor-specific `GeneratedRoute`, but it is also unnecessarily incorporated in the peer's export policy. This introduces two categories of bugs (though the second is minor):
- The route-map can block advertising the generated default route (should not be possible)
- The exported route may fail to adhere to some settings in the BGP neighbor/process. For example, if the route-map sets a community but `send-community` is not set for the neighbor,  the route should be sent without communities. However, the step to enforce `send-community` settings comes before the export policy in the Batfish pipeline, so Batfish sends the route with whatever communities the route-map sets.

Fixed this issue by replacing the export policy's call statement of the route-map with a `ReturnTrue` (i.e., the generated default route is now unconditionally permitted by the export policy).


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7031)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7038, opened 2021-06-02
Cumulus FRR assign LLAs to point-to-point OSPF interfaces without addresses

arifogel:
- enables forwarding across unnumbered links

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java
@@ -406,6 +406,7 @@ public class CumulusConcatenatedConfiguration extends VendorConfiguration {
   @VisibleForTesting
   static void populateLoopbackProperties(
       @Nullable InterfacesInterface vsLoopback, org.batfish.datamodel.Interface viLoopback) {
+    viLoopback.setInterfaceType(InterfaceType.LOOPBACK);
     if (vsLoopback != null && vsLoopback.getClagVxlanAnycastIp() != null) {
       // Just assume CLAG is correctly configured and comes up
       viLoopback.setAllAddresses(
--- a/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java
@@ -65,6 +65,7 @@ import org.batfish.datamodel.LinkLocalAddress;
 import org.batfish.datamodel.MacAddress;
 import org.batfish.datamodel.Prefix;
 import org.batfish.datamodel.SwitchportMode;
+import org.batfish.datamodel.ospf.OspfNetworkType;
 import org.batfish.datamodel.vendor_family.cumulus.CumulusFamily;
 import org.batfish.representation.cumulus.CumulusPortsConfiguration.PortSettings;
 import org.batfish.vendor.VendorConfiguration;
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7038)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java, line 226 at r1](https://reviewable.io/reviews/batfish/batfish/7038#-MbD29efC_5XKsvHf1S6:-MbD29efC_5XKsvHf1S7:bogd606) ([raw file](https://github.com/batfish/batfish/blob/f6dc904dbf182cfbea14a862fb59871caef2a565/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java#L226)):*
> ```Java
>         .forEach(
>             (iname, iface) -> {
>               if (iface.getAllAddresses().size() == 0
> ```

It feels like there is a missing `&& <some vs property>` that indicates this interface was configured for unnumbered.

Above, the function was labeled something like `isUsedForBgpUnnumbered(ifacename, bgp VS model)`. I imagine something generic there.
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java, line 226 at r1](https://reviewable.io/reviews/batfish/batfish/7038#-MbD2tXb0aMWwIwtB-SY:-MbD2tXb0aMWwIwtB-SZ:b-xs9wy2) ([raw file](https://github.com/batfish/batfish/blob/f6dc904dbf182cfbea14a862fb59871caef2a565/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java#L226)):*
> ```Java
>         .forEach(
>             (iname, iface) -> {
>               if (iface.getAllAddresses().size() == 0
> ```

From discussion: seems like we should add an LLA unconditionally? Otherwise there might not be an L3 edge.

Also from discussion: it seems like there should not even be able to be a concrete address assigned. Unnumbered should be mutually exclusive with concrete.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7038)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7038#-MbDLG4T7juJxjeIs4wC:-MbDLG4U2-8-myprqYRS:bkpqc80) (no related file):*
After batfish/batfish#7039

---

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7038)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java, line 226 at r1](https://reviewable.io/reviews/batfish/batfish/7038#-MbD2tXb0aMWwIwtB-SY:-MbDLQPaAAbMaqXrDlgh:bwdh87b) ([raw file](https://github.com/batfish/batfish/blob/f6dc904dbf182cfbea14a862fb59871caef2a565/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java#L226)):*

From discussion: seems like we should add an LLA unconditionally? Otherwise there might not be an L3 edge.

Also from discussion: it seems like there should not even be able to be a concrete address assigned. Unnumbered should be mutually exclusive with concrete.
</blockquote></details>

I don't believe we said they should be mutually exclusive. An interface could win one unnumbered and address and lose another. In that case it would have a concrete address, but still may need the LLA for L3 edge.

In any case, now adding the LLA even for winning interface.
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java, line 226 at r1](https://reviewable.io/reviews/batfish/batfish/7038#-MbD29efC_5XKsvHf1S6:-MbDLciF-3zEklmmR_k2:b-6tevt6) ([raw file](https://github.com/batfish/batfish/blob/f6dc904dbf182cfbea14a862fb59871caef2a565/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java#L226)):*

It feels like there is a missing `&& <some vs property>` that indicates this interface was configured for unnumbered.

Above, the function was labeled something like `isUsedForBgpUnnumbered(ifacename, bgp VS model)`. I imagine something generic there.
</blockquote></details>

Now relying on ospfAddresses being non-empty. This happens when at least 1 address was assigned in VS interfaces or frr file.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7038)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7040, opened 2021-06-02
FRR: ignore invalid bandwidths, if present in runtime data

dhalperi:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7040)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java, line 568 at r1](https://reviewable.io/reviews/batfish/batfish/7040#-MbD_oXT06UMGcrElHFr:-MbD_oXT06UMGcrElHFs:b-6a5dsm) ([raw file](https://github.com/batfish/batfish/blob/d4f6180d5dd8c09860fc2d19122fcb64712053fa/projects/batfish/src/main/java/org/batfish/representation/cumulus/CumulusConcatenatedConfiguration.java#L568)):*
> ```Java
>               value, iface));
>       return false;
>     } else if (value > 1000e9) {
> ```

nit: personally I would prefer to put this in a named constant.

PR #7043, opened 2021-06-03
ibdp/Node: use Optional#flatMap correctly

dhalperi:
We accidentally made an `Optional<Optional<Rib>>` that could be absent, or
present with a Rib, or present with an absent Rib. We meant to chain the
optionality, which is done with `flatMap`.


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7043)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7052, opened 2021-06-07
Arista: fix handling of undefined continue targets

dhalperi:
* If continue target is undefined, the continue line is ignored
* If the continue target is before this term, then the continue line is ignored
* Also move reference tracking to parser, and add warnings

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java
@@ -36,6 +36,7 @@ import static org.batfish.representation.arista.AristaStructureType.POLICY_MAP;
 import static org.batfish.representation.arista.AristaStructureType.PREFIX6_LIST;
 import static org.batfish.representation.arista.AristaStructureType.PREFIX_LIST;
 import static org.batfish.representation.arista.AristaStructureType.ROUTE_MAP;
+import static org.batfish.representation.arista.AristaStructureType.ROUTE_MAP_ENTRY;
 import static org.batfish.representation.arista.AristaStructureType.SERVICE_TEMPLATE;
 import static org.batfish.representation.arista.AristaStructureType.TRACK;
 import static org.batfish.representation.arista.AristaStructureType.VXLAN;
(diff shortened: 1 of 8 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7052)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java, line 4096 at r1](https://reviewable.io/reviews/batfish/batfish/7052#-MbceuL59bZyC9jHn2m7:-MbceuL59bZyC9jHn2m8:b-ms5kmt) ([raw file](https://github.com/batfish/batfish/blob/994be46d83f6b0bb2a7b6e482ddbed9cf3362be1/projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java#L4096)):*
> ```Java
>             String.format(
>                 "Route-map %s entry %d: continue %d introduces a loop",
>                 _currentRouteMap.getName(), _currentRouteMapClause.getSeqNum(), target));
> ```

the entries are applied in numerical order, not order of appearance? (or at least `show run` always orders them numerically?)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7052)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java, line 4096 at r1](https://reviewable.io/reviews/batfish/batfish/7052#-MbceuL59bZyC9jHn2m7:-MbcmUwh4H6jE1qyP2Vq:b-9ulmcy) ([raw file](https://github.com/batfish/batfish/blob/994be46d83f6b0bb2a7b6e482ddbed9cf3362be1/projects/batfish/src/main/java/org/batfish/grammar/arista/AristaControlPlaneExtractor.java#L4096)):*

the entries are applied in numerical order, not order of appearance? (or at least `show run` always orders them numerically?)
</blockquote></details>

Yes and yes.

PR #7054, opened 2021-06-08
dataplane: parallelize across virtual routers, not nodes

dhalperi:
1. Streaming `nodes.values()` will always chunk similar nodes together.
   This means imbalanced work, as in the common case nodes with similar
   names have similar complexity (number of routes, etc).
2. `parallelStream().flatMap()` does not parallelize over the flat-mapped
   things.

To improve both of these, materialize a randomized list of all virtual
routers. This means that we can parallelize over the entire list, and
remove the name-based locality during parallelization.

After #7053

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/StreamUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/StreamUtil.java
@@ -1,0 +1,29 @@
+package org.batfish.common.util;
+
+import com.google.common.collect.ImmutableList;
+import java.util.Collections;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7054)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @corinaminer)

dhalperi (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7054)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7056, opened 2021-06-09
Fix BGP export bug; never set next hop IP to 0.0.0.0

corinaminer:
Fixes a bug where BGP aggregate routes could be exported with NHIP 0.0.0.0 and thus dropped upon import as unroutable. The bug would surface in an IBGP context where the original route had NHIP 0.0.0.0. The logic to update the next hop IP looks like this:

```
    // Skip setting our own next hop if it has already been set by the routing policy
    if (routeBuilder.getNextHopIp().equals(UNSET_ROUTE_NEXT_HOP_IP)) {
      if (isEbgp) {
        routeBuilder.setNextHopIp(localIp);
      } else { // iBGP session
        /*
        Note: implementation of next-hop-self in the general case is delegated to routing
        policy.
        If original route has next-hop ip, preserve it. If not, set our own.
        */
        routeBuilder.setNextHopIp(
            originalRouteNhip.equals(UNSET_ROUTE_NEXT_HOP_IP) ? localIp : originalRouteNhip);
      }
    }
```

See updates in `XrGrammarTest.testBgpAggregateWithLearnedSuppressedRoutes` for a demonstration of the bug (wasn't already occurring because that test was previously EBGP-only).

Not strictly necessary to impose the NHIP ≠ 0.0.0.0 invariant in this PR, but it doesn't seem to mess anything up and will help prevent issues like this in the future.


dhalperi (commented):
> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7056)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 263 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiGGdRCRWiwdanUvAi:-MbiGGdRCRWiwdanUvAj:b-bhh7un) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L263)):*
> ```
> Ip nextHopIp
> ```
Should this just be a `NextHop`, and the logic can go in the caller?

If pushed to the below function, would also get rid of duplicate logic (that is also not fully tested in both paths).
___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 287 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiGRv0401LbTSaH9Em:-MbiGRv0401LbTSaH9En:bek0nqa) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L287)):*
> ```
> Ip nextHopIp
> ```
here, I guess?
___
*[projects/batfish/src/test/java/org/batfish/dataplane/protocols/BgpProtocolHelperTransformBgpRouteOnExportTest.java, line 178 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiHgUO971q9k7_Wsqu:-MbiHgUO971q9k7_Wsqv:b7j6xzw) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/projects/batfish/src/test/java/org/batfish/dataplane/protocols/BgpProtocolHelperTransformBgpRouteOnExportTest.java#L178)):*
> ```
> // Can't use non-concrete IPs for local IP or original route NHIP because both may be used
> ```
nit: unnecessary in code. I probably would have FYI'ed the review.
___
*[tests/basic/routes.ref, line 475 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiI0VA30jd_oFILhSh:-MbiI0VA30jd_oFILhSi:b8vjar4) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/tests/basic/routes.ref#L475)):*
> ```
>           "name" : "as2core2"
>         },
>         "VRF" : "default",
> ```

FYI this is going to change things downstream, i think.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7056)* status: 13 of 15 files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 263 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiGGdRCRWiwdanUvAi:-Mbj2M9fEdoMq4jGULrp:bcne0a6) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L263)):*

> ```
> Ip nextHopIp
> ```
Should this just be a `NextHop`, and the logic can go in the caller?

If pushed to the below function, would also get rid of duplicate logic (that is also not fully tested in both paths).
</blockquote></details>

Yep, good call
___
*[projects/batfish/src/test/java/org/batfish/dataplane/protocols/BgpProtocolHelperTransformBgpRouteOnExportTest.java, line 178 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiHgUO971q9k7_Wsqu:-Mbj2TJbC5-1LN9Ab5QP:b1o2hm8) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/projects/batfish/src/test/java/org/batfish/dataplane/protocols/BgpProtocolHelperTransformBgpRouteOnExportTest.java#L178)):*

> ```
> // Can't use non-concrete IPs for local IP or original route NHIP because both may be used
> ```
nit: unnecessary in code. I probably would have FYI'ed the review.
</blockquote></details>

fair, cleaned up
___
*[tests/basic/routes.ref, line 475 at r1](https://reviewable.io/reviews/batfish/batfish/7056#-MbiI0VA30jd_oFILhSh:-Mbj2aPHEgj9x-k6-Cq4:b-o5fi9n) ([raw file](https://github.com/batfish/batfish/blob/0737a9643eea3e9eab74a63a80d33b4801a439ef/tests/basic/routes.ref#L475)):*

FYI this is going to change things downstream, i think.
</blockquote></details>

ack. i'll hold off merging until i've scoped it out.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7056)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7059, opened 2021-06-09
XR: remove unused routing policy from VS ifaces

sfraint:
Routing policy attached to XR interfaces isn't populated anywhere, I suspect it was just copied from IOS.

Remove it since it is unused (and if it were used in the future, it would conflict with ACL based forwarding policies).


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7059)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7065, opened 2021-06-14
SearchRoutePolicies: Better handling of conditionals

millstein:
Updates the handling of if statements in the symbolic route-map analysis so that it only analyzes feasible branches.  Some guards are statically resolved (notably CallExprContext and CallStatementContext), so analyzing both branches is unnecessary.  Skipping analysis of the infeasible branch also avoids a possible failure to answer the question due to reaching route-map features not currently supported.  Further, it allows us to properly account for updates to things like the default action within the one feasible branch, which previously were being ignored.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/TransferParam.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/TransferParam.java
@@ -134,6 +134,12 @@ public class TransferParam<T extends IDeepCopy<T>> {
     return ret;
   }
 
+  public TransferParam<T> setDefaultsFrom(TransferParam<T> updatedParam) {
+    return setDefaultAccept(updatedParam._defaultAccept)
+        .setDefaultPolicy(updatedParam._defaultPolicy)
+        .setDefaultAcceptLocal(updatedParam._defaultAcceptLocal);
+  }
+
   public TransferParam<T> enterScope(String name) {
     TransferParam<T> ret = new TransferParam<>(this);
     ret._scopes = ret._scopes.plus(name);
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
@@ -510,7 +510,7 @@ public class TransferBDD {
       // Symbolically execute each branch if it is feasible.
       // Some guards are statically resolved (e.g. CallExprContext and StatementExprContext), which
       // means that only one branch will be analyzed.  Skipping analysis of the other branch avoids
-      // causing an error unnecessarily if we reach a route-map construct that is not currently
+      // signaling an error unnecessarily if we reach a route-map construct that is not currently
       // modelled.  It also allows us to properly account for updates to things like the default
       // action that occur within the one feasible branch (see below for more on this).
       if (!guard.isZero()) {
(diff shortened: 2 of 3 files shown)


arifogel (commented):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7065)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1033 at r2](https://reviewable.io/reviews/batfish/batfish/7065#-McGGVMY7C3wwtuBW_3I:-McGGVMY7C3wwtuBW_3J:b5x80u2) ([raw file](https://github.com/batfish/batfish/blob/89c7aae7e1cb7fafd6ad3e47a67c5ad733ca50c5/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1033)):*
> ```Java
> 
>   @Test
>   public void testConditionalDefaultAction() {
> ```

Was this test failing before? If so, what was preventing analysis of the false branch?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7065)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1033 at r2](https://reviewable.io/reviews/batfish/batfish/7065#-McGGVMY7C3wwtuBW_3I:-McGKchn6fBzT1cv6sFp:b-9merw4) ([raw file](https://github.com/batfish/batfish/blob/89c7aae7e1cb7fafd6ad3e47a67c5ad733ca50c5/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1033)):*

Was this test failing before? If so, what was preventing analysis of the false branch?
</blockquote></details>

Yes it was failing before because both branches would be analyzed but the update to the default action in the false branch would be ignored when producing the state of the analysis after the "if".  Therefore when reaching the second statement, we would think that the default action is still to reject.

That is still a limitation today, but in the special case where only one branch is feasible we now only analyze that branch and then properly reflect the update to the default action in the resulting analysis state.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7065)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1033 at r2](https://reviewable.io/reviews/batfish/batfish/7065#-McGGVMY7C3wwtuBW_3I:-McGUCPT-9dMa_qBl9RS:bbhokhx) ([raw file](https://github.com/batfish/batfish/blob/89c7aae7e1cb7fafd6ad3e47a67c5ad733ca50c5/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1033)):*

Yes it was failing before because both branches would be analyzed but the update to the default action in the false branch would be ignored when producing the state of the analysis after the "if".  Therefore when reaching the second statement, we would think that the default action is still to reject.

That is still a limitation today, but in the special case where only one branch is feasible we now only analyze that branch and then properly reflect the update to the default action in the resulting analysis state.
</blockquote></details>

Right, I forget you added the defaults stuff in this PR.

Regarding the other issue, can you add a test for that? E.g. that when one branch is infeasible and also has stuff that can't be analyzed currently, the other branch is still analyzed?

arifogel (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7065)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

arifogel (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7065#-:-McGzRcV08Uli3plR4dg:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7067, opened 2021-06-15
IOS-XR and VI support for BGP aggregates activated from BGP RIB

arifogel:
- add VI/iBDP support for activating aggregate-address via routes
present in BGP RIB rather than main RIB
- add IOS-XR conversion to support the same
- currently support: summary-only
- TODO:
  - IOS-XR aggregate-address options:
    - route-policy
    - as-set
  - VI support for transformation of aggregates based on properties
    of contributors


corinaminer (commented):
> Reviewed 5 of 11 files at r1, 2 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 6 of 17 files reviewed, 11 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1167 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGb54r0yfE4VzMhP38:-McGb54r0yfE4VzMhP39:b-hvnpjz) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1167)):*
> ```Java
>    */
>   void initBgpAggregateRoutes() {
>     if (!_exportFromBgpRib) {
> ```

Would it make more sense to use `if (_aggregates.isEmpty())` here? That should always be true if `!_exportFromBgpRib`, and sometimes true otherwise (for Cisco-style devices with no aggregates)
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1179 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGfXbl72ARz9zZWLSg:-McGfXbl72ARz9zZWLSh:b8oxrik) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1179)):*
> ```Java
>     for (Bgpv4Route potentialContributor : currentRoutes) {
>       Prefix contribNet = potentialContributor.getNetwork();
>       if (contribNet.equals(Prefix.ZERO)) {
> ```

Should it also skip aggregate routes? That would make sense as long as aggregates don't overlap.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1184 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGg0cA-mboR3JJDU7t:-McGg0cA-mboR3JJDU7u:b-o62je7) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1184)):*
> ```Java
>       BgpAggregate aggregate =
>           _aggregates
>               .longestPrefixMatch(contribNet.getStartIp(), contribNet.getPrefixLength() - 1)
> ```

if we don't think aggregates can ever overlap, might be nice to assert this has at most one element. would refactor something like:
```
aggs = _aggregates.lpm(...)
if aggs is empty, continue
assert aggs.size() == 1
agg = aggs.get(0)
```

But if they can overlap, then we have to activate all aggregates in the set, not just the first.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1201 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGqBt76JKgD5e-kqxq:-McGqBt76JKgD5e-kqxr:bebgm4z) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1201)):*
> ```Java
>       }
>     }
>     // TODO: use local BGP cost instead once available
> ```

what do you mean? isn't `getAdminCost` what we want to be using here?
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGrB6q83yUmekWEXvJ:-McGrB6q83yUmekWEXvK:b6ibu2b) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*
> ```Java
>                 _bgpv4DeltaBuilder.from(_bgpv4Rib.mergeRouteGetDelta(action.getRoute()));
>               }
>             });
> ```

you can just do `_bgpv4DeltaBuilder.from(aggDeltaBuilder.build())`, i believe
___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 327 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGrqxn6Sgn43szN52-:-McGrqxn6Sgn43szN520:b-sclcq3) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L327)):*
> ```Java
>             /*
>              * Note: Origin type and originator IP should get overwritten by export policy,
>              * but are needed initially
> ```

this note isn't needed; now that the returned route is going into the BGP RIB rather than being immediately exported, naturally we need to set these properties here.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java, line 817 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGuC1y3nhU214AhjyC:-McGuC1y3nhU214AhjyD:bgxmcsl) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java#L817)):*
> ```Java
>       RoutingPolicy genPolicy = generateGenerationPolicy(c, vrfName, prefix6);
>       // Should be local bgp admin, but defaults are the same and this code will disappear when IPv6
>       // is implemented.
> ```

this whole section is pretty wrong now. i'm tempted to say just delete it. seems absurd to add a 
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java, line 276 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGuogi2AEYFRcJFTM2:-McGuogi2AEYFRcJFTM3:b-54hd56) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java#L276)):*
> ```Java
>   @JsonProperty(PROP_AGGREGATES)
>   private void setAggregates(List<BgpAggregate> aggregates) {
>     aggregates.forEach(this::addAggregate);
> ```

this is unintuitive, i would expect `setAggregates` to clear existing aggregates. plus this approach is going to create one immutable map per aggregate (and throw away all but the last one).

i realize this is a private method for json deserialization only, but why not `_aggregates = toImmutableMap(aggregates, ...)`?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 48 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGiDwF3T-JgjT-Ev3T:-McGiDwF3T-JgjT-Ev3U:b5j2lm1) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L48)):*
> ```Java
>    * that are rejected by the policy should be discarded. This policy also can additionally suppress
>    * or unsuppress contributing routes. However it cannot modify the suppressed status of
>    * non-contributing routes.
> ```

please separate this comment into a description of what the generation policy currently does (limit which potential contributors can activate an aggregate) and what it should eventually do (modify aggregate, un/suppress contributors)
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 69 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGjTkW4F_hxdkICbwx:-McGjTkX7awkXIZWu0Lb:b-t0yfn1) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L69)):*
> ```Java
>    * This policy is used to determine which potential contributors are nominally suppressed. The
>    * suppressed status determined by this policy may be overridden by generation policy, but only
>    * when {@link #getGenerationPolicy} accepts the potential contributor.
> ```

to clarify -- a potential contributor that matches the suppression policy will be suppressed even if it doesn't match the generation policy (and therefore doesn't activate the aggregate)? Is this still true if the aggregate isn't activated at all?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 74 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGl9KjB2mlPi198UOb:-McGl9KjB2mlPi198UOc:bttp7lg) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L74)):*
> ```Java
>    */
>   @JsonProperty(PROP_SUPPRESSION_POLICY)
>   public @Nullable String getSuppressionPolicy() {
> ```

given that we only support `summary-only` suppression right now, i probably would've leaned towards a `suppressContributors` boolean for the first pass on this, but we'll need a policy for it eventually so this is fine
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 99 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGlr3204qqoz0dqVuE:-McGlr3204qqoz0dqVuF:b-saqls) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L99)):*
> ```Java
>   @Override
>   public String toString() {
>     return MoreObjects.toStringHelper(this)
> ```

Add `.omitNullValues()`. Also we usually use the json PROPs for the field names (e.g. `add(PROP_ATTRIBUTE_POLICY, _attributePolicy)`)

corinaminer (commented):
> Reviewed 3 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 9 of 17 files reviewed, 11 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java, line 817 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGuC1y3nhU214AhjyC:-McGyt2REbo7ebId3MMe:b-ud8c2l) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java#L817)):*

this whole section is pretty wrong now. i'm tempted to say just delete it. seems absurd to add a 
</blockquote></details>

...field for IPv6 aggregates to `BgpProcess` at this point

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 9 of 17 files reviewed, 10 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1167 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGb54r0yfE4VzMhP38:-McGzoOo2xujkWnfEQfP:b-rf0vfu) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1167)):*

Would it make more sense to use `if (_aggregates.isEmpty())` here? That should always be true if `!_exportFromBgpRib`, and sometimes true otherwise (for Cisco-style devices with no aggregates)
</blockquote></details>

There is no such function, and I'm loathe to update the `PrefixTrieMultimap` API here.
Instead, recorded a boolean at construction time that is the same thing.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1179 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGfXbl72ARz9zZWLSg:-McH4GY28q-iJAmKj-oK:b-r3m0tj) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1179)):*

Should it also skip aggregate routes? That would make sense as long as aggregates don't overlap.
</blockquote></details>

no, because aggregates are perfectly valid contributors to other aggregates. confirmed on device.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1184 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGg0cA-mboR3JJDU7t:-McH-vDF1pAeYpQq_kz2:b-t610k8) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1184)):*

if we don't think aggregates can ever overlap, might be nice to assert this has at most one element. would refactor something like:
```
aggs = _aggregates.lpm(...)
if aggs is empty, continue
assert aggs.size() == 1
agg = aggs.get(0)
```

But if they can overlap, then we have to activate all aggregates in the set, not just the first.
</blockquote></details>

It can already only have one element because the `PrefixTrieMultimap` is populated from a set of prefixes. 
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1201 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGqBt76JKgD5e-kqxq:-McH09Uj1XffEfl0rbhF:b365tva) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1201)):*

what do you mean? isn't `getAdminCost` what we want to be using here?
</blockquote></details>

There are 3 different configurable admin costs in a BGP process: eBGP, iBGP, and local. We don't currently support changing admin of local routes. Different vendors variously consider things like aggregates, redistributed routes, and vrf leaked routes to be either iBGP or local (at least wrt admin distance). You can see this when you use non-default values.

By default, local and iBGP have the same admin distance (200), so this is not obvious at first glance. I strongly suspect that aggregates
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGrB6q83yUmekWEXvJ:-McH0Y7e8dgfbMD2eln_:b2ku9ma) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*

you can just do `_bgpv4DeltaBuilder.from(aggDeltaBuilder.build())`, i believe
</blockquote></details>

Yes, but then I'd want an assertion that that delta is the same as the delta that was applied to the BGP RIB, and then nothing has been saved.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 327 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGrqxn6Sgn43szN52-:-McH0jrPA9zx6jB9PApI:b-hfpv0l) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L327)):*

this note isn't needed; now that the returned route is going into the BGP RIB rather than being immediately exported, naturally we need to set these properties here.
</blockquote></details>

removed
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java, line 817 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGuC1y3nhU214AhjyC:-McH0r8s7bcT6S_s3ix1:bjmbn12) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConfiguration.java#L817)):*

...field for IPv6 aggregates to `BgpProcess` at this point
</blockquote></details>

Fair enough. Gone.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java, line 276 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGuogi2AEYFRcJFTM2:-McH33T--qNItEYys30i:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java#L276)):*

this is unintuitive, i would expect `setAggregates` to clear existing aggregates. plus this approach is going to create one immutable map per aggregate (and throw away all but the last one).

i realize this is a private method for json deserialization only, but why not `_aggregates = toImmutableMap(aggregates, ...)`?
</blockquote></details>

done
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 99 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGlr3204qqoz0dqVuE:-McH3oyD4g4kYnjWrn63:b7ctx3k) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L99)):*

Add `.omitNullValues()`. Also we usually use the json PROPs for the field names (e.g. `add(PROP_ATTRIBUTE_POLICY, _attributePolicy)`)
</blockquote></details>

This is a generated debugging convenience function that affects nothing, but sure.

corinaminer (commented):
> Reviewed 2 of 11 files at r1, 2 of 7 files at r2, 4 of 7 files at r3, 4 of 5 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: all files reviewed, 7 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1167 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGb54r0yfE4VzMhP38:-McHEVY1CHpkRMDtdxs8:b12jrsh) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1167)):*

There is no such function, and I'm loathe to update the `PrefixTrieMultimap` API here.
Instead, recorded a boolean at construction time that is the same thing.
</blockquote></details>

could just use `_process.getAggregates().isEmpty()`, but this is ok too.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1201 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGqBt76JKgD5e-kqxq:-McHEmuz8Yy9snLQRKUh:b-88zkrw) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1201)):*

There are 3 different configurable admin costs in a BGP process: eBGP, iBGP, and local. We don't currently support changing admin of local routes. Different vendors variously consider things like aggregates, redistributed routes, and vrf leaked routes to be either iBGP or local (at least wrt admin distance). You can see this when you use non-default values.

By default, local and iBGP have the same admin distance (200), so this is not obvious at first glance. I strongly suspect that aggregates
</blockquote></details>

you probably had something more to say here, but i'm convinced
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGrB6q83yUmekWEXvJ:-McHF2-o2LNkprWXqMqm:b-cqsugj) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*

Yes, but then I'd want an assertion that that delta is the same as the delta that was applied to the BGP RIB, and then nothing has been saved.
</blockquote></details>

Oh whoops, totally missed that you were merging the changes into the bgp rib here.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 343 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH1hwQ1Tpq0DsExuvy:-McH1hwQ1Tpq0DsExuvz:blbxzlg) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L343)):*
> ```Java
>       return routePolicy;
>     }
>     String name = generateAggregateInheritPolicyName(routePolicy);
> ```

for now this policy always has the exact same effect as just applying `routePolicy` directly, right? I'd just use `routePolicy` directly, with a TODO to implement as-set and generate a customized generation policy. unclear when we'll be implementing support for as-set.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 351 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH3g8A4dtC5zfywY_o:-McH3g8A4dtC5zfywY_p:brvlfb1) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L351)):*
> ```Java
>         .setName(name)
>         .setOwner(c)
>         .addStatement(new CallStatement(routePolicy))
> ```

what does this do if `routePolicy` is null?
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-aggregate-address, line 23 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH3FTFB5LlVCWbqGRp:-McH3FTFB5LlVCWbqGRq:b684szo) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-aggregate-address#L23)):*
> ```
>   aggregate-address 2.2.0.0/16 summary-only route-policy gen2
>   aggregate-address 3.1.0.0/16 as-set
>   aggregate-address 3.2.0.0/16 as-set route-policy gen3
> ```

are `summary-only` and `as-set` mutually exclusive?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 99 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGlr3204qqoz0dqVuE:-McHGOIODPYhDTHSMdK2:b-e0jncc) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L99)):*

This is a generated debugging convenience function that affects nothing, but sure.
</blockquote></details>

hey, i like debugging to be convenient ¯\_(ツ)_/¯
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java, line 153 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH0apTA4IzIlIgiCdB:-McH0apTA4IzIlIgiCdC:b-66nzgw) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L153)):*
> ```Java
>   public static String generateAggregateInheritPolicyName(@Nullable String contributorPolicyName) {
>     return String.format(
>         "~AGG_INHERIT~%s", contributorPolicyName != null ? contributorPolicyName : "");
> ```

nit: would prefer generated routing policy names to consistently end with a tilde
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/BgpAggregateTest.java, line 34 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH0ryuDDmgojUS666A:-McH0ryuDDmgojUS666B:bnhgnyr) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/BgpAggregateTest.java#L34)):*
> ```Java
>         .addEqualityGroup(BgpAggregate.of(Prefix.ZERO, "a", null, null))
>         .addEqualityGroup(BgpAggregate.of(Prefix.ZERO, "a", "b", null))
>         .addEqualityGroup(BgpAggregate.of(Prefix.ZERO, "a", "b", "c"))
> ```

missing a group with a different prefix

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: all files reviewed, 8 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7067#-McLOfvEBWF3APrCaOq2:-McLOfvEBWF3APrCaOq3:bxulbom) (no related file):*
Need to fix aggregate activation per offline discussion

---
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1167 at r1](https://reviewable.io/reviews/batfish/batfish/7067#-McGb54r0yfE4VzMhP38:-McLNpKpDhQQDo1J8boi:bugpcsy) ([raw file](https://github.com/batfish/batfish/blob/549e80d5b06886f75885b7da3601d250c6f133af/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1167)):*

could just use `_process.getAggregates().isEmpty()`, but this is ok too.
</blockquote></details>

I considered that, but I was trying to avoid referring to `BgpProcess` after constructor. But it turns out we actually do that it in a bunch of places.
Since the field doesn't really save much, switched to your version.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 343 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH1hwQ1Tpq0DsExuvy:-McLLqeRA36WWSleqDfm:banigyo) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L343)):*

for now this policy always has the exact same effect as just applying `routePolicy` directly, right? I'd just use `routePolicy` directly, with a TODO to implement as-set and generate a customized generation policy. unclear when we'll be implementing support for as-set.
</blockquote></details>

left some skeleton in place, but no longer generating policy.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 351 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH3g8A4dtC5zfywY_o:-McLLtHR4dbocaFmNf8L:b-2yiccj) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L351)):*

what does this do if `routePolicy` is null?
</blockquote></details>

Good catch!

Exits false:
```
    if (policy == null) {
      environment.setError(true);
      return Result.builder().setBooleanValue(false).build();
    }
```

Moot now, since removed.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-aggregate-address, line 23 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH3FTFB5LlVCWbqGRp:-McLMDPJC_7ek5cfVJcs:b-88rury) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/test/resources/org/batfish/grammar/cisco_xr/testconfigs/xr-aggregate-address#L23)):*

are `summary-only` and `as-set` mutually exclusive?
</blockquote></details>

They are completely orthogonal, so I didn't feel the need to try combinations.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 48 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGiDwF3T-JgjT-Ev3T:-McLMOfq6Rdfc0Sj2UDA:bj4fvy0) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L48)):*

please separate this comment into a description of what the generation policy currently does (limit which potential contributors can activate an aggregate) and what it should eventually do (modify aggregate, un/suppress contributors)
</blockquote></details>

How's this?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 69 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGjTkW4F_hxdkICbwx:-McLMWCw2GvzKBLuLcNa:bq7hp05) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L69)):*

to clarify -- a potential contributor that matches the suppression policy will be suppressed even if it doesn't match the generation policy (and therefore doesn't activate the aggregate)? Is this still true if the aggregate isn't activated at all?
</blockquote></details>

Nothing is ever suppressed by an inactive aggregate.
But a route may be suppressed (by passing suppression policy) without being a contributor (contributor means a route passes generation policy) if the aggregate is activated by some other route.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java, line 153 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH0apTA4IzIlIgiCdB:-McLNRKkCoQ8huMRu9jM:bw97m8g) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L153)):*

nit: would prefer generated routing policy names to consistently end with a tilde
</blockquote></details>

moot, since changed to identity for now
___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/BgpAggregateTest.java, line 34 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH0ryuDDmgojUS666A:-McLNVjb226e_zuJ85XX:b-adf9wd) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/BgpAggregateTest.java#L34)):*

missing a group with a different prefix
</blockquote></details>

how embarrassing, fixed

corinaminer (commented):
> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: all files reviewed, 5 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 343 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH1hwQ1Tpq0DsExuvy:-McLUYVV0DJzzmoBIqdH:bnno1th) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L343)):*

left some skeleton in place, but no longer generating policy.
</blockquote></details>

i think `generateAggregateInheritPolicyName` should just be deleted for now -- it's too confusing for a name-generation function to be nullable (and it's not marked as such).
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 348 at r6](https://reviewable.io/reviews/batfish/batfish/7067#-McLVm261IpRqQsaUdYO:-McLVm279hKodTxgMFDX:b9i4w05) ([raw file](https://github.com/batfish/batfish/blob/ebe179e44868eb7e0c72c051b19e3b83530d6b6d/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L348)):*
> ```Java
>     }
>     // TODO: implement as-set, make generateAggregateInheritPolicyName not be identity.
>     throw new NotImplementedException();
> ```

Should never throw in conversion because of an unsupported feature. At most, file a warning about as-set if we're not already doing that in extraction, but this function should always return a valid or null generation policy. I think it's just going to boil down to `return routePolicy` for now
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 48 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGiDwF3T-JgjT-Ev3T:-McLTjW06ltZ8ygp-IoV:bjtm9te) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L48)):*

How's this?
</blockquote></details>

works for me
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 69 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGjTkW4F_hxdkICbwx:-McLTrlw9AqaYhNfCxTr:b-7t6oc6) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L69)):*

Nothing is ever suppressed by an inactive aggregate.
But a route may be suppressed (by passing suppression policy) without being a contributor (contributor means a route passes generation policy) if the aggregate is activated by some other route.
</blockquote></details>

ugh. once we implement suppression policy in dataplane we'll have to be careful to avoid suppressing routes if the aggregate isn't activated.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: all files reviewed, 5 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 343 at r3](https://reviewable.io/reviews/batfish/batfish/7067#-McH1hwQ1Tpq0DsExuvy:-McLkJnD4et_QNkV0XIZ:b-7zx38l) ([raw file](https://github.com/batfish/batfish/blob/e4a5cb594a79ab4eb03c31cc85ca99a2b8b83d22/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L343)):*

i think `generateAggregateInheritPolicyName` should just be deleted for now -- it's too confusing for a name-generation function to be nullable (and it's not marked as such).
</blockquote></details>

fine fine, removed it all
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 348 at r6](https://reviewable.io/reviews/batfish/batfish/7067#-McLVm261IpRqQsaUdYO:-McLkL6X08WKercb_WpH:b-3sdpja) ([raw file](https://github.com/batfish/batfish/blob/ebe179e44868eb7e0c72c051b19e3b83530d6b6d/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L348)):*

Should never throw in conversion because of an unsupported feature. At most, file a warning about as-set if we're not already doing that in extraction, but this function should always return a valid or null generation policy. I think it's just going to boil down to `return routePolicy` for now
</blockquote></details>

moot, since removed
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java, line 69 at r2](https://reviewable.io/reviews/batfish/batfish/7067#-McGjTkW4F_hxdkICbwx:-McLkTm6DFlVVxLMYzyv:bm3tww1) ([raw file](https://github.com/batfish/batfish/blob/f5967d30778368bca9fb9abaf01966561c9e6cb8/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpAggregate.java#L69)):*

ugh. once we implement suppression policy in dataplane we'll have to be careful to avoid suppressing routes if the aggregate isn't activated.
</blockquote></details>

I have an idea for that, but definitely way out of scope for now.

corinaminer (commented):
> Reviewed 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel, @corinaminer, and @dhalperi)

corinaminer (commented):
:lgtm: after one more test. other comments are all nonblocking nits

> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067#-:-McNHjW-6NKr5FvTNFvE:b-8hqh8j)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNCLaD3cXce1IYTHNT:-McNCLaD3cXce1IYTHNU:b-p2gchd) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*
> ```Java
>           Collection<Bgpv4Route> potentialContributors =
>               potentialContributorsByAggregatePrefix.get(aggNet);
>           BgpAggregate moreGeneralAggregate = getMostSpecificAggregate(aggNet).orElse(null);
> ```

just for legibility, could you move this line down to where it's used? like right above `if (moreGeneralAggregate != null)`
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1297 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNAqtbDQipzaENAbFr:-McNAqtbDQipzaENAbFs:br7jl81) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1297)):*
> ```Java
>             potentialContributingPrefix.getPrefixLength() - 1)
>         .stream()
>         .findFirst();
> ```

perhaps add a comment as to why at most one can match
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 757 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNGusq464APZZn2rec:-McNGusq464APZZn2red:bgtj6mw) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L757)):*
> ```Java
>         containsInAnyOrder(
>             hasPrefix(Prefix.strict("1.0.0.0/16")), hasPrefix(Prefix.strict("1.0.0.0/24"))));
>   }
> ```

I would add one more test where the general aggregate can only be generated by some third prefix length that won't be present (say 28) just to prove that the generation policy is having an effect at all.
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 762 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNDn92DGKYFAsVmvmu:-McNDn92DGKYFAsVmvmv:bgjtzc5) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L762)):*
> ```Java
>    * - Sets up a process with 2 aggregates: 1.0.0.0/16 and 1.0.0.0/24.
>    * - Redistributes connected route 1.0.0.0/32 into BGP
>    * - Calls initBgpAggregates and returns the actions in resulting RIB delta.
> ```

could include a note of what the parameter does
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 783 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNFyht8IdPxCaiDfzK:-McNFyht8IdPxCaiDfzL:bcd1o12) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L783)):*
> ```Java
> 
>     // Create policy only allowing /32s to contribute to /16 aggregate
>     String onlyAccept32s = "onlyAccept32s";
> ```

comment & name not technically correct (dependent on parameter)
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 795 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNFHbl2nTcAW4wlvSd:-McNFHbm1H80VAqryW_Z:b-ch3ets) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L795)):*
> ```Java
>                             new PrefixRange(
>                                 Prefix.ZERO,
>                                 SubRange.singleton(prefixLengthContributingToAggregate16))))),
> ```

hmm if i saw this outside of a test i would definitely say to add a `static PrefixRange ofLength(int length)` to `PrefixRange`... this is just a test though :speak_no_evil:

Nonblocking :angel:

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 16 of 18 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNCLaD3cXce1IYTHNT:-McQ8GFWEpySumKbR0pI:b6knfuv) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*

just for legibility, could you move this line down to where it's used? like right above `if (moreGeneralAggregate != null)`
</blockquote></details>

I'd rather not since that location is inside a loop.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1297 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNAqtbDQipzaENAbFr:-McQ8cI473BlcJNJJGtb:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1297)):*

perhaps add a comment as to why at most one can match
</blockquote></details>

done
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 757 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNGusq464APZZn2rec:-McQ9BPp3tlkEliLc2i6:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L757)):*

I would add one more test where the general aggregate can only be generated by some third prefix length that won't be present (say 28) just to prove that the generation policy is having an effect at all.
</blockquote></details>

done
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 762 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNDn92DGKYFAsVmvmu:-McQAN8u-jbEyy_RIb4g:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L762)):*

could include a note of what the parameter does
</blockquote></details>

done
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 783 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNFyht8IdPxCaiDfzK:-McQAd_gENCFn7AhnKqP:b-7ql1td) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L783)):*

comment & name not technically correct (dependent on parameter)
</blockquote></details>

fixed
___
*[projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java, line 795 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNFHbl2nTcAW4wlvSd:-McQAjDz2ZWT5EQbQB3n:b-ags3r1) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpRoutingProcessTest.java#L795)):*

hmm if i saw this outside of a test i would definitely say to add a `static PrefixRange ofLength(int length)` to `PrefixRange`... this is just a test though :speak_no_evil:

Nonblocking :angel:
</blockquote></details>

We can do a structural search and replace later.

dhalperi:
-reviewer:@dhalperi

corinaminer (commented):
> Reviewed 4 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 17 of 18 files reviewed, 1 unresolved discussion (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNCLaD3cXce1IYTHNT:-McQGTrd08RufvXUK5dg:bicnzoo) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*

I'd rather not since that location is inside a loop.
</blockquote></details>

no, i'm suggesting moving it to after the loop. feel free to resolve though.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 17 of 19 files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 1224 at r8](https://reviewable.io/reviews/batfish/batfish/7067#-McNCLaD3cXce1IYTHNT:-McQIvfA1lF7oLyCtr0z:b-bhfbfv) ([raw file](https://github.com/batfish/batfish/blob/0c2f7299847f5bb6131b7537ac6de18277b5d4fc/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L1224)):*

no, i'm suggesting moving it to after the loop. feel free to resolve though.
</blockquote></details>

Oh I misread the code. Moved.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 16 of 19 files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[tests/basic/genDp.ref, line 16 at r10](https://reviewable.io/reviews/batfish/batfish/7067#-McQJCTm0W8PnCNTfTIB:-McQJCTm0W8PnCNTfTIC:bept1c9) ([raw file](https://github.com/batfish/batfish/blob/c42c6e9bc075521febbbdee5c48d06a233263260/tests/basic/genDp.ref#L16)):*
> ```
>         "9" : 78,
>         "10" : 78,
>         "11" : 78
> ```

I have no idea why there is an extra iteration.
I really don't see which parts of this PR could possible affect this ref.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 16 of 19 files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[tests/basic/genDp.ref, line 16 at r10](https://reviewable.io/reviews/batfish/batfish/7067#-McQJCTm0W8PnCNTfTIB:-McQL7hA56CdEvPpirtc:b-mnmxhe) ([raw file](https://github.com/batfish/batfish/blob/c42c6e9bc075521febbbdee5c48d06a233263260/tests/basic/genDp.ref#L16)):*

I have no idea why there is an extra iteration.
I really don't see which parts of this PR could possible affect this ref.
</blockquote></details>

Must be based on new guard in `initBgpAggregateRoutesLegacy`. Will be moot after follow-on PR for IOS BGP aggregates.

arifogel (commented):
> Reviewed 2 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: 18 of 19 files reviewed, all discussions resolved (waiting on @corinaminer)

arifogel (commented):
> Reviewed 1 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7067)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #7074, opened 2021-06-17
SearchRoutePolicies: Proper handling of updates to the default actions

millstein:
Updates the symbolic route analysis to track a BDD representing the conditions under which the default action is true, rather than just a boolean as previously done, and similarly for the default local action.  This allows us to properly handle updates to the default action that occur within conditional statements.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
@@ -75,7 +75,13 @@ import org.batfish.datamodel.routing_policy.statement.SetOspfMetricType;
 import org.batfish.datamodel.routing_policy.statement.SetTag;
 import org.batfish.datamodel.routing_policy.statement.Statement;
 import org.batfish.datamodel.routing_policy.statement.Statements.StaticStatement;
-import org.batfish.minesweeper.*;
+import org.batfish.minesweeper.CommunityVar;
+import org.batfish.minesweeper.Graph;
+import org.batfish.minesweeper.IDeepCopy;
+import org.batfish.minesweeper.OspfType;
+import org.batfish.minesweeper.Protocol;
+import org.batfish.minesweeper.SymbolicAsPathRegex;
+import org.batfish.minesweeper.SymbolicRegex;
 import org.batfish.minesweeper.bdd.CommunitySetMatchExprToBDD.Arg;
 import org.batfish.minesweeper.utils.PrefixUtils;
 
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
@@ -504,8 +504,7 @@ public class TransferBDD {
       // Some guards are statically resolved (e.g. CallExprContext and CallStatementContext), which
       // means that only one branch will be analyzed.  Skipping analysis of the other branch avoids
       // signaling an error unnecessarily if we reach a route-map construct that is not currently
-      // modelled.  It also allows us to properly account for updates to things like the default
-      // action that occur within the one feasible branch (see below for more on this).
+      // modelled.
       if (!guard.isZero()) {
         curP.debug("True Branch");
         // copy the current BDDRoute so we can separately track any updates on the two branches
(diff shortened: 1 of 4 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7074)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 561 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McQy0vW5pSqNgi-C9Nl:-McQxPqAD055cRSyMpHI:bwumq28) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L561)):*
Isn't this a regression?

Below you had previously written:
> ` // (unlikely but seems possible) that the "if" statement returns from within its guard`

Why can't that be handled now? Related to default policy tracking?
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 706 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McR--eEApggPukUnOsY:-McR--eEApggPukUnOsZ:bt80f2) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L706)):*
> ```Java
>     BDDRoute outAnnouncements = result.getFirst();
> 
>     assertTrue(acceptedAnnouncements.isZero());
> ```

What was happening before this PR?
I'm having trouble understanding how the changes apply to this test.
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1139 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McR-diT5jP77OUDhyrU:-McR-diT5jP77OUDhyrV:b-8twbh7) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1139)):*
> ```Java
>     BDDRoute outAnnouncements = result.getFirst();
> 
>     assertTrue(acceptedAnnouncements.isZero());
> ```

I suspect this was passing before? Is this test just for sanity?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7074)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 561 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McQy0vW5pSqNgi-C9Nl:-McR4gGPEBex6-YMXJcA:bersqr3) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L561)):*

Isn't this a regression?

Below you had previously written:
> ` // (unlikely but seems possible) that the "if" statement returns from within its guard`

Why can't that be handled now? Related to default policy tracking?
</blockquote></details>

This is unrelated to default tracking.  I realized that what was being done before wasn't actually correct.  The boolean returned from analyzing an expression represents the truth value of that expression, but it was being used previously also as the conditions under which the policy exits and accepts.  I believe that we would have to separately track the latter when analyzing expressions in order to support these kinds of early returns.  That could be done but it's not a priority at the moment.
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 706 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McR--eEApggPukUnOsY:-McR69BG5SzWarnoonMf:buzcz6y) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L706)):*

What was happening before this PR?
I'm having trouble understanding how the changes apply to this test.
</blockquote></details>

Yes, this is unrelated to default-action tracking.  Because I realized the issue with early returns from the conditional guard, I simplified the logic for computing the final state of the analysis after a conditional so wanted to check this particular case of an exit before a stateful conditional guard.  It was working properly before but this particular scenario was not being tested.
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1139 at r1](https://reviewable.io/reviews/batfish/batfish/7074#-McR-diT5jP77OUDhyrU:-McR6eoHDh8kdf04hP-F:b-f22uwu) ([raw file](https://github.com/batfish/batfish/blob/5621a9dee8905387b4a9e1172d4c3cce1984cead/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1139)):*

I suspect this was passing before? Is this test just for sanity?
</blockquote></details>

Yes it was passing before but only because the SetDefaultActionAccept within the branch was simply being ignored.  Now it's being analyzed but we still have to recognize when it's unreachable.

PR #7091, opened 2021-06-22
TestRoutePolicies: Don't ignore route tags

millstein:
Fixes a bug in TestRoutePolicies where the tags on the input/output routes were being ignored and replaced with the default value (0).


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7091)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7094, opened 2021-06-23
SearchRoutePolicies: Use TestRoutePolicies to produce the results

millstein:
Currently SearchRoutePolicies produces a model of the logical constraints representing the results of the analysis and then converts that model into the concrete input and output routes that are returned to the user.  This PR instead uses the model to obtain only the concrete input route and then uses TestRoutePolicies to simulate this route and produce the output route.  Doing so allows us to also return other useful information that TestRoutePolicies produces, notably (soon-to-be-added) execution trace information.


ratulm (commented):
> Reviewed 6 of 7 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7094)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7094#-McrBvPH4_QmwNWFSwsu:-McrBvPH4_QmwNWFSwsv:baalqqx) (no related file):*
org.batfish.minesweeper.question.searchroutepolicies.Result can be killed now, right?

---
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java, line 292 at r2](https://reviewable.io/reviews/batfish/batfish/7094#-McrCuU-6zGHvufb7_2z:-McrCuU-6zGHvufb7_3-:b-yoc5jf) ([raw file](https://github.com/batfish/batfish/blob/faf42586387dab24dc0feff2c76f3cce57ccb56a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L292)):*
> ```Java
>               policy.getOwner().getHostname(),
>               policy.getName());
>       TableAnswerElement ae = new TestRoutePoliciesAnswerer(trp, _batfish).answer(snapshot);
> ```

Instead of creating a question and running it end-to-end (and then asserting on the result), perhaps better to call an internal function in TRP? The rationale is that this leaves us free to interpret question parameters, which are user facing, the way it makes sense from a user perspective. The semantics of internal helpers is more black-n-white.

To make this happen, you may to make some TRP function public and/or define a new static function. (combination of testPolicy and toRow in TRP there is close to what you want.)

ratulm (commented):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7094)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7101, opened 2021-06-24
AbstractRoute: unbox metric

anothermattbrown:
getMetric used to return a Long, but implementations never return null, and clients don't null-check. Now it returns a long.


dhalperi (commented):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7101)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7130, opened 2021-07-01
XR: L2vpn conversion

sfraint:
This PR adds _some_ conversion for XR `l2vpn`.

----

`l2vpn` on XR devices can be configured in a way that is functionally ~the same as IOS VLAN interfaces & switchport mode trunk interfaces (with only a single allowed VLAN). As of this PR, when this is the case, Batfish will model the `l2vpn` as VLAN and trunk interfaces.

However, if the `l2vpn` is not configured to function like VLAN & single-VLAN-trunk interfaces, then the associated interfaces will be set as inactive and conversion warnings will be added to surface this (since Batfish can't fully model other behaviors yet).


dhalperi:
Subsumed by more recent work on new bridge domains: @arifogel

PR #7159, opened 2021-07-13
TransferBDD: Support formats that use output attributes

millstein:
Updates the symbolic route analysis to properly handle configuration formats that use output attributes for matching, rather than the attributes of the original route.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Environment.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Environment.java
@@ -40,7 +40,6 @@ public class Environment {
    * several fields.
    */
   public static Builder builder(@Nonnull Configuration c) {
-    ConfigurationFormat format = c.getConfigurationFormat();
     return new Builder()
         .setAsPathAccessLists(c.getAsPathAccessLists())
         .setAsPathExprs(c.getAsPathExprs())
@@ -54,10 +53,19 @@ public class Environment {
         .setRouteFilterLists(c.getRouteFilterLists())
         .setRoute6FilterLists(c.getRoute6FilterLists())
         .setRoutingPolicies(c.getRoutingPolicies())
-        .setUseOutputAttributes(
-            format == ConfigurationFormat.JUNIPER
-                || format == ConfigurationFormat.JUNIPER_SWITCH
-                || format == ConfigurationFormat.FLAT_JUNIPER);
+        .setUseOutputAttributes(useOutputAttributes(c));
+  }
+
+  /**
+   * Indicates whether the given {@link Configuration} should use the output route attributes
+   * instead of the original route attributes during route-policy evaluation, based on the
+   * configuration's format.
+   */
+  public static boolean useOutputAttributes(@Nonnull Configuration c) {
+    ConfigurationFormat format = c.getConfigurationFormat();
+    return format == ConfigurationFormat.JUNIPER
+        || format == ConfigurationFormat.JUNIPER_SWITCH
+        || format == ConfigurationFormat.FLAT_JUNIPER;
   }
 
   public enum Direction {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Environment.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Environment.java
@@ -53,15 +53,15 @@ public class Environment {
         .setRouteFilterLists(c.getRouteFilterLists())
         .setRoute6FilterLists(c.getRoute6FilterLists())
         .setRoutingPolicies(c.getRoutingPolicies())
-        .setUseOutputAttributes(useOutputAttributes(c));
+        .setUseOutputAttributes(useOutputAttributesFor(c));
   }
 
   /**
-   * Indicates whether the given {@link Configuration} should use the output route attributes
-   * instead of the original route attributes during route-policy evaluation, based on the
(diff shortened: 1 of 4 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7159)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 672 at r1](https://reviewable.io/reviews/batfish/batfish/7159#-MeWzFtHCEWggP56Zobb:-MeWzFtHCEWggP56Zobc:bu8bwx4) ([raw file](https://github.com/batfish/batfish/blob/197dbe00481ea20919accaf9dc601d785525a8b0/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L672)):*
> ```Java
>           sc.getCommunitySetExpr();
>       CommunityAPDispositions dispositions =
>           setExpr.accept(new SetCommunitiesVisitor(), new Arg(this, _originalRoute));
> ```

Why doesn't this use `routeForMaching`?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7159)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 672 at r1](https://reviewable.io/reviews/batfish/batfish/7159#-MeWzFtHCEWggP56Zobb:-MeX0OfB3W1e68Ja0IMh:b-atrd59) ([raw file](https://github.com/batfish/batfish/blob/197dbe00481ea20919accaf9dc601d785525a8b0/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L672)):*

Why doesn't this use `routeForMaching`?
</blockquote></details>

In this PR I added a comment about this at the top of the SetCommunitiesVisitor class, but I can put one here too.  In this situation the route is just being used to obtain the atomic predicates associated with each community regex (indirectly via a BDD, since we already have visitors that convert things to BDDs), so it needs to be a "fresh" route rather than the current one.  When we actually do the updates via the updateCommunities() function called below, we use the current route.

Separately, I do have a semantic question about how the actual update should work.  Consider an IOS route policy that does "set 30:40" followed by "add 40:40".  This becomes in our VI model (roughly) SetCommunities(30:40) followed by SetCommunities(InputCommunities U 40:40).  In my modeling we always consider InputCommunities to be the current route in this scenario, so we end up with both communities 30:40 and 40:40.  But SetCommunities::execute will treat InputCommunities as the original route and hence end up with just 40:40.  Which is correct?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7159)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 672 at r1](https://reviewable.io/reviews/batfish/batfish/7159#-MeWzFtHCEWggP56Zobb:-MeX2dsJ46B_g1MsbxnW:bvcoj9k) ([raw file](https://github.com/batfish/batfish/blob/197dbe00481ea20919accaf9dc601d785525a8b0/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L672)):*

In this PR I added a comment about this at the top of the SetCommunitiesVisitor class, but I can put one here too.  In this situation the route is just being used to obtain the atomic predicates associated with each community regex (indirectly via a BDD, since we already have visitors that convert things to BDDs), so it needs to be a "fresh" route rather than the current one.  When we actually do the updates via the updateCommunities() function called below, we use the current route.

Separately, I do have a semantic question about how the actual update should work.  Consider an IOS route policy that does "set 30:40" followed by "add 40:40".  This becomes in our VI model (roughly) SetCommunities(30:40) followed by SetCommunities(InputCommunities U 40:40).  In my modeling we always consider InputCommunities to be the current route in this scenario, so we end up with both communities 30:40 and 40:40.  But SetCommunities::execute will treat InputCommunities as the original route and hence end up with just 40:40.  Which is correct?
</blockquote></details>

I think each device has its own semantics for this.
A set followed by an add is dumb. An operator should just use a set.
IIRC some devices actually only allow either add or set, and may replace the other.
But a delete may be followed by an add or a set. So the problem remains.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7159)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 672 at r1](https://reviewable.io/reviews/batfish/batfish/7159#-MeWzFtHCEWggP56Zobb:-MeX6ecZArIOWE3CNggE:b2041vi) ([raw file](https://github.com/batfish/batfish/blob/197dbe00481ea20919accaf9dc601d785525a8b0/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L672)):*

I think each device has its own semantics for this.
A set followed by an add is dumb. An operator should just use a set.
IIRC some devices actually only allow either add or set, and may replace the other.
But a delete may be followed by an add or a set. So the problem remains.
</blockquote></details>

Ok, thanks.  Regarding your original question, I added a comment here.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7159)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7162, opened 2021-07-13
BroadcastL3Adjacencies: initial draft

dhalperi:
An L3Adjacencies implementation that models all interfaces' L1, L2, L3
configuration. This differs from `GlobalBroadcastNoPointToPoint` in that
interfaces with incompatible L2 configuration are not considered connected.

This differs from `HybridL3Adjacencies` in that it models every interface as
being connected to some Ethernet hub, whether or not it appears in the L1
topology. By using a flood-fill-style algorithm, performance is also greatly
improved.

This algorithm is not yet plumbed into any accessible path, as it cannot handle
all networks.

* Does not handle VxLAN or multiple L2 subinterfaces yet


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7162)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7176, opened 2021-07-16
Main RIB track and update routes whose network contains next hop

arifogel:
- If a route in the main RIB contains its own next hop within its
  network, it may only be activated if that next hop ip resolves
  to a more specific route


arifogel:
Solution is not general enough.

PR #7191, opened 2021-07-21
Track concrete structure types

ratulm:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7191)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @ratulm and @sfraint)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java, line 145 at r1](https://reviewable.io/reviews/batfish/batfish/7191#-Mf759mV8tvOjZ1_vArY:-Mf759mV8tvOjZ1_vArZ:b4xtmo6) ([raw file](https://github.com/batfish/batfish/blob/4427e0a36e9bea566ed8329d08ee18e294745bd4/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/VendorConfiguration.java#L145)):*
> ```Java
>           }
>         });
>     _concreteStructureTypes
> ```

I don't believe this is safe -- the abstraction is per-reference type.

Also, not all vendors use this way of tracking, I believe -- do not think it's good to do generically.

Instead, we should look at the vendor structure enum classes itself -- and only use the ones that are defined that way.

ratulm:
Will take another tack.

PR #7195, opened 2021-07-22
ForwardingAnalysis: use ARP IP instead of top-level route's next-hop IP

anothermattbrown:
ForwardingAnalysis distinguishes between ARP errors for owned vs unowned IPs for the purposes of assigning FlowDisposition. Previously we checked whether the top-level route's next-hop IP was owned, instead of the resolved ARP IP. This change fixes that, and also rewrites the classification logic to make clear that a single route can resolve to both an owned and an unowned  ARP IP (also a bit more efficient).


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7195)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7199, opened 2021-07-26
Layer1Topology: flexible matching for interfaces

dhalperi:
1. Introduce `InterfaceUtil` to be a single place to handle logic
2. Plumb it into cleanLayer1Topology. The raw Layer1Topology is the only one
   that should contain non-canonical names.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7199)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

PR #7210, opened 2021-07-30
L3AdjacencyComputer: do not connect interfaces with L1 to global hub

dhalperi:
The activeLogicalL1Topology does not include edges where one or both sides are
inactive or not present in the snapshot. For deciding whether an interface is
attached to the global hub, however, we need to consider these removed edges.
Fall back to the logicalL1Topology, which does include all (logical) interfaces
mentioned in any L1 topology and present in the snapshot.


ratulm (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7210)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7234, opened 2021-08-09
SearchRoutePolicies: Remove the metric field

millstein:
The symbolic route analysis previously tracked a (symbolic) value for the MED and also separately one for the metric.  Following the way the routing simulation works we no longer have this distinction and instead only track one symbolic value (the MED).  This PR also removes legacy best-effort support in the symbolic route analysis for PrependAsPath that depended on the metric -- that support is not useful or correct for our purposes, and handling PrependAsPath properly requires more care and thought and is not a priority currently.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7234)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #7241, opened 2021-08-17
Create ConversionContext and make accessible in conversion

corinaminer:
Nothing in ConversionContext class yet, just a placeholder.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java
@@ -212,20 +212,19 @@ public class FileBasedStorage implements StorageProvider {
     }
   }
 
-  @Nullable
   @Override
-  public ConversionContext loadConversionContext(NetworkSnapshot snapshot) {
+  public @Nonnull ConversionContext loadConversionContext(NetworkSnapshot snapshot)
+      throws IOException {
     Path ccPath = getConversionContextPath(snapshot.getNetwork(), snapshot.getSnapshot());
     if (!Files.exists(ccPath)) {
-      return null;
+      throw new FileNotFoundException();
     }
     try {
       return deserializeObject(ccPath, ConversionContext.class);
     } catch (BatfishException e) {
-      _logger.errorf(
-          "Failed to deserialize ConversionContext: %s", Throwables.getStackTraceAsString(e));
-      LOGGER.error("Failed to deserialize ConversionContext", e);
-      return null;
+      throw new IOException(
+          String.format(
+              "Failed to deserialize ConversionContext: %s", Throwables.getStackTraceAsString(e)));
     }
   }
 
@@ -485,7 +484,9 @@ public class FileBasedStorage implements StorageProvider {
     serializeObject(conversionContext, ccPath);
   }
 
-  private @Nonnull Path getConversionContextPath(NetworkId network, SnapshotId snapshot) {
+  @VisibleForTesting
+  @Nonnull
+  Path getConversionContextPath(NetworkId network, SnapshotId snapshot) {
     return getSnapshotOutputDir(network, snapshot).resolve(RELPATH_CONVERSION_CONTEXT);
   }
 
@@ -595,7 +596,8 @@ public class FileBasedStorage implements StorageProvider {
    * Writes a single object of the given class to the given file. Uses the {@link FileBasedStorage}
    * default file encoding including serialization format and compression.
    */
-  private void serializeObject(Serializable object, Path outputFile) {
+  @VisibleForTesting
+  void serializeObject(Serializable object, Path outputFile) {
     Path sanitizedOutputFile = validatePath(outputFile);
     try {
       Path tmpFile = Files.createTempFile(null, null);
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java
@@ -97,6 +97,7 @@ import org.batfish.identifiers.QuestionId;
 import org.batfish.identifiers.SnapshotId;
 import org.batfish.referencelibrary.ReferenceLibrary;
 import org.batfish.role.NodeRolesData;
+import org.batfish.vendor.ConversionContext;
 import org.batfish.vendor.VendorConfiguration;
 
 /** A utility class that abstracts the underlying file system storage used by Batfish. */
@@ -124,6 +125,7 @@ public class FileBasedStorage implements StorageProvider {
   private static final String RELPATH_METADATA_FILE = "metadata.json";
   private static final String RELPATH_FORK_REQUEST_FILE = "fork_request";
   private static final String RELPATH_ENV_TOPOLOGY_FILE = "env_topology";
+  private static final String RELPATH_CONVERSION_CONTEXT = "conversion_context";
   private static final String RELPATH_CONVERT_ANSWER_PATH = "convert_answer";
   private static final String RELPATH_ANSWERS_DIR = "answers";
   private static final String RELPATH_ANSWER_METADATA = "answer_metadata.json";
@@ -206,6 +208,23 @@ public class FileBasedStorage implements StorageProvider {
     try {
       return deserializeObjects(namesByPath, Configuration.class);
     } catch (BatfishException e) {
+      return null;
+    }
+  }
+
+  @Nullable
+  @Override
+  public ConversionContext loadConversionContext(NetworkSnapshot snapshot) {
+    Path ccPath = getConversionContextPath(snapshot.getNetwork(), snapshot.getSnapshot());
+    if (!Files.exists(ccPath)) {
+      return null;
+    }
+    try {
+      return deserializeObject(ccPath, ConversionContext.class);
+    } catch (BatfishException e) {
+      _logger.errorf(
+          "Failed to deserialize ConversionContext: %s", Throwables.getStackTraceAsString(e));
+      LOGGER.error("Failed to deserialize ConversionContext", e);
       return null;
     }
   }
@@ -456,6 +475,18 @@ public class FileBasedStorage implements StorageProvider {
             configurations.size(), snapshot);
 
     storeConfigurations(outputDir, batchName, configurations);
+  }
+
+  @Override
+  public void storeConversionContext(ConversionContext conversionContext, NetworkSnapshot snapshot)
+      throws IOException {
+    Path ccPath = getConversionContextPath(snapshot.getNetwork(), snapshot.getSnapshot());
+    mkdirs(ccPath.getParent());
+    serializeObject(conversionContext, ccPath);
+  }
+
+  private @Nonnull Path getConversionContextPath(NetworkId network, SnapshotId snapshot) {
+    return getSnapshotOutputDir(network, snapshot).resolve(RELPATH_CONVERSION_CONTEXT);
   }
 
   private @Nonnull Path getConvertAnswerPath(NetworkId network, SnapshotId snapshot) {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java
@@ -818,10 +818,11 @@ public interface StorageProvider {
   /**
    * Loads the {@link ConversionContext} for the given {@link NetworkSnapshot}, if present.
    *
-   * @throws IOException if there is an error
+   * @throws FileNotFoundException if there is no serialized {@link ConversionContext}
+   * @throws IOException if there is an error deserializing
    */
-  @Nullable
-  ConversionContext loadConversionContext(NetworkSnapshot snapshot);
+  @Nonnull
+  ConversionContext loadConversionContext(NetworkSnapshot snapshot) throws IOException;
 
   /**
    * Stores the {@link ConversionContext} for the given {@link NetworkSnapshot}.
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java
@@ -45,6 +45,7 @@ import org.batfish.identifiers.QuestionId;
 import org.batfish.identifiers.SnapshotId;
 import org.batfish.referencelibrary.ReferenceLibrary;
 import org.batfish.role.NodeRolesData;
+import org.batfish.vendor.ConversionContext;
 import org.batfish.vendor.VendorConfiguration;
 
 /** Storage backend for loading and storing persistent data used by Batfish */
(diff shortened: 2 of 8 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @arifogel and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7241#-MhGUDhS0k3zsEPchkJM:-MhGUDhS0k3zsEPchkJN:b-ss37y2) (no related file):*
FYI, I thought it might make sense to put runtime data in `ConversionContext`. Didn't do so in this PR since we haven't discussed and it can wait indefinitely.

---

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 4 unresolved discussions (waiting on @corinaminer and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7241#-MhKLPeu5z_6WIuhWucH:-MhKLPeu5z_6WIuhWucI:bi2dydc) (no related file):*
Please add tests to `FileBasedStorageTest`

---
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java, line 217 at r1](https://reviewable.io/reviews/batfish/batfish/7241#-MhKIo1Z-5dDn3RvM5Wz:-MhKIo1Z-5dDn3RvM5X-:bvx6ick) ([raw file](https://github.com/batfish/batfish/blob/7e3f10f6d5a842e4f9c6f3708781446097ae982d/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java#L217)):*
> ```Java
>   @Nullable
>   @Override
>   public ConversionContext loadConversionContext(NetworkSnapshot snapshot) {
> ```

New APIs should not throw `BatfishException`. Instead, throw `IOException` and document in javadoc in `StorageProvider`. Just wrap the `BatfishException` in `IOException` when deserialization fails.
Throw `FileNotFoundException` specifically instead of returning `null` when file does not exist.
Callers should be responsible for recovery in case of `FileNotFoundException`, which should be distinguishable from deserialization failure.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java, line 821 at r1](https://reviewable.io/reviews/batfish/batfish/7241#-MhKK2JLBDYwY9S1OpFH:-MhKK2JLBDYwY9S1OpFI:b-ektmwp) ([raw file](https://github.com/batfish/batfish/blob/7e3f10f6d5a842e4f9c6f3708781446097ae982d/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java#L821)):*
> ```Java
>    * Loads the {@link ConversionContext} for the given {@link NetworkSnapshot}, if present.
>    *
>    * @throws IOException if there is an error
> ```

Put `throws FileNotFoundException, IOException` in the signature here (former might not be required).
Add javadoc `@throws FileNotFoundException if conversion context file does not exist` 
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java, line 5 at r1](https://reviewable.io/reviews/batfish/batfish/7241#-MhKKzoi4Q0DXn9ziuZQ:-MhKKzoi4Q0DXn9ziuZR:b-szionk) ([raw file](https://github.com/batfish/batfish/blob/7e3f10f6d5a842e4f9c6f3708781446097ae982d/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java#L5)):*
> ```Java
> import java.io.Serializable;
> 
> public class ConversionContext implements Serializable {}
> ```

nit: `final`

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: 3 of 8 files reviewed, all discussions resolved (waiting on @arifogel and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7241#-MhKLPeu5z_6WIuhWucH:-MhKShOi9eTlgpyI2tSm:bb7f2mf) (no related file):*

Please add tests to `FileBasedStorageTest`
</blockquote></details>

done

---
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java, line 217 at r1](https://reviewable.io/reviews/batfish/batfish/7241#-MhKIo1Z-5dDn3RvM5Wz:-MhKSY691w1b_K8GyJlW:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/7e3f10f6d5a842e4f9c6f3708781446097ae982d/projects/batfish-common-protocol/src/main/java/org/batfish/storage/FileBasedStorage.java#L217)):*

New APIs should not throw `BatfishException`. Instead, throw `IOException` and document in javadoc in `StorageProvider`. Just wrap the `BatfishException` in `IOException` when deserialization fails.
Throw `FileNotFoundException` specifically instead of returning `null` when file does not exist.
Callers should be responsible for recovery in case of `FileNotFoundException`, which should be distinguishable from deserialization failure.


</blockquote></details>

done
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java, line 821 at r1](https://reviewable.io/reviews/batfish/batfish/7241#-MhKK2JLBDYwY9S1OpFH:-MhKSa9mBVulu5kdsmWD:b-voq4fv) ([raw file](https://github.com/batfish/batfish/blob/7e3f10f6d5a842e4f9c6f3708781446097ae982d/projects/batfish-common-protocol/src/main/java/org/batfish/storage/StorageProvider.java#L821)):*

Put `throws FileNotFoundException, IOException` in the signature here (former might not be required).
Add javadoc `@throws FileNotFoundException if conversion context file does not exist` 
</blockquote></details>

done. `FileNotFoundException` is a child of `IOException` so not necessary in signature.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2391 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKZa5b3m4B0V435ie4:-MhKZa5c5l7mq2SoxBnJ:bw3zass) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2391)):*
> ```Java
>       } catch (FileNotFoundException ignored) {
>       } catch (IOException e) {
>         LOGGER.warn("Failed to load conversion context: {}", Throwables.getStackTraceAsString(e));
> ```

Is it really OK to keep going if this fails? When it's absent? Why would that ever happen?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java, line 8 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhK_HyZ9w7E0mBKn3TU:-MhK_HyZ9w7E0mBKn3TV:b32rgds) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java#L8)):*
> ```Java
>   @Override
>   public boolean equals(Object obj) {
>     return obj instanceof ConversionContext;
> ```

This is really an unsafe and misleading implementation given its likely future content. Furthermore, it won't be the case that you'll be able to properly test equality of all the future fields. If this is just for the tests you added, I'd suggest simply testing `instanceOf` instead of `equalTo`, and getting rid of these overrides.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2389 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKaflU7MqWozwGOJTb:-MhKaflVCGxqEDhJhbIL:bp8z9yx) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2389)):*
> ```Java
>       try {
>         conversionContext = _storage.loadConversionContext(snapshot);
>       } catch (FileNotFoundException ignored) {
> ```

I think a comment inside the empty catch block will satisfy checkstyle, but not sure.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2391 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKZa5b3m4B0V435ie4:-MhKao4Q7rILxusU91qd:b-6yu1wc) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2391)):*

Is it really OK to keep going if this fails? When it's absent? Why would that ever happen?
</blockquote></details>

To clarify: If you don't get a context now, then you'll likely get a crash later in conversion when a `VendorConfiguration` expects it to be present. This violates the fail early principle.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: 5 of 8 files reviewed, 1 unresolved discussion (waiting on @arifogel and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2389 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKaflU7MqWozwGOJTb:-MhKeeZI6rgH04vdaaWs:b63e958) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2389)):*

I think a comment inside the empty catch block will satisfy checkstyle, but not sure.
</blockquote></details>

ack, trying it
___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2391 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKZa5b3m4B0V435ie4:-MhKe-8xBqvlUGH2WOyz:bi84c16) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2391)):*

To clarify: If you don't get a context now, then you'll likely get a crash later in conversion when a `VendorConfiguration` expects it to be present. This violates the fail early principle.
</blockquote></details>

if deserialization fails, that probably means there's a stale serialized `ConversionContext` that can't be deserialized to the current version of `ConversionContext`. Not sure if/when that could happen (would have to re-convert without reparsing), but if it does, ignoring the stale `ConversionContext` seems like the best option to me.

I expect `ConversionContext` will be absent most of the time, since there's no point in creating or serializing an empty one. I think it's OK to let `conversionContext` be null in these cases; it's nullable in `VendorConfiguration`. I'm using the same pattern as for `runtimeData`, which is allowed to be null in the constructor for `ConvertConfigurationJob` (although it's unnecessarily guaranteed to be nonnull below).
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java, line 8 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhK_HyZ9w7E0mBKn3TU:-MhKgFtC1x0Zp6kZ78IJ:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java#L8)):*

This is really an unsafe and misleading implementation given its likely future content. Furthermore, it won't be the case that you'll be able to properly test equality of all the future fields. If this is just for the tests you added, I'd suggest simply testing `instanceOf` instead of `equalTo`, and getting rid of these overrides.
</blockquote></details>

done

arifogel (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2391 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKZa5b3m4B0V435ie4:-MhKjMWbC1RVBwx8erGb:brgd1ag) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2391)):*
> Not sure if/when that could happen (would have to re-convert without reparsing)

I am arguing that this cannot happen via user APIs. Making it `null` will hurt the debugging experience in case of an actual deserialization problem. For checkpoint gateways that don't actually _need_ management info, you will never even notice there is a problem unless you meticulously inspect the logger output, or see that the gateways are not behaving as expected

> I expect ConversionContext will be absent most of the time, since there's no point in creating or serializing an empty one. 

It is _more_ work in terms of code complexity to not (de)serialize an empty conversion context. The computation time is negligible.
Further, it means that in `CheckpointGatewayConfiguration` you won't be able to tell the difference between missing/unset management info during conversion. If it is always nonnull during conversion, then you can assert on that in `toVendorIndependentConfigurations`.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: 5 of 8 files reviewed, all discussions resolved (waiting on @arifogel and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 2391 at r2](https://reviewable.io/reviews/batfish/batfish/7241#-MhKZa5b3m4B0V435ie4:-MhL6A64CfdvujRLyTm0:b-976fmo) ([raw file](https://github.com/batfish/batfish/blob/809a1bbe59d71329f83ae3730c28262ab234dc82/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L2391)):*

> Not sure if/when that could happen (would have to re-convert without reparsing)

I am arguing that this cannot happen via user APIs. Making it `null` will hurt the debugging experience in case of an actual deserialization problem. For checkpoint gateways that don't actually _need_ management info, you will never even notice there is a problem unless you meticulously inspect the logger output, or see that the gateways are not behaving as expected

> I expect ConversionContext will be absent most of the time, since there's no point in creating or serializing an empty one. 

It is _more_ work in terms of code complexity to not (de)serialize an empty conversion context. The computation time is negligible.
Further, it means that in `CheckpointGatewayConfiguration` you won't be able to tell the difference between missing/unset management info during conversion. If it is always nonnull during conversion, then you can assert on that in `toVendorIndependentConfigurations`.
</blockquote></details>

done, i think

arifogel (commented):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 54 at r4](https://reviewable.io/reviews/batfish/batfish/7241#-MhLYncQAb6XuxXvwzz5:-MhLYncQAb6XuxXvwzz6:b-tt0tf6) ([raw file](https://github.com/batfish/batfish/blob/fe4f97061f1213c5d812dbed470957a188699704/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L54)):*
> ```Java
>   public ConvertConfigurationJob(
>       Settings settings,
>       @Nullable ConversionContext conversionContext,
> ```

Will this ever be null now?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7241)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 54 at r4](https://reviewable.io/reviews/batfish/batfish/7241#-MhLYncQAb6XuxXvwzz5:-MhLZYhz2e4czTdKjsLk:b-69nrlu) ([raw file](https://github.com/batfish/batfish/blob/fe4f97061f1213c5d812dbed470957a188699704/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L54)):*

Will this ever be null now?
</blockquote></details>

possibly in tests. following the pattern for `runtimeData`, which is also guaranteed nonnull when this is called by `Batfish`

PR #7248, opened 2021-08-18
junit.bzl: compile test sources only once

dhalperi:
We need a java_library of the tests in order to run PMD. Previously,
we compiled the sources twice: once in the java_test (passing sources as sources)
and once in the java_library for pmdlib.

Now, generate the testlib first and use the same lib for both java_tests (deps)
and pmd_test (lib). This saves a compilation.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7248)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7266, opened 2021-08-24
CP: Convert two types of IP spaces

corinaminer:
Converts all `AddressRange` and `Network` objects referenced in NAT rulebases to named IP spaces.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/BUILD
+++ b/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/BUILD
@@ -11,6 +11,7 @@ java_library(
     ),
     deps = [
         "//projects/batfish-common-protocol:common",
+        "//projects/batfish/src/main/java/org/batfish/vendor/check_point_management",
         "@maven//:com_google_code_findbugs_jsr305",
         "@maven//:com_google_guava_guava",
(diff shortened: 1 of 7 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7266)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConversions.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/7266#-MhuG-vYCbCNP1M5SVMP:-MhuG-vYCbCNP1M5SVMQ:b-ggxwa6) ([raw file](https://github.com/batfish/batfish/blob/98baa72dd32f82e4aef6a36a2c024d6d388e4a8d/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConversions.java#L30)):*
> ```Java
>     // TODO Network objects also have a "mask-length4" that we don't currently extract.
>     //  If network objects always represent valid Prefixes, it may be simpler to extract
>     //  that instead of subnet-mask and convert the network to a PrefixIpSpace.
> ```

should i just make this change in this PR? Doubt network objects could be true wildcards (i.e. invalid prefixes) considering they have a length field

PR #7271, opened 2021-08-27
Checkpoint gateway conversion use correct package

arifogel:
- add test infrastructure to use pre-populated ConversionContext

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/vendor/ConversionContext.java
@@ -21,4 +21,8 @@ public final class ConversionContext implements Serializable {
   }
 
   private @Nullable VendorSupplementalInformation _checkpointManagementConfiguration;
+
+  public boolean isEmpty() {
+    return _checkpointManagementConfiguration == null;
+  }
 }
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -2453,10 +2453,12 @@ public class Batfish extends PluginConsumer implements IBatfish {
     LOGGER.info("\n*** SERIALIZING CONVERSION CONTEXT ***\n");
     ConversionContext conversionContext = new ConversionContext();
     conversionContext.setCheckpointManagementConfiguration(cpMgmtConfig);
-    try {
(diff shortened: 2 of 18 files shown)


corinaminer (commented):
> Reviewed 3 of 16 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7271)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java, line 184 at r1](https://reviewable.io/reviews/batfish/batfish/7271#-Mi42xPd4wiULlDmjfzB:-Mi7jusw3_eDrgBEpK9R:bt4kn4u) ([raw file](https://github.com/batfish/batfish/blob/26a51f9ef73cf91e6d244cbffc2697a5884e6bc6/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java#L184)):*

There could plausibly be a discrepancy. In the end, the most specific entry is the one that matters. So this should do the right thing.
</blockquote></details>

seems plausible that, for example, the cluster policy's NAT rulebase could get inherited if the cluster member's policy has no NAT rulebase. But we can figure that out later.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7271)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java, line 184 at r1](https://reviewable.io/reviews/batfish/batfish/7271#-Mi42xPd4wiULlDmjfzB:-Mi8xDQ56jsR2dEU0eg5:b-u8t0xz) ([raw file](https://github.com/batfish/batfish/blob/26a51f9ef73cf91e6d244cbffc2697a5884e6bc6/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java#L184)):*

seems plausible that, for example, the cluster policy's NAT rulebase could get inherited if the cluster member's policy has no NAT rulebase. But we can figure that out later.
</blockquote></details>

Yeah, but the inheritance could be reflected in the show-gateways-and-servers output for the member when that happens. Let's defer for now.

PR #7273, opened 2021-08-27
IOS-XR: fix some issues with prefix-sets

dhalperi:
1. Add null annotations, so we get warnings in users.
2. Make the visitor more readable, and fix a bug where mixed prefix-sets would
   not get converted.
3. Add a test that v4 addresses in combo prefix-sets are permitted.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7273)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer and @sfraint)

PR #7296, opened 2021-08-31
MatchBgpSessionType: a new routing policy BooleanExpr

dhalperi:
Returns true only when being evaluated in the context of a specific type of BGP Session.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/CommunityStructuresVerifier.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/CommunityStructuresVerifier.java
@@ -66,6 +66,7 @@ import org.batfish.datamodel.routing_policy.expr.HasRoute;
 import org.batfish.datamodel.routing_policy.expr.HasRoute6;
 import org.batfish.datamodel.routing_policy.expr.LegacyMatchAsPath;
 import org.batfish.datamodel.routing_policy.expr.MainRib;
+import org.batfish.datamodel.routing_policy.expr.MatchBgpSessionType;
 import org.batfish.datamodel.routing_policy.expr.MatchColor;
 import org.batfish.datamodel.routing_policy.expr.MatchIp6AccessList;
 import org.batfish.datamodel.routing_policy.expr.MatchIpv4;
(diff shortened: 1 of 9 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7296)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchBgpSessionType.java, line 102 at r1](https://reviewable.io/reviews/batfish/batfish/7296#-MiTqouQEB1JEDlVUjCL:-MiTqouQEB1JEDlVUjCM:b-wnh8ri) ([raw file](https://github.com/batfish/batfish/blob/33bbbe82c440603598e6e369bf8e1efd007b09d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchBgpSessionType.java#L102)):*
> ```Java
>     int hash = 0;
>     for (Type t : _types) {
>       hash += t.ordinal();
> ```

nit: This can have collisions, but not with only two enum members (unless ordinal starts at 0). Did you mean to get hash of an ordinal somewhere, or multiply something by 31 before adding?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7296)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchBgpSessionType.java, line 102 at r1](https://reviewable.io/reviews/batfish/batfish/7296#-MiTqouQEB1JEDlVUjCL:-MiTt6w87_2KpmWGNwZ2:b-tg7f1o) ([raw file](https://github.com/batfish/batfish/blob/33bbbe82c440603598e6e369bf8e1efd007b09d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/expr/MatchBgpSessionType.java#L102)):*
> multiply something by 31 before adding

A key thing to remember is that this is a set, and we just copied the standard Set hash algorithm (also indicated in comment): sum all the hash codes.

The usual time that you multiply by 31 is when you're making it order sensitive. Sets can't do that.

> Did you mean to get hash of an ordinal somewhere,

The hash of an int is the int itself, so this is inlined.

> This can have collisions,

Collisions are okay, especially as we are highly unlikely to have `Set<MatchBgpSessionType>` or similar.

--

I'm assuming we could, yeah, map each ordinal to 2^ord (as long as fewer than 31 enum values), so that you can distinguish every set, but seems unnecessary.

PR #7346, opened 2021-09-08
Checkpoint: fix handling of ServiceTcp port ranges

dhalperi:
This hotfix assumes that IntegerSpace parse just works.
We should probably have VS special handling, unless we can
confirm that the syntax used on device always confirms to
that accepted by IntegerSpace.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7346)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @sfraint)

PR #7351, opened 2021-09-08
Checkpoint: test type hierarchy

dhalperi:
JSON serialization tests should always test the class with @JsonTypeInfo on it.
This is because that is what the known type will be when deserializing these
fields as part of another object.


sfraint (commented):
> Reviewed 25 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7351)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #7360, opened 2021-09-09
CheckPoint: convert `ServiceGroup`s

sfraint:


arifogel (commented):
> Reviewed 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7360)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java, line 45 at r3](https://reviewable.io/reviews/batfish/batfish/7360#-MjBri2n3_kT_YLkraL-:-MjBri2n3_kT_YLkraL0:bsry2le) ([raw file](https://github.com/batfish/batfish/blob/0d77c638472aebf0da5adf2b951582ca481c6845/projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java#L45)):*
> ```Java
>             .filter(Service.class::isInstance)
>             .map(Service.class::cast)
>             .map(s -> s.accept(this))
> ```

This seems like the wrong solution.
Can you instead use a reference instead of reconverting the referenced services? Then you won't need to do any of this loop checking (it should be done in post-processing checks instead).

OK to do in follow-on.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7360)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7360)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java, line 45 at r3](https://reviewable.io/reviews/batfish/batfish/7360#-MjBri2n3_kT_YLkraL-:-MjBuA7Z6brUV2BWwUEa:bdmovg9) ([raw file](https://github.com/batfish/batfish/blob/0d77c638472aebf0da5adf2b951582ca481c6845/projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java#L45)):*

This seems like the wrong solution.
Can you instead use a reference instead of reconverting the referenced services? Then you won't need to do any of this loop checking (it should be done in post-processing checks instead).

OK to do in follow-on.
</blockquote></details>

would require a lot more refactoring, if doing that, should be done separately

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7360)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java, line 45 at r3](https://reviewable.io/reviews/batfish/batfish/7360#-MjBri2n3_kT_YLkraL-:-MjBuYwp87XQhwiFfe6a:b-9oeshb) ([raw file](https://github.com/batfish/batfish/blob/0d77c638472aebf0da5adf2b951582ca481c6845/projects/batfish/src/main/java/org/batfish/vendor/check_point_management/ServiceToMatchExpr.java#L45)):*

would require a lot more refactoring, if doing that, should be done separately
</blockquote></details>

As discussed offline, we don't actually have references of this type. So out of scope, definitely.

Issue #6502, opened 2020-12-12
NX-OS: Support "match interface" in route-maps


corinaminer:
`match interface` is unsupported as of #6500. We need to implement two new BooleanExprs to match next hop IP and next hop interface.

Documentation here: https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/unicast/configuration/guide/l3_cli_nxos/l3_rpm.html
"Matches any routes that have their next hop out one of the configured interfaces."

PR #7378, opened 2021-09-11
MatchInterface: implement simple version

dhalperi:
Interface routes (OSPF, Connected, Local, Static) will be matched, but not
resolved routes that happen to use those interface. Probably on most vendors
the latter is actually the correct semantics; however this route-map is most
often used to match local routes so it should get us going in the right
direction.

This is for batfish/batfish#6502


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7378)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7378)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #7410, opened 2021-09-20
Pick the VRRP winner deterministically

ratulm:
A hotfix to make VRRP winner deterministic. 

Non-determinism can happen if interfaces from separate groups get combined together because they have the same group Id. Will send a proper fix later that distinguishes different groups.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import javax.annotation.Nonnull;
+import org.batfish.common.util.ComparableStructure;
 import org.batfish.datamodel.AclIpSpace;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7410)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel and @ratulm)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java, line 280 at r1](https://reviewable.io/reviews/batfish/batfish/7410#-Mk4BtU0ENXGzdVjFRdF:-Mk4BtU0ENXGzdVjFRdG:b-nqzqk3) ([raw file](https://github.com/batfish/batfish/blob/71d8b6a123ec119750e9d07cc510668018943ca8/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java#L280)):*
> ```
>                           .thenComparing(o -> o.getOwner().getHostname())
>                           .thenComparing(ComparableStructure::getName));
> ```

`.thenComparing(NodeInterfacePair::of)` is probably better. Already use that comparator.

PR #7414, opened 2021-09-21
Disambiguate VRRP groups using L3 adjacencies

ratulm:
1. Creates a new method that disambiguates VRRP IP ownership. Only used by RoutesAnswerer at the moment. 

2. Some refactoring for testability


dhalperi:
I think at this point @anothermattbrown best understands IpOwners

ratulm:
Let me figure out the full path forward

PR #7416, opened 2021-09-21
IpOwners: VRRP IP disambiguation

ratulm:
1. Disambiguate VRRP groups using L3 adjacencies. Needed with virtual IPs and groups names are reused.

2. Clean up IpOwners class to eliminate the possibility of clients getting different sources of truth. 

@anothermattbrown - I need to write a test or two but let me know if this makes sense otherwise.


anothermattbrown (commented):
> Reviewed 14 of 18 files at r1, 6 of 6 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: 19 of 21 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @ratulm)

___
*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 898 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9Uzr66kjAYdXFJ5qs:-Mk9Uzr66kjAYdXFJ5qt:br0sg5q) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L898)):*
> ```
>     // Needed to prevent a loop -- if configurations are not loaded, accessing topology info will
>     // load them, which will iin turn load topologies for serialization.
>     batfish.loadConfigurations(batfish.getSnapshot());
> ```
this seems problematic. is there a better way? is `batfish.getTopologyProvider()` the call that will crash? if yes, should we fill the cache there to prevent the crash?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java, line 411 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9TSkZ77BUZEHFh6Xp:-Mk9TSkZ77BUZEHFh6Xq:b-js3g1x) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java#L411)):*
> ```
> partitionVrrpCandidates
> ```
this strikes me as less efficient than it could be, though I realize you're running into API limitations. It won't matter as long as the number of interfaces/domains is always small. Do you have a sense for how large they could be?

Also, nit: the return type doesn't need to use sets; all you do is iterate over them.

anothermattbrown (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: all files reviewed, 2 unresolved discussions (waiting on @ratulm)

anothermattbrown (commented):
> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: all files reviewed, 2 unresolved discussions (waiting on @ratulm)

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @ratulm)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java, line 411 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9TSkZ77BUZEHFh6Xp:-Mk9gZ2g8bnjE0uXJjqm:b-7hhhot) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java#L411)):*

> ```
> partitionVrrpCandidates
> ```
this strikes me as less efficient than it could be, though I realize you're running into API limitations. It won't matter as long as the number of interfaces/domains is always small. Do you have a sense for how large they could be?

Also, nit: the return type doesn't need to use sets; all you do is iterate over them.
</blockquote></details>

I expect small numbers in that set -- definitely less than 10. 

Curious: what might you do if all you have is the ability to get a pairwise yes/no?

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 898 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9Uzr66kjAYdXFJ5qs:-MkAABmx3vb7S0UsW-EL:bga26xv) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L898)):*

> ```
>     // Needed to prevent a loop -- if configurations are not loaded, accessing topology info will
>     // load them, which will iin turn load topologies for serialization.
>     batfish.loadConfigurations(batfish.getSnapshot());
> ```
this seems problematic. is there a better way? is `batfish.getTopologyProvider()` the call that will crash? if yes, should we fill the cache there to prevent the crash?
</blockquote></details>

if this is just a quirk of `BatfishTestUtils.getBatfishFromTestrigText` and it needs to stay as-is (for other clients), it may be sufficient to document the pitfall and move on. Or could add a variant of the method that calls loadConfigurations before returning 
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java, line 411 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9TSkZ77BUZEHFh6Xp:-MkA9mLY588oGCiCYPAi:b-28e18b) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java#L411)):*

I expect small numbers in that set -- definitely less than 10. 

Curious: what might you do if all you have is the ability to get a pairwise yes/no?
</blockquote></details>

Yeah we're pretty limited with that API. Sounds like perf it won't matter much here

ratulm:
Wow, this branch got plastered due to my rebase attempts!

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: 0 of 46 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java, line 898 at r3](https://reviewable.io/reviews/batfish/batfish/7416#-Mk9Uzr66kjAYdXFJ5qs:-MkBCJKsCfPVy0fIIDFs:brpkomh) ([raw file](https://github.com/batfish/batfish/blob/a6b840eaf1e5b6a15251439607dc1193397fbe84/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L898)):*

if this is just a quirk of `BatfishTestUtils.getBatfishFromTestrigText` and it needs to stay as-is (for other clients), it may be sufficient to document the pitfall and move on. Or could add a variant of the method that calls loadConfigurations before returning 
</blockquote></details>

That other PR makes the dependency explicit.

anothermattbrown (commented):
> Reviewed 20 of 24 files at r8, 21 of 22 files at r11, 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7416)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7420, opened 2021-09-22
IBatfish: Another load config that refuses to initialize snapshot

ratulm:
A variant of Batfish::loadConfiguration that will not parse uninitialized snapshots. Call it from topology provider, which helps make sure that topology provider does not try to initialize snapshots. Topology provider is not expected to work for uninitialized snapshots.

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/e2e/l1/asymmetric/L1AsymmetricTest.java
+++ b/projects/allinone/src/test/java/org/batfish/e2e/l1/asymmetric/L1AsymmetricTest.java
@@ -27,6 +27,7 @@ public class L1AsymmetricTest {
                 .setLayer1TopologyPrefix(prefix + "/batfish")
                 .build(),
             _folder);
+    batfish.loadConfigurations(batfish.getSnapshot());
     return batfish.getTopologyProvider().getInitialLayer3Topology(batfish.getSnapshot());
   }
 
--- a/projects/allinone/src/test/java/org/batfish/e2e/layer2ntc/Layer2NtcTest.java
+++ b/projects/allinone/src/test/java/org/batfish/e2e/layer2ntc/Layer2NtcTest.java
@@ -36,6 +36,7 @@ public class Layer2NtcTest {
                 .setLayer1TopologyPrefix(prefix + '/' + nameOfBatfishFolderContainingL1)
                 .build(),
             _folder);
+    batfish.loadConfigurations(batfish.getSnapshot());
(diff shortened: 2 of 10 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7420)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 1200 at r1](https://reviewable.io/reviews/batfish/batfish/7420#-MkAVHse5o6-ytX7fml5:-MkAVHsfAVwdYDle8UsT:brixm7u) ([raw file](https://github.com/batfish/batfish/blob/80abede8e0de24f587ce22df8b5bd39bb6c1e911/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L1200)):*
> ```Java
> 
>   @Override
>   public Optional<SortedMap<String, Configuration>> loadSerializedConfigurations(
> ```

`loadSerializedConfigurations` seems too much about how it works (load serialized objects), rather than what it does.

We don't even have to use `load` here - may well not load if cached, for example.

* getReadyConfigurations
* getExistingConfigurations

...
?
___
*[projects/batfish/src/main/java/org/batfish/main/Batfish.java, line 1228 at r1](https://reviewable.io/reviews/batfish/batfish/7420#-MkAVvRn6M_vmK06UdJ4:-MkAVvRn6M_vmK06UdJ5:bmcxxve) ([raw file](https://github.com/batfish/batfish/blob/80abede8e0de24f587ce22df8b5bd39bb6c1e911/projects/batfish/src/main/java/org/batfish/main/Batfish.java#L1228)):*
> ```Java
>       if (configurations != null) {
>         _logger.debugf("Loaded configurations for %s off disk", snapshot);
>       } else if (parseIfNeeded) {
> ```

move this up, right after the cache miss. if !, return optioanl.empty().

Early exit makes the rest of the code simpler.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7420)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7420)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7435, opened 2021-09-24
Skip nonlocal interface dependency tracking for likely VPCs

anothermattbrown:
We don't model virtual port-channels (VPCs), and our dependency tracker can't express their dependencies. So if we encounter a portchannel that looks like a VPC (i.e. it has two neighbor interfaces, on different devices, and its members have neighbors on those same devices), don't track the dependencies on the neighbor portchannels.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7435)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7449, opened 2021-09-28
CP auto hide NAT: Match internal traffic without transforming

corinaminer:
Automatic hide NAT rules configured on `AddressRange` and `Network` objects should match all traffic whose source IP is in that object's IP space, but should only transform traffic destined outside that IP space.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckPointGatewayConfiguration.java
@@ -6,7 +6,7 @@ import static org.batfish.common.util.CollectionUtil.toImmutableMap;
 import static org.batfish.datamodel.FirewallSessionInterfaceInfo.Action.POST_NAT_FIB_LOOKUP;
 import static org.batfish.vendor.check_point_gateway.representation.CheckPointGatewayConversions.aclName;
 import static org.batfish.vendor.check_point_gateway.representation.CheckPointGatewayConversions.toIpAccessLists;
-import static org.batfish.vendor.check_point_gateway.representation.CheckpointNatConversions.automaticHideRuleTransformationFunction;
+import static org.batfish.vendor.check_point_gateway.representation.CheckpointNatConversions.automaticHideRuleTransformationFunctions;
 import static org.batfish.vendor.check_point_gateway.representation.CheckpointNatConversions.automaticStaticRuleTransformation;
 import static org.batfish.vendor.check_point_gateway.representation.CheckpointNatConversions.getManualNatRules;
 import static org.batfish.vendor.check_point_gateway.representation.CheckpointNatConversions.getOutgoingTransformations;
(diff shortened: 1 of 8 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7449)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java, line 378 at r1](https://reviewable.io/reviews/batfish/batfish/7449#-Mke6EKv9ttTnupzjsL5:-Mke6EKv9ttTnupzjsL6:bjl8q4c) ([raw file](https://github.com/batfish/batfish/blob/5bed736ecfc9dd1dc98a9b3f78c6752b40285109/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java#L378)):*
> ```Java
>    * the NAT settings cannot be converted.
>    */
>   static @Nonnull List<Function<Ip, Transformation>> automaticHideRuleTransformationFunctions(
> ```

Maybe I'm missing something, or the code is elsewhere.
Where do you handle IPv6?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7449)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java, line 378 at r1](https://reviewable.io/reviews/batfish/batfish/7449#-Mke6EKv9ttTnupzjsL5:-Mke7T1q6m-eGdoTi5XT:b-m8yzhj) ([raw file](https://github.com/batfish/batfish/blob/5bed736ecfc9dd1dc98a9b3f78c6752b40285109/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java#L378)):*

Maybe I'm missing something, or the code is elsewhere.
Where do you handle IPv6?
</blockquote></details>

I am not sure how a `NatSettings` representing an IPv6 hide rule would look. If the `hide-behind` was an IPv6 address, that would parse as an `UnhandledNatHideBehind`, and this function would return an empty list.

Is there a way to tell from the current VS datamodel whether `NatSettings` with method `HIDE` represent an IPv6 rule? If not, I'd push this to another PR as it's only tangentially related to this one.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7449)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java, line 378 at r1](https://reviewable.io/reviews/batfish/batfish/7449#-Mke6EKv9ttTnupzjsL5:-Mkhj60F-4aiXVIH1LDb:b-rgejqi) ([raw file](https://github.com/batfish/batfish/blob/5bed736ecfc9dd1dc98a9b3f78c6752b40285109/projects/batfish/src/main/java/org/batfish/vendor/check_point_gateway/representation/CheckpointNatConversions.java#L378)):*

I am not sure how a `NatSettings` representing an IPv6 hide rule would look. If the `hide-behind` was an IPv6 address, that would parse as an `UnhandledNatHideBehind`, and this function would return an empty list.

Is there a way to tell from the current VS datamodel whether `NatSettings` with method `HIDE` represent an IPv6 rule? If not, I'd push this to another PR as it's only tangentially related to this one.
</blockquote></details>

I think I was just confused on my first review. If you filter out ipv6 hosts/networks/etc before getting this far, then you can get by here with nothing (or add assertions for sanity). If you don't filter before you get here, this PR is not the place to do it.

PR #7453, opened 2021-09-29
ISP modeling via BGP peers

ratulm:
Includes https://github.com/batfish/batfish/pull/7475


dhalperi (commented):
> Reviewed 3 of 3 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7453)* status: all files reviewed, 7 unresolved discussions (waiting on @ratulm)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYeEaCHY-zPqAWBlD:-MkmYeEaCHY-zPqAWBlE:b-h4bq1l) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java#L22)):*
> ```Java
>   private static final String PROP_INTERFACE_NAMES = "interfaceNames";
> 
>   public static class Layer1 {
> ```

This makes me nervous. for the following reasons:

1. why multiple interface names on the same host, rather than multiple interfaces anywhere?
2. Since interface names is a `String`, I don't see where we'd evolve to add things like layer2 information. Having separate layer-1 and a layer-2 structures that users would have to keep in sync and that we'd have to join (and provide good debugging warnings./ error messages for) seems like a lot of work on both sides.

(I might just punt this for now, if we don't need it right away).
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmXXfeAVpb3JOb5NFp:-MkmXXfeAVpb3JOb5NFq:bhzpzv) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java#L15)):*
> ```
> peer address
> ```
sanity check: we don't need to handle dynamic peers because no one would do that with an ISP.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmXczf6dPpY3I_FMM7:-MkmXczf6dPpY3I_FMM8:bki8bj7) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java#L15)):*
> ```
> either an interface or peer address
> ```
I'd probably put peer address first, and clarify interface as (for an unnumbered peer)
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java, line 17 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmXlYr1fIcGXWGLWPM:-MkmXlYr1fIcGXWGLWPN:bw1sh89) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java#L17)):*
> ```
> Using an interface means that all BGP peers
>  * with (static) update source as the interface are being specified.
> ```

I do not think we should support this wildcarding. I'd propose that this class only be able to indicate a single peer, in this case a single unnumbered peer. I think that approach makes the most sense with the rest of the model.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java, line 67 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYOQdEwaI-_Y-9MAC:-MkmYOQdEwaI-_Y-9MAD:b-5qq7qs) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java#L67)):*
<details><summary><i alt="summary">Quoted 10 lines of code…</i></summary>

> ```
>     checkArgument(
>         iface != null || peerAddress != null,
>         "Either one of %s or %s should be specified",
>         PROP_INTERFACE,
>         PROP_PEER_ADDRESS);
>     checkArgument(
>         iface == null || peerAddress == null,
>         "Both %s or %s should not be specified",
>         PROP_INTERFACE,
>         PROP_PEER_ADDRESS);
> ```
</details>


These checks should go in the main constructor?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BorderInterfaceInfo.java, line 14 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmVNI63UwJ2hz1Fy1J:-MkmVNI63UwJ2hz1Fy1K:b-s9a46a) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BorderInterfaceInfo.java#L14)):*
Taking a stab at making it mildly more precise?

> ```
> simple L3 interface (e.g., Ethernet1),
> ```

This specification supports a physical layer-3 interface (e.g., Ethernet1), a layer-3 aggregate interface (e.g., Port-Channel1), or a layer-3 sub-interface of a physical or aggregate interface where the parent interface is directly wired to the ISP.

As {@link BorderInterface} can only be used for directly-connected ISPs, multiple BGP sessions on the interface or subinterfaces must be to the same directly-connected ISP.

Batfish will automatically create new layer-1 edges between the physical interfaces corresponding to each {@link BorderInterface} and its ISP. If any of these interfaces already have a layer-1 edge, ISP modeling will fail.

(for this peer? for the snapshot? something else?)
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java, line 14 at r2](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYTQTBVbt9R99a9wp:-MkmYTQTBVbt9R99a9wq:b-jzeo0f) ([raw file](https://github.com/batfish/batfish/blob/b0b048e306a5fb189b905ba9e001ab0bc9e0306c/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java#L14)):*
> ```Java
> 
> /** Configuration required to create ISPs for a given network snapshot. */
> public class IspConfiguration {
> ```

I guess here is where the primary user documentation goes? (Future work?)
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java, line 36 at r2](https://reviewable.io/reviews/batfish/batfish/7453#-MkmXDVTExP-TxZOMPSL:-MkmXDVTExP-TxZOMPSM:bjx5p0u) ([raw file](https://github.com/batfish/batfish/blob/b0b048e306a5fb189b905ba9e001ab0bc9e0306c/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java#L36)):*
> ```Java
>     this(borderInterfaces, ImmutableList.of(), filter, ispNodeInfos);
>   }
> ```

Please do not add new constructors, but rather update callers to provide all arguments. IntelliJ add parameter refactoring can help here.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7453)* status: all files reviewed, 2 unresolved discussions (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYeEaCHY-zPqAWBlD:-MknEYCHBh5pvYbiYz3F:bo1j1id) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java#L22)):*

This makes me nervous. for the following reasons:

1. why multiple interface names on the same host, rather than multiple interfaces anywhere?
2. Since interface names is a `String`, I don't see where we'd evolve to add things like layer2 information. Having separate layer-1 and a layer-2 structures that users would have to keep in sync and that we'd have to join (and provide good debugging warnings./ error messages for) seems like a lot of work on both sides.

(I might just punt this for now, if we don't need it right away).
</blockquote></details>

2. I was thinking that you'd do something completely separate for L2 connectivity, not spread information across L1 and L2. This allows L1 spec to be just wiring. If you want to be fancier than wiring, do something else entirely (which I don't know how to define yet and not something we need right away). 

1. I went back and forth on this a bit: Given it is L1 only and for enabling one session, will the ISP connect to multiple nodes ever?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmXXfeAVpb3JOb5NFp:-MknCqmtEtxL6muxHMp4:by39y7b) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerInfo.java#L15)):*

> ```
> peer address
> ```
sanity check: we don't need to handle dynamic peers because no one would do that with an ISP.
</blockquote></details>

Yeah, seems like an odd thing to do. Modeling without additional information also requires us to make up stuff like IP address on the other side, which has other implications. If we do move to support dynamic peers, that information might need to be made explicit. 
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java, line 14 at r2](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYTQTBVbt9R99a9wp:-MknBeSk2k8beZKdx9Tf:bau16xs) ([raw file](https://github.com/batfish/batfish/blob/b0b048e306a5fb189b905ba9e001ab0bc9e0306c/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/IspConfiguration.java#L14)):*

I guess here is where the primary user documentation goes? (Future work?)
</blockquote></details>

 Yeah.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7453)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYeEaCHY-zPqAWBlD:-MknScaMBcf1ZlIyf3Vs:b-ddt4gt) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java#L22)):*

2. I was thinking that you'd do something completely separate for L2 connectivity, not spread information across L1 and L2. This allows L1 spec to be just wiring. If you want to be fancier than wiring, do something else entirely (which I don't know how to define yet and not something we need right away). 

1. I went back and forth on this a bit: Given it is L1 only and for enabling one session, will the ISP connect to multiple nodes ever?
</blockquote></details>

"3/" I'm not sure why there wouldn't be. E.g., I connect to my backbone from two different border switches?

2/ I'm not sure what you are able to do with only the L1 connectivity. E.g., you can't create the ISP-side interfaces...

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7453)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java, line 22 at r1](https://reviewable.io/reviews/batfish/batfish/7453#-MkmYeEaCHY-zPqAWBlD:-Mkn_gsx7E9cK2xmADeP:bhtglnj) ([raw file](https://github.com/batfish/batfish/blob/20ef5eed8e548d05b2f9e57c79e7351c3627dfec/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/isp_configuration/BgpPeerConnectivity.java#L22)):*

"3/" I'm not sure why there wouldn't be. E.g., I connect to my backbone from two different border switches?

2/ I'm not sure what you are able to do with only the L1 connectivity. E.g., you can't create the ISP-side interfaces...
</blockquote></details>

Let's iterate offline

PR #7462, opened 2021-09-29
RangesIpSpace: an IpSpace implementation suitable for IP address ranges

dhalperi:
There are a number of places within Batfish where we need an IpSpace
corresponding to a set of IP addresses, a set of Prefixes, a set of Ranges, or
a combination of the above.

Today, we have two strategies for this: IpWildcardSetIpSpace or AclIpSpace.
Both of these are memory-intensive, encoding-intensive, and have O(n)
membership checks. Instead, introduce an IpSpace backed by a LongSpace. The
resulting structure merges contiguous ranges, stores them efficiently in a
a RangeSet, and provides logarithmic lookup of an IP address.

Not connected anywhere yet.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/IpSpaceToBDD.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/IpSpaceToBDD.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static org.batfish.common.util.CollectionUtil.toImmutableMap;
 
 import com.google.common.base.Suppliers;
+import com.google.common.collect.BoundType;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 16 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7462)* status: 0 of 16 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7462#-MkrcAvCAlTQrJZzH0AQ:-MkrcAvCAlTQrJZzH0AR:b-g86k51) (no related file):*
Working: if we're not careful, this will make performance worse. As it stands:

1/ complement is slow (bad implementation in NumberSpace)
2/ individual range->BDD calls are not cached, which makes a lot of redundant work compared to IpWildcardIpSpace, which materializes the BDD conversions of its members.

---

PR #7464, opened 2021-09-30
Prefix: elide copy in longestCommonPrefix

dhalperi:
We already have an Ip with value l1, and picking the minimum is more likely to
be the base of the returned prefix.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.google.common.collect.Ordering;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.Optional;
@@ -130,7 +131,7 @@ public final class Prefix implements Comparable<Prefix>, Serializable {
     long l2 = p2.getStartIp().asLong();
     long oneAtFirstDifferentBit = Long.highestOneBit(l1 ^ l2);
     int lengthInCommon = MAX_PREFIX_LENGTH - 1 - Long.numberOfTrailingZeros(oneAtFirstDifferentBit);
-    return Prefix.create(Ip.create(l1), lengthInCommon);
+    return Prefix.create(Ordering.natural().min(p1.getStartIp(), p2.getStartIp()), lengthInCommon);
   }
 
   @Override


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7464)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @anothermattbrown, @arifogel, and @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java, line 133 at r1](https://reviewable.io/reviews/batfish/batfish/7464#-Mkrb1ML6oMvQd0y8euI:-Mkrb1ML6oMvQd0y8euJ:b-blgxmc) ([raw file](https://github.com/batfish/batfish/blob/0325786bb5265ac77dafa21e5626302e3e46684f/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java#L133)):*
> ```
> Ip.create(l1)
> ```
FYI this is just `p1.getStartIp()`, no need to recreate.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7464)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7464)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @arifogel, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java, line 134 at r1](https://reviewable.io/reviews/batfish/batfish/7464#-Mkrphrj4c8zN3TuxAeQ:-Mkrphrj4c8zN3TuxAeR:b-qhd1t5) ([raw file](https://github.com/batfish/batfish/blob/da611b7d9ca9e78906f13fc4af81e557681d22dc/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java#L134)):*
> ```Java
>     long oneAtFirstDifferentBit = Long.highestOneBit(l1 ^ l2);
>     int lengthInCommon = MAX_PREFIX_LENGTH - 1 - Long.numberOfTrailingZeros(oneAtFirstDifferentBit);
>     return Prefix.create(Ordering.natural().min(p1.getStartIp(), p2.getStartIp()), lengthInCommon);
> ```

I don't think you need to take the min -- i think adding `/lengthInCommon` will produce the same prefix for either start ip

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7464)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java, line 134 at r1](https://reviewable.io/reviews/batfish/batfish/7464#-Mkrphrj4c8zN3TuxAeQ:-MkrqJFd6rc6J9--f3XR:b-d9pzc9) ([raw file](https://github.com/batfish/batfish/blob/da611b7d9ca9e78906f13fc4af81e557681d22dc/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Prefix.java#L134)):*

I don't think you need to take the min -- i think adding `/lengthInCommon` will produce the same prefix for either start ip
</blockquote></details>

Inside `Prefix.create`, it will canonicalize the `ip` (any Ip in the prefix) into `startIp`, aka the lowest Ip in the prefix.

Using the min increases the probability that that canonicalization will be a no-op, hence not create a new Ip.

Example: merging

1.0.1.0/24 and 1.0.0.0/24 into 1.0.0.0/23. If we pass in 1.0.1.0, then the Prefix will have to create 1.0.0.0. If we pass in 1.0.0.0, it will save an Ip creation.

PR #7466, opened 2021-09-30
A10: handle hybrid vlan syntax

sfraint:
Handle `vlan` syntax where interfaces are referenced by a list and range on the same line, which appears to be allowed in `acos v2`. Note: list and range references were previously allowed, but only on separate lines.

e.g.
```
vlan 3
 untagged ethernet 1 ethernet 3 to 4
!
```
(which corresponds to `Ethernet 1`, `Ethernet 3`, and `Ethernet 4`)


dhalperi (commented):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7466)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

PR #7468, opened 2021-09-30
AclIpSpace: optimize complement of complement

dhalperi:
Because AclIpSpace#intersection and AclIpSpace#difference both use complement,
it's not uncommon that we complement a complement. Optimize in this case to
the original AclIpSpace.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7468)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7501, opened 2021-10-06
AllActiveSources location specifier

ratulm:


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7501)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7501#-MlLkeNwAiim-ooOl5BR:-MlLkeNwAiim-ooOl5BS:b-odagkt) (no related file):*
Is this needed? Looking here, it seems only active interfaces are considered sources:
https://github.com/batfish/batfish/blob/master/projects/batfish-common-protocol/src/main/java/org/batfish/specifier/LocationInfoUtils.java#L98-L104

---

PR #7619, opened 2021-10-29
DataplaneUtil: parallelism and redundancy improvements

dhalperi:
1. VirtualRouter#getBgpRoutes/getEvpnRoutes can be expensive.
   (The result is recomputed in most cases.) So only call it once,
   and use the prior result.
2. The Bgp/Evpn routes tables do not need empty cells.
3. Parallelize computation of tables, especially when the
   "cell value" function can be expensive.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7619)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @sfraint)

PR #7624, opened 2021-10-30
PacketPolicyToBdd: rewrite for efficiency

anothermattbrown:
PacketPolicyToBdd now generates intermediate graph states. It makes some effort to produce efficient graph without relying on optimization (which is not run by default). In particular, avoid generating long chains of nodes (with single in- and out-edges), avoid identity transitions when possible, etc.


dhalperi (commented):
> Reviewed 3 of 17 files at r1, 1 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 3 of 18 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

___
*[projects/symbolic/src/main/java/org/batfish/symbolic/state/StateExprVisitor.java, line 53 at r1](https://reviewable.io/reviews/batfish/batfish/7624#-MnDiV6zDA3FXhPf0BNI:-MnDiV6zDA3FXhPf0BNJ:bhzr0b9) ([raw file](https://github.com/batfish/batfish/blob/3de5ad00dc2f8a2d014ada46f8fd35f7f12b7d0e/projects/symbolic/src/main/java/org/batfish/symbolic/state/StateExprVisitor.java#L53)):*
> ```Java
>   R visitOriginateVrf(OriginateVrf originateVrf);
> 
>   R visitPbrFibLookup(PbrFibLookup pbrFibLookup);
> ```

I'm confused how this is still here but `PbrFibLookup` state is deleted.

And similar issues in BDDRAF

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 3 of 18 files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/symbolic/src/main/java/org/batfish/symbolic/state/StateExprVisitor.java, line 53 at r1](https://reviewable.io/reviews/batfish/batfish/7624#-MnDiV6zDA3FXhPf0BNI:-MnDigcbAKgIiOVqtlTZ:b-29f2kt) ([raw file](https://github.com/batfish/batfish/blob/3de5ad00dc2f8a2d014ada46f8fd35f7f12b7d0e/projects/symbolic/src/main/java/org/batfish/symbolic/state/StateExprVisitor.java#L53)):*

I'm confused how this is still here but `PbrFibLookup` state is deleted.

And similar issues in BDDRAF
</blockquote></details>

Lol NM.

dhalperi (commented):
> Reviewed 4 of 17 files at r1, 5 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 12 of 18 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 909 at r2](https://reviewable.io/reviews/batfish/batfish/7624#-MnDjl6e2mA9FL1xSNzR:-MnDjl6e2mA9FL1xSNzS:bhw27gz) ([raw file](https://github.com/batfish/batfish/blob/13f10a8cfaf6d0b41234910c8847796c3f9dd73b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L909)):*
> ```
> ipAccessListToBddForNode(config)
> ```
worth extracting this and similar out of loop? `config` is the same in this lambda IIUC.
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 2011 at r2](https://reviewable.io/reviews/batfish/batfish/7624#-MnDkjsR3k0j4bqLUiTq:-MnDkjsS3YmzlnvIWzIM:b-9a7430) ([raw file](https://github.com/batfish/batfish/blob/13f10a8cfaf6d0b41234910c8847796c3f9dd73b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L2011)):*
> ```Java
>     public Stream<Edge> visitFibLookupOverrideLookupIp(FibLookupOverrideLookupIp fibLookup) {
>       // Currently unsupported
>       return Stream.of();
> ```

already not supported, not needed because of better design elsewhere, new bug, or something else?

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 12 of 18 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 2011 at r2](https://reviewable.io/reviews/batfish/batfish/7624#-MnDkjsR3k0j4bqLUiTq:-MnDvzm99h2e9RogckeS:bo8f94m) ([raw file](https://github.com/batfish/batfish/blob/13f10a8cfaf6d0b41234910c8847796c3f9dd73b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L2011)):*

already not supported, not needed because of better design elsewhere, new bug, or something else?
</blockquote></details>

already not supported. this comment was in PacketPolicyToBdd, but we'll be better able to implement it here

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 12 of 18 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java, line 909 at r2](https://reviewable.io/reviews/batfish/batfish/7624#-MnDjl6e2mA9FL1xSNzR:-MnDwRAT0Tnirz4cm6Cr:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/13f10a8cfaf6d0b41234910c8847796c3f9dd73b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityAnalysisFactory.java#L909)):*

> ```
> ipAccessListToBddForNode(config)
> ```
worth extracting this and similar out of loop? `config` is the same in this lambda IIUC.
</blockquote></details>

done

dhalperi (commented):
> Reviewed 3 of 17 files at r1, 1 of 7 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7624#-MnEL_cZ1ju_xeZSEkb_:-MnEL_cZ1ju_xeZSEkba:b-nr1dzs) (no related file):*
Overall lovely, but a few qs inline.

---
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 74 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEGAAK5dICysx53eXR:-MnEGAAK5dICysx53eXS:b-k908an) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L74)):*
> ```
> PacketPolicyToBdd
> ```
nit: `BddPacketPolicy`
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 75 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEGHIkCyoMReXqrbdM:-MnEGHIkCyoMReXqrbdN:b-jich9x) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L75)):*
> ```
>  Examine the result of the conversion using methods such as
> ```
nit: did you mean to delete the whole sentence?
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 202 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEK5dPA7OxIktsCegr:-MnEK5dPA7OxIktsCegs:bydrkn3) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L202)):*
> ```
> _pathConstraint.getFactory().one();
> ```

I'm surely missing something, but wondering why would this not be `_pathConstraint.or(elseConstraint)`? (Which I guess would be original path constraint, which we lost above?)
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 227 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEL95k7e-J4EKBGU22:-MnEL95k7e-J4EKBGU23:bd2nq2i) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L227)):*
> ```
> _pathConstraint.diff(permitBdd)
> ```
Fine as-is. but wondering about the tradeoffs vs `permitBdd.not()` (or using PermitAndDenyBdds). I guess simpler BDD vs more likely to get to ZERO?
___
*[projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java, line 42 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnE9krQ42rwT6bdu0DX:-MnE9krR6HPkQ4Rczdv7:bvda25) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java#L42)):*
> ```
> edge
> ```

i'm worried about later confusion. One could reasonably guess this could be `any Transition` as `IDENTITY Transition`.

Consider renaming to `identityEdge`, but it's fine as-is.
___
*[projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java, line 43 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnE9DaLApR8f7Ynjg2s:-MnE9DaLApR8f7Ynjg2t:bokdts6) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java#L43)):*
> ```Java
>    */
>   public static Matcher<Edge> edge(StateExpr preState, StateExpr postState) {
>     return allOf(
> ```

FYI I tend to chain these - simpler code, and fewer places to fixup later.

e.g., this could be `return edge(preState, postState, IDENTITY)`
___
*[projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java, line 16 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEHPkeA2nLziYnugMR:-MnEHPkeA2nLziYnugMS:b8a7nu0) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java#L16)):*
> ```Java
>     _hostname = hostname;
>     _policyName = policyName;
>     _action = action;
> ```

It looks to me like there's an inbuilt assumption that the same `PacketPolicy` is used in only one VRF. Is that enforced anywhere?

It doesn't seem obvious to me from a VS standpoint that this is required.

(E.g., if `_action` is FibLookup, that is "in the vrf of the interface", right?)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: 14 of 18 files reviewed, 4 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 202 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEK5dPA7OxIktsCegr:-MnIEQQi-BmRkFgIaAjZ:b-261i2l) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L202)):*

> ```
> _pathConstraint.getFactory().one();
> ```

I'm surely missing something, but wondering why would this not be `_pathConstraint.or(elseConstraint)`? (Which I guess would be original path constraint, which we lost above?)
</blockquote></details>

I think the variable name was misleading. Renamed, and added a comment. The `or` is the captured in the two edges we just created. I don't think  we want to keep propagating the `or` constraint forward, as it could lead to larger BDDs that are more expensive to operate on. We may be missing some opportunities to prune, but I'm assuming that the common case is that all (or at least most) paths through the policy are viable. 

My strategy here is to generate graphs that don't have identity edges or states with single in and out edges that are both constraints. There may be opportunities for further optimizations, but I'm happy to leave those to the optimizer (unless they're simple and efficient to handle here).
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 227 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEL95k7e-J4EKBGU22:-MnIGja_2oz4qm3hlvMu:b74p4ei) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L227)):*

> ```
> _pathConstraint.diff(permitBdd)
> ```
Fine as-is. but wondering about the tradeoffs vs `permitBdd.not()` (or using PermitAndDenyBdds). I guess simpler BDD vs more likely to get to ZERO?
</blockquote></details>

I probably don't understand your comment, but we have to apply `_nextEdgeConstraint` somewhere for correctness. The choices are to add another state before branching for the the filter or not. I don't know if either is strictly better in all cases. 
___
*[projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java, line 42 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnE9krQ42rwT6bdu0DX:-MnIHuMn9z343p_kSmKq:bwg588l) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java#L42)):*

> ```
> edge
> ```

i'm worried about later confusion. One could reasonably guess this could be `any Transition` as `IDENTITY Transition`.

Consider renaming to `identityEdge`, but it's fine as-is.
</blockquote></details>

removed the method
___
*[projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java, line 43 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnE9DaLApR8f7Ynjg2s:-MnIHy-67pAYhl4mNmhO:bwg588l) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/test/java/org/batfish/bddreachability/EdgeMatchers.java#L43)):*

FYI I tend to chain these - simpler code, and fewer places to fixup later.

e.g., this could be `return edge(preState, postState, IDENTITY)`
</blockquote></details>

removed the method
___
*[projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java, line 16 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEHPkeA2nLziYnugMR:-MnIJaRm8bPA0t8_j-4s:b-ukyzrp) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java#L16)):*

It looks to me like there's an inbuilt assumption that the same `PacketPolicy` is used in only one VRF. Is that enforced anywhere?

It doesn't seem obvious to me from a VS standpoint that this is required.

(E.g., if `_action` is FibLookup, that is "in the vrf of the interface", right?)
</blockquote></details>

yikes, good catch. thanks

dhalperi (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7624)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 202 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEK5dPA7OxIktsCegr:-MnIqnfN6YK1FBGQTHOf:b-9dudkk) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L202)):*

I think the variable name was misleading. Renamed, and added a comment. The `or` is the captured in the two edges we just created. I don't think  we want to keep propagating the `or` constraint forward, as it could lead to larger BDDs that are more expensive to operate on. We may be missing some opportunities to prune, but I'm assuming that the common case is that all (or at least most) paths through the policy are viable. 

My strategy here is to generate graphs that don't have identity edges or states with single in and out edges that are both constraints. There may be opportunities for further optimizations, but I'm happy to leave those to the optimizer (unless they're simple and efficient to handle here).
</blockquote></details>

Makes sense now, with new name and comment :)
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 227 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEL95k7e-J4EKBGU22:-MnIrSnlEGYFE5A30atP:b-2yaltt) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L227)):*

I probably don't understand your comment, but we have to apply `_nextEdgeConstraint` somewhere for correctness. The choices are to add another state before branching for the the filter or not. I don't know if either is strictly better in all cases. 
</blockquote></details>

Makes sense now with new name and comment
___
*[projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java, line 16 at r3](https://reviewable.io/reviews/batfish/batfish/7624#-MnEHPkeA2nLziYnugMR:-MnIrXOWCEde61l6mqeq:bwzkccl) ([raw file](https://github.com/batfish/batfish/blob/ed041bdb2b4235dc597d84320b3352b8c509430f/projects/symbolic/src/main/java/org/batfish/symbolic/state/PacketPolicyAction.java#L16)):*

yikes, good catch. thanks
</blockquote></details>

This is still not resolved, right? Leaving open.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 56 at r5](https://reviewable.io/reviews/batfish/batfish/7624#-MnRlQI6AaYC3SU9qSk3:-MnRlQI74GbQ6NeTS0dX:b-5aw8xi) ([raw file](https://github.com/batfish/batfish/blob/f1b7382404d554d1c9b7b6a8e278c2472d8c9a97/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L56)):*
> ```Java
>   private PacketPolicyStatement _currentStatement;
> 
>   public static final class BddPacketPolicy {
> ```

might be worth documenting the semantics of this thing:

* every edge has as its preState one of the PacketPolicy states in this vrf
* every edge has as its postState one of the other PP states or a PPA.

(?)

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 206 at r5](https://reviewable.io/reviews/batfish/batfish/7624#-MnRoCegCvElX2CEgIAP:-MnRoCegCvElX2CEgIAQ:b-addeg1) ([raw file](https://github.com/batfish/batfish/blob/f1b7382404d554d1c9b7b6a8e278c2472d8c9a97/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L206)):*
> ```Java
>       // initialize pathConstraint for then branch
>       _nextEdgeConstraint = thenConstraint;
>       visitStatements(ifStmt.getTrueStatements());
> ```

Seems like if the else branch is false we don't need to create a new state, even if then falls through.

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 217 at r5](https://reviewable.io/reviews/batfish/batfish/7624#-MnRpKkf6T_AqbJw3NDS:-MnRpKkg1noES6cfEtCv:b-mcgy15) ([raw file](https://github.com/batfish/batfish/blob/f1b7382404d554d1c9b7b6a8e278c2472d8c9a97/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L217)):*
> ```Java
>         _nextEdgeConstraint = _nextEdgeConstraint.getFactory().one();
>       } else {
>         _nextEdgeConstraint = elseConstraint;
> ```

The then branch doesn't fall through, so we need to continue in the current statement (`inSt`). That means we need to reset `currentStatement()`?

dhalperi:
___
*[projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java, line 215 at r5](https://reviewable.io/reviews/batfish/batfish/7624#-MnRpeahB_3QXTVWAQnV:-MnRpeahB_3QXTVWAQnW:b-ww68dl) ([raw file](https://github.com/batfish/batfish/blob/f1b7382404d554d1c9b7b6a8e278c2472d8c9a97/projects/batfish/src/main/java/org/batfish/bddreachability/PacketPolicyToBdd.java#L215)):*
> ```Java
>         addEdge(inSt, nextSt, elseConstraint);
>         addEdge(fallThroughSt, nextSt, _nextEdgeConstraint);
>         _nextEdgeConstraint = _nextEdgeConstraint.getFactory().one();
> ```

Documenting a discussion:

My belief is that after this code runs: `currentStatement()` is `nextSt`. So could this instead be [then falls through, or guard is not matched]: `_nextEdgeConstraint = _nextEdgeConstraint.or(elseConstraint);`, which would propagate a complicated BDD but might allow us to find some unreachable states during construction ("Unreachable NAT rules", for example.)

But this may be harmful: it trades (significant) complexity in the graph for possibility of optimization. The propagation algorithm will instead find zero bdds quickly (in the forward direction at least) if they exist.

PR #7633, opened 2021-11-01
PAN: parse active-active-device-binding and warn if needed

dhalperi:


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7633)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2554 at r1](https://reviewable.io/reviews/batfish/batfish/7633#-MnS9f-M18u678noXwCp:-MnS9f-M18u678noXwCq:b-8xjovg) ([raw file](https://github.com/batfish/batfish/blob/83a67c908cb68a891312120f1602a73ea1d49ac5/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2554)):*
> ```
> warn(ctx, "Batfish currently models this as active-active-device-binding both");
> ```
Our modeling is correct if the same rule is repeated for both `0` and `1`, as is recommended in the docs and what we've seen in practice. Could we suppress the warning in that case? I guess it would require making this a conversion-time warning.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7633)* status: all files reviewed, 1 unresolved discussion (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java, line 2554 at r1](https://reviewable.io/reviews/batfish/batfish/7633#-MnS9f-M18u678noXwCp:-MnSAfUYD_kCzVW8kihz:bcio2j1) ([raw file](https://github.com/batfish/batfish/blob/83a67c908cb68a891312120f1602a73ea1d49ac5/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2554)):*

> ```
> warn(ctx, "Batfish currently models this as active-active-device-binding both");
> ```
Our modeling is correct if the same rule is repeated for both `0` and `1`, as is recommended in the docs and what we've seen in practice. Could we suppress the warning in that case? I guess it would require making this a conversion-time warning.
</blockquote></details>

Hmm. I guess we'd have to find some rule equivalence test. (And also, not include both versions of the rule)?

Maybe @sfraint knows what's up and can take this over.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7633)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7633#-MnXu_OS0roYuVWZ08Va:-MnXu_OZ20ddRijjQZ7s:bwi7k11) (no related file):*
see https://github.com/batfish/batfish/pull/7650, supersedes this PR

---

PR #7644, opened 2021-11-02
FibEntry: only track top level route for now

dhalperi:
The resolution steps will likely be needed later, but right now only the top
level route is used.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7644)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7652, opened 2021-11-03
Routes: remove confusing Next_Hop column

### LINKED ISSUE #1862: Routes2: better strategy for determining next-hop-node
Currently we take next hop IP of the route and attempt to pull out a node using computed `ipOwners`. 
If there are more and one owner of the ip, we pick the lexicographically lower node to display in the table.
This works great in the common case (1 owner per IP) but is not always correct.

dhalperi:
Fix #1862. tl;dr: this column is based on IP Ownership of next hop IP, but this has many problems

- Route with no next hop IP has no hostname to put here
- Route with next hop IP outside the snapshot has no hostname to put here
- In networks with IP reuse, there could be multiple owners of the next hop
  IP, so no hostname to put here. (Traceroute, say, is too expensive to run
  inside this question and may not even reach one of the owners.)
- Routes with a 'virtual next hop IP' that is only used to resolve an interface
  for routing, do not have a next hop IP that is owned in the snapshot. So no
  hostname to put here.
- The hostname used is based on next hop IP, but this is by no means guaranteed
  to be the next (L3) node to route/forward the packet.

Future work will replace this column with something more directly
related to forwarding, and/or add another question to get at the FIBs
instead.


progwriter (commented):
:star_struck: 

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7652#-:-MnayAqv5NdPjTYfZpHk:be5ck22)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7652)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

PR #7659, opened 2021-11-06
Edges: add EdgeType.USER_PROVIDED_LAYER1

anothermattbrown:
Returns the canonicalized user-provided layer1 topology.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyProvider.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyProvider.java
@@ -80,6 +80,16 @@ public interface TopologyProvider {
   }
 
   /**
+   * Returns the user-provided {@link Layer1Topology}, potentially modified to clean up {@link
+   * Layer1Node nodes} that indicated interfaces in non-canonical form.
+   */
+  default @Nonnull Optional<Layer1Topology> getUserProvidedLayer1Topology(
+      NetworkSnapshot networkSnapshot) {
+    return Optional.of(getLayer1Topologies(networkSnapshot).getUserProvidedL1())
+        .filter(l1 -> !l1.isEmpty());
+  }
+
+  /**
    * Returns the layer-3 {@link Topology} corresponding to the converged {@link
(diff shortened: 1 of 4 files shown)


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7659)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7659#-MnmfigU97NX6b_-iD6W:-MnmfigU97NX6b_-iD6X:byu2b3v) (no related file):*
So .... :) 

There are many types of L1 edges -- raw user, cleaned user, synthesized L1 (ISP modeling), combined L1, logical L1, ... these are overlapping sets. We've also talked about "imputed L1 edges" -- back propagating L3 inference. (There is also the active/inactive dimension to things, but a separable concern. likely)

Is a separate question how we should be handling this or should we add an L1EdgeType parameter and column (like we do for BGP RIBs) and users can specify the categories they want? That may be more extensible. 

(l1edge is itself a type at the moment but that fact is hidden from users.) 

---

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7659)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7659#-MnmfigU97NX6b_-iD6W:-Mnmkxq227tcCMmYGfek:b-2olzb1) (no related file):*

So .... :) 

There are many types of L1 edges -- raw user, cleaned user, synthesized L1 (ISP modeling), combined L1, logical L1, ... these are overlapping sets. We've also talked about "imputed L1 edges" -- back propagating L3 inference. (There is also the active/inactive dimension to things, but a separable concern. likely)

Is a separate question how we should be handling this or should we add an L1EdgeType parameter and column (like we do for BGP RIBs) and users can specify the categories they want? That may be more extensible. 

(l1edge is itself a type at the moment but that fact is hidden from users.) 

</blockquote></details>

Discussed offline: This is a net positive for now. 

---

PR #7660, opened 2021-11-06
NXOS BGP use independent network policy

arifogel:
- use indendependent network policy for `network` statements instead of
  combined redistribution policy
- allows multiple BGP RIB entries to be produced from single route
  applicable to both a `network` and a `redistribute` statement per
  device behavior


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7660)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7665, opened 2021-11-11
Arista: BGP no redistribute, route-map variant

ratulm:
1. "no redistribute" syntax permits route-map specification (which is ignored by the box). 
2. add tests
3. add missing extractor logic for "no redistribute attached-host"


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7665)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7672, opened 2021-11-12
FRR: Dynamic BGP neighbors

ratulm:


projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java
```diff
@@ -1198,6 +1208,32 @@ public void exitSbb_max_med_administrative(Sbb_max_med_administrativeContext ctx
...
+              .getNeighbors()
+              .computeIfAbsent(
+                  listenRange.toString(), (name) -> new BgpPassive6Neighbor(name, listenRange));
+    }
+    bgpNeighbor.setPeerGroup(ctx.name.getText());
+  }
+
+  @Override
+  public void exitSbbl_limit(Sbbl_limitContext ctx) {
+    warn(ctx, "Batfish does not limit the number sessions for passive BGP neighbors");
```
ton31337, line 1234: passive command shouldn't be bundled with bgp listen.

ratulm: Passive command is not handled in this PR. 

What are the semantics of the "listen range" command? FRR documentation (https://docs.frrouting.org/en/latest/bgp.html#clicmd-bgp-listen-range-A.B.C.D-M-X-X-X-X-M-peer-group-PGNAME) is thin, but my understanding is that the router will accept connections from the specified prefix, and it will not initiate connections. Such listening peers are "passive" in batfish terminology (some vendors call them passive, and some call them dynamic).

Is my understanding correct?

ton31337: Correct, just as a reader I was misreading the warning message :) I'd say the `passive` keyword should be replaced with `dynamic`.

projects/batfish/src/main/java/org/batfish/representation/cumulus/BgpPassiveNeighbor.java
```diff
@@ -0,0 +1,18 @@
+package org.batfish.representation.cumulus;
+
+import javax.annotation.Nonnull;
+import org.batfish.datamodel.Prefix;
+
+/** Passive BGP neighbor identified by a listen range prefix. */
+public class BgpPassiveNeighbor extends BgpNeighbor {
```
ton31337, line 7: I'd call it BgpDynamicNeighbor instead to avoid confusion with the `passive` command in the future.

projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java
ratulm: I see - happy to change. We try to use vendor-specific terminology in this part of the code. 

Do you have a pointer handy that talks about FRR terminology?

ton31337: Basically, the terminology is kinda a copy/paste from the Cisco in FRR regarding `bgp listen` command.

ton31337: https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/iproute_bgp/configuration/xe-16/irg-xe-16-book/bgp-dynamic-neighbors.html (BGP Dynamic Neighbors)

projects/batfish/src/main/java/org/batfish/representation/cumulus/BgpPassiveNeighbor.java
ratulm: What exactly does the `passive` command do? Is it independent of the neighbor being dynamic? 

Related: Can an unnumbered peer be passive?

ton31337: Yes, it's absolutely independent command. Can be unnumbered as well.

projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java
ratulm: Ok, thanks. Changed the terminology.

arifogel (commented):
> Reviewed 5 of 10 files at r1, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7672)* status: all files reviewed, 2 unresolved discussions (waiting on @ratulm and @ton31337)

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7672)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel and @ton31337)

___
*[projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4, line 212 at r1](https://reviewable.io/reviews/batfish/batfish/7672#-MoHemJH7Asup-Q8a7sp:-MoHemJH7Asup-Q8a7sq:b-hz50j1) ([raw file](https://github.com/batfish/batfish/blob/998a4d146cf72f451933e3c62540f67fc29e3ca6/projects/batfish/src/main/antlr4/org/batfish/grammar/cumulus_frr/CumulusFrrLexer.g4#L212)):*
> ```ANTLR
> 
> IBGP: 'ibgp';
> ```

made alphabetic
___
*[projects/batfish/src/main/java/org/batfish/representation/cumulus/BgpPassiveNeighbor.java, line 7 at r2](https://reviewable.io/reviews/batfish/batfish/7672#-MoHegG4pkVeOg3oGbWd-r2-7:-MoLAec5-1Or7nFNUmqS:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/53410335a63706cdaa51e6e6fa9f9938b379c273/projects/batfish/src/main/java/org/batfish/representation/cumulus/BgpPassiveNeighbor.java#L7)):*
<details><summary><i>Previously, ton31337 wrote…</i></summary><blockquote>

Yes, it's absolutely independent command. Can be unnumbered as well. 
</blockquote></details>

Done.

PR #7675, opened 2021-11-12
FRR: Warn that TTL is not enforced if TTL number is ommited

ton31337:
Because now it confuses.

```
ebgp-multihop 123  [sbn_property sbn_ip sb_neighbor bgp_inner s_b...  Neighbor recognized as ebgp-multihop, but dist
```

Signed-off-by: Donatas Abraitis <donatas.abraitis@gmail.com>

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cumulus_frr/CumulusFrrConfigurationBuilder.java
@@ -1296,7 +1296,7 @@ public class CumulusFrrConfigurationBuilder extends CumulusFrrParserBaseListener
 
   @Override
   public void exitSbnp_ebgp_multihop(Sbnp_ebgp_multihopContext ctx) {
-    if (ctx.num != null) {
+    if (ctx.num == null) {
       warn(
           ctx.getParent(),
           "Neighbor recognized as ebgp-multihop, but distance limit is not enforced");


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7675)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ratulm and @ton31337)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7675#-MoK94zPB8OYjqXXM7OS:-MoK94zPB8OYjqXXM7OT:bkae6v) (no related file):*
Unless I am missing something, this change does not seem right.  When an ebgp session is multihop, Batfish does not enforce how far the peer can be. So, if the config has TTL it in it, it is not honored. The warning before the PR is making the user aware of it.  If the TTL is not in the config, Batfish's behavior is fully accurate. 

By this PR, the warning will disappear when the TTL is present and appear when it is not present. That is backwards because Batfish behavior is accurate for the no limit case. 

Perhaps the warning was misleading? In that case, welcome improvements there. 

---

ton31337:
In FRR, if you enter `neighbor X ebgp-multihop` it's the same as `neighbor X ebgp-multihop 255`.

ratulm:
Right. so Batfish implements `neighbor X ebgp-multihop` and, prior to this PR, warns if config has `neighbor X ebgp-multihop 3` (because the TTL limit is not implemented). Is that not right?

ton31337:
It should warn only if `neighbor X ebgp-multihop` specified, but not warn if `neighbor X ebgp-multihop <num>`.

PR #7676, opened 2021-11-12
FRR: Redistribution into BGP

ton31337:


ratulm:
@ton31337 - please add comments to this PR about what problem it is fixing.

ton31337:
All good is here, a bit confused.

PR #7688, opened 2021-11-15
Refactor VRF leaking config

corinaminer:
No functional changes, just reorganizes VRF leaking VI model. Creates an umbrella class that can contain either BGP VRF-leaking configs or main RIB VRF-leaking configs, and separate those types of VRF-leaking configs into separate classes. Adds a field for an instance of this class in `Vrf` and removes the old field that contained a list of leak configs.


arifogel (commented):
> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7688)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 1869 at r1](https://reviewable.io/reviews/batfish/batfish/7688#-MoX23ilAOeiRfdeJbX5:-MoX23im5hDObZxntG66:bsxogrs) ([raw file](https://github.com/batfish/batfish/blob/1ef874b19ec610ae232ecc23ceeeb05bffde3181/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L1869)):*
> ```Java
>                       // admin)
>                       .setAdmin(DEFAULT_EBGP_ADMIN)
>                       // TODO: this should be export RTs, not single import RT
> ```

Are you leaving it like this in this PR because this is functionally what was happening before?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java, line 49 at r1](https://reviewable.io/reviews/batfish/batfish/7688#-MoX3PMtBTgFJ7FcfIg8:-MoX3PMtBTgFJ7FcfIg9:b-5nl39o) ([raw file](https://github.com/batfish/batfish/blob/1ef874b19ec610ae232ecc23ceeeb05bffde3181/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java#L49)):*
> ```Java
>    */
>   @JsonProperty(PROP_LEAK_AS_BGP)
>   public boolean getLeakAsBgp() {
> ```

Why is this here? All vendors I am aware of have the option to leak via BGP or directly. They are not mutually exclusive to my knowledge. Why not just handle the different leak types completely independently?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7688)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java, line 1869 at r1](https://reviewable.io/reviews/batfish/batfish/7688#-MoX23ilAOeiRfdeJbX5:-MoZkGRPCXV6F1c-gkns:b-mun4sb) ([raw file](https://github.com/batfish/batfish/blob/1ef874b19ec610ae232ecc23ceeeb05bffde3181/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConversions.java#L1869)):*

Are you leaving it like this in this PR because this is functionally what was happening before?
</blockquote></details>

yes, copied code and todo from the helper function it was in before
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java, line 49 at r1](https://reviewable.io/reviews/batfish/batfish/7688#-MoX3PMtBTgFJ7FcfIg8:-MoZkYD25HPHBDFALP4h:bsptx0s) ([raw file](https://github.com/batfish/batfish/blob/1ef874b19ec610ae232ecc23ceeeb05bffde3181/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java#L49)):*
> All vendors I am aware of have the option to leak via BGP or directly

We don't support that. So far in Batfish they are mutually exclusive -- Juniper only leaks directly between main RIBs, while Cisco-like devices leak only between BGP RIBs. theoretically both could be supported by a single vendor, and if we move to support that, this can go away.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7688)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java, line 49 at r1](https://reviewable.io/reviews/batfish/batfish/7688#-MoX3PMtBTgFJ7FcfIg8:-Mo_4gGXCQZKJ9nJd0jp:b69bklo) ([raw file](https://github.com/batfish/batfish/blob/1ef874b19ec610ae232ecc23ceeeb05bffde3181/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/VrfLeakConfig.java#L49)):*

> All vendors I am aware of have the option to leak via BGP or directly

We don't support that. So far in Batfish they are mutually exclusive -- Juniper only leaks directly between main RIBs, while Cisco-like devices leak only between BGP RIBs. theoretically both could be supported by a single vendor, and if we move to support that, this can go away.
</blockquote></details>

Let's defer, but actually do it.

PR #7690, opened 2021-11-15
Only allow single BGP entry from `network` and `redistribute`

arifogel:
- `network` and `redistribute` statements may only contribute up to one route each to the BGP RIB for a given prefix
- track the mechanism by whcih BGP entries are introduced:
  - LEARNED - from a neighbor
  - GENERATED - via an aggregate or otherwise not directly redistributed from an existing route
  - NETWORK - from a `network` statement
  - REDISTRIBUTE - from a `redistribute` statement
- `network` and `redistribute` statements may have independent semantics per vendor on which route they prefer contribute
  - depending on vendor, each may prefer routes with lowest or highest NHIP
- on vendors where you may get a route for a prefix from both `network` and `redistribute`, support the ability to strictly prefer one or the other mechanism, such that the less preferred mechanism goes to backup

Follow-on work:
- properly distinguish REDISTRIBUTE and NETWORK mechanisms on devices that do not have an independent network policy
- apply correct BGP route attributes on devices where they come from more than one redistributable route for the same prefix
- support restricting routes from consideration where one with a lower or higher nhip exists but does not pass policy


corinaminer (commented):
> Reviewed 18 of 18 files at r1, 13 of 13 files at r2, 20 of 20 files at r3, 8 of 8 files at r4, 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: all files reviewed, 16 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 360 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_zlDc97lbj2jDgmKl:-Mo_zlDc97lbj2jDgmKm:b-lshg0u) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L360)):*
> ```Java
>         .setProtocol(protocol)
>         .setSrcProtocol(route.getProtocol())
>         .setOriginMechanism(REDISTRIBUTE)
> ```

Does it make sense to set an origin mechanism here? At this point we don't know if this route is being converted for a redistribute or network statement or for export. I think it'd be better to leave the origin mechanism null here and require the caller to set it. that way we will catch if any callers forget to update it before building.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java, line 132 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa0Dcl-58Ghv83LuWe:-Moa0Dcl-58Ghv83LuWf:bu9ji1q) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java#L132)):*
> ```Java
>         assert _localOriginationTypeTieBreaker == PREFER_REDISTRIBUTE;
>         return lhs.getOriginMechanism() == REDISTRIBUTE ? 1 : -1;
>       }
> ```

How does origin mechanism `GENERATED` factor in here? Have you seen cases where generated routes are always preferred over redistribute or network routes, or vice versa?

If not, i'd avoid determining preference from origin mechanism if either route is generated.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java, line 137 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa038-Ejje1jVWa3ab:-Moa038-Ejje1jVWa3ac:b-rtagwg) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java#L137)):*
> ```Java
>         Comparator
> 
>             // Prefer
> ```

delete
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 140 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa2E0D1RvqHZcurE-D:-Moa2E0D1RvqHZcurE-E:bk48eed) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L140)):*
> ```Java
>         nextHopIpTieBreaker == HIGHEST_NEXT_HOP_IP ? Comparator.reverseOrder() : Ip::compareTo;
>     return Comparator.comparing(Bgpv4Route::getNextHopIp, nhipComparator)
>         .thenComparing(Bgpv4Route::getSrcProtocol);
> ```

what's the comparator for source protocols? is the source protocol preference the same across vendors?
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 184 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoawZUZ4PHpdmib0ngW:-MoawZUZ4PHpdmib0ngX:bblynkh) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L184)):*
> ```Java
>                 route.getOriginMechanism(),
>                 o ->
>                     Multimaps.newSortedSetMultimap(
> ```

Seems like we may want to be able to keep multiple copies of the same route in this map. Suppose two main RIB routes result in identical BGP local routes. withdrawal of one from the main rib should not remove the corresponding BGP local route.

This may be a rare case since it would require two equally-preferred main RIB routes with the same NHIP, though.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 193 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoatsOvDJTThN3B_Y6R:-MoatsOvDJTThN3B_Y6S:b-am3tnn) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L193)):*
> ```Java
>             ? Optional.empty()
>             : Optional.of(routesForMechanismAndPrefix.first());
>     routesForMechanismAndPrefix.add(route);
> ```

Do NHIP and source protocol take precedence over the attributes compared in `comparePreference`? Or is this just assuming those attributes will be equivalent for local routes from the same origin mechanism?
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 203 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moas8390SQGBLPSBi4o:-Moas8390SQGBLPSBi4p:b-jet7c1) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L203)):*
> ```Java
>         RibDelta.<Bgpv4Route>builder().from(super.mergeRouteGetDelta(route));
>     // if old best was present, remove it
>     maybeOldBest.ifPresent(oldBest -> delta.from(super.removeRouteGetDelta(oldBest)));
> ```

There is no vendor that can have more than one local BGP route for the same origination mechanism?
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-MoavM_P7LxCdq8tmZuH:b-3qzjis) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*
> ```Java
>     }
>     // remove the route. no effect if already absent.
>     delta.from(super.removeRouteGetDelta(route));
> ```

Agree with your comments that this will give the correct result, but seems at least as simple and may make for easier debugging to do:
```
oldBest = routesForMechanismAndPrefix.first();
if (!routesForMechanismAndPrefix.remove(route)) {
    return empty;
} else if (oldBest.equals(route)) {
    delta.from(merge new best route);
}
delta.from(remove route);
```
___
*[projects/batfish/src/main/java/org/batfish/representation/aws/Utils.java, line 79 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobWrwJ4skNSEiU_19P:-MobWrwJ4skNSEiU_19Q:bsgvksy) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/aws/Utils.java#L79)):*
> ```Java
>         .setLocalOriginationTypeTieBreaker(LocalOriginationTypeTieBreaker.NO_PREFERENCE)
>         .setNetworkNextHopIpTieBreaker(HIGHEST_NEXT_HOP_IP)
>         .setRedistributeNextHopIpTieBreaker(HIGHEST_NEXT_HOP_IP)
> ```

verified?
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 951 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobUp0h7sQOryhRjG6S:-MobUp0h7sQOryhRjG6T:bsu09yc) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L951)):*
> ```Java
>       Configuration c, BgpProcess proc, String vrfName) {
>     Ip bgpRouterId = getBgpRouterId(c, vrfName, proc);
>     // TODO: surely this is customizable
> ```

please remove this comment as well
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 953 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobV6vG8AJuilXcLyIv:-MobV6vG8AJuilXcLyIw:b-3xgton) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L953)):*
> ```Java
>     // TODO: surely this is customizable
>     org.batfish.datamodel.BgpProcess newBgpProcess =
>         // TODO: customizable admin distances
> ```

do you know whether there are actually knobs for this?
___
*[projects/batfish/src/test/java/org/batfish/dataplane/rib/AbstractRibTest.java, line 476 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moay6uJ1gObbG_3WAVD:-Moay6uJ1gObbG_3WAVE:b4acshl) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/test/java/org/batfish/dataplane/rib/AbstractRibTest.java#L476)):*
> ```Java
>             null,
>             null,
>             null);
> ```

didn't think the tiebreakers were allowed to be null, surprised this doesn't throw. Same in other rib tests
___
*[projects/batfish/src/test/java/org/batfish/dataplane/rib/Bgpv4RibTest.java, line 1015 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Mob6TOGCaDzCroOZksu:-Mob6TOGCaDzCroOZksv:bdjsovz) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/test/java/org/batfish/dataplane/rib/Bgpv4RibTest.java#L1015)):*
> ```Java
>           containsInAnyOrder(
>               RouteAdvertisement.adding(highestNhipRoute),
>               RouteAdvertisement.withdrawing(lowestNhipRoute)));
> ```

Seems like you could just run this suite once, and outside of that just test the comparators themselves and make sure the right comparators are used in the right places. but no real need to redo.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 5 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_ucCG-6IBN_-12C16:-Mo_ucCG-6IBN_-12C17:bdn2rk4) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L5)):*
> ```Java
> /**
>  * Configuration-wise, how a route got into the BGP RIB. Used in tie-breaking on some vendors, and
>  * to distinguish which next-hop-IP preference to sue for locally-originated routes.
> ```

typo: sue -> use
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 13 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_uUK29rdgEoc9opcF:-Mo_uUK29rdgEoc9opcG:b-3kqgfq) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L13)):*
> ```Java
>   LEARNED,
>   /** Locally originated via a {@code network} statement. */
>   NETWORK,
> ```

which of `GENERATED` and `NETWORK` will be used for A10 network statement routes, which may be created independent of main rib routes?
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_v5mI0OIW5OD0Lb9h:-Mo_v5mI0OIW5OD0Lb9i:b5uxvn0) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L16)):*
> or via export from main RIB on devices that do so

What are you referring to here? Devices that export directly from the main RIB do not install the resulting routes in their BGP RIB. Do you mean those routes are sent to neighbors with origin mechanism `REDISTRIBUTE`? Looks like it's set to `LEARNED` in `transformBgpRoutePreExport`.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java, line 116 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_vryjCCIpu07uLrRZ:-Mo_vryjCCIpu07uLrR_:bftqdad) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java#L116)):*
> ```Java
>         firstNonNull(nextHopIp, Route.UNSET_ROUTE_NEXT_HOP_IP),
>         originatorIp,
>         firstNonNull(originMechanism, LEARNED),
> ```

why is `LEARNED` the default? why may the origin mechanism be null? Same goes for handling of origin mechanism in `build()`
___
*[projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java, line 210 at r4](https://reviewable.io/reviews/batfish/batfish/7690#-MobY9wJEWuiwAqgSaeQ:-MobY9wJEWuiwAqgSaeR:boj7e8r) ([raw file](https://github.com/batfish/batfish/blob/91b6340cd7f4c70f1c27d410c32e671b31429c10/projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java#L210)):*
> ```Java
>         _communities,
>         _originProtocol,
>         _originType == null ? 0 : _originType.ordinal(),
> ```

isn't 0 a possible ordinal value? thought the move was `enumVal == null ? null : enumVal.ordinal`

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: 50 of 68 files reviewed, 13 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 360 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_zlDc97lbj2jDgmKl:-MoewIVDBQCJksSFeLXz:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L360)):*

Does it make sense to set an origin mechanism here? At this point we don't know if this route is being converted for a redistribute or network statement or for export. I think it'd be better to leave the origin mechanism null here and require the caller to set it. that way we will catch if any callers forget to update it before building.
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java, line 132 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa0Dcl-58Ghv83LuWe:-MoewkMrEY0zc7vH5lqW:b-1l69nx) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java#L132)):*

How does origin mechanism `GENERATED` factor in here? Have you seen cases where generated routes are always preferred over redistribute or network routes, or vice versa?

If not, i'd avoid determining preference from origin mechanism if either route is generated.
</blockquote></details>

Ah you caught a bug I introduced in the middle of my PR. Changed the outer guard.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java, line 137 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa038-Ejje1jVWa3ab:-MoexpGZ1UBgG3kDsLZd:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java#L137)):*

delete
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 140 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moa2E0D1RvqHZcurE-D:-MoekFZa0Ys2EW4vljmH:bdq1c85) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L140)):*

what's the comparator for source protocols? is the source protocol preference the same across vendors?
</blockquote></details>

Doesn't matter at all. All that matters is that only one protocol gets redistributed when there are more than one with the same NHIP. AFAIK it uses enum ordering by default.
The only real use case for this is NX-OS, which redistributes either a local route or a connected route for a /32 connected route (and there is no way to tell which one, since that information is not preserved or distinguishable in a route-map match statement).
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 184 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoawZUZ4PHpdmib0ngW:-MoekrPR6dHc26y2Iqig:b8dt6jq) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L184)):*

Seems like we may want to be able to keep multiple copies of the same route in this map. Suppose two main RIB routes result in identical BGP local routes. withdrawal of one from the main rib should not remove the corresponding BGP local route.

This may be a rare case since it would require two equally-preferred main RIB routes with the same NHIP, though.
</blockquote></details>

The solution here is not multiple copies, but either tracking source route or checking whether the route can be recomputed during a remove. IMO we should defer for now, though this is a real issue. It is most vulnerable when the NHIP is cleared, since all contributors will likely map to the same local route. Note this is analagous to an existing issue preventing proper implementation of add-path, and in fact affects ALL redistribution across iBDP.
Added a TODO
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 193 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoatsOvDJTThN3B_Y6R:-MoemGIM3HEdOpb28GjQ:b-yhkzci) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L193)):*

Do NHIP and source protocol take precedence over the attributes compared in `comparePreference`? Or is this just assuming those attributes will be equivalent for local routes from the same origin mechanism?
</blockquote></details>

The purpose is to remove from consideration routes with non-best NHIPs, even if they would be transformed to better routes according to BGP path selection. Their other attributes do not matter. They do not compete if they do not have optimal NHIP.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 203 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moas8390SQGBLPSBi4o:-MoemX363Ouil_aM89SM:b-u6o6bh) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L203)):*

There is no vendor that can have more than one local BGP route for the same origination mechanism?
</blockquote></details>

Not among any of the vendors I tested.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-Moen6IvDpADO7wJUhn4:b8p4b45) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*

Agree with your comments that this will give the correct result, but seems at least as simple and may make for easier debugging to do:
```
oldBest = routesForMechanismAndPrefix.first();
if (!routesForMechanismAndPrefix.remove(route)) {
    return empty;
} else if (oldBest.equals(route)) {
    delta.from(merge new best route);
}
delta.from(remove route);
```
</blockquote></details>

Your code will place a withdraw above a merge, which I was trying to avoid.
___
*[projects/batfish/src/main/java/org/batfish/representation/aws/Utils.java, line 79 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobWrwJ4skNSEiU_19P:-MoenrSK4dRz9Oqt09UM:bv9r8vd) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/aws/Utils.java#L79)):*

verified?
</blockquote></details>

doesn't export from bgp rib, so values don't matter. Added comment.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 951 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobUp0h7sQOryhRjG6S:-Moeo5SC3qg6PABh_WVH:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L951)):*

please remove this comment as well
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java, line 953 at r3](https://reviewable.io/reviews/batfish/batfish/7690#-MobV6vG8AJuilXcLyIv:-Moeo6YF9LadoH1VEuR3:bl7rc3l) ([raw file](https://github.com/batfish/batfish/blob/7b3bf76742847e2a3f7677b909f88e7bddccd18f/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java#L953)):*

do you know whether there are actually knobs for this?
</blockquote></details>

checking which of these TODOs can be removed is well out of scope. All of this crap is copy-pasted mostly across cisco-like vendors.
___
*[projects/batfish/src/test/java/org/batfish/dataplane/rib/AbstractRibTest.java, line 476 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moay6uJ1gObbG_3WAVD:-MoeoagX97dUnAmdW3p4:b-1oxts8) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/test/java/org/batfish/dataplane/rib/AbstractRibTest.java#L476)):*

didn't think the tiebreakers were allowed to be null, surprised this doesn't throw. Same in other rib tests
</blockquote></details>

Not throwing for me for whatever reason. Fixed.
___
*[projects/batfish/src/test/java/org/batfish/dataplane/rib/Bgpv4RibTest.java, line 1015 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Mob6TOGCaDzCroOZksu:-Moeq4wh1BqQU9UWY62C:b-c2cqi2) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/test/java/org/batfish/dataplane/rib/Bgpv4RibTest.java#L1015)):*

Seems like you could just run this suite once, and outside of that just test the comparators themselves and make sure the right comparators are used in the right places. but no real need to redo.
</blockquote></details>

yeah i know it's ugly, but after some thought I decided it would take too long to refactor/reorganize.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 5 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_ucCG-6IBN_-12C16:-MoeqBotDJ3KqiWNHs_C:b-7ql1td) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L5)):*

typo: sue -> use
</blockquote></details>

fixed
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 13 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_uUK29rdgEoc9opcF:-MoeqDqD3gLrh016kioO:bh23la8) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L13)):*

which of `GENERATED` and `NETWORK` will be used for A10 network statement routes, which may be created independent of main rib routes?
</blockquote></details>

They don't exist now. I'd use `NETWORK` just since it's a `network` statement. But it doesn't really matter cause you'll only ever get one route out of it.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_v5mI0OIW5OD0Lb9h:-MoeqcFN8P_jvDf9wdWB:bbgvdje) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L16)):*

> or via export from main RIB on devices that do so

What are you referring to here? Devices that export directly from the main RIB do not install the resulting routes in their BGP RIB. Do you mean those routes are sent to neighbors with origin mechanism `REDISTRIBUTE`? Looks like it's set to `LEARNED` in `transformBgpRoutePreExport`.
</blockquote></details>

It's transiently `REDISTRIBUTE`. Only visible in debugging during that interval. Doesn't really matter what you set it to.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java, line 116 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_vryjCCIpu07uLrRZ:-Moeqowe8Yg1YXUn6LnY:b-jeyno2) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java#L116)):*

why is `LEARNED` the default? why may the origin mechanism be null? Same goes for handling of origin mechanism in `build()`
</blockquote></details>

This class is for an input field to some questions. mechanism will be null if user does not input it.
I didn't want to force breaking changes to existing user questions. We haven't yet decided if/how to expose this to users.
This is backwards compatible. We can address later if we want to force users to specify this.
___
*[projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java, line 210 at r4](https://reviewable.io/reviews/batfish/batfish/7690#-MobY9wJEWuiwAqgSaeQ:-MoevYRADBMShfJ99AUn:bcyf6nl) ([raw file](https://github.com/batfish/batfish/blob/91b6340cd7f4c70f1c27d410c32e671b31429c10/projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java#L210)):*

isn't 0 a possible ordinal value? thought the move was `enumVal == null ? null : enumVal.ordinal`
</blockquote></details>

You're overthinking this. Worst case is a hash collision. Also, the hash of null is 0.
`Arrays.hashCode`:
```
    public static int hashCode(Object a[]) {
        if (a == null)
            return 0;
```

corinaminer (commented):
> Reviewed 11 of 15 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: all files reviewed, 6 unresolved discussions (waiting on @arifogel, @corinaminer, and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 184 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoawZUZ4PHpdmib0ngW:-MofEpag14TbgfuMtjqr:b-xbvwxe) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L184)):*

The solution here is not multiple copies, but either tracking source route or checking whether the route can be recomputed during a remove. IMO we should defer for now, though this is a real issue. It is most vulnerable when the NHIP is cleared, since all contributors will likely map to the same local route. Note this is analagous to an existing issue preventing proper implementation of add-path, and in fact affects ALL redistribution across iBDP.
Added a TODO
</blockquote></details>

I don't quite see why allowing duplicates in `_localRoutes` wouldn't fix it, but i'm fine with deferring with a todo.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 203 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-Moas8390SQGBLPSBi4o:-MofDJ5XCvPLJjGKAS_N:b-q74txj) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L203)):*

Not among any of the vendors I tested.
</blockquote></details>

great!
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-MofEKDxDgUTicK3CYxj:bv4ub95) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*

Your code will place a withdraw above a merge, which I was trying to avoid.
</blockquote></details>

no it doesn't (note that `routesForMechanismAndPrefix.remove(route)` is just removing from the set, not calling `super.removeRouteGetDelta`)
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 222 at r7](https://reviewable.io/reviews/batfish/batfish/7690#-MofFSwk7Xu2U7cq0wXh:-MofFSwk7Xu2U7cq0wXi:b-6oossm) ([raw file](https://github.com/batfish/batfish/blob/2d7b5eff268ef9ca27402db5be80becd730e4b4e/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L222)):*
> ```Java
>     int ebgpAdmin = RoutingProtocol.BGP.getDefaultAdministrativeCost(c.getConfigurationFormat());
>     int ibgpAdmin = RoutingProtocol.IBGP.getDefaultAdministrativeCost(c.getConfigurationFormat());
>     int localAdmin = ibgpAdmin; // TODO: is this correct?
> ```

please preserve this todo, unless you checked it
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java, line 241 at r7](https://reviewable.io/reviews/batfish/batfish/7690#-MofAGOy5lG0boX6rJs6:-MofAGOy5lG0boX6rJs7:b-g1n9ax) ([raw file](https://github.com/batfish/batfish/blob/1d72c80afd23e96c66ab35c54636297e8f612840/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java#L241)):*
> ```Java
> 
>   /** Constructs a BgpProcess with the given router ID and admin costs */
>   @VisibleForTesting
> ```

all uses really switched over to builder? :star2: 
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_v5mI0OIW5OD0Lb9h:-MofAXNSB3sIYt1VX5qN:b-rqtnui) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L16)):*

It's transiently `REDISTRIBUTE`. Only visible in debugging during that interval. Doesn't really matter what you set it to.
</blockquote></details>

In that case, why set the origin mechanism to `LEARNED` in `transformBgpRoutePreExport`? That seems inconsistent. I think all exported routes should either keep their original origin mechanism or get marked as `LEARNED` -- i'm fine with either.

But if we choose the former (or with the current setup), routes need to be marked `LEARNED` upon import. I don't see where that's happening -- am i just missing it?

corinaminer (commented):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @sfraint)

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: 67 of 68 files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 184 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoawZUZ4PHpdmib0ngW:-MofGxYDAlLMlL7k1ZNT:bsu97zu) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L184)):*

I don't quite see why allowing duplicates in `_localRoutes` wouldn't fix it, but i'm fine with deferring with a todo.
</blockquote></details>

It would be complicated to guarantee that that you get the same number of transformed withdrawals. If anything goes wrong, we will have a very hard-to-track-down bug living in iBDP. Definitely defer.
___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-MofHYnH3BkIjvS7QEHk:bn4y5g5) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*

no it doesn't (note that `routesForMechanismAndPrefix.remove(route)` is just removing from the set, not calling `super.removeRouteGetDelta`)
</blockquote></details>

Right, my mistake.
Anyway the code for `merge new best route` in your suggestion still needs to check for the presence of a remaining route for the prefix, and grab the best one.
Seems just as complicated to me.
___
*[projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java, line 222 at r7](https://reviewable.io/reviews/batfish/batfish/7690#-MofFSwk7Xu2U7cq0wXh:-MofIYcmD-jzwrk4cHcg:b-pr0a90) ([raw file](https://github.com/batfish/batfish/blob/2d7b5eff268ef9ca27402db5be80becd730e4b4e/projects/batfish/src/main/java/org/batfish/representation/fortios/FortiosBgpConversions.java#L222)):*

please preserve this todo, unless you checked it
</blockquote></details>

added the todo to the new constant.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java, line 241 at r7](https://reviewable.io/reviews/batfish/batfish/7690#-MofAGOy5lG0boX6rJs6:-MofIk2kEs3zYD8yLj34:bu7j827) ([raw file](https://github.com/batfish/batfish/blob/1d72c80afd23e96c66ab35c54636297e8f612840/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpProcess.java#L241)):*

all uses really switched over to builder? :star2: 
</blockquote></details>

Actually yeah.
Only remaining user was `testBuilder`. Moved that over to builder and removed this constructor.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java, line 16 at r1](https://reviewable.io/reviews/batfish/batfish/7690#-Mo_v5mI0OIW5OD0Lb9h:-MofJn7Z5E8xhi3BOyV4:bvcc5aa) ([raw file](https://github.com/batfish/batfish/blob/34f6015a9765ac5a478a2e208bf4a800eefee7e2/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/OriginMechanism.java#L16)):*

In that case, why set the origin mechanism to `LEARNED` in `transformBgpRoutePreExport`? That seems inconsistent. I think all exported routes should either keep their original origin mechanism or get marked as `LEARNED` -- i'm fine with either.

But if we choose the former (or with the current setup), routes need to be marked `LEARNED` upon import. I don't see where that's happening -- am i just missing it?
</blockquote></details>

For debugging, they should be marked LEARNED on import, not pre-export. Fixed.

corinaminer (commented):
> Reviewed 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-MofcTUf3bT6Wh6UPC_F:bihhyna) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*

Right, my mistake.
Anyway the code for `merge new best route` in your suggestion still needs to check for the presence of a remaining route for the prefix, and grab the best one.
Seems just as complicated to me.


</blockquote></details>

the main difference is that my version won't call merge/remove on the rib tree if nothing changed.

Actually, we know that `route` will be in the rib tree iff it is currently best, right? So it could be:
```
oldBest = routesForMechanismAndPrefix.first();
if (!routesForMechanismAndPrefix.remove(route)) {
    return empty;
} else if (oldBest.equals(route)) {
    // This is the only case in which the rib tree should be affected
    delta.from(merge new best route);
    delta.from(remove route);
}
```
That said, seems like this is the last open discussion, so not worth holding up the PR.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7690)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java, line 234 at r2](https://reviewable.io/reviews/batfish/batfish/7690#-MoavM_P7LxCdq8tmZuG:-MogLtFn15nWjLE8NsGI:b-42nrb9) ([raw file](https://github.com/batfish/batfish/blob/1c3cd6a0cdf15e11f7c4b19e00cc4bd82c4da193/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java#L234)):*

the main difference is that my version won't call merge/remove on the rib tree if nothing changed.

Actually, we know that `route` will be in the rib tree iff it is currently best, right? So it could be:
```
oldBest = routesForMechanismAndPrefix.first();
if (!routesForMechanismAndPrefix.remove(route)) {
    return empty;
} else if (oldBest.equals(route)) {
    // This is the only case in which the rib tree should be affected
    delta.from(merge new best route);
    delta.from(remove route);
}
```
That said, seems like this is the last open discussion, so not worth holding up the PR.
</blockquote></details>

Your version does a route equality check, which is on par with the expensive part of merge/remove.
I think the difference is trivial. If you want to do a small cleanup PR to make it prettier (and a tiny bit more performant) that's fine. But I'm going to merge now.

PR #7696, opened 2021-11-15
validation for NODE_NAME variable type

anothermattbrown:
Previously, all queries for NODE_NAME were considered VALID. Now only matching names are valid, all other queries are NO_MATCH. We could/should distinguish between NO_MATCH and INVALID, but punting on that for now.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7696)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7731, opened 2021-11-30
Fix Layer2Vni creation and usage

arifogel:
- fixes issue where vxlan flood lists are not updated due to missing
  `Layer2Vni`s on default VRF.
- always put Layer2Vnis in default vrf
- fix conversion for EOS
- fix enum hashing in Layer2Vni hashCode

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/Layer2Vni.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/Layer2Vni.java
@@ -141,7 +141,13 @@ public final class Layer2Vni implements Vni {
   @Override
   public int hashCode() {
     return Objects.hash(
-        _bumTransportMethod, _bumTransportIps, _sourceAddress, _udpPort, _vlan, _vni, _srcVrf);
+        _bumTransportMethod.ordinal(),
+        _bumTransportIps,
+        _sourceAddress,
+        _udpPort,
+        _vlan,
+        _vni,
+        _srcVrf);
   }
 
   @Override
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -79,7 +79,6 @@ import org.batfish.datamodel.Prefix;
 import org.batfish.datamodel.PrefixTrieMultiMap;
 import org.batfish.datamodel.Route;
 import org.batfish.datamodel.RoutingProtocol;
-import org.batfish.datamodel.Vrf;
 import org.batfish.datamodel.bgp.AddressFamily;
 import org.batfish.datamodel.bgp.AddressFamily.Type;
 import org.batfish.datamodel.bgp.BgpAggregate;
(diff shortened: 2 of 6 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7731)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 744 at r1](https://reviewable.io/reviews/batfish/batfish/7731#-MpirKh-E6ljVKQGGxdB:-MpirKh0BWO9zHKaSBOA:b-sl1yi4) ([raw file](https://github.com/batfish/batfish/blob/e726df368937dc532d4a08985b8609e664898494/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L744)):*
> ```Java
>                 // Merge into our own RIBs
>                 RibDelta<EvpnType3Route> d = _evpnType3Rib.mergeRouteGetDelta(route);
>                 _evpnType3DeltaBuilder.from(d);
> ```

Is there a reason `_evpnType3DeltaBuilder` is no longer affected by this?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7731)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 744 at r1](https://reviewable.io/reviews/batfish/batfish/7731#-MpirKh-E6ljVKQGGxdB:-Mpj43Hy4LnMb5I5eb8W:bl06nd2) ([raw file](https://github.com/batfish/batfish/blob/e726df368937dc532d4a08985b8609e664898494/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L744)):*

Is there a reason `_evpnType3DeltaBuilder` is no longer affected by this?
</blockquote></details>

Well the true branch was never visited in my experiments, so I just removed it.

Looking now, it seems `_evpnType3DeltaBuilder` is only used to populate `_evpnType3DeltaPrev`, which itself is only used in `isDirty()`. But `_evpnInitializationDelta` is also checked in `isDirty`, so it seems redundant to modify both. At some point we need to clean up the initialization stuff, since it is currently associated with a TODO:
```
  private void sendOutRoutesToNewEvpnEdges(Map<String, Node> allNodes, NetworkConfigurations nc) {
    for (EdgeId edge : _evpnEdgesWentUp) {
      // TODO: _localType3Routes is not enough
      //    Ideally we need to re-send all EVPN routes we have to new neighbors
      sendEvpnType3RoutesToEdge(edge.reverse(), _localType3Routes, nc, allNodes);
      sendEvpnType5RoutesToEdge(edge.reverse(), _evpnType5Advertisements.build(), nc, allNodes);
    }
  }
```
But we can do that later. For now, I believe it is fine to leave `_evpnType3DeltaBuilder` alone.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7731)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java, line 744 at r1](https://reviewable.io/reviews/batfish/batfish/7731#-MpirKh-E6ljVKQGGxdB:-MpjU8E9BCIZTcnIJCVi:b-plh31) ([raw file](https://github.com/batfish/batfish/blob/e726df368937dc532d4a08985b8609e664898494/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L744)):*

Well the true branch was never visited in my experiments, so I just removed it.

Looking now, it seems `_evpnType3DeltaBuilder` is only used to populate `_evpnType3DeltaPrev`, which itself is only used in `isDirty()`. But `_evpnInitializationDelta` is also checked in `isDirty`, so it seems redundant to modify both. At some point we need to clean up the initialization stuff, since it is currently associated with a TODO:
```
  private void sendOutRoutesToNewEvpnEdges(Map<String, Node> allNodes, NetworkConfigurations nc) {
    for (EdgeId edge : _evpnEdgesWentUp) {
      // TODO: _localType3Routes is not enough
      //    Ideally we need to re-send all EVPN routes we have to new neighbors
      sendEvpnType3RoutesToEdge(edge.reverse(), _localType3Routes, nc, allNodes);
      sendEvpnType5RoutesToEdge(edge.reverse(), _evpnType5Advertisements.build(), nc, allNodes);
    }
  }
```
But we can do that later. For now, I believe it is fine to leave `_evpnType3DeltaBuilder` alone.
</blockquote></details>

ok, makes sense that checking `_evpnInitializationDelta` in `isDirty` should make it unnecessary to modify `_evpnType3DeltaBuilder` here.

PR #7732, opened 2021-11-30
EVPN: Use NVE IP as NHIP for originated routes

corinaminer:
Local EVPN routes don't appear to have any NHIP attached, but when they're sent out, the sender sets their NHIP to the IP of the NVE (Network Virtualization Edge): https://datatracker.ietf.org/doc/html/rfc8365#section-5.1.3

In this PR we add the NVE IP as an attribute of `EvpnAddressFamily`, so every peer sending EVPN routes has an IP to use as NHIP for its local EVPN routes.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/EvpnAddressFamily.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/EvpnAddressFamily.java
@@ -13,6 +13,7 @@ import java.util.SortedSet;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
+import org.batfish.datamodel.Ip;
 
 /** Configuration settings for EVPN address family */
 @ParametersAreNonnullByDefault
@@ -21,10 +22,12 @@ public final class EvpnAddressFamily extends AddressFamily {
   private static final String PROP_L2_VNIS = "l2Vnis";
   private static final String PROP_L3_VNIS = "l3Vnis";
   private static final String PROP_PROPAGATE_UNMATCHED = "propagateUnmatched";
+  private static final String PROP_NVE_IP = "nveIp";
 
   @Nonnull private final SortedSet<Layer2VniConfig> _l2VNIs;
   @Nonnull private final SortedSet<Layer3VniConfig> _l3VNIs;
   private final boolean _propagateUnmatched;
+  @Nullable private final Ip _nveIp;
 
   protected EvpnAddressFamily(
       // super fields
@@ -37,6 +40,7 @@ public final class EvpnAddressFamily extends AddressFamily {
       // local fields
       Set<Layer2VniConfig> l2Vnis,
       Set<Layer3VniConfig> l3Vnis,
+      @Nullable Ip nveIp,
       boolean propagateUnmatched) {
     super(
         addressFamilyCapabilities,
@@ -47,6 +51,7 @@ public final class EvpnAddressFamily extends AddressFamily {
         routeReflectorClient);
     _l2VNIs = ImmutableSortedSet.copyOf(l2Vnis);
     _l3VNIs = ImmutableSortedSet.copyOf(l3Vnis);
+    _nveIp = nveIp;
     _propagateUnmatched = propagateUnmatched;
   }
 
@@ -63,6 +68,7 @@ public final class EvpnAddressFamily extends AddressFamily {
       // local fields
       @Nullable @JsonProperty(PROP_L2_VNIS) Set<Layer2VniConfig> l2Vnis,
       @Nullable @JsonProperty(PROP_L3_VNIS) Set<Layer3VniConfig> l3Vnis,
+      @Nullable @JsonProperty(PROP_NVE_IP) Ip nveIp,
       @Nullable @JsonProperty(PROP_PROPAGATE_UNMATCHED) Boolean propagateUnmatched) {
     checkArgument(propagateUnmatched != null, "Missing %s", PROP_PROPAGATE_UNMATCHED);
     return new Builder()
@@ -74,6 +80,7 @@ public final class EvpnAddressFamily extends AddressFamily {
         .setRouteReflectorClient(firstNonNull(routeReflectorClient, Boolean.FALSE))
         .setL2Vnis(firstNonNull(l2Vnis, ImmutableSortedSet.of()))
         .setL3Vnis(firstNonNull(l3Vnis, ImmutableSortedSet.of()))
+        .setNveIp(nveIp)
         .setPropagateUnmatched(propagateUnmatched)
         .build();
   }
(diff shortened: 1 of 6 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 399 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj6lnM99v6QjLjo9JB:-Mpj6lnM99v6QjLjo9JC:b-2nejkw) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L399)):*
> ```Java
>       if (originalRouteNhip.equals(UNSET_ROUTE_NEXT_HOP_IP)) {
>         EvpnAddressFamily evpnFamily = (EvpnAddressFamily) af;
>         if (evpnFamily.getNveIp() != null) {
> ```

Not sure how this could happen. Maybe an assert would be better. Perhaps we should check that it is available as a condition of originating evpn routes in the first place.
___
*[projects/batfish/src/main/java/org/batfish/representation/arista/AristaConversions.java, line 787 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj7f4b-xc6mmh5y-po:-Mpj7f4b-xc6mmh5y-pp:b55fcfa) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/representation/arista/AristaConversions.java#L787)):*
> ```Java
>   }
> 
>   static Optional<Ip> getSourceInterfaceIp(
> ```

nit: `@Nonnull`

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 399 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj6lnM99v6QjLjo9JB:-Mpj80iJCPKR-TetORpI:b-pfpnuk) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L399)):*

Not sure how this could happen. Maybe an assert would be better. Perhaps we should check that it is available as a condition of originating evpn routes in the first place.
</blockquote></details>

That is, I'm pretty sure we don't want to set the NHIP to something arbitrary later in the pipeline if it doesn't get set here. Perhaps that is the source of the weird NHIP you told me about offline.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 399 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj6lnM99v6QjLjo9JB:-MpjT0I783r16w4wGJcv:b-90fxka) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L399)):*

That is, I'm pretty sure we don't want to set the NHIP to something arbitrary later in the pipeline if it doesn't get set here. Perhaps that is the source of the weird NHIP you told me about offline.
</blockquote></details>

Well, for now it's going to happen for vendors other than Arista if they have EVPN routes. I agree the NHIPs will (likely) be wrong in that case, but it has to be set to something for the route to build. Is it better to explicitly set it to `NextHopDiscard`, which is also wrong?

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/main/java/org/batfish/representation/arista/AristaConversions.java, line 787 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj7f4b-xc6mmh5y-po:-MpmIiFzAdrSk3pIwUUj:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/representation/arista/AristaConversions.java#L787)):*

nit: `@Nonnull`
</blockquote></details>

done

arifogel (commented):
> Reviewed 3 of 3 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 399 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj6lnM99v6QjLjo9JB:-MpmKaB8Bp6qVG_b8Iqc:bs1ljmi) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L399)):*

Well, for now it's going to happen for vendors other than Arista if they have EVPN routes. I agree the NHIPs will (likely) be wrong in that case, but it has to be set to something for the route to build. Is it better to explicitly set it to `NextHopDiscard`, which is also wrong?
</blockquote></details>

Why does the route have to build? I'm suggesting you don't get this far if the AF has no nve IP. Not even sure the sessions should come up. Could test.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7732)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java, line 399 at r1](https://reviewable.io/reviews/batfish/batfish/7732#-Mpj6lnM99v6QjLjo9JB:-MpmObP04tK-NIPnkciv:b-k9epcr) ([raw file](https://github.com/batfish/batfish/blob/95495942274d8eb7bc1a9ac3b63f7b36b2aa5bba/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java#L399)):*

Why does the route have to build? I'm suggesting you don't get this far if the AF has no nve IP. Not even sure the sessions should come up. Could test.
</blockquote></details>

Per offline discussion, seems like in general it is possible to write something that isn't an NVE IP. Until we have resources to test edge case behavior, this is good enough.

PR #7738, opened 2021-12-01
NX-OS: Set NVE IP for EVPN address families

corinaminer:
For now, assuming that devices only have one NVE, as that seems to be the common case and it is unclear which NVE's source interface address should be used for EVPN NHIPs if there are multiple.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7738)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7744, opened 2021-12-02
A10: create interfaces that are vlan members

sfraint:
Vlan member interfaces may not show up later in configs. So, create them at reference time.

This should only apply to `ethernet` interfaces, `trunk` interfaces will have other configuration that is required to occur elsewhere.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7744)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7750, opened 2021-12-03
Frr: Parse route-map set src

ratulm:
The feature is described here (separately from other route map statements): http://docs.frrouting.org/en/stable-8.0/zebra.html#clicmd-set-src-ADDRESS


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7750)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7742, opened 2021-12-01
Multifile parsing jobs 2

ratulm:
@arifogel - new changes are in the second commit

PR #7766, opened 2021-12-08
Granular status for multifile parsing

ratulm:
In #7742, ParseStatus was a job-level, not file-level. That has some undesirable side-effects for multi-file jobs, such as, parseStatus question not returning the nodes produces by the file and not being able to report the status separately for files. 

This PR makes ParseStatus file-level. It also make VendorConfiguration class aware of any other files that went into its production, which makes for accurate file to node mapping.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java
@@ -11,10 +11,7 @@ public enum ParseStatus {
   EMPTY,
   /** Batfish has encountered an unrecoverable error during parsing */
   FAILED,
-  /**
-   * File was ignored either explicitly by the user or it was of a multi-file bundle that was not
-   * properly packaged.
-   */
+  /** File was explicitly ignore by the user */
   IGNORED,
   /** File is part of an unused overlay configuration */
   ORPHANED,
@@ -22,6 +19,8 @@ public enum ParseStatus {
   PARTIALLY_UNRECOGNIZED,
   /** File was fully parsed */
   PASSED,
+  /** File was packaged in an unexpected manner in the snapshot */
+  UNEXPECTED_PACKAGING,
   /** Batfish could not detect the file format */
   UNKNOWN,
   /** Batfish does not support the format/vendor config in the file */
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java
@@ -11,7 +11,10 @@ public enum ParseStatus {
   EMPTY,
   /** Batfish has encountered an unrecoverable error during parsing */
   FAILED,
-  /** File was explicitly ignored by the user */
+  /**
+   * File was ignored either explicitly by the user or it was of a multi-file bundle that was not
+   * properly packaged.
+   */
   IGNORED,
   /** File is part of an unused overlay configuration */
   ORPHANED,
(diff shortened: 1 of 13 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7766)* status: all files reviewed, 5 unresolved discussions (waiting on @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java, line 132 at r1](https://reviewable.io/reviews/batfish/batfish/7766#-MqNC_Te0NWuYlXJzZI-:-MqNC_Te0NWuYlXJzZI0:bcqtjsv) ([raw file](https://github.com/batfish/batfish/blob/a25527b65611052a3cdd9455c3cebb31cdbed324/projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java#L132)):*
> ```Java
>     }
> 
>     public void setParseStatus(ParseStatus parseStatus) {
> ```

Two constructors that differ in parseStatus, and also a setter? Can this be simplified? You still using all of these?
___
*[projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java, line 241 at r1](https://reviewable.io/reviews/batfish/batfish/7766#-MqNDabz0BKL_QPmgp9c:-MqNDabz0BKL_QPmgp9d:b-1btfyk) ([raw file](https://github.com/batfish/batfish/blob/a25527b65611052a3cdd9455c3cebb31cdbed324/projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java#L241)):*
> ```Java
> 
>   /**
>    * Returns the {@link VendorConfiguration vendor configuration} corresponding to the current file
> ```

What is the "current file"?
___
*[projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java, line 259 at r1](https://reviewable.io/reviews/batfish/batfish/7766#-MqNCzLaEvEQFAYg_Wdp:-MqNCzLaEvEQFAYg_Wdq:b-xn0iwd) ([raw file](https://github.com/batfish/batfish/blob/a25527b65611052a3cdd9455c3cebb31cdbed324/projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java#L259)):*
> ```Java
> 
>       // pull out the common case
>       String filename = _fileTexts.keySet().iterator().next();
> ```

I get what you're doing here, but maybe better to just use a helper to enforce invariant.
Like `getOnlyFilename()` that calls `getOnlyElement`, used in everything below but sonic.
___
*[projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java, line 701 at r1](https://reviewable.io/reviews/batfish/batfish/7766#-MqNGv6xCc87ovQVSxBH:-MqNGv6xCc87ovQVSxBI:b4tppwt) ([raw file](https://github.com/batfish/batfish/blob/a25527b65611052a3cdd9455c3cebb31cdbed324/projects/batfish/src/main/java/org/batfish/job/ParseVendorConfigurationJob.java#L701)):*
> ```Java
>    * Parses the given file and returns a {@link ParseResult} for this job.
>    *
>    * <p>The returned {@link ParseResult} will always have a valid ParseStatus for each file. It may
> ```

nit: `{@link ParseStatus}`
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java, line 15 at r1](https://reviewable.io/reviews/batfish/batfish/7766#-MqNIKqLEp2hOjWNnkLt:-MqNIKqM1QYc_QeR-p4D:b-y8cm4w) ([raw file](https://github.com/batfish/batfish/blob/a25527b65611052a3cdd9455c3cebb31cdbed324/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/answers/ParseStatus.java#L15)):*
> ```Java
>   FAILED,
>   /**
>    * File was ignored either explicitly by the user or it was of a multi-file bundle that was not
> ```

I don't think you should generalize this enum constant. It was specifically meant to show something the user intended. Maybe you want UNEXPECTED, or IGNORED_UNEXPECTED, or something like that?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7766)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

PR #7771, opened 2021-12-09
AclLineMatchExprs: add VALID_FLOWS, port BDDRepresentativePicker on it

dhalperi:
It is useful to have a single place we define the set of valid flows, so that
we can post-process, pre-process, or otherwise filter sets of flows down to
sets of valid flows.

Move to AclLineMatchExprs from private inside BDDPacket, and move BDDPacket
onto it.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7771)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7777, opened 2021-12-09
Flow: cleanup tcp flags

dhalperi:
If we want to keep json backwards compat, we can do that in get/deserialize
no reason to keep it in our code.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7777)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java, line 306 at r1](https://reviewable.io/reviews/batfish/batfish/7777#-MqWMQ7o05w0u0lNmV2j:-MqWMQ7o05w0u0lNmV2k:b-c4uksl) ([raw file](https://github.com/batfish/batfish/blob/9b6f4b0dceb35b7852b636f57b70ec496a4d587f/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java#L306)):*
> ```Java
>     }
> 
>     public Builder setTcpFlagsAck(Integer tcpFlagsAck) {
> ```

FYI this was implicitly nullable, but would crash as the field was `int`.

PR #7783, opened 2021-12-10
AclLineMatchExprs: switch new tcp flows to syn-only

dhalperi:
1. Instead of not established, make it affirmatively syn
2. Make it use all tcp bits, so that reachability and other headerspace searches
   do not turn other bits on.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7792, opened 2021-12-12
FRR: Factor out conversion

ratulm:
Move most FRR-related conversion to its own method (so, it can be shared), and document what should have been done before that method is called. 

Also fix method annotations along the way.


arifogel (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7792)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7815, opened 2021-12-16
Split VxlanTopology into layer2 and layer3 VNI nodes

arifogel:
- add VniLayer enum to VxlanNode
- only check edges between l2 VNI VxlanNodes when deciding whether to recompute l3 adjacencies
- only use l2 VNI VxlanNodes within computation of l3 adjacencies
- only use l3 VNI edges for computation of l3 edges between generated nve interfaces

Follow-on work:
- creation of l3 VxlanNodes during data plane computation
  - create flood lists for L3 VNIs
- fix flood list calculation for L2 VNIs (remove inapplicable VTEPs)
- surfacing VNI layer in VXLAN topology answer

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java
@@ -350,7 +350,8 @@ public final class TopologyUtil {
       @Nonnull Layer1Topology layer1LogicalTopology,
       VxlanTopology vxlanTopology,
       @Nonnull Map<String, Configuration> configurations) {
-    if (layer1LogicalTopology.isEmpty() && vxlanTopology.getGraph().edges().isEmpty()) {
+    if (layer1LogicalTopology.isEmpty()
+        && !vxlanTopology.getLayer2VniEdges().findAny().isPresent()) {
       return Layer2Topology.EMPTY;
     }
     Layer2Topology.Builder l2TopologyBuilder = Layer2Topology.builder();
@@ -365,7 +366,8 @@ public final class TopologyUtil {
             .map(Layer1Node::getHostname)
             .collect(ImmutableSet.toImmutableSet());
     Set<String> nodesWithVxlan =
-        vxlanTopology.getGraph().edges().stream()
+        vxlanTopology
+            .getLayer2VniEdges()
             .flatMap(edge -> Stream.of(edge.nodeU(), edge.nodeV()))
             .map(VxlanNode::getHostname)
             .collect(ImmutableSet.toImmutableSet());
@@ -413,7 +415,7 @@ public final class TopologyUtil {
    */
   @VisibleForTesting
   static @Nonnull Stream<Layer2Edge> computeVniInterNodeEdges(VxlanTopology vxlanTopology) {
-    return vxlanTopology.getGraph().edges().stream().flatMap(TopologyUtil::toVniVniEdges);
+    return vxlanTopology.getLayer2VniEdges().flatMap(TopologyUtil::toVniVniEdges);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


corinaminer (commented):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7815)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java, line 354 at r1](https://reviewable.io/reviews/batfish/batfish/7815#-Mr-vmkwAdk2zZjAY602:-Mr-vmkwAdk2zZjAY603:b68pydc) ([raw file](https://github.com/batfish/batfish/blob/9df7ef2667405250aeb57f43d2d3eef2699ac3ce/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java#L354)):*
> ```Java
>       @Nonnull Map<String, Configuration> configurations) {
>     if (layer1LogicalTopology.isEmpty()
>         && !vxlanTopology.getLayer2VniEdges().findAny().isPresent()) {
> ```

nit: `[layer 2 edge stream].count() == 0` may be simpler
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/VxlanTopology.java, line 72 at r1](https://reviewable.io/reviews/batfish/batfish/7815#-Mr-uqij5Icdf7bkOSqE:-Mr-uqij5Icdf7bkOSqF:b-7ikt2x) ([raw file](https://github.com/batfish/batfish/blob/9df7ef2667405250aeb57f43d2d3eef2699ac3ce/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/VxlanTopology.java#L72)):*
> ```Java
>             endpointPair ->
>                 endpointPair.nodeU().getVniLayer() == LAYER_3
>                     && endpointPair.nodeV().getVniLayer() == LAYER_3);
> ```

There should never be an edge between a layer 2 and layer 3 node, right? If these get methods filtered on only `nodeU().getVniLayer() == LAYER_X`, do you figure it could cause hard-to-diagnose bugs?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7815)* status: all files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java, line 354 at r1](https://reviewable.io/reviews/batfish/batfish/7815#-Mr-vmkwAdk2zZjAY602:-Mr05xoC7saSUwwj8zlA:bnvp1vl) ([raw file](https://github.com/batfish/batfish/blob/9df7ef2667405250aeb57f43d2d3eef2699ac3ce/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/TopologyUtil.java#L354)):*

nit: `[layer 2 edge stream].count() == 0` may be simpler
</blockquote></details>

I believe that will iterate over the whole stream, and thus be more expensive.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/VxlanTopology.java, line 72 at r1](https://reviewable.io/reviews/batfish/batfish/7815#-Mr-uqij5Icdf7bkOSqE:-Mr069J58zEv6GsILiA-:bcaz3o1) ([raw file](https://github.com/batfish/batfish/blob/9df7ef2667405250aeb57f43d2d3eef2699ac3ce/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/vxlan/VxlanTopology.java#L72)):*

There should never be an edge between a layer 2 and layer 3 node, right? If these get methods filtered on only `nodeU().getVniLayer() == LAYER_X`, do you figure it could cause hard-to-diagnose bugs?
</blockquote></details>

yes.
if we want to catch that, may want an assert somewhere else in the pipeline. I'd skip for now, since no nodes are currently even generated from layer3 VNIs.

PR #7820, opened 2021-12-16
BDDReachabilityUtils: use orAll and traverse incrementally

dhalperi:
This is two changes that can be separated out if desired.

1. Rather than "pushing BDDs to neighbor states" one at a time, instead we stage all the BDDs and `orAll` them into a single new BDD to unioned with the existing reachable set for a state. Using `orAll` saves in computation. It can also push BDDs farther in fewer iterations (if A and B are both dirty, and A->B, and A is processed first, then processing B will handle A's output in this same iteration).

2. Since the newly-learned reachability (overapproximation, since some may be contained in `prior`) is "free", switch to an incremental traversal algorithm. Instead of feeding everything reachable in a state forward, feed only the learned reachability. This seems to converge faster, but induces more garbage collection than 1.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java
+++ b/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java
@@ -5,8 +5,11 @@ import static com.google.common.collect.ImmutableTable.toImmutableTable;
 import static org.batfish.common.util.CollectionUtil.toImmutableMap;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableTable;
+import com.google.common.collect.SetMultimap;
+import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.google.common.collect.Table;
 import com.google.common.collect.Tables;
@@ -14,7 +17,6 @@ import io.opentracing.Scope;
 import io.opentracing.Span;
 import io.opentracing.util.GlobalTracer;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -22,6 +24,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Stream;
 import net.sf.javabdd.BDD;
+import net.sf.javabdd.BDDFactory;
 import org.batfish.bddreachability.transition.Transition;
 import org.batfish.bddreachability.transition.Transitions;
 import org.batfish.common.BatfishException;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7820)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java, line 87 at r2](https://reviewable.io/reviews/batfish/batfish/7820#-Mr3NbE5CBxooeTGEsNn:-Mr3NbE5CBxooeTGEsNo:beo8ux6) ([raw file](https://github.com/batfish/batfish/blob/82dfaac6c09a1260f6b48e8ecce63c6a83f628cc/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L87)):*
> ```Java
>               BDD prior = reachableSets.get(dirtyState);
>               BDD learned = factory.orAll(inputs);
>               BDD newValue = prior == null ? learned : learned.or(prior);
> ```

Probably it would be better to do this as 2 `orAll` ops. The `newValue` one first, and exit early. The `learned` one second, after the `if`, which will benefit from caching of the first one.



_Code quote:_
```Java
              BDD learned = factory.orAll(inputs);
              BDD newValue = prior == null ? learned : learned.or(prior);
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7820)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7820#-Mr3NxpY5_6lwSC-X49U:-Mr3NxpY5_6lwSC-X49V:b-ldzbu0) (no related file):*
FYI the build failure is just PMD complaining that BDD library defined `.equals` incorrectly.

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7820)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java, line 73 at r1](https://reviewable.io/reviews/batfish/batfish/7820#-Mr3X5DRDE16LgYdB5gc:-Mr3X5DRDE16LgYdB5gd:bh8f729) ([raw file](https://github.com/batfish/batfish/blob/cce377569287434abd4ea7bdb9e5c835ea67f5f4/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L73)):*
> ```Java
>             dirtyState -> {
>               Map<StateExpr, Transition> dirtyStateEdges = edges.row(dirtyState);
>               if (dirtyStateEdges == null) {
> ```

FYI this was never true b/c Table gives you an empty map, not null, for missing key.



_Code quote:_
```Java
dirtyStateEdges == null
```

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7820)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7826, opened 2021-12-16
JFactory: don't flush cache when GC frees BDDs

anothermattbrown:
Instead, walk over the cache and delete entries for freed BDDs. This is safe since we do it as part of GC, i.e. before any new BDDs are allocate: we know which BDDs are freed vs not.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7826)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7829, opened 2021-12-17
Produce VXLAN edges from type-5 routes

arifogel:
- each iteration, update `Layer3Vni`s' learned VTEP IPs based on next
hop VTEP routes in the main RIB with the same VNI
- construct VXLAN edges between for each Layer3Vni pair containing each
other's source IP in their learned VTEP IPs

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DataPlane.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/DataPlane.java
@@ -7,6 +7,7 @@ import java.util.Set;
 import java.util.SortedMap;
 import javax.annotation.Nonnull;
 import org.batfish.datamodel.vxlan.Layer2Vni;
+import org.batfish.datamodel.vxlan.Layer3Vni;
 
 public interface DataPlane extends Serializable {
 
@@ -52,4 +53,12 @@ public interface DataPlane extends Serializable {
    */
   @Nonnull
   Table<String, String, Set<Layer2Vni>> getLayer2Vnis();
+
+  /**
+   * Return {@link Layer3Vni} for each node/VRF. Returned settings are based on the vni settings in
+   * a {@link Vrf}, but may include additional information obtained during dataplane computation,
+   * such as updated VTEPs due to EVPN route exchange.
+   */
+  @Nonnull
+  Table<String, String, Set<Layer3Vni>> getLayer3Vnis();
 }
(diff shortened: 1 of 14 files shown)


dhalperi (commented):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7829)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java, line 175 at r1](https://reviewable.io/reviews/batfish/batfish/7829#-Mr9P69B2GUOgC5MJsRJ:-Mr9P69B2GUOgC5MJsRK:buf0x76) ([raw file](https://github.com/batfish/batfish/blob/5ea38ab1684aad41823ed416a21d6125a4eff722/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java#L175)):*
> ```Java
>   @Nonnull
>   public <V extends Vni> Optional<V> getVniSettings(
>       String hostname, int vni, Function<Vrf, Map<Integer, V>> vniGetter) {
> ```

it seems weird to expose this as the API, rather than 2 named getters. But okay as-is if you want to keep it this way.



_Code quote:_
```Java
  public <V extends Vni> Optional<V> getVniSettings(
      String hostname, int vni, Function<Vrf, Map<Integer, V>> vniGetter) {
```

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7829)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java, line 175 at r1](https://reviewable.io/reviews/batfish/batfish/7829#-Mr9P69B2GUOgC5MJsRJ:-Mr9cwZh7mJuGAtrHZYz:bgmte4w) ([raw file](https://github.com/batfish/batfish/blob/5ea38ab1684aad41823ed416a21d6125a4eff722/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/NetworkConfigurations.java#L175)):*

it seems weird to expose this as the API, rather than 2 named getters. But okay as-is if you want to keep it this way.
</blockquote></details>

I started out with it cause I thought I could templatize more of the functions. If I were going to clean it up, I'd probably change the whole call stack. Not worth it right now IMO.
Merging as is.

PR #7839, opened 2021-12-20
Don't use parallel stream with LinkedList

ratulm:
Using parallelStream with LinkedList is not thread safe (`forEach(jobs::add)`). So, made it sequential stream. The perf gain of parallel stream is unclear in this context.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7839)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7841, opened 2021-12-20
Sonic: L3 interfaces should not be switchports

ratulm:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7841)* status: all files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/vendor/sonic/representation/SonicConversions.java, line 142 at r1](https://reviewable.io/reviews/batfish/batfish/7841#-MrO6KNB9hIIRnNRtAML:-MrO6KNB9hIIRnNRtAMM:b-u68l07) ([raw file](https://github.com/batfish/batfish/blob/1c1068a5f5453303abe690cd4b83da925e9ed60d/projects/batfish/src/main/java/org/batfish/vendor/sonic/representation/SonicConversions.java#L142)):*
> ```Java
>               memberInterface.setSwitchportMode(SwitchportMode.TRUNK);
>             }
>             memberInterface.setNativeVlan(vlanId);
> ```

Why would you set these two trunk properties on an interface that isn't a switchport? This violates VI invariants.
Same question for access vlan below.

PR #7844, opened 2021-12-20
Frr: Rename router bgp rules

ratulm:
Because `bgp` can occur at the top-level (not just under router bgp), and brings things in like with frr/ospf from a naming standpoint. 

No functional change.


arifogel (commented):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7844)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7862, opened 2021-12-25
NX-OS: Prefer physical interfaces for BGP router ids

ratulm:
After loopbacks, physical interfaces are preferred over other interfaces:
https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/unicast/configuration/guide/l3_cli_nxos/l3_bgp.html#94583

```
By default, Cisco NX-OS sets the router ID to the IPv4 address of a loopback interface on the router. 
If no loopback interface is configured on the router, the software chooses the highest IPv4 address 
configured to a physical interface on the router to represent the BGP router ID.
```

("highest IPv4" in here is wrong. real world shows lowest.)

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/Conversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/Conversions.java
@@ -177,6 +177,22 @@ public final class Conversions {
       return lowestLoopback.get();
     }
 
+    // Next, NX-OS prefers "physical" interfaces.
+    Optional<Ip> lowestPhysical =
+        interfaces.stream()
+            .filter(
+                i ->
+                    // all three are physical interfaces for this computation
+                    i.getInterfaceType() == InterfaceType.AGGREGATED
+                        || i.getInterfaceType() == InterfaceType.AGGREGATE_CHILD
+                        || i.getInterfaceType() == InterfaceType.PHYSICAL)
+            .map(org.batfish.datamodel.Interface::getConcreteAddress)
+            .map(ConcreteInterfaceAddress::getIp)
+            .min(Comparator.naturalOrder());
+    if (lowestPhysical.isPresent()) {
+      return lowestPhysical.get();
+    }
+
     // Finally, NX-OS uses the first non-loopback interface defined in the vrf, assuming no loopback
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7862)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @arifogel and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7862#-Mrp2fSr4kkmc8_Pk_Bv:-Mrp2fSr4kkmc8_Pk_Bw:bnqae5f) (no related file):*
Please make a file like https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/bgp/nxos-bgp-router-id .

That file can be copied and pasted into a fresh NX-OS VM, committed and reloaded, and then the router IDs will be shown.

Presumably need an IRB and a physical interface where the IRB has a lower IP and is up, but loses? (after commit and reload -- it is nondeterministic if commands are entered live)

---

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7862)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @arifogel and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7862#-Mrp2fSr4kkmc8_Pk_Bv:-Mrp2iWs1e-ertVY3szR:bssgc58) (no related file):*

Please make a file like https://github.com/batfish/batfish/blob/master/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/bgp/nxos-bgp-router-id .

That file can be copied and pasted into a fresh NX-OS VM, committed and reloaded, and then the router IDs will be shown.

Presumably need an IRB and a physical interface where the IRB has a lower IP and is up, but loses? (after commit and reload -- it is nondeterministic if commands are entered live)
</blockquote></details>

(and make sure the behavior when doing that in a VM is as expected, of course)

---

ratulm:
I couldn't repro the behavior in this PR in a lab. (Not seeing consistency and stickiness past reboot.) Closing until I can figure this out.

PR #7880, opened 2021-12-31
VRRP question: Filtering by address

ratulm:
And add active interface columns and better column descriptions.


ratulm:
Closing in favor of https://github.com/batfish/batfish/pull/7889

PR #7892, opened 2022-01-04
NXOS: Support removing static routes in VRF context

corinaminer:
This was already supported for static routes in the default VRF, and the extraction method used for the `no_ip_route` rule already removes from `_currentVrf` rather than the default VRF, so all that was needed was to add that rule in the `no ip` rule for VRFs.


dhalperi (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7892#-:-MsY6KK94ITz3_BQ_U3M:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #7898, opened 2022-01-05
NX-OS: better handling for `[no] ip route`

corinaminer:
In NX-OS, two static route definition lines will only result in different routes if key properties (prefix or next hop info) differ. For example, these two lines will create different routes:
```
ip route 10.0.1.0/24 10.0.1.1
ip route 10.0.1.0/24 Ethernet1/1 10.0.1.1
```
But these two lines will result in one route with the properties listed in the second line:
```
ip route 10.0.15.0/24 10.0.1.1
ip route 10.0.15.0/24 10.0.1.1 track 500 name foo tag 1000 5
```

Additionally, a `no ip route` command will remove the route with the specified key properties, ignoring any other properties on the route or in the `no` command, so the second line here will work:
```
ip route 10.0.1.0/24 10.0.1.1 track 500 tag 10
no ip route 10.0.1.0/24 10.0.1.1 name foo tag 1000 5
```

Batfish currently creates a new route for every (valid) `ip route` command, and only removes routes via `no ip route` if the `no` line exactly matches the existing route. To update the behavior, this PR:
- Changes the NX-OS VS model to store static routes keyed on a new `StaticRouteKey` object that includes all the key properties, rather than keying on prefix
- Makes the non-key property fields in `StaticRoute` nonfinal, since they may be modified
- Updates behavior in extraction for `ip route` and `no ip route` accordingly

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java
@@ -203,7 +203,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Range;
 import com.google.common.collect.Table;
 import com.google.common.primitives.Ints;
-import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7898)* status: all files reviewed, 3 unresolved discussions (waiting on @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java, line 18 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKuZ3CZqSountb5LO:-MscKuZ3CZqSountb5LP:b-8acvuu) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java#L18)):*
> ```Java
> 
>   /** Uniquely identifies a static route. */
>   public static final class StaticRouteKey implements Serializable {
> ```

missing tests for this, I think. Mainly equals tests.
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKxmv4umOHH0m0M6w:-MscKxmv4umOHH0m0M6x:by8oqgb) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java#L30)):*
> ```Java
>         @Nullable String nextHopInterface,
>         @Nullable Ip nextHopIp,
>         @Nullable String nextHopVrf) {
> ```

Is it worth checking invariants here, or at least documenting them?

Presumably discard and nhint can't both be present, for example.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route, line 70 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKXF_3KJ6ltRMyBzA:-MscKXF_3KJ6ltRMyBzB:b-uqfttq) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route#L70)):*
> ```
> ip route 10.0.15.0/24 10.255.1.254 name foo
> ip route 10.0.15.0/24 10.255.1.254 name bar
> ip route 10.0.15.0/24 10.255.1.254 track 500 tag 1000 5
> ```

confirming: after this line, `name` is not cleared on the box? That would surprise me.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7898)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java, line 18 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKuZ3CZqSountb5LO:-Msd-KlpEX8e6eJS7JAw:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java#L18)):*

missing tests for this, I think. Mainly equals tests.
</blockquote></details>

done
___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java, line 30 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKxmv4umOHH0m0M6w:-Mscvq9P5t5xSaEcGJwC:bhlh6tw) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/main/java/org/batfish/representation/cisco_nxos/StaticRoute.java#L30)):*

Is it worth checking invariants here, or at least documenting them?

Presumably discard and nhint can't both be present, for example.
</blockquote></details>

yeah that's reasonable. Factored out the checks that already happen for `StaticRoute`.
___
*[projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route, line 70 at r1](https://reviewable.io/reviews/batfish/batfish/7898#-MscKXF_3KJ6ltRMyBzA:-Mscrmv6DmDxblMQiBJk:b-ayrs) ([raw file](https://github.com/batfish/batfish/blob/0ed4f4b809b8af02b57d323ee2aba7c4bd699193/projects/batfish/src/test/resources/org/batfish/grammar/cisco_nxos/testconfigs/nxos_static_route#L70)):*

confirming: after this line, `name` is not cleared on the box? That would surprise me.
</blockquote></details>

correct, I tested that. All of these non-key attributes accumulate and replace. Afaict you can't clear them without removing the route.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7898)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

PR #7925, opened 2022-01-12
Cleanup track evaluation

arifogel:
- for tracking an interface, evaluator now returns true when interface is up
- TrackAction now has property getNegateTrack()
  - when true, client of TrackMethod evaluator should negate the result evaluation when deciding
    whether to evaluate TrackAction

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
@@ -344,7 +344,8 @@ public final class IpOwners {
         .forEach(
             (trackName, action) -> {
               TrackMethod trackMethod = trackMethods.get(trackName);
-              if (trackMethod != null && trackMethod.accept(trackMethodEvaluator)) {
+              if (trackMethod != null
+                  && (trackMethod.accept(trackMethodEvaluator) ^ action.getNegateTrack())) {
                 action.accept(hsrpEvaluator);
               }
             });
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/IpOwners.java
@@ -344,8 +344,7 @@ public final class IpOwners {
         .forEach(
             (trackName, action) -> {
               TrackMethod trackMethod = trackMethods.get(trackName);
-              if (trackMethod != null
-                  && (trackMethod.accept(trackMethodEvaluator) ^ action.getNegateTrack())) {
+              if (trackMethod != null && (trackMethod.accept(trackMethodEvaluator))) {
                 action.accept(hsrpEvaluator);
               }
             });
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Names.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Names.java
@@ -92,6 +92,10 @@ public final class Names {
         type.getExplanation());
   }
 
+  public static String generatedNegatedTrackMethodId(String trackMethodId) {
+    return String.format("~!%s~", trackMethodId);
+  }
+
   public static String generatedBgpCommonExportPolicyName(String vrf) {
     return String.format("~BGP_COMMON_EXPORT_POLICY:%s~", vrf);
   }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java
@@ -4,6 +4,7 @@ import static com.google.common.base.MoreObjects.toStringHelper;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
 import javax.annotation.Nullable;
 
 /** Decrements HSRP/VRRP priority by a given subtrahend when the tracked object is not available */
@@ -11,10 +12,14 @@ public class DecrementPriority implements TrackAction {
   private static final String PROP_SUBTRAHEND = "subtrahend";
 
   private final int _subtrahend;
+  private final boolean _negateTrack;
 
   @JsonCreator
-  public DecrementPriority(@JsonProperty(PROP_SUBTRAHEND) int subtrahend) {
+  public DecrementPriority(
+      @JsonProperty(PROP_SUBTRAHEND) int subtrahend,
+      @JsonProperty(PROP_NEGATE_TRACK) boolean negateTrack) {
     _subtrahend = subtrahend;
+    _negateTrack = negateTrack;
   }
 
   @Override
@@ -25,7 +30,8 @@ public class DecrementPriority implements TrackAction {
     if (!(obj instanceof DecrementPriority)) {
       return false;
     }
-    return _subtrahend == ((DecrementPriority) obj)._subtrahend;
+    DecrementPriority that = (DecrementPriority) obj;
+    return _subtrahend == that._subtrahend && _negateTrack == that._negateTrack;
   }
 
   @JsonProperty(PROP_SUBTRAHEND)
@@ -33,9 +39,15 @@ public class DecrementPriority implements TrackAction {
     return _subtrahend;
   }
 
+  @JsonProperty(PROP_NEGATE_TRACK)
+  @Override
+  public boolean getNegateTrack() {
+    return _negateTrack;
+  }
+
   @Override
   public int hashCode() {
-    return _subtrahend;
+    return Objects.hash(_subtrahend, _negateTrack);
   }
 
   @Override
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java
@@ -12,14 +12,10 @@ public class DecrementPriority implements TrackAction {
   private static final String PROP_SUBTRAHEND = "subtrahend";
 
   private final int _subtrahend;
-  private final boolean _negateTrack;
 
   @JsonCreator
-  public DecrementPriority(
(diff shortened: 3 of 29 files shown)


dhalperi (commented):
> Reviewed 2 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @arifogel and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtB28HVEIKq-6zDes6A:brxkbzu) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*
> ```Java
> /** An action to take when a {@link TrackMethod} has been triggered (is down). */
> @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "class")
> public interface TrackAction extends Serializable {
> ```

This seems like it belongs in the track, not the action, right?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtB7Xk01o3sZlaa_4ht:b9u8i0u) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*

This seems like it belongs in the track, not the action, right?
</blockquote></details>

DO you mean `TrackMethod`? It cannot, since the method is associated with a single index. That is, two callers could refer to the same `TrackMethod` by index, but only one caller would want to negate. The `TrackAction`, in contrast, is associated with the caller.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: 2 of 16 files reviewed, 1 unresolved discussion (waiting on @arifogel and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtBEg8rEaT50YNwHs8P:b-m00prs) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*

DO you mean `TrackMethod`? It cannot, since the method is associated with a single index. That is, two callers could refer to the same `TrackMethod` by index, but only one caller would want to negate. The `TrackAction`, in contrast, is associated with the caller.
</blockquote></details>

I could think of:

1. Add `TrackMethod` `Negate(TrackMethod inner)`. On a VS-basis, create up to two different tracks (one negated, one not) depending on the VS use semantics. Note that as `StaticRoute#_track` is a string, nothing is keeping us from doing, say, `100-up` and `100-down` for VS `track 100`.

2. Add `TrackReference(String idx, boolean negate)`. Replace current `String _track` with `TrackReference` in VI.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: 0 of 30 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtBHDBo1WEUGIFRnxSz:bl7oz6r) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*

I could think of:

1. Add `TrackMethod` `Negate(TrackMethod inner)`. On a VS-basis, create up to two different tracks (one negated, one not) depending on the VS use semantics. Note that as `StaticRoute#_track` is a string, nothing is keeping us from doing, say, `100-up` and `100-down` for VS `track 100`.

2. Add `TrackReference(String idx, boolean negate)`. Replace current `String _track` with `TrackReference` in VI.
</blockquote></details>

Went with a combination. See how you like it.

dhalperi (commented):
> Reviewed 30 of 30 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java, line 40 at r3](https://reviewable.io/reviews/batfish/batfish/7925#-MtEQnnK5ZZwHw6shJbQ:-MtEQnnK5ZZwHw6shJbR:bh6nrmt) ([raw file](https://github.com/batfish/batfish/blob/b04fe020d9e6a885ebf35cee7f101a45db77b461/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java#L40)):*
> ```Java
>   @Override
>   public int hashCode() {
>     return Objects.hash(_subtrahend);
> ```

why this change? It's a no-op on int btw.

If needed, use `Integer.hashCode` to avoid some work.



_Code quote:_
```Java
Objects.hash(_subtrahend);
```
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/PredicateTrackMethodEvaluator.java, line 36 at r3](https://reviewable.io/reviews/batfish/batfish/7925#-MtEQRUi3HitYw3Qhits:-MtEQRUi3HitYw3Qhitt:bd2w8u9) ([raw file](https://github.com/batfish/batfish/blob/b04fe020d9e6a885ebf35cee7f101a45db77b461/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/PredicateTrackMethodEvaluator.java#L36)):*
> ```Java
>       return false;
>     }
>     return trackedInterface.getActive() && !trackedInterface.getBlacklisted();
> ```

(FYI) why is this necessary? Seems like `active` should already have taken this into account.



_Code quote:_
```Java
 && !trackedInterface.getBlacklisted()
```
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtERnO33-PIDXeIGdOM:b-9ug5iw) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*

Went with a combination. See how you like it.
</blockquote></details>

Interesting. is there a use case for Reference outside of Negated?

Fine either way.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: 29 of 30 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java, line 40 at r3](https://reviewable.io/reviews/batfish/batfish/7925#-MtEQnnK5ZZwHw6shJbQ:-MtEciFTAjoSV1qoBP49:b2zr20m) ([raw file](https://github.com/batfish/batfish/blob/b04fe020d9e6a885ebf35cee7f101a45db77b461/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/DecrementPriority.java#L40)):*

why this change? It's a no-op on int btw.

If needed, use `Integer.hashCode` to avoid some work.
</blockquote></details>

Oh that was a mindless revert. Fixed
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/PredicateTrackMethodEvaluator.java, line 36 at r3](https://reviewable.io/reviews/batfish/batfish/7925#-MtEQRUi3HitYw3Qhits:-MtEcuuq35abOqrKW1Hc:b-nlsn3w) ([raw file](https://github.com/batfish/batfish/blob/b04fe020d9e6a885ebf35cee7f101a45db77b461/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/PredicateTrackMethodEvaluator.java#L36)):*

(FYI) why is this necessary? Seems like `active` should already have taken this into account.
</blockquote></details>

Didn't dig into invariants.
Looing now, there are several places in the code that check active and blacklisted together.
E.g. in `IpOwners.computeIpInterfaceOwners`:
```
                  if ((!i.getActive() || i.getBlacklisted()) && excludeInactive) {
                    return;
                  }
```
 I think if there is an invariant that blacklisted -> !active, then, we should fix this in a separate PR for all such usages.

Note this may be ripe for fixing when we add granularity to interface status.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java, line 8 at r1](https://reviewable.io/reviews/batfish/batfish/7925#-MtB28HU5v8Pazo104vG:-MtEdnX-EchP6whtk3CE:b-x6ox78) ([raw file](https://github.com/batfish/batfish/blob/a224fb830cc1eee0ab4e4dd1985ecadd082c4399/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/tracking/TrackAction.java#L8)):*

Interesting. is there a use case for Reference outside of Negated?

Fine either way.
</blockquote></details>

Just future-proofing in case e.g. XR or Juniper allow boolean combinations or other complex crap.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7925)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

PR #7929, opened 2022-01-13
Junos: Proper linking of implicitly created interfaces

ratulm:
Under some conditions (see test file) BF automatically creates interfaces. Prior to this PR, when such an interface was a sub-interface, it was not being properly linked to its parent. In particular, the parent's units map had the sub-interface but the sub-interface did not have the parent configured. The partial linking was later preventing the parent from being set and then causing an NPE.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7929)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ratulm)

PR #7932, opened 2022-01-13
Junos: Ignore protocol OAM

ratulm:


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7932)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/ignored-protocols, line 4 at r1](https://reviewable.io/reviews/batfish/batfish/7932#-MtK6DE46-SvvMSxM1s-:-MtKATDh9SmzI-eZvsE0:b-669jor) ([raw file](https://github.com/batfish/batfish/blob/7ae98cd5afc929394b9bea39fd33c6bf3a57b00e/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/ignored-protocols#L4)):*

Fine, but curious why not?
</blockquote></details>

Mainly because I didn't have example legal lines handy for all of those. I didn't think putting garbage was worth it.

PR #7936, opened 2022-01-14
NX-OS: Trunk interfaces' allowed VLANs must be active

corinaminer:
Previously, if no allowed VLANs were specified for an interface in trunk mode, we allowed the full valid range (1-4094). Actually it should default to all active declared VLANs.

Similarly, if allowed VLANs are specified, any that are not active declared VLANs should be ignored.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java
@@ -5928,14 +5928,13 @@ public final class CiscoNxosControlPlaneExtractor extends CiscoNxosParserBaseLis
     }
     _currentInterfaces.forEach(
         iface -> {
-          if (iface.getSwitchportMode() == null) {
-            // NX-OS has these commands in show run all even for interfaces in other modes.
-            iface.setSwitchportMode(SwitchportMode.TRUNK);
-          }
           if (ctx.ADD() != null) {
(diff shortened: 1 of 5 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7936)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java, line 5933 at r1](https://reviewable.io/reviews/batfish/batfish/7936#-MtPRpbgC7QHPXqoANl7:-MtPRpbgC7QHPXqoANl8:bvm8kmb) ([raw file](https://github.com/batfish/batfish/blob/5855fac8e130ecf71cebe09d99e8c2cff9b27426/projects/batfish/src/main/java/org/batfish/grammar/cisco_nxos/CiscoNxosControlPlaneExtractor.java#L5933)):*
> ```Java
>           if (iface.getSwitchportMode() == null) {
>             // NX-OS has these commands in show run all even for interfaces in other modes.
>             iface.setSwitchportMode(SwitchportMode.TRUNK);
> ```

fyi: It does let you configure allowed VLANs on non-trunk interfaces, but this does not automatically put the interface in trunk mode (`show vlan` shows that such interfaces are not associated with their specified VLANs).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7936)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

Issue #1644, opened 2018-06-12
Support OSPF NSSA external default route origination


arifogel:
Need to support originating a default route into NSSA as external OSPF route.

dhalperi:
@progwriter any update here?

progwriter:
No, need much better support for all the ospf area knobs in VI.

jeffkala:
Running into this currently, is there a workaround or update?

arifogel:
Unfortunately @progwriter is no longer active on this repo.

Can you provide a minimal set of configs that showcase this issue? What route do you expect to see where that isn't showing up in Batfish? What are your reproduction steps?

If needed, you can scrub sensitive information using [netconan](https://github.com/intentionet/netconan), available on pypi.

jeffkala:
At a simplified level:

core-rtr <-------> local-rtr1

core-rtr
```
 router ospf 1
   bfd default
   router-id 10.1.1.1
   area 0.0.0.90 stub no-summary
   default-information originate
 ```
 
 local-rtr1
 ```
router ospf 1
   bfd
   router-id 10.1.1.6
   area 0.0.0.90 stub
 ```
  
  This is simplified if more information is needed I can attempt to provide it.
  
  What I'm noticing is when checking the routes in local-rtr1 it does not have a default route.  Because of this all tracing questions fails with no_route from this router.  Looking at the details of the devices in batfish they accurately knew that core-rtr1 is a ospf NSSA and local-rtr1 is a STUB, just expected that batfish would properly inject that default route into the routes.
  
  Hopefully this is enough details let me know if more are needed.

jeffkala:
@arifogel is this enough detail to proceed?  Happy to help however I can.

arifogel:
Thanks @jeffkala, but I still need a bit more information.

To start, I need to know the vendor language. Is it CIsco IOS? NX-OS? Arista EOS (if EOS, give version)? FRR?

Also, need some more config to be able to test in Batfish. If you already have a snapshot zip you are feeding, I could use that. You may anonymize the files and include only relevant ones (but make sure network wildcards and OSPF area numbers are not garbled).

In particular, what is missing from the config for testing is:
- interface config, especially addressing and any OSPF settings
- area assignment of interface(s)
- any other areas (e.g. does core-rtr also participate in area 0?) necessary for this configuration to work
- any static routes that overlap with networks communicated over OSPF, including default route if applicable

jeffkala:
[...]

PR #7943, opened 2022-01-16
Arista EOS fix OSPF network statement processing

arifogel:
- convert wildcard networks to prefixes, since only valid wildcards are
inverse of valid subnet mask
- remove OspfWildcardNetwork, and simplify interface ospf network area
assignment to just checking if ip is in network prefix
- fixes batfish/batfish#1644


ratulm (commented):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7943)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer and @dhalperi)

PR #7961, opened 2022-01-20
Junos: Check VRF when using lo0-based BGP local IP

ratulm:
This PR makes two small changes to inferring local Ip for BGP neighbors:
1. Make sure that the neighbor is in the same VRF as lo0 (default)
2. Do not pick 127.0.01

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/Interface.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/Interface.java
@@ -317,6 +317,10 @@ public class Interface implements Serializable {
     return _routingInstance;
   }
 
+  public @Nonnull InterfaceType getType() {
+    return _type;
+  }
+
   public Map<String, Interface> getUnits() {
     return _units;
   }
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
@@ -344,7 +344,7 @@ public final class JuniperConfiguration extends VendorConfiguration {
   /** Maximum IS-IS route cost if wide-metrics-only is not set */
   @VisibleForTesting static final int MAX_ISIS_COST_WITHOUT_WIDE_METRICS = 63;
 
-  private static final String FIRST_LOOPBACK_INTERFACE_NAME = "lo0";
+  @VisibleForTesting static final String FIRST_LOOPBACK_INTERFACE_NAME = "lo0";
 
   private static String communityRegexToJavaRegex(String regex) {
     String out = regex;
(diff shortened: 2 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7961)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @arifogel and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7961#-MtrhzGY9q8RAYAJCtU1:-MtrhzGY9q8RAYAJCtU2:b-jowvlb) (no related file):*
I don't think this is necessary. The [Juniper article](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/local-address-edit-protocols-bgp.html) is long and confusing, but I'm homing in on these parts:

```
For protocols in which the local address is unconstrained by the protocol specification, for example IBGP and multihop EBGP, if you do not configure a specific local address when configuring the protocol, the local address is chosen using the same methods as other locally generated IP packets.
...
If you do not configure a local address, BGP uses the routing device’s source address selection rules to set the local address.
```

That is what we do by default in VI if not set in VS.

I think the Juniper docs continually use `lo0` (it's the default loopback), but of course you can have more than one routing instance, and each of them should have its own loopback. Using literally only `lo0` would be .. insane?

I think "peer with a loopback in IBGP or multihop EBGP" is just working with best practices and not at all what happens.

tl;dr: I'd say just remove this code in VS and let VI do the right thing.

---

saparikh:
I am pretty sure Juniper allows you to create multiple logical units under `lo0`. If you need a loopback interface in a VRF, folks typically create a new unit for each VRF that needs on and assigns an IP to each logical unit.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7961)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @arifogel, @dhalperi, and @ratulm)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/7961#-MvPgvww5DbtSjQ7qatN:-MvPgvww5DbtSjQ7qatO:b-knmr2p) (no related file):*
Need to figure out what to do about inactive interfaces (before merging)

---

dhalperi (commented):
> Reviewed 6 of 7 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7961)* status: all files reviewed, 3 unresolved discussions (waiting on @arifogel and @ratulm)

___
*[projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java, line 771 at r3](https://reviewable.io/reviews/batfish/batfish/7961#-MvPtXh8C0qa497FOQjP:-MvPtXh8C0qa497FOQjQ:b-7zgjy7) ([raw file](https://github.com/batfish/batfish/blob/331a916198a88fa837230a871bab096510c186af/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L771)):*
> ```Java
>     // in a routing instance
>     if (loopback.isPresent()
>         && !loopback.get().getPrimaryAddress().getIp().equals(Ip.parse("127.0.0.1"))) {
> ```

Will this crash if there is no primary address configured?



_Code quote:_
```Java
!loopback.get().getPrimaryAddress().getIp()
```
___
*[projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java, line 899 at r4](https://reviewable.io/reviews/batfish/batfish/7961#-MvPuCBW-MfNuKQXvD1B:-MvPuCBW-MfNuKQXvD1C:bwiyirw) ([raw file](https://github.com/batfish/batfish/blob/e5e8e98fe055e33ca817fef632715c54bb87b1af/projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java#L899)):*
> ```Java
>       ri.getInterfaces().put(em0.getName(), em0);
> 
>       // l0 has legal (not 127.0.0.1) address
> ```

lo0 has no address?

PR #7977, opened 2022-01-31
ExplicitAsPathSet: port off it and delete

dhalperi:
The only user was Juniper, which I was able to port onto the legacy code by removing
the need for a prepended space.

Then delete and fixup some tests.


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @millstein)

___
*[projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java, line 108 at r2](https://reviewable.io/reviews/batfish/batfish/7977#-MuhyLJe-N3sp6T_CKqR:-MuhyLJe-N3sp6T_CKqS:b4bk99a) ([raw file](https://github.com/batfish/batfish/blob/62848e7e56b0aecc2b0a6f4f10dc31a0f9e3b1f0/projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java#L108)):*
> ```Java
>   @SuppressSubnodes
>   Rule T_Dot() {
>     return Sequence(Ch('.'), push(MAYBE_SEPARATOR + "\\d+"));
> ```

I'm confused about how this works.  Is the idea that you are over-approximating the actual allowed grammar, but it's ok since you'll always match against well-formed AS-paths?  The regex ".*8075.*" previously translated to "^ \d+8075 \d+".  What does it translate to now?  "^(^| )\d+8075(^| )\d+" doesn't seem like it helps.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @millstein)

___
*[projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java, line 108 at r2](https://reviewable.io/reviews/batfish/batfish/7977#-MuhyLJe-N3sp6T_CKqR:-MuhzSxRED0F3F0rgGiF:bxk9tt4) ([raw file](https://github.com/batfish/batfish/blob/62848e7e56b0aecc2b0a6f4f10dc31a0f9e3b1f0/projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java#L108)):*

I'm confused about how this works.  Is the idea that you are over-approximating the actual allowed grammar, but it's ok since you'll always match against well-formed AS-paths?  The regex ".*8075.*" previously translated to "^ \d+8075 \d+".  What does it translate to now?  "^(^| )\d+8075(^| )\d+" doesn't seem like it helps.
</blockquote></details>

The two translations should end with $.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java, line 108 at r2](https://reviewable.io/reviews/batfish/batfish/7977#-MuhyLJe-N3sp6T_CKqR:-Mui4Tvo5YJ6AdAVoTzg:b-eh75pp) ([raw file](https://github.com/batfish/batfish/blob/62848e7e56b0aecc2b0a6f4f10dc31a0f9e3b1f0/projects/batfish/src/main/java/org/batfish/representation/juniper/AsPathRegex.java#L108)):*

The two translations should end with $.
</blockquote></details>

Talked offline, but for posterity:

The construct `(^| )` is intended to mean "require a space, unless at start of the word". At the front of the regex, this will not require a space to have been prepended, so we can use `AsPathMatchRegex` (which does not prepend a space when testing an AsPathString) instead of `ExplicitAsPathSet` (which does). In the middle of a regex, this will be equivalent to just space.

`^^` is legal and works, so `^(^| )\d+$` works as a representation for, e.g.,  the Juniper regex `.`.

millstein (commented):
> Reviewed 13 of 14 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7977)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #7984, opened 2022-02-01
SearchRoutePolicies: handle the Java regex format for Juniper AS-path regexes

millstein:
Addresses two problems in SearchRoutePolicies for handling AS-path regexes from Juniper devices:
* Previously the \d character class for digits was not handled properly.
* A recent update to parsing for Juniper AS-path regexes introduces the possibility of having two start-of-string ^ characters in a row, which requires special handling.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/7984)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8014, opened 2022-02-08
Juniper fix apply-groups processing bug

arifogel:
- We generate `WILDCARD_ARTIFACT` instead of `WILDCARD` tokens during preprocessor lexing
- An exception is thrown when intermediate preprocessor line fails to parse due to lexing of
  `WILDCARD_ARTIFACT`

Solution: make parser allow `WILDCARD_ARTIFACT` wherever `WILDCARD` is allowed

Also: fix handling of quoted wildcards and their application to quoted strings.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8014)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8042, opened 2022-02-13
Cisco IOS-XR improve interface name lexing and parsing

arifogel:
- stricter lexing
- lex as a single token
- compile patterns once
- remove interface_name_unstructured
- ditch interface ranges, which do not exist on XR
  - there is some kind of apply-group syntax instead


dhalperi (commented):
> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8042)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8043, opened 2022-02-13
Cisco IOS-XR convert HSRP tracks

arifogel:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8043)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java, line 200 at r1](https://reviewable.io/reviews/batfish/batfish/8043#-MvoPH-i7S1KLj9UA7AI:-MvoSNXM4fa-nYUNVWP0:b-q58qdz) ([raw file](https://github.com/batfish/batfish/blob/53cb4f11eb0d2b50cb775cb90049063ba5bcdec7/projects/batfish/src/main/java/org/batfish/representation/cisco_xr/CiscoXrConversions.java#L200)):*

Why be opaque with this. Why `~`, etc.?

how about just `%s is down`?
</blockquote></details>

Well I surrounded in tildes to avoid clashing with manually-specified track names. But as it turns out (just checked), you cannot have spaces in such a name on XR (quoted strings not allowed).  So %s is down is fine. Also I assume by your comment that we don't care about hiding these generated track names via the usual surrounded-by-tildes heuristic.

Done.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8043)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #8073, opened 2022-02-24
BgpRoute: move properties into interned class

dhalperi:
If many routes are received from the same neighbor,
all of them will share the same attributes.

Save (possibly a lot of) RAM in such cases.

On one smallish network, results after computing the dataplane:

Before:
1250631 BgpRoute objects, 104 bytes each, 130M RAM

After:
1250631 BgpRoute objects, 48 bytes each, 60M RAM
 148124 BgpRouteAttribute objects, 80 bytes each, 11.M RAM

This seems like the sweet spot, at least in this network. If we move
receivedFromIp up to BgpRoute, it reduces to about 117K BgpRouteAttribute
objects, but the additional overhead on BgpRoute is not worth it.

1250631 BgpRoute objects, 56 bytes each, 70M RAM
 117000 BgpRouteAttribute objects, 72 bytes each, 8.4M RAM

There may be networks where this additional move is beneficial, e.g.,
high ECMP and ASN reuse (so AS Paths are the same, but receivedFromIp are
different).

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
@@ -9,6 +9,7 @@ import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
+import java.io.Serializable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8073)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @corinaminer, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java, line 178 at r1](https://reviewable.io/reviews/batfish/batfish/8073#-Mwh6H612G7AxMo_Or9f:-Mwh6H626WZKRNB-yYsg:bsxwqkv) ([raw file](https://github.com/batfish/batfish/blob/34efdf748ec856bfe7bcf25b5d7aeb273105ab65/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java#L178)):*
> ```Java
>   //   (100 bytes seems smallest possible object size, would be 100 MiB total).
>   private static final LoadingCache<BgpRouteAttributes, BgpRouteAttributes> ATTRIBUTE_CACHE =
>       Caffeine.newBuilder().softValues().maximumSize(1 << 20).build(a -> a);
> ```

1 << 20 is 1Mibbion (sorry) entries, right? How does this relate to the comments above? Did you mean to update maximum size above to `2^20`?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8073)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @anothermattbrown, @arifogel, and @corinaminer)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java, line 178 at r1](https://reviewable.io/reviews/batfish/batfish/8073#-Mwh6H612G7AxMo_Or9f:-MwiJNRc8vw8xS-87HCc:b-p6iwvm) ([raw file](https://github.com/batfish/batfish/blob/34efdf748ec856bfe7bcf25b5d7aeb273105ab65/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java#L178)):*

1 << 20 is 1Mibbion (sorry) entries, right? How does this relate to the comments above? Did you mean to update maximum size above to `2^20`?
</blockquote></details>

Copy-paste.

Issue #8061, opened 2022-02-18
ARISTA: VXLAN multicast group misinterpreted as multicast flood group


rmcmilli:
Hi batfish devs, 

Currently when loading an Arista EVPN VXLAN configuration, `vxlan vrf <name> multicast group <group ip>` is misinterpreted as a flood group configuration, clearing the unicast flood groups. 

A snippet of said configuration: 
```
interface Vxlan1
   vxlan source-interface Loopback1
   vxlan vrf blue multicast-group 239.10.10.10
   vxlan udp-port 4789
   vxlan virtual-router encapsulation mac-address mlag-system-id
```

Resulting `vxlanVniProperties`
```	Node	VRF	VNI	Local_VTEP_IP	Multicast_Group	VLAN	VTEP_Flood_List	VXLAN_Port
0	ny5-tor1-1	default	10104	10.254.2.1	239.10.10.10	104	None	4789
1	ny5-tor5-1	default	10103	10.254.2.5	239.10.10.10	103	None	4789
2	ny5-tor1-2	default	10104	10.254.2.1	239.10.10.10	104	None	4789
3	ny5-tor2-2	default	10104	10.254.2.2	239.10.10.10	104	None	4789
4	ny5-tor1-1	default	10100	10.254.2.1	239.10.10.10	100	None	4789
5	ny5-tor1-1	default	10102	10.254.2.1	239.10.10.10	102	None	4789
6	ny5-tor2-1	default	10100	10.254.2.2	239.10.10.10	100	None	4789
7	ny5-tor2-2	default	10103	10.254.2.2	239.10.10.10	103	None	4789
8	ny5-tor1-1	default	10103	10.254.2.1	239.10.10.10	103	None	4789
9	ny5-tor1-2	default	10100	10.254.2.1	239.10.10.10	100	None	4789
10	ny5-tor1-2	default	10102	10.254.2.1	239.10.10.10	102	None	4789
11	ny5-tor2-2	default	10100	10.254.2.2	239.10.10.10	100	None	4789
12	ny5-tor2-2	default	10102	10.254.2.2	239.10.10.10	102	None	4789
13	ny5-tor2-2	default	10003	10.254.2.2	239.10.10.10	3	None	4789
14	ny5-tor1-1	default	10003	10.254.2.1	239.10.10.10	3	None	4789
15	ny5-tor2-1	default	10003	10.254.2.2	239.10.10.10	3	None	4789
[...]

PR #8085, opened 2022-02-26
Arista EOS do not treat vxlan multicast-groups as VTEPs

arifogel:
- Fixes batfish/batfish#8061
- Multicast-groups are not supported for BUM traffic on EOS up to at least 4.23
  - They are only for BM traffic, and should not displace unicast VTEPs
- Even on 4.25 which apparently allows multicast VTEPs, they are not configured
  via the `multicast-group` command.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8085)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #8090, opened 2022-02-26
AclIpSpace: fewer passes, missing universe optimization

dhalperi:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8090)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @anothermattbrown)

___
*[projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/AclIpSpaceTest.java, line 58 at r1](https://reviewable.io/reviews/batfish/batfish/8090#-Mx5nvI08H5ypsFbCpe1:-Mx5nvI08H5ypsFbCpe2:b-wc7f5p) ([raw file](https://github.com/batfish/batfish/blob/2ab1d1c6f744f6420b81483b05da0d2f5db3f714/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/AclIpSpaceTest.java#L58)):*
> ```Java
>     assertThat(union(EmptyIpSpace.INSTANCE, ipSpace), equalTo(ipSpace));
>     assertThat(union(UniverseIpSpace.INSTANCE, ipSpace), equalTo(UniverseIpSpace.INSTANCE));
>     assertThat(union(ipSpace, UniverseIpSpace.INSTANCE), equalTo(UniverseIpSpace.INSTANCE));
> ```

FYI this is the one that fails on master.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8090)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java, line 256 at r1](https://reviewable.io/reviews/batfish/batfish/8090#-Mx5pfu53DEjZSfsS5lQ:-Mx5pfu53DEjZSfsS5lR:br0eaom) ([raw file](https://github.com/batfish/batfish/blob/2ab1d1c6f744f6420b81483b05da0d2f5db3f714/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java#L256)):*
> ```Java
>     // In one pass, determine if the iterable contains a universe, contains anything, and filter out
>     // null/empty spaces. Have to do this complicated algorithm to properly flatten AclIpSpaces.
>     boolean[] hasUniverse = new boolean[] {false};
> ```

Why this instead of AtomicBoolean (which we usually use for this kind of thing)?



_Code quote:_
```Java
    boolean[] hasUniverse = new boolean[] {false};
    boolean[] hasAnything = new boolean[] {false};
```
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java, line 267 at r1](https://reviewable.io/reviews/batfish/batfish/8090#-Mx5p3HZE7s9yec1xSsZ:-Mx5p3HZE7s9yec1xSs_:b-ze2o8l) ([raw file](https://github.com/batfish/batfish/blob/2ab1d1c6f744f6420b81483b05da0d2f5db3f714/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java#L267)):*
> ```Java
>                   if (s instanceof UniverseIpSpace) {
>                     hasUniverse[0] = true;
>                     return false; // can't fully exit from here, but don't include in array
> ```

java9 has a takeWhile operator we can use to break out early (though we may be unlikely to do that very often in practice) https://docs.oracle.com/javase/9/docs/api/java/util/stream/Stream.html#takeWhile-java.util.function.Predicate-



_Code quote:_
```Java
return false; // can't fully exit from here, but don't include in array
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8090)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java, line 256 at r1](https://reviewable.io/reviews/batfish/batfish/8090#-Mx5pfu53DEjZSfsS5lQ:-Mx5rXEyDBbG5xjG3N2z:bt8qbaq) ([raw file](https://github.com/batfish/batfish/blob/2ab1d1c6f744f6420b81483b05da0d2f5db3f714/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java#L256)):*

Why this instead of AtomicBoolean (which we usually use for this kind of thing)?
</blockquote></details>

AtomicBoolean is better when you have parallelism - threadsafe, with the overhead that introduces.

This is better if you're single threaded.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java, line 267 at r1](https://reviewable.io/reviews/batfish/batfish/8090#-Mx5p3HZE7s9yec1xSsZ:-Mx5s5AS2kQd1jW3z7s7:b-x0pdu9) ([raw file](https://github.com/batfish/batfish/blob/2ab1d1c6f744f6420b81483b05da0d2f5db3f714/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AclIpSpace.java#L267)):*

java9 has a takeWhile operator we can use to break out early (though we may be unlikely to do that very often in practice) https://docs.oracle.com/javase/9/docs/api/java/util/stream/Stream.html#takeWhile-java.util.function.Predicate-
</blockquote></details>

TIL!

Java 1.9+ tho, which Batfish currently still allows.

PR #8108, opened 2022-03-02
TraceTree: make constructor public

anothermattbrown:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8108)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/trace/TraceTree.java, line 26 at r1](https://reviewable.io/reviews/batfish/batfish/8108#-MxBl8gx-xwfRFb0LjbO:-MxBl8gx-xwfRFb0LjbP:b-3bkvrc) ([raw file](https://github.com/batfish/batfish/blob/40e04266c3f8e800e642322d7ce3896cbf875c09/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/trace/TraceTree.java#L26)):*
> ```Java
>   private final @Nonnull List<TraceTree> _children;
> 
>   public TraceTree(TraceElement traceElement, List<TraceTree> children) {
> ```

Missing the context -- there's now an expectation that TraceTrees will be produced by something other than Tracer?

PR #8111, opened 2022-03-03
PAN: extract security rule categories

sfraint:
Extract [`category`](https://docs.paloaltonetworks.com/pan-os/9-0/pan-os-admin/url-filtering/custom-url-categories.html) attached to security rules.

These are not used in the VI model.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8111)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/CustomUrlCategoryReference.java, line 19 at r1](https://reviewable.io/reviews/batfish/batfish/8111#-MxQw2ybCVqIfnLYfc4z:-MxQw2ybCVqIfnLYfc5-:b-4bkkiw) ([raw file](https://github.com/batfish/batfish/blob/9c4cab661733be262165fc53e4ae4258fa7996c4/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CustomUrlCategoryReference.java#L19)):*
> ```Java
> 
>   @Override
>   public int compareTo(CustomUrlCategoryReference o) {
> ```

Where is this used?
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/SecurityRule.java, line 95 at r1](https://reviewable.io/reviews/batfish/batfish/8111#-MxQwNb34_4PIM4mvfri:-MxQwNb34_4PIM4mvfrj:b-92e0cc) ([raw file](https://github.com/batfish/batfish/blob/9c4cab661733be262165fc53e4ae4258fa7996c4/projects/batfish/src/main/java/org/batfish/representation/palo_alto/SecurityRule.java#L95)):*
> ```Java
> 
>   public void addCategory(String category) {
>     _category =
> ```

Usually when I write a function like this, I do nothing if the item is already present. Otherwise,  it will crash if the item is repeated due to the size expectation. Is that guarded against elsewhere? If you do want to prevent this function being called twice for the same item, you should javadoc as much.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8111)* status: 4 of 11 files reviewed, 2 unresolved discussions (waiting on @arifogel and @corinaminer)

___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/CustomUrlCategoryReference.java, line 19 at r1](https://reviewable.io/reviews/batfish/batfish/8111#-MxQw2ybCVqIfnLYfc4z:-Mxei49k1byxpl8Mn_wK:b-u52s2r) ([raw file](https://github.com/batfish/batfish/blob/9c4cab661733be262165fc53e4ae4258fa7996c4/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CustomUrlCategoryReference.java#L19)):*

Where is this used?
</blockquote></details>

not used, removed
___
*[projects/batfish/src/main/java/org/batfish/representation/palo_alto/SecurityRule.java, line 95 at r1](https://reviewable.io/reviews/batfish/batfish/8111#-MxQwNb34_4PIM4mvfri:-MxeiMjqAwsG-ugEqqpE:b-7ql1td) ([raw file](https://github.com/batfish/batfish/blob/9c4cab661733be262165fc53e4ae4258fa7996c4/projects/batfish/src/main/java/org/batfish/representation/palo_alto/SecurityRule.java#L95)):*

Usually when I write a function like this, I do nothing if the item is already present. Otherwise,  it will crash if the item is repeated due to the size expectation. Is that guarded against elsewhere? If you do want to prevent this function being called twice for the same item, you should javadoc as much.
</blockquote></details>

fixed

arifogel (commented):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8111)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #8115, opened 2022-03-05
PrefixTreeMultiMap: sensible public API

dhalperi:
1. Don't expose internal nodes (nodes for branching, but with no data)
2. Don't let users initialize an internal node
3. Clean up all the tests.

I was unable to come up with a reason internal nodes need to be visible to
callers, and it seems confusing -- otherwise, naive users will think that
internal nodes always have data.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/FibImpl.java
@@ -98,7 +98,7 @@ public final class FibImpl implements Fib {
 
   public <R extends AbstractRouteDecorator> FibImpl(
       GenericRib<R> rib, ResolutionRestriction<R> restriction) {
-    _root = new PrefixTrieMultiMap<>(Prefix.ZERO);
+    _root = new PrefixTrieMultiMap<>();
     rib.getTypedRoutes().stream()
         .map(AbstractRouteDecorator::getAbstractRoute)
         .filter(r -> !r.getNonForwarding())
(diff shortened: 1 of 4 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8115)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMultiMap.java, line 334 at r1](https://reviewable.io/reviews/batfish/batfish/8115#-MxQzrCaDgHk3KX0CN9R:-MxQzrCaDgHk3KX0CN9S:b-yhvsr0) ([raw file](https://github.com/batfish/batfish/blob/00c43980195907cccd6b704d1250eca32abd6da1/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMultiMap.java#L334)):*
> ```Java
>     Consumer<Node<T>> nodeConsumer =
>         node -> {
>           if (!node._elements.isEmpty()) {
> ```

I was hesitant about this change, but I suppose it's fine since client tests are passing. Certainly more sensible.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMultiMap.java, line 536 at r1](https://reviewable.io/reviews/batfish/batfish/8115#-MxQyJEw9mXmR9JhX1jf:-MxQy2gz6FQZnjZNpYU4:bl51n5y) ([raw file](https://github.com/batfish/batfish/blob/00c43980195907cccd6b704d1250eca32abd6da1/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/PrefixTrieMultiMap.java#L536)):*
> ```Java
> 
>   /**
>    * Returns {@code true} iff there is any intersection between the prefixes that are keys of this
> ```

nit: this was true before, but the javadoc should say "non-empty intersection".

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8115)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

PR #8117, opened 2022-03-05
Overhaul new L3 adjacencies computation

arifogel:

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java
@@ -10,6 +10,7 @@ import org.batfish.datamodel.collections.NodeInterfacePair;
 @ParametersAreNonnullByDefault
 public interface L3Adjacencies extends Serializable {
   boolean USE_NEW_METHOD = false;
+  boolean USE_NEW_NEW_METHOD = true;
 
   /**
    * Return whether the two interfaces are in the same broadcast domain.
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/BridgeDomainL3Adjacencies.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/BridgeDomainL3Adjacencies.java
@@ -1,0 +1,60 @@
+package org.batfish.common.topology.bridge_domain;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import org.batfish.common.topology.L3Adjacencies;
+import org.batfish.common.topology.Layer1Topologies;
+import org.batfish.common.topology.PointToPointComputer;
+import org.batfish.common.topology.PointToPointInterfaces;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.collections.NodeInterfacePair;
+import org.batfish.datamodel.vxlan.VxlanTopology;
+
+public final class BridgeDomainL3Adjacencies implements L3Adjacencies {
+
+  public static @Nonnull BridgeDomainL3Adjacencies create(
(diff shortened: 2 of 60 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8117)* status: 0 of 60 files reviewed, 2 unresolved discussions (waiting on @arifogel and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8117#-MxeRxG34C-bXc3hajUE:-MxeRxG34C-bXc3hajUF:b-5n6f81) (no related file):*
Still wading in and trying to grok this, but some high-level questions/suggestions:

1) What motivated the representational changes? E.g. reifying state functions, typing changes, etc. Were they needed for the functional changes in this PR (bridge domains)? 
2) Why did you implement this from scratch instead of modifying the existing code? It's harder to see what is changing, that the old tests still pass, etc.

---
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/Search.java, line 26 at r1](https://reviewable.io/reviews/batfish/batfish/8117#-MxeVdFT6N__FraSNzK-:-MxeVdFT6N__FraSNzK0:b-8ed6t9) ([raw file](https://github.com/batfish/batfish/blob/8a08f3272c4a63d056e15d660f7c4ce95dd5a65b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/Search.java#L26)):*
> ```Java
> import org.batfish.common.topology.bridge_domain.node.Node;
> 
> final class Search {
> ```

this class could use some comments



_Code quote:_
```Java
final class Search {
```

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8117)* status: 0 of 60 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8117#-MxeRxG34C-bXc3hajUE:-MxemIy31Z3tQT7bys8E:b-115k5i) (no related file):*
> What motivated the representational changes? E.g. reifying state functions, typing changes, etc. 

We reified the state functions to make it possible to perform a symbolic analysis, optimizations etc. at a later time. The edge typing is there to force the developer to justify state primitives allowed between two types of nodes.

> Were they needed for the functional changes in this PR (bridge domains)?
Breaking apart PhysicalInterface into PhysicalInterface and L2Interface was needed. The reification of edge functions was a bit orthogonal, but it certainly made it easier to understand how state evolves during a search.

> Why did you implement this from scratch instead of modifying the existing code?

This is actually mostly copied from the "old new" l3 adjacencies algorithm at package `org.batfish.common.topology.broadcast`:
- node types are copied and extended
- edges and functions are new
- `L3AdjacencyComputer` is copied almost wholesale

> It's harder to see what is changing, that the old tests still pass, etc.

This was my mistake structuring the PR. I should have broken it into multiple commits, with the first being a wholesale copy+rename of the old package, and then subsequent incremental commits for the new stuff. If you have the branch checked out, you can follow the constructors of the old and new `L3AdjacencyComputer` classes to see how they diverge.

I will note that when the bit is flipped to use new l3 adjacencies _before_ this PR, some OSS `batfish` tests do not pass. However, with the bit flipped on top of the code from this PR, all OSS `batfish` tests pass.

---

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8117)* status: 0 of 60 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/Search.java, line 26 at r1](https://reviewable.io/reviews/batfish/batfish/8117#-MxeVdFT6N__FraSNzK-:-Mxf5VlF9guO4A0oP4xF:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/8a08f3272c4a63d056e15d660f7c4ce95dd5a65b/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/Search.java#L26)):*

this class could use some comments
</blockquote></details>

done

arifogel:
closing to make a better structured PR

PR #8119, opened 2022-03-05
CompletionMetadataUtils: remove performance bottleneck

dhalperi:
TreeRangeSet::removeAll does not optimize for the span() of the set
from which elements are being removed. Simplify the code and reduce
runtime. This basically turns O(n) into O(log n) work, where n can
be very large (# owned IPs in the network).

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java
@@ -94,10 +94,8 @@ public final class CompletionMetadataUtils {
   }
 
   static RangeSet<Ip> unownedSubnetHostIps(Prefix prefix, RangeSet<Ip> ownedIps) {
-    TreeRangeSet<Ip> result = TreeRangeSet.create();
-    result.add(Range.closed(prefix.getFirstHostIp(), prefix.getLastHostIp()));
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8119)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @anothermattbrown)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java, line 98 at r1](https://reviewable.io/reviews/batfish/batfish/8119#-MxP2Q4g7JNpLv2mEV2n:-MxP2Q4g7JNpLv2mEV2o:br92enm) ([raw file](https://github.com/batfish/batfish/blob/e89794c546878edd373c3dc16fd3e51a1f26ec79/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/CompletionMetadataUtils.java#L98)):*
> ```Java
>   static RangeSet<Ip> unownedSubnetHostIps(Prefix prefix, RangeSet<Ip> ownedIps) {
>     Range<Ip> prefixRange = Range.closed(prefix.getFirstHostIp(), prefix.getLastHostIp());
>     return ImmutableRangeSet.copyOf(ownedIps.complement().subRangeSet(prefixRange));
> ```

aside: both `complement` and `subRangeSet` are O(1) view operations

PR #8130, opened 2022-03-09
Overhaul new L3 adjacencies computation

arifogel:


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 0 of 81 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8130#-Mxl4Y343oNHa8RMZkqG:-Mxl4Y343oNHa8RMZkqH:b-uaz9z7) (no related file):*
Base to r1 is a functional copy of the "old new" l3adjacencies algorithm with some movement and renaming of classes to make the real implementation diff - r2 - legible

---

anothermattbrown (commented):
> Reviewed 39 of 39 files at r1, 9 of 78 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 12 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java, line 13 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlQPH23RHZFh5lhzcc:-MxlQPH23RHZFh5lhzcd:bsjy5ea) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java#L13)):*
> ```Java
> public interface L3Adjacencies extends Serializable {
>   boolean USE_NEW_METHOD = false;
>   boolean USE_NEW_NEW_METHOD = true;
> ```

perhaps it's time to switch to a version number scheme?



_Code quote:_
```Java
  boolean USE_NEW_METHOD = false;
  boolean USE_NEW_NEW_METHOD = true;
```
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxlVjKxEM1_I50fchZl:b1f4ye4) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*
> ```Java
> import org.batfish.common.topology.bridge_domain.edge.PhysicalToL2;
> 
> /** Helper interface to allow simplification of {@link PopTagImpl}. */
> ```

I'm not familiar with this pattern. Can you explain what this is doing?



_Code quote:_
```Java
/** Helper interface to allow simplification of {@link PopTagImpl}. */
public interface PopTag extends PhysicalToL2.Function, L2ToBridgeDomain.Function {}
```

anothermattbrown (commented):
> Reviewed 39 of 39 files at r1, 9 of 78 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 12 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

anothermattbrown (commented):
> Reviewed 39 of 39 files at r1, 9 of 78 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 12 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java, line 13 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlQPH23RHZFh5lhzcc:-MxlePnW1YFmzg3Fvpmr:b-3qmicf) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/L3Adjacencies.java#L13)):*

perhaps it's time to switch to a version number scheme?
</blockquote></details>

Switched to enum. I _guess_ it's cleaner.
___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxlecPt-oShtFhKvUce:bcxomln) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*

I'm not familiar with this pattern. Can you explain what this is doing?
</blockquote></details>

If you pop 0 tags for instance, that can be simplified as `Identity`. But `Identity` is not a `PopTagImpl`. So by putting the constraints on who can use PopTag here in a separate interface, `PopTagImpl.of` can return `Identity` since the function returns a `PopTag`.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-Mxln2B05vkAHE_8rw5B:b668uh9) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*

If you pop 0 tags for instance, that can be simplified as `Identity`. But `Identity` is not a `PopTagImpl`. So by putting the constraints on who can use PopTag here in a separate interface, `PopTagImpl.of` can return `Identity` since the function returns a `PopTag`.
</blockquote></details>

To expand:
I there were only a `PopTag` class, then the creator function would have to return a `StateFunction` to support simplification. But a`general StateFunction` does not conform to the constraints of specific functions allowable on particular edge types.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-Mxlnbd655gqGXYf7oFU:bbdaa5k) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*

To expand:
I there were only a `PopTag` class, then the creator function would have to return a `StateFunction` to support simplification. But a`general StateFunction` does not conform to the constraints of specific functions allowable on particular edge types.
</blockquote></details>

Final word; I tried rigging this all up with generics, but it's impossible in Java. due to type erasure. Similar to the reason why I couldn't have a single `Compose` class while still constraining the functions allowed on each edge.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxlpabD-X5Wty7exELp:b-xpivam) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*
> PopTagImpl.of can return Identity since the function returns a PopTag

Is there value in having it returning a `PopTag` instead of a `StateFunction` (this is what we do for the related problem in Transitions.java for BDD)? These empty interfaces aren't providing any useful functionality to the client code. It just seems to be a sanity check that `PopTag.of` only returns functions that have been blessed, as far as I can tell.

> I tried rigging this all up with generics

The previous version did this with edges carrying data of different types. I'm comfortable with the new universally-typed approach, but what changed that made the heterogeneously-typed approach no longer work?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxlwHvt8p9iAwk5Gyru:b-kjpjyz) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*
> Is there value in having it returning a PopTag instead of a StateFunction

>  It just seems to be a sanity check that `PopTag.of` only returns functions that have been blessed, as far as I can tell.

That is the value and the purpose.

> The previous version did this with edges carrying data of different types. I'm comfortable with the new universally-typed approach, but what changed that made the heterogeneously-typed approach no longer work?

Both approaches work. The main improvement here is reifying the functions so they can be visited. This was just the approach that made sense to me. And IMO it's more legible without introducing any runtime overhead.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @arifogel, and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxmYXQd5jEGcqiXQSYT:b-97d8il) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*
> It just seems to be a sanity check that PopTag.of only returns functions that have been blessed, as far as I can tell.
> 
> That is the value and the purpose.

I think it would be simpler and easier to understand without this -- clear comments and tests should be sufficient.  But this should at least be clearly commented, with javadoc links to the method it's governing.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8130)* status: 9 of 81 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java, line 6 at r2](https://reviewable.io/reviews/batfish/batfish/8130#-MxlVjKxEM1_I50fchZk:-MxpZTll3eD05ZnXbWsH:b8w1zar) ([raw file](https://github.com/batfish/batfish/blob/9553e52eb290b253b57edf394db1741a756d9494/projects/batfish-common-protocol/src/main/java/org/batfish/common/topology/bridge_domain/function/PopTag.java#L6)):*

> It just seems to be a sanity check that PopTag.of only returns functions that have been blessed, as far as I can tell.
> 
> That is the value and the purpose.

I think it would be simpler and easier to understand without this -- clear comments and tests should be sufficient.  But this should at least be clearly commented, with javadoc links to the method it's governing. 
</blockquote></details>

I reorganized the files in a fashion that I believe makes such comments unnecessary. The impl classes are now inner classes of the respective interfaces.

dhalperi:
Seems this PR is premature -- the core redesign that needs to be reviewed is not included. We'll sync back after.

PR #8136, opened 2022-03-11
Unify and standarize enter/exit/l3 interface configuration

dhalperi:
It is common in Batfish to want to distinguish whether an interface

* is active and L3
* can originate IP (v4) traffic
* can receive IP (v4) traffic
* can send IP (v4) traffic

However, the various logic around this is scattered and inconsistent. For example, traceroute would not let flows
enter L2 interfaces, but reachability and searchFilters would if they were active.

1. Create `Interface` helpers `isActiveL3`, `canOriginateIpTraffic`, `canReceiveIpTraffic`, `canSendIpTraffic`, with
    our current rules.
2. Reify Enter/SourceInterfaces with only interfaces that can receive traffic
3. Reify ExitInterfaces with only interfaces that can send traffic
4. Ensure that only interfaces that `canOriginateIpTraffic` are used as origination points.

and fix up some tests.


anothermattbrown (commented):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8136)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

Issue #8134, opened 2022-03-11
ASA crash when static route next hop is 0.0.0.0


jhammond-git:
```
Type                                                                      Parse error
Details           java.lang.IllegalArgumentException: NextHopIp must be a valid concrete IP address. Received 0.0.0.0\n\tat com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)\n\tat org.batfish.datamodel.route.nh.NextHopInterface.<init>(NextHopInterface.java:126)\n\tat org.batfish.datamodel.route.nh.NextHopInterface.of(NextHopInterface.java:69)\n\tat org.batfish.grammar.cisco_asa.AsaControlPlaneExtractor.exitS_route(AsaControlPlaneExtractor.java:7985)\n\tat org.batfish.grammar.cisco_asa...
Line_Text                                                                 route UNPROTECTED <IP> 255.255.255.248 0.0.0.0 1\n
Parser_Context                                                            [s_route stanza asa_configuration]
```
This apparently stopped the parsing engine as when I removed the offending lines, my issue list went from 308 to 600.

dhalperi:
What does this mean? 0.0.0.0 is not a valid IP.

Is it shorthand for discard or something?

jhammond-git:
That's a good question.  I presumed it meant "default route".  But apparently, in this context it can mean "connected route", which I suppose can translate to "null route" if the IP or network isn't actually connected.

jhammond-git:
I confirmed with another engineer here that it's a way to get multiple networks as connected per interface on those devices.

PR #8137, opened 2022-03-11
ASA: handle next-hop interface routes

dhalperi:
Fix batfish/batfish#8134


dhalperi:
Trying to kick github.

PR #8144, opened 2022-03-15
Routes: return warnings in table answer

anothermattbrown:
When the input node/vrf matches nothing, return an empty table with a warning instead of the warning alone.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8144)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #7741, opened 2021-12-01
Add Warnings.Settings class

ratulm:
Helps communicate the configuration of new Warnings objects  without having to pass three booleans.

PR #8145, opened 2022-03-15
Warnings: remove transient from Settings, fix possible NPE

dhalperi:
This was introduced in batfish/batfish#7741 to replace 3 transient booleans.

The problem: transient boolean almost certainly deserializes as false,
while transient class deserializes as null. So before we were not crashing
because values were just false; now they are null and crashing is happening.

Easiest fix seems to be to simply keep the settings around.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8145)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8159, opened 2022-03-17
ColumnFilter: allow exact matching

dhalperi:
The default behavor of substring matching is preserved.

Some use cases, however, call for exact matching. Enable this, and additionally
be graceful in whether the user is supposed to know exactly how JSON strings are
stringified.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java
@@ -6,6 +6,7 @@ import static java.util.Objects.requireNonNull;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.re2j.Pattern;
 import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8159)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @SLarkworthy)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java, line 78 at r1](https://reviewable.io/reviews/batfish/batfish/8159#-MyOVRAvCzPUAEMTVYao:-MyOVRAvCzPUAEMTVYap:b-f8mstw) ([raw file](https://github.com/batfish/batfish/blob/6c95d38db3411d6f1345b093ee960322d108ed7d/projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java#L78)):*
> ```Java
>       // either.
>       return rowText.equalsIgnoreCase(_filterText)
>           || rowText.equalsIgnoreCase('"' + _filterText + '"');
> ```

since this `matches` method can be called multiple (potentially many) times, it may make sense to cache this as a field



_Code quote:_
```Java
'"' + _filterText + '"'
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8159)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @SLarkworthy)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java, line 78 at r1](https://reviewable.io/reviews/batfish/batfish/8159#-MyOVRAvCzPUAEMTVYao:-MyOhC187LqpQWFpIaBH:b1lxbo7) ([raw file](https://github.com/batfish/batfish/blob/6c95d38db3411d6f1345b093ee960322d108ed7d/projects/batfish-common-protocol/src/main/java/org/batfish/common/ColumnFilter.java#L78)):*

since this `matches` method can be called multiple (potentially many) times, it may make sense to cache this as a field
</blockquote></details>

Switched to regexes (compiled once) which just seems cleaner and likely faster.

used re2j, the faster and not-dos-able regex library.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8159)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SLarkworthy)

PR #8170, opened 2022-03-21
Cumulus NCLU disable invalid interfaces that are bond slaves with L3

arifogel:
- a bond slave (aggregate member) cannot have L3 settings


arifogel:
wrong approach

PR #8171, opened 2022-03-21
Juniper disable invalid and parent-down interfaces during conversion

arifogel:
- disable irb interfaces that have no vlan set
- disable units whose parent is disabled


arifogel:
wrong approach

PR #8191, opened 2022-03-24
JFactory: intern the zero and one BDDImpl objects

anothermattbrown:
We create many many copies of these, so it makes sense to intern them.

### DIFF
--- a/projects/bdd/src/main/java/net/sf/javabdd/JFactory.java
+++ b/projects/bdd/src/main/java/net/sf/javabdd/JFactory.java
@@ -450,6 +450,10 @@ public class JFactory extends BDDFactory {
 
     @Override
     public void free() {
+      if(_index == BDDZERO || _index == BDDONE) {
+        // zero, one BDDs are interned.
+        return;
+      }
       bdd_delref(_index);
       _index = INVALID_BDD;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8191)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

___
*[projects/bdd/src/main/java/net/sf/javabdd/JFactory.java, line 119 at r1](https://reviewable.io/reviews/batfish/batfish/8191#-MyyIV3Q6qcTkOKGf9qi:-MyyIV3Q6qcTkOKGf9qj:b-feaeyp) ([raw file](https://github.com/batfish/batfish/blob/4b50f896ba95d14665e98c494baca522b8f4465e/projects/bdd/src/main/java/net/sf/javabdd/JFactory.java#L119)):*
> ```Java
>     if(id == BDDONE){
>       return _one;
>     }
> ```

also make `free` a no-op?

anothermattbrown:
nevermind, the `*With` methods mutate the objects. e.g. `zero.orWith(one)` would change zero to one.

PR #8209, opened 2022-03-31
IpSpaceToBDD: make it safe for callers to free result

dhalperi:


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8209)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8209#-MzWeAg4BguWhR44zZUz:-MzWgWHw2G78qO4mfzFx:b-p6g193) (no related file):*

FYI worth discussing if this is worth doing. Needed to really make use of it:

* same change for other toBDDs
* ability to clear cache / free BDDs when no longer in use
* ability to free BDDs that are evicted from cache
* actually push freeing down into users

That said, it will likely make it easier to free transient garbage in graph construction if the callers always own the returned BDDs.
</blockquote></details>

Independently of being able to reclaim BDDs, it seems worth doing now that we use mutating BDD methods more to ensure the cached BDD doesn't get mutated by client code.

---

PR #8218, opened 2022-04-05
Verify interface references in OSPF areas

arifogel:
- remove references to undefined interfaces in OSPF areas during `finalizeConfiguration`

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfArea.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ospf/OspfArea.java
@@ -210,6 +210,19 @@ public class OspfArea implements Serializable {
   @Nonnull private final SortedMap<Prefix, OspfAreaSummary> _summaries;
   @Nullable private final String _summaryFilter;
 
+  public @Nonnull Builder toBuilder() {
+    return builder()
+        .setNumber(_areaNumber)
+        .setInjectDefaultRoute(_injectDefaultRoute)
+        .setInterfaces(_interfaces)
+        .setMetricOfDefaultRoute(_metricOfDefaultRoute)
+        .setNssa(_nssa)
+        .setStub(_stub)
+        .setStubType(_stubType)
+        .setSummaries(_summaries)
+        .setSummaryFilter(_summaryFilter);
+  }
+
   public OspfArea(
       long areaNumber,
       boolean injectDefaultRoute,
--- a/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
+++ b/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
@@ -81,6 +81,8 @@ import org.batfish.datamodel.answers.ConvertConfigurationAnswerElement;
 import org.batfish.datamodel.bgp.community.CommunityStructuresVerifier;
 import org.batfish.datamodel.collections.NodeInterfacePair;
 import org.batfish.datamodel.hsrp.HsrpGroup;
+import org.batfish.datamodel.ospf.OspfArea;
+import org.batfish.datamodel.ospf.OspfProcess;
 import org.batfish.datamodel.packet_policy.PacketPolicy;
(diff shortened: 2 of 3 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8218)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 516 at r1](https://reviewable.io/reviews/batfish/batfish/8218#-Mzvvs1r2W9wGHQPthhb:-Mzvvs1r2W9wGHQPthhc:bfrp3zw) ([raw file](https://github.com/batfish/batfish/blob/809229d174c9b9a23cab725a6261c6f30bcb5ed5/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L516)):*
> ```Java
>                             .filter(name -> c.getAllInterfaces().containsKey(name))
>                             .collect(ImmutableSortedSet.toImmutableSortedSet(naturalOrder())))
>                     .build();
> ```

just to check -- we want to keep the `OspfArea` even if all its interfaces turn out to be undefined?
___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 522 at r1](https://reviewable.io/reviews/batfish/batfish/8218#-MzvvSof21o3AYAYHlkl:-MzvvSof21o3AYAYHlkm:by6nbqx) ([raw file](https://github.com/batfish/batfish/blob/809229d174c9b9a23cab725a6261c6f30bcb5ed5/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L522)):*
> ```Java
>                 .forEach(a -> areasBuilder.put(a.getAreaNumber(), a));
>             areasBuilder.put(newArea.getAreaNumber(), newArea);
>             proc.setAreas(areasBuilder.build());
> ```

why not have just one `areasBuilder` per process, created before looping through areas, and then call `setAreas` after it's fully populated? Then you also wouldn't need to copy `proc.getAreas()` for the areas loop.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8218)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @sfraint)

___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 516 at r1](https://reviewable.io/reviews/batfish/batfish/8218#-Mzvvs1r2W9wGHQPthhb:-Mzw-874EcApjKpd7pP4:b2kr9h8) ([raw file](https://github.com/batfish/batfish/blob/809229d174c9b9a23cab725a6261c6f30bcb5ed5/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L516)):*

just to check -- we want to keep the `OspfArea` even if all its interfaces turn out to be undefined?
</blockquote></details>

Not sure it matters, but that's a bit out of scope for my intention. The main goal here is preventing references to undefined interfaces.
The forthcoming new l3 adjacencies PR will remove interfaces for more reasons, so I need this PR to prevent crashes in certain places while iterating. Since this portion is separable, I split it out into this separate PR.
___
*[projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java, line 522 at r1](https://reviewable.io/reviews/batfish/batfish/8218#-MzvvSof21o3AYAYHlkl:-Mzw-NgI7NzZ4d_sk3u2:b-8icl9e) ([raw file](https://github.com/batfish/batfish/blob/809229d174c9b9a23cab725a6261c6f30bcb5ed5/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java#L522)):*

why not have just one `areasBuilder` per process, created before looping through areas, and then call `setAreas` after it's fully populated? Then you also wouldn't need to copy `proc.getAreas()` for the areas loop.
</blockquote></details>

I'm trying to minimize work in the typical case of no undefined references.
I wouldn't worry too much about optimizing here, since we really are supposed to prevent this code from modifying configs by enforcing invariants during conversion.

PR #8222, opened 2022-04-06
BDDRoute: admin dist is only 8 bits (unsigned)

dhalperi:
Max value is 255

commit-id:63458a00

---

**Stack**:
- #8224
- #8223
- #8222 ⬅

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDIcmpCode.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDIcmpCode.java
@@ -28,7 +28,7 @@ public final class BDDIcmpCode {
    * @param satAssignment a satisfying assignment (i.e. produced by fullSat, allSat, etc)
    */
   public int satAssignmentToValue(BDD satAssignment) {
-    return _var.satAssignmentToLong(satAssignment).intValue();
+    return _var.satAssignmentToInt(satAssignment);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8222)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDPacketLength.java, line 64 at r1](https://reviewable.io/reviews/batfish/batfish/8222#-N--0uMy8K89AbDTooFx:-N--0uMy8K89AbDTooFy:b-3yk6ao) ([raw file](https://github.com/batfish/batfish/blob/e2ec49e928e7e9ae67edf8bef38b2cf58627b9d9/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDPacketLength.java#L64)):*
> ```Java
>   public int satAssignmentToValue(BitSet satAssignment) {
>     // the sat assignment will be the minimum value of the original BDD. If the BDD did not
>     // constrain packet length, the sat assignment will have value 0, which is below the minimum
> ```

Should we revisit this? It's unsound if the BDD explicitly constrained the packet length to be <20. Admittedly unlikely, but we do already add constraints on packet length in `FlowPreference`



_Code quote:_
```Java
    // the sat assignment will be the minimum value of the original BDD. If the BDD did not
    // constrain packet length, the sat assignment will have value 0, which is below the minimum
    // value of 20. Rather than maintaining this minimum in every BDD (expensive), we do it here.
    // Note that we can get weird behavior is clients constrain the underlying BDDs bits directly
    // (i.e. without going through this class).
    return Integer.max(_var.satAssignmentToInt(satAssignment), 20);
```

PR #8228, opened 2022-04-07
SearchRoutePolicies: Better handling of unsupported features

millstein:
Previously the symbolic route analysis (TransferBDD) threw an exception and failed when encountering a route-policy feature that it does not support.  Now it logs such features but continues the analysis, and later when SearchRoutePolicies obtains results it makes sure to only produce answers that don't rely on unsupported features.  This is accomplished by adding an extra BDD to the BDDRoute so that the symbolic route analysis can track the conditions under which such features are encountered.


anothermattbrown (commented):
> Reviewed 11 of 12 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: all files reviewed, 4 unresolved discussions (waiting on @millstein)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java, line 491 at r3](https://reviewable.io/reviews/batfish/batfish/8228#-N-ATQ7yCzVlzp5plgPX:-N-ATQ7yCzVlzp5plgPY:bnnhlai) ([raw file](https://github.com/batfish/batfish/blob/efa6e1d4ed763b85749b09b4e64dadad4e88328e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L491)):*
> ```Java
>                   ? Arrays.hashCode(_asPathRegexAtomicPredicates)
>                   : 0);
>       result = 31 * result + (_unsupported != null ? _unsupported.hashCode() : 0);
> ```

it seems like the only way `_unspported` can become `null` is if client code calls `setUnsupported(null)`? As far as I can tell your code never does that -- is there a valid use case for it? If not, we should just disallow it.



_Code quote:_
```Java
_unsupported != null ? _unsupported.hashCode() : 0
```
___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 1076 at r3](https://reviewable.io/reviews/batfish/batfish/8228#-N-AV2NN4rlafDP7AjMH:-N-AV2NN4rlafDP7AjMI:bumzlvp) ([raw file](https://github.com/batfish/batfish/blob/efa6e1d4ed763b85749b09b4e64dadad4e88328e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1076)):*
> ```Java
>     TransferResult result = state.getTransferResult();
>     BDD currUnsupported = curP.getData().getUnsupported();
>     BDD newValue = ite(unreachable(result), currUnsupported, _factory.one());
> ```

I found this somewhat confusing. May be simpler to write this as:
```
BDD reachUnsupportedStatement = unreachable(result).not();
BDD newValue = currUnsupported.or(reachUnsupportedStatement);
```

Adding a comment to that effect would help too :)



_Code quote:_
```Java
ite(unreachable(result), currUnsupported, _factory.one());
```
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1270 at r3](https://reviewable.io/reviews/batfish/batfish/8228#-N-AZKJ13grLowCVORQ3:-N-AZKJ13grLowCVORQ4:b-2dxkyj) ([raw file](https://github.com/batfish/batfish/blob/efa6e1d4ed763b85749b09b4e64dadad4e88328e/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1270)):*
> ```Java
>   @Test
>   public void testUnsupportedMatchNextHop() {
>     _policyBuilder.addStatement(
> ```

Could you add an ExitAccept statement after the If? That would make it clear that we won't return results from either branch since we don't which branch any route will take.



_Code quote:_
```Java
    _policyBuilder.addStatement(
        new If(
            new MatchPrefixSet(
                new IpPrefix(NextHopIp.instance(), new LiteralInt(8)),
                new ExplicitPrefixSet(
                    new PrefixSpace(
                        ImmutableList.of(PrefixRange.fromPrefix(Prefix.parse("1.0.0.0/8")))))),
            ImmutableList.of(new StaticStatement(Statements.ExitAccept))));
```
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 2215 at r3](https://reviewable.io/reviews/batfish/batfish/8228#-N-AXOcsAMvF_poytJUY:-N-AXOcsAMvF_poytJUZ:bg1afu6) ([raw file](https://github.com/batfish/batfish/blob/efa6e1d4ed763b85749b09b4e64dadad4e88328e/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L2215)):*
> ```Java
> 
>     BDDRoute outRoute = anyRoute(tbdd.getFactory());
>     outRoute.setUnsupported(tbdd.getFactory().one());
> ```

it seems like you're only testing for `unsupported` equal to zero or one. Can you add a more interesting case too?



_Code quote:_
```Java
outRoute.setUnsupported(tbdd.getFactory().one());
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: 10 of 13 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java, line 1076 at r3](https://reviewable.io/reviews/batfish/batfish/8228#-N-AV2NN4rlafDP7AjMH:-N-BVkva6Igb6IFv9Ji8:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/efa6e1d4ed763b85749b09b4e64dadad4e88328e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1076)):*

I found this somewhat confusing. May be simpler to write this as:
```
BDD reachUnsupportedStatement = unreachable(result).not();
BDD newValue = currUnsupported.or(reachUnsupportedStatement);
```

Adding a comment to that effect would help too :)

</blockquote></details>

Done.

anothermattbrown (commented):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1290 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q7bcB90fqEUQCc0mV:-N-Q7bcB90fqEUQCc0mW:bhs0zc5) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1290)):*
> ```Java
>     BDDRoute outRoute = anyRoute(tbdd.getFactory());
>     outRoute.setUnsupported(tbdd.getFactory().one());
>     assertEquals(acceptedAnnouncements, tbdd.getFactory().one());
> ```

Hmm I did not expect this to change. But maybe this is not the right place for the test I had in mind, since we don't exclude unsupported flow from acceptedAnnouncements. Maybe a similar test in the answerer (where the exclusion is done) would be better.



_Code quote:_
```Java
    assertEquals(acceptedAnnouncements, tbdd.getFactory().one());
```
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java, line 1480 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q8gSZA7yGSraxCJhP:-N-Q8gSZA7yGSraxCJhQ:bki8k0y) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L1480)):*
> ```Java
> 
>   @Test
>   public void testUnsupportedStatementAvoid() {
> ```

I think what I'm looking for is a version of this where the unsupported statement is an `If`, so we don't return results from either branch.



_Code quote:_
```Java
testUnsupportedStatementAvoid(
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java, line 1290 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q7bcB90fqEUQCc0mV:-N-Q9bBO3LtyPKpa9JPP:b-pw6sn5) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1290)):*

Hmm I did not expect this to change. But maybe this is not the right place for the test I had in mind, since we don't exclude unsupported flow from acceptedAnnouncements. Maybe a similar test in the answerer (where the exclusion is done) would be better.
</blockquote></details>

Yes, this changes for the reason you said (it also surprised me at first!).  But the important point is that the unsupported BDD is still ONE, even with this change, which shows that we indeed don't return results from either branch.  That's in contrast to what happens with the new test that I added based on your other comment, testPartialUnsupportedStatement, where only one branch of the if encounters an unsupported statement.
___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java, line 1480 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q8gSZA7yGSraxCJhP:-N-QAokJ4VJ8_NWTqslf:b20618v) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L1480)):*

I think what I'm looking for is a version of this where the unsupported statement is an `If`, so we don't return results from either branch.
</blockquote></details>

I believe that's exactly what's happening -- that logic is encoded in the unsupported BDD of the output BDDRoute.  (I had originally thought about a scheme that updates the acceptedAnnouncements BDD, but that doesn't work, or at least I couldn't figure out how to make it work.)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java, line 1480 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q8gSZA7yGSraxCJhP:-N-ThMITASpdzmA-SnEf:b-27vws0) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L1480)):*

I believe that's exactly what's happening -- that logic is encoded in the unsupported BDD of the output BDDRoute.  (I had originally thought about a scheme that updates the acceptedAnnouncements BDD, but that doesn't work, or at least I couldn't figure out how to make it work.)
</blockquote></details>

Reading the comment in the input route below, I thought the unsupported statement here was the `PrependAsPath`. The output route is being returned by the top-level `ExitAccept` statement, i.e. the else branch of the `If`, right?

What I had in mind was a test where the unsupported statement is the `If` itself, not a statement within its then branch, e.g. the one you use in the TransferBDD test we've been discussing:

```
            new If(
                new MatchPrefixSet(
                    new IpPrefix(NextHopIp.instance(), new LiteralInt(8)),
                    new ExplicitPrefixSet(
                        new PrefixSpace(
                            ImmutableList.of(PrefixRange.fromPrefix(Prefix.parse("1.0.0.0/8")))))),
```

In this case, there would be no way to avoid the unsupported `If` statement, so we would return no rows, right?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java, line 1480 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q8gSZA7yGSraxCJhP:-N-UFIEaEGwfdEvydkXX:b29mwv4) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L1480)):*

Reading the comment in the input route below, I thought the unsupported statement here was the `PrependAsPath`. The output route is being returned by the top-level `ExitAccept` statement, i.e. the else branch of the `If`, right?

What I had in mind was a test where the unsupported statement is the `If` itself, not a statement within its then branch, e.g. the one you use in the TransferBDD test we've been discussing:

```
            new If(
                new MatchPrefixSet(
                    new IpPrefix(NextHopIp.instance(), new LiteralInt(8)),
                    new ExplicitPrefixSet(
                        new PrefixSpace(
                            ImmutableList.of(PrefixRange.fromPrefix(Prefix.parse("1.0.0.0/8")))))),
```

In this case, there would be no way to avoid the unsupported `If` statement, so we would return no rows, right?
</blockquote></details>

Oh sorry, I misunderstood which test you were referring to here.  I've added another test that I believe addresses your comment.

anothermattbrown (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8228)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

___
*[projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java, line 1480 at r4](https://reviewable.io/reviews/batfish/batfish/8228#-N-Q8gSZA7yGSraxCJhP:-N-UUO_y96PumJfWYwFj:bmyr7pm) ([raw file](https://github.com/batfish/batfish/blob/a84623cdb86ed58935adbd83884c937877604fd1/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L1480)):*

Oh sorry, I misunderstood which test you were referring to here.  I've added another test that I believe addresses your comment.
</blockquote></details>

yep, that's what I was looking for. thanks!

PR #8246, opened 2022-04-14
New bridge-domain-based L3 adjacencies model

arifogel:

Other notable features:
- disable each interface with subinterfaces that allow overlapping input tags
- add modeling support for individual trunk interfaces attached to multiple bridge domains
- Add InterfaceTopology to each interface encompassing tag-handling and bridging configuration
  - Must be populated prior by end of conversion
  - Add legacy helpers to populate interface topology for existing vendors
  - Replaces switchport/encapuslation settings as source of truth for behavior
    - Existing settings preserved as informational


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8246)* status: 0 of 131 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @corinaminer, @dhalperi, and @sfraint)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8246#-N-_g_FTBzTM3Z2284f_:-N-_g_FTBzTM3Z2284fa:bvbdbtc) (no related file):*
self-merge only

---

Issue #8266, opened 2022-04-25
Route-maps incorrectly flagged as undefined in Cisco IOSXE > router bgp


network-dave:
Hello,

I did found at least 1 occurence where a route-map was incorrectly flagged as unused by q.undefinedReferences() ;

```
router bgp XXXXX
  redistribute static route-map STATIC-TO-BGP

route-map STATIC-TO-BGP permit 10
 set local-preference 200

```

Kind regards,

David

dhalperi:
Hi @network-dave - Having a hard time reproducing this with the snippet you gave. Can you share more details?

See the attached screenshot:

<img width="1183" alt="Screen Shot 2022-04-25 at 3 58 29 PM" src="https://user-images.githubusercontent.com/526415/165188038-26c11d4e-e51e-45f4-bf14-5bea78fba7e5.png">

dhalperi:
Notebook attached:
[gh-8266.ipynb.zip](https://github.com/batfish/batfish/files/8558523/gh-8266.ipynb.zip)

network-dave:
Hi Dan @dhalperi, 

I confirm the behaviour when loading the full config in a Jupyter notebook :

![image](https://user-images.githubusercontent.com/60117841/165279935-511f0ce1-e07f-4e8b-b336-d2ae03d4dd68.png)


What is the best way to send you the fully anonymized configuration ?


Kr,



David

dhalperi:
Hi @network-dave if you want to share data, the two easiest ways are probably to Slack me on the Batfish Slack channel or email it to dan@intentionet.com. We often use `netconan` for this: https://pypi.org/project/netconan/ and you can add whatever is behind `xxxx` to the `--sensitive-words` flag.

dhalperi:
In this case, Batfish incorrectly thought that `set extcommunity rt <16-bit>:<32-bit> additive` was incorrect, which caused the next line (defining the route-map in question) to be lost. Fixed by batfish/batfish#8271.

PR #8271, opened 2022-04-26
IOS: fix parsing of rt extended communities

dhalperi:
Both GA and LA can be 16-bit or 32-bit, as long as they aren't both 32-bit.
See ExtendedCommunity#target for how this is signaled on the wire.

Fix batfish/batfish#8266.

commit-id:4d8b2632

**Issue references**: 

 - Fix batfish/batfish#8266


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8271)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

PR #8283, opened 2022-04-29
Junos:  Vxlan and Evpn VendorConfig

jeffkala:
Another set of updates.  Additional PRs will be needed to continue working through the remaining logic.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 12 unresolved discussions (waiting on @jeffkala)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8283#-N15hYOd0oljcznrozqQ:-N15hYOd0oljcznrozqR:bu1kn3m) (no related file):*
Please add test and testconfig for extended vni list with literal values to cover your implementation of that.

---
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4207 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15fJ025OMFVM4NhW5T:-N15fJ034Qk6kRr2Ab56:b5wlkxf) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4207)):*
> ```Java
> 
>   @Override
>   public void exitE_encapsulation(E_encapsulationContext ctx) {
> ```

The definition of e_encapsulation is:
```
e_encapsulation
:
    ENCAPSULATION VXLAN
;
```
`ctx.ENCAPSUALTION() != null` will always be `true`, so there is no need to check that.

Did you intend to check whether `ctx.VXLAN() != null` instead?
Is there another value other than `vxlan` valid here?
If not, then I would make change `String _encapsulation` in `Evpn` to `boolean _encapsuationVlan`, and just set it to `true` here as the sole line in this function.
Otherwise, you should make an enum of the acceptable values, and make encapsulation be of that enum type rather than `String`.

Either way, no need to extract `ctx.VXLAN().getText()`.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4216 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15gOUK39byEdBhqdkB:-N15gOUL5XSioJlQlTtz:b-46nr4o) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4216)):*
> ```Java
>   public void exitE_extended_vni_list(E_extended_vni_listContext ctx) {
>     if (ctx.ALL() != null) {
>       _currentLogicalSystem.getEvpn().setExtendedVniAll(ctx.ALL().getText());
> ```

Either `Evpn._extendedVniAll` should be a boolean (like above comment about encapsulation), or there should be a unified polymorphic `_extendedVni` property of say type `ExtendedVniList`, where the subtypes are something like:
- `LiteralExtendedVniList`
- `ExtendedVniAll` (singleton class)
The benefit of the latter pattern is it won't allow both `all` and a literal `list` simultaneously, and there will only be one property to check during conversion.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4221 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15hu2s5jW4EMpx3A58:-N15hu2s5jW4EMpx3A59:bk3s4za) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4221)):*
> ```Java
>       for (Vni_rangeContext vni : ctx.vni_range()) {
>         if (vni.getText().contains("-")) {
>           continue;
> ```

Looks like you're not handling this. You should add an explicit warning, or at least `todo(ctx)`.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4223 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15iN5T2dP1ez5UTBRT:-N15iN5T2dP1ez5UTBRU:b-5ema8z) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4223)):*
> ```Java
>           continue;
>         } else {
>           vnis.add(Integer.parseInt(vni.getText()));
> ```

You should make use this function instead of manually parsing:
```
  private @Nonnull Optional<Integer> toInteger(
      ParserRuleContext messageCtx, Vni_numberContext ctx) {
    return toIntegerInSpace(messageCtx, ctx, VNI_NUMBER_RANGE, "vni");
  }

```
Then if the returned `Optional` is empty, you should `continue` (a warning will already have been generated by `toInteger`.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4233 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15hhh_8v95K23EOLGI:-N15hhh_8v95K23EOLGJ:b-a9zlv0) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4233)):*
> ```Java
>   public void exitE_multicast_mode(E_multicast_modeContext ctx) {
>     if (ctx.INGRESS_REPLICATION() != null) {
>       _currentLogicalSystem.getEvpn().setMulticastMode(ctx.INGRESS_REPLICATION().getText());
> ```

`Evpn._multicastMode` should be an enum with members `INGRESS_REPLICATION` and `CLIENT` rather than a string.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15irlsAOgHrXqZMKlP:-N15irlsAOgHrXqZMKlQ:b-cpcvbc) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L29)):*
> ```Java
>   }
> 
>   public @Nullable List<Integer> getExtendedVniList() {
> ```

To support ranges without having to enumerate every VNI, better to use an `IntegerSpace` than `List<Integer>`. Note that `IntegerSpace` is immutable, so you have to build a new one including the contents of the old one when updating.
So you'll need a separate function in this class for updating, since you can't modify a returned `IntegerSpace`.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2031 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15l3Cr8QWTzJ1rDtFa:-N15l3Cr8QWTzJ1rDtFb:b-xarb4e) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2031)):*
> ```Java
>               Layer2Vni.builder()
>                   .setVni(vxlan.getVniId())
>                   .setVlan(vxlan.getVlanId())
> ```

Is it possible for getvlanId() to return `null`? If so, should handle that and exit early to avoid NPE.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2038 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15kHxH583B_qubMYjQ:-N15kHxH583B_qubMYjR:b-34z7ka) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2038)):*
> ```Java
>           _c.getDefaultVrf().addLayer2Vni(vniSettings);
>         } else {
>           continue;
> ```

This else continue branch is redundant.
But as a general style comment, we usually prefer to exit early.
So you could do instead above:
```
if (lvlxan.getVniId() == null) {
  continue;
}
// code for when it's not null, without extra indentation.
```

Similarly in the l3 case you can exit early if `vtepSource == null`, and avoid the extra indentation that follows.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2051 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15kp8wDU6lOFALpT0Z:-N15kp8wDU6lOFALpT0_:b-lc467b) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2051)):*
> ```Java
>                           .getInterfaces()
>                           .get(l3Interface)
>                           .getPrimaryAddress()
> ```

If either
- `get(l3Interface)` returns `null` because interface does not exist
- `getPrimaryAddress()` returns `null` because none was assigned
you will get an NPE (or builder exception).

You should check each of these first before building an l3vni, exiting early if necessary to avoid the NPE.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N15jfYZBj9JjY-705nv:byyxaqw) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
> ```Java
>                           .getIp())
>                   .setUdpPort(Vni.DEFAULT_UDP_PORT)
>                   .setSrcVrf(_masterLogicalSystem.getDefaultRoutingInstance().getName())
> ```

In general, this should be the vrf of the interface, not necessarily the default vrf.
For Juniper, I believe that means you should use the name of the routing instance that owns the interface.
Note that in contrast, converted l2vnis always must be in the default vrf (Batfish quirk, vrfs don't really mean anything for l2vnis anyway).
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java` line 124 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15lsePBp05mvZwTTLl:-N15lsePBp05mvZwTTLm:bymt7fh) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java#L124)):*
> ```Java
>     _dnsServers = new TreeSet<>();
>     _dscpAliases = new TreeMap<>();
>     _evpn = new Evpn();
> ```

I think it would be better to leave this null in constructor, and instead instantiate this with a setter method from `ConfigurationBuilder` via:

```
@Override
public void enterP_evpn(P_evpnContext ctx) {
  if (_currentLogicalSystem.getEvpn() == null) {
    _currentLogicalSystem.setEvpn(new Evpn()); // <-- instantiate here
  }
}
```
(create the method it if it doesn't exist)

Otherwise every logical system will have an `Evpn` object, which creates more overhead.

This also means you'll need to handle it being null in `JuniperConfiguration`.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4207 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15fJ025OMFVM4NhW5T:-N1AfND9DeexqS_AY64W:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4207)):*

The definition of e_encapsulation is:
```
e_encapsulation
:
    ENCAPSULATION VXLAN
;
```
`ctx.ENCAPSUALTION() != null` will always be `true`, so there is no need to check that.

Did you intend to check whether `ctx.VXLAN() != null` instead?
Is there another value other than `vxlan` valid here?
If not, then I would make change `String _encapsulation` in `Evpn` to `boolean _encapsuationVlan`, and just set it to `true` here as the sole line in this function.
Otherwise, you should make an enum of the acceptable values, and make encapsulation be of that enum type rather than `String`.

Either way, no need to extract `ctx.VXLAN().getText()`.
</blockquote></details>

Will do



_Code quote:_
```Java
E_encapsulationContext
```
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4216 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15gOUK39byEdBhqdkB:-N1AkjRP5lgvwcScREva:bbo7p6r) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4216)):*

Either `Evpn._extendedVniAll` should be a boolean (like above comment about encapsulation), or there should be a unified polymorphic `_extendedVni` property of say type `ExtendedVniList`, where the subtypes are something like:
- `LiteralExtendedVniList`
- `ExtendedVniAll` (singleton class)
The benefit of the latter pattern is it won't allow both `all` and a literal `list` simultaneously, and there will only be one property to check during conversion.
</blockquote></details>

Changed vniAll to a boolean
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4233 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15hhh_8v95K23EOLGI:-N1AgyAS0uwfquY--CtR:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4233)):*

`Evpn._multicastMode` should be an enum with members `INGRESS_REPLICATION` and `CLIENT` rather than a string.
</blockquote></details>

Will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4221 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15hu2s5jW4EMpx3A58:-N1G1XPh1rzZy5udOEtv:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4221)):*

Looks like you're not handling this. You should add an explicit warning, or at least `todo(ctx)`.
</blockquote></details>

Will do
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4223 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15iN5T2dP1ez5UTBRT:-N1G1VeZ183ohSFsUdmj:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4223)):*

You should make use this function instead of manually parsing:
```
  private @Nonnull Optional<Integer> toInteger(
      ParserRuleContext messageCtx, Vni_numberContext ctx) {
    return toIntegerInSpace(messageCtx, ctx, VNI_NUMBER_RANGE, "vni");
  }

```
Then if the returned `Optional` is empty, you should `continue` (a warning will already have been generated by `toInteger`.
</blockquote></details>

Will do
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15irlsAOgHrXqZMKlP:-N1G1RhV4ly5k5Jo4FvY:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L29)):*

To support ranges without having to enumerate every VNI, better to use an `IntegerSpace` than `List<Integer>`. Note that `IntegerSpace` is immutable, so you have to build a new one including the contents of the old one when updating.
So you'll need a separate function in this class for updating, since you can't modify a returned `IntegerSpace`.
</blockquote></details>

Will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2031 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15l3Cr8QWTzJ1rDtFa:-N1eKka4EhaPaSZtD5eV:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2031)):*

Is it possible for getvlanId() to return `null`? If so, should handle that and exit early to avoid NPE.
</blockquote></details>

Will do
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2038 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15kHxH583B_qubMYjQ:-N1eKmKP3fhJ-RBVGAv1:bebrt1w) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2038)):*

This else continue branch is redundant.
But as a general style comment, we usually prefer to exit early.
So you could do instead above:
```
if (lvlxan.getVniId() == null) {
  continue;
}
// code for when it's not null, without extra indentation.
```

Similarly in the l3 case you can exit early if `vtepSource == null`, and avoid the extra indentation that follows.
</blockquote></details>

Will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2051 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15kp8wDU6lOFALpT0Z:-N1eLk3o8l6_5844Oxk4:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2051)):*

If either
- `get(l3Interface)` returns `null` because interface does not exist
- `getPrimaryAddress()` returns `null` because none was assigned
you will get an NPE (or builder exception).

You should check each of these first before building an l3vni, exiting early if necessary to avoid the NPE.
</blockquote></details>

will do
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1eLrNdCs4ZJQqgNai-:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*

In general, this should be the vrf of the interface, not necessarily the default vrf.
For Juniper, I believe that means you should use the name of the routing instance that owns the interface.
Note that in contrast, converted l2vnis always must be in the default vrf (Batfish quirk, vrfs don't really mean anything for l2vnis anyway).
</blockquote></details>

will do
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java` line 124 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15lsePBp05mvZwTTLl:-N1eLvmp48vvB0IDfuq_:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java#L124)):*

I think it would be better to leave this null in constructor, and instead instantiate this with a setter method from `ConfigurationBuilder` via:

```
@Override
public void enterP_evpn(P_evpnContext ctx) {
  if (_currentLogicalSystem.getEvpn() == null) {
    _currentLogicalSystem.setEvpn(new Evpn()); // <-- instantiate here
  }
}
```
(create the method it if it doesn't exist)

Otherwise every logical system will have an `Evpn` object, which creates more overhead.

This also means you'll need to handle it being null in `JuniperConfiguration`.

</blockquote></details>

will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 6 of 14 files reviewed, 8 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2031 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15l3Cr8QWTzJ1rDtFa:-N1eVIt5--iazIZkB1Um:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2031)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will do
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1eVKvg2LIy795Gq0dD:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 14 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4226 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1iulvJ7_po8aN60Ms-:-N1iulvJ7_po8aN60Ms0:b7eket5) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4226)):*
> ```Java
>       _currentLogicalSystem.getEvpn().setEncapsulation(EvpnEncapsulation.MPLS);
>     } else {
>       throw new BatfishException(
> ```

This branch will never be reached based on the grammar. Instead, can just make `else if` branch above an `else` branch in which you `assert ctx.MPLS() != null` instead of  guarding on it.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4241 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1ivWl70hObeup0y5Pg:-N1ivWl70hObeup0y5Ph:b-yuti0v) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4241)):*
> ```Java
>           List temp =
>               IntStream.range(
>                       Integer.parseInt(vni.getText().split("-")[0]),
> ```

If you wanted to keep this behavor:
- No need to call `split` twice; can store the result once in an array, and then index off of it twice.
- no need to create a list - can just call `forEach` on a stream.

However, I'd be very worried about storing individual elements for a large range of VNIs. The time order of creating the vni list will  be O(average size of range * number of ranges). If you use `IntegerSpace`instead of iterating over each VNI in the range, the time would be O(number of ranges) (constant time per range).

And either way, you should be making use of a validating extraction function to get the ends of the vni range, rather than doing raw text processing in this function. In particular, need VNIs out of valid range, as well as invalid ranges with end before start.

Can adapt this VLAN range extracting function for VNI ranges by just changing the second parameter type.
```
  private @Nonnull Optional<SubRange> toSubRange(
      ParserRuleContext messageCtx, Vlan_rangeContext ctx) {
    Optional<Integer> rangeStart = toInteger(messageCtx, ctx.start);
    if (!rangeStart.isPresent()) {
      return Optional.empty();
    }
    Optional<Integer> rangeEnd = ctx.end == null ? rangeStart : toInteger(messageCtx, ctx.end);
    if (!rangeEnd.isPresent()) {
      return Optional.empty();
    }
    return Optional.of(new SubRange(rangeStart.get(), rangeEnd.get()));
  }
```
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4265 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1izDqiD8bdULSECDmf:-N1izDqiD8bdULSECDmg:bpd4nin) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4265)):*
> ```Java
>       _currentLogicalSystem.getEvpn().setMulticastMode(MulticastModeOptions.INGRESS_REPLICATION);
>     } else {
>       _currentLogicalSystem.getEvpn().setMulticastMode(MulticastModeOptions.CLIENT);
> ```

Please insert this line in this branch: `assert ctx.CLIENT() != null` to make sure someone notices if the grammar gets changed to add a third alternative.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 12 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1j-92REl21xX2JOKC0:-N1j-92REl21xX2JOKC1:b-u42mlc) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L12)):*
> ```Java
> public final class Evpn implements Serializable {
> 
>   private Enum _multicast_mode;
> ```

This should be of stricter type `MulticastModeOptions`, not any generic `Enum`.
Applies to setter and getter as well.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 15 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1j-FCS2XtoXZ7r0BLq:-N1j-FCS2XtoXZ7r0BLr:be1pn1u) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L15)):*
> ```Java
>   private @Nullable Boolean _extended_vni_all;
>   private @Nullable List<Integer> _extended_vni_list;
>   private @Nullable Enum _encapsulation;
> ```

This should be of stricter type `EvpnEncapsulation`, not any generic `Enum`.
Applies to setter and getter as well.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1j11sP0X-ZLhObahMz:b-31bh19) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The vrf you add this vni to must be the same as the src vrf vrf you set in the l3vni. Please fix here and make sure you have unit test for it.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2024 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1j-yVCEbe-SxQe7lEG:-N1j-yVD0CHid9zee3Ll:bnpmj57) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2024)):*
> ```Java
>   private void convertVxlan() {
>     for (Vlan vxlan : _masterLogicalSystem.getNamedVlans().values()) {
>       if (vxlan.getVlanId() == null) {
> ```

This early exit should only apply if l3 interface is null, i.e. this is an l2vni, right? So shouldn't you move this to the top of the second branch below?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1j0ec_DVbES4_mCNt8:-N1j0ec_DVbES4_mCNt9:bg7ndnc) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
> ```Java
>                 .getDefaultRoutingInstance()
>                 .getInterfaces()
>                 .get(l3Interface)
> ```

Nit: can factor out a variable for this chain up to `get(l3Interface)` to use in this guard and the one above.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 14 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4241 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1ivWl70hObeup0y5Pg:-N1o7BqBAsLdK3DCDyW4:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4241)):*

If you wanted to keep this behavor:
- No need to call `split` twice; can store the result once in an array, and then index off of it twice.
- no need to create a list - can just call `forEach` on a stream.

However, I'd be very worried about storing individual elements for a large range of VNIs. The time order of creating the vni list will  be O(average size of range * number of ranges). If you use `IntegerSpace`instead of iterating over each VNI in the range, the time would be O(number of ranges) (constant time per range).

And either way, you should be making use of a validating extraction function to get the ends of the vni range, rather than doing raw text processing in this function. In particular, need VNIs out of valid range, as well as invalid ranges with end before start.

Can adapt this VLAN range extracting function for VNI ranges by just changing the second parameter type.
```
  private @Nonnull Optional<SubRange> toSubRange(
      ParserRuleContext messageCtx, Vlan_rangeContext ctx) {
    Optional<Integer> rangeStart = toInteger(messageCtx, ctx.start);
    if (!rangeStart.isPresent()) {
      return Optional.empty();
    }
    Optional<Integer> rangeEnd = ctx.end == null ? rangeStart : toInteger(messageCtx, ctx.end);
    if (!rangeEnd.isPresent()) {
      return Optional.empty();
    }
    return Optional.of(new SubRange(rangeStart.get(), rangeEnd.get()));
  }
```
</blockquote></details>

will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 10 of 14 files reviewed, 7 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4223 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15iN5T2dP1ez5UTBRT:-N1o9EnDADWa6-k1UWuv:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4223)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will do
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4233 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15hhh_8v95K23EOLGI:-N1o9Frx9cae8um7z6n8:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4233)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will do
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4241 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1ivWl70hObeup0y5Pg:-N1o9DrIAlRji9s_INxW:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4241)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1o9L5HEfbMLd0sGrn6:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 5 unresolved discussions (waiting on @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4216 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15gOUK39byEdBhqdkB:-N1oCnXFA8Ui-tH5cLom:bgqeckx) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4216)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Changed vniAll to a boolean
</blockquote></details>

OK for now, but at some point (doesn't need to be in this PR) we will want conflict handling if user both sets `all` and a literal list.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4251 at r3](https://reviewable.io/reviews/batfish/batfish/8283#-N1oCKRpAgfUixHPJBac:-N1oCKRpAgfUixHPJBad:bn71vds) ([raw file](https://github.com/batfish/batfish/blob/7da8aac1171db6b67d9a9709254ad1d28054792b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4251)):*
> ```Java
>           vnis.add(
>               IntegerSpace.of(
>                   new SubRange(maybeVniRange.get().getStart(), maybeVniRange.get().getEnd())));
> ```

Here you are unncessarily cloning a `SubRange`.
The type of `maybeVniRange.get()` is already a `SubRange` - just use that.

Also see other comment about using single IntegerSpace instead of list of IntegerSpacess.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15irlsAOgHrXqZMKlP:-N1o98gm9SQ74jQG9ea9:b5da54h) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L29)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will do
</blockquote></details>

Sorry let me clarify: use a single `IntegerSpace`, not a list of them. If you need to update it after it has been set, you will need to set a new IntegerSpace that is the union of the old one and what you are adding, e.g. 
```
evpn.setExtendedVniList(IntegerSpace.builder().including(evpn.getExtendedVniList()).including(andAlsoTheseNewVnis).build());
```
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1oBQa90fmZvKZ5AI_Q:-N1oBQa90fmZvKZ5AI_R:blwdoxx) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L25)):*
> ```Java
>   }
> 
>   public @Nullable Boolean getExtendedVniAll() {
> ```

Juniper docs I saw say that you cannot use both all and a literal list. Best I can tell you have not explicitly handled the case 
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1oAQcG3QB17fzeiTqQ:bw3ad08) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Where is the unit test?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1j0ec_DVbES4_mCNt8:-N1o9VoO6XR4598srXaM:b-nov6im) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

will do
</blockquote></details>

nit: just use a single local var for the interface, no need to store the map of interfaces, i.e do:
```
Interface iface = _masterLogicalSystem.getDefaultRoutingInstance().getInterfaces().get(l3Interface);
```
and use `iface` below instead of `allInterfaces.get(l3Interface)`.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 4 unresolved discussions (waiting on @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8283#-N1oBQa90fmZvKZ5AI_Q:-N1oD4ya97aGIG8Nr-oO:b6h2ymb) ([raw file](https://github.com/batfish/batfish/blob/ec568c201934e19c33e0d56c7276386891c3898a/projects/batfish/src/main/java/org/batfish/representation/juniper/Evpn.java#L25)):*

Juniper docs I saw say that you cannot use both all and a literal list. Best I can tell you have not explicitly handled the case 
</blockquote></details>

Sorry, ignore, didn't mean to publish this.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 4 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4251 at r3](https://reviewable.io/reviews/batfish/batfish/8283#-N1oCKRpAgfUixHPJBac:-N1p06JeBHOvNIImEUBu:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/7da8aac1171db6b67d9a9709254ad1d28054792b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4251)):*

Here you are unncessarily cloning a `SubRange`.
The type of `maybeVniRange.get()` is already a `SubRange` - just use that.

Also see other comment about using single IntegerSpace instead of list of IntegerSpacess.
</blockquote></details>

Done.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 10 of 14 files reviewed, 4 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N1sO62-D2qh_P3olNhb:b-mpznqe) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*

Where is the unit test?
</blockquote></details>

Where/what file should this test live?  I'm struggling to find the right place for the unit test.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 10 of 14 files reviewed, 4 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N2HOgPy9rJuzLqbnoij:bl0vjt0) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Where/what file should this test live?  I'm struggling to find the right place for the unit test.
</blockquote></details>

I'm working on adding the test to JuniperConfigurationTest.java which appears to be the accurate place.  Looking through other examples in the file I'm not immediately clear what specifically I should be testing here?  Just that when a l3interface isn't defined on a vlan that its being set as l2vni?

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 10 of 14 files reviewed, 4 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2054 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15jfYZBj9JjY-705nu:-N2r-2GzE71qjcXSTN5E:b-1jyzyz) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2054)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

I'm working on adding the test to JuniperConfigurationTest.java which appears to be the accurate place.  Looking through other examples in the file I'm not immediately clear what specifically I should be testing here?  Just that when a l3interface isn't defined on a vlan that its being set as l2vni?
</blockquote></details>

If it is an extraction/conversion test, the test should go in FlatJuniperGrammarTest.java. We usually test here that config to VS model and/or to VI model is correct.

ratulm (commented):
> Reviewed 4 of 4 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 6 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java` line 1585 at r5](https://reviewable.io/reviews/batfish/batfish/8283#-N3W9PltAF_10IN69G54:-N3W9PltAF_10IN69G55:b1lmyfs) ([raw file](https://github.com/batfish/batfish/blob/6a779e90b13308e2b8900e59ba78a0635aba50a6/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L1585)):*
> ```Java
>   @Test
>   public void testEvpnEncapsulationExtraction() {
>     JuniperConfiguration c = parseJuniperConfig("juniper-evpn-multicast-ingress");
> ```

This file has a lot of content whose extraction is not being tested. Would a smaller test case make sense?
___
*[`projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java` line 143 at r5](https://reviewable.io/reviews/batfish/batfish/8283#-N3WAEpu2FkabaG4Cq-v:-N3WAEpu2FkabaG4Cq-w:b-r0gidh) ([raw file](https://github.com/batfish/batfish/blob/6a779e90b13308e2b8900e59ba78a0635aba50a6/projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java#L143)):*
> ```Java
> 
>   @Test
>   public void testL3VniVrf() {
> ```

Javadoc about what this is/not testing will be helpful.

ratulm (commented):
> Dismissed @arifogel from a discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 5 unresolved discussions (waiting on @arifogel and @jeffkala)

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 5 unresolved discussions (waiting on @arifogel and @jeffkala)

___
*[`projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java` line 1585 at r5](https://reviewable.io/reviews/batfish/batfish/8283#-N3W9PltAF_10IN69G54:-N3WEKyr693RryzEs3HF:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/6a779e90b13308e2b8900e59ba78a0635aba50a6/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L1585)):*

This file has a lot of content whose extraction is not being tested. Would a smaller test case make sense?
</blockquote></details>

will do
___
*[`projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java` line 143 at r5](https://reviewable.io/reviews/batfish/batfish/8283#-N3WAEpu2FkabaG4Cq-v:-N3WEMod9R0hPNmD4JJv:b-d7bhn0) ([raw file](https://github.com/batfish/batfish/blob/6a779e90b13308e2b8900e59ba78a0635aba50a6/projects/batfish/src/test/java/org/batfish/representation/juniper/JuniperConfigurationTest.java#L143)):*

Javadoc about what this is/not testing will be helpful.
</blockquote></details>

will do

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: all files reviewed, 2 unresolved discussions (waiting on @jeffkala)

___
*[`projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java` line 1585 at r5](https://reviewable.io/reviews/batfish/batfish/8283#-N3W9PltAF_10IN69G54:-N3ZXGOR1QISI55_FM8r:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/6a779e90b13308e2b8900e59ba78a0635aba50a6/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L1585)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

will do
</blockquote></details>

done

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 14 of 15 files reviewed, 2 unresolved discussions (waiting on @jeffkala and @ratulm)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4216 at r1](https://reviewable.io/reviews/batfish/batfish/8283#-N15gOUK39byEdBhqdkB:-N3ZYyE85vAGqSj3m3nP:b249r76) ([raw file](https://github.com/batfish/batfish/blob/f59c4e77d0c830001f42a824fa190403987c085c/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4216)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

working on getting a lab device, but dont think you can set both of these.
</blockquote></details>

can only be one or the other.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 18 of 19 files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[`tests/parsing-tests/unit-tests-warnings.ref` line 1129 at r8](https://reviewable.io/reviews/batfish/batfish/8283#-N3ffIvD8888jzjiod0D:-N3gLCDh99CZsgYYI9FE:baa9881) ([raw file](https://github.com/batfish/batfish/blob/c13a249790a06f6d19730a2d5f1995b693fe79a6/tests/parsing-tests/unit-tests-warnings.ref#L1129)):*

We are getting two warnings per line. Is that expected? Easy to prevent?
</blockquote></details>

Trying to figure out where this is coming from, will update shortly.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 18 of 19 files reviewed, 1 unresolved discussion (waiting on @ratulm)

___
*[`tests/parsing-tests/unit-tests-warnings.ref` line 1129 at r8](https://reviewable.io/reviews/batfish/batfish/8283#-N3ffIvD8888jzjiod0D:-N3gWe2x-bVoFhQmxhbY:b-sncs7r) ([raw file](https://github.com/batfish/batfish/blob/c13a249790a06f6d19730a2d5f1995b693fe79a6/tests/parsing-tests/unit-tests-warnings.ref#L1129)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Trying to figure out where this is coming from, will update shortly.
</blockquote></details>

appears to be caused by this
```
  @Override
  public void exitEvo_vrf_target(Evo_vrf_targetContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_auto(Evovt_autoContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_community(Evovt_communityContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_export(Evovt_exportContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_import(Evovt_importContext ctx) {
    todo(ctx);
  }
```
would it be best to only do the todo ctx on Evo_vrf_target?  I'm trying to find the history but a previous review requested I add these.

ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: 18 of 19 files reviewed, 1 unresolved discussion (waiting on @jeffkala and @ratulm)

___
*[`tests/parsing-tests/unit-tests-warnings.ref` line 1129 at r8](https://reviewable.io/reviews/batfish/batfish/8283#-N3ffIvD8888jzjiod0D:-N3gdisjCxJpWKGEsXbT:b623os1) ([raw file](https://github.com/batfish/batfish/blob/c13a249790a06f6d19730a2d5f1995b693fe79a6/tests/parsing-tests/unit-tests-warnings.ref#L1129)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

appears to be caused by this
```
  @Override
  public void exitEvo_vrf_target(Evo_vrf_targetContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_auto(Evovt_autoContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_community(Evovt_communityContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_export(Evovt_exportContext ctx) {
    todo(ctx);
  }

  @Override
  public void exitEvovt_import(Evovt_importContext ctx) {
    todo(ctx);
  }
```
would it be best to only do the todo ctx on Evo_vrf_target?  I'm trying to find the history but a previous review requested I add these.
</blockquote></details>

If nothing that follows `evo_vrf_target` in the grammar is supported, then, yes, enough to just have a todo there (and delete all the other methods as well).

ratulm (commented):
> Reviewed 1 of 1 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8283)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jeffkala)

Issue #8275, opened 2022-04-27
BGP template peer-session incorrectly flagged as unused in Cisco IOSXE > router bgp


network-dave:
Hello,

I found at least 1 occurrence of an inherited BGP peer-session template incorrectly flagged as unused by .q.unusedStructures():

```
router bgp 64512
 template peer-policy PP-VPNV4-AGGREGATION-PE-RR
  advertisement-interval 0
  next-hop-self
  maximum-prefix 20000
  send-community both
 exit-peer-policy
 !
 template peer-session PS-AGGREGATION-PE-RR
  password XXXXXX
  inherit peer-session PS-RRC-GENERAL
 exit-peer-session
 !
 template peer-session PS-RRC-GENERAL
  update-source Loopback0
  timers 30 90
 exit-peer-session
```

yields:

```

              Structure_Type                           Structure_Name                                               Source_Lines
0  bgp template peer-session                           PS-RRC-GENERAL                    configs/config:[1279, 1280, 1281, 1282]

```

Kind regards,

David

PR #8300, opened 2022-05-07
IOS: indirect inheritance for peer sessions

ratulm:
Fixes https://github.com/batfish/batfish/issues/8275

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4
@@ -997,6 +997,7 @@ template_peer_session_rb_stanza
    TEMPLATE PEER_SESSION name = variable_permissive NEWLINE
    (
       bgp_tail
+      | inherit_peer_session_bgp_tail
       | remote_as_bgp_tail
    )*
    (
--- a/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java
@@ -5974,6 +5974,8 @@ public class CiscoControlPlaneExtractor extends CiscoParserBaseListener
       _currentIpPeerGroup.setPeerSession(sessionName);
     } else if (_currentNamedPeerGroup != null) {
       _currentNamedPeerGroup.setPeerSession(sessionName);
+    } else if (_currentPeerSession != null) {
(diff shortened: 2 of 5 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8300)* status: all files reviewed, 2 unresolved discussions (waiting on @corinaminer and @ratulm)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java` line 5977 at r1](https://reviewable.io/reviews/batfish/batfish/8300#-N1eG_Ve1KCYoO0cEsfx:-N1eG_Ve1KCYoO0cEsfy:bbybi4e) ([raw file](https://github.com/batfish/batfish/blob/de74f45199d6c598ef5c95d07da2e9dbb6c60746/projects/batfish/src/main/java/org/batfish/grammar/cisco/CiscoControlPlaneExtractor.java#L5977)):*
> ```Java
>     } else if (_currentNamedPeerGroup != null) {
>       _currentNamedPeerGroup.setPeerSession(sessionName);
>     } else if (_currentPeerSession != null) {
> ```

FYI from https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst3850/software/release/16-11/configuration_guide/rtng/b_1611_rtng_3850_cg/configuring_bgp.html#task_pmr_wsb_fhb

it appears that the inheritance depth for peer session templates is limited to 7, and for neighbors it is limited to 8 (1 plus max inheritance depth of peer-session template). Shallower inherited settings take priority as you might expect. Might be a warning about exceeding max depth in conversion, or at least a TODO comment there. Optional.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java` line 6954 at r1](https://reviewable.io/reviews/batfish/batfish/8300#-N1eHPKhAjofk3kAwFBm:-N1eHPKhAjofk3kAwFBn:b-6knmss) ([raw file](https://github.com/batfish/batfish/blob/de74f45199d6c598ef5c95d07da2e9dbb6c60746/projects/batfish/src/test/java/org/batfish/grammar/cisco/CiscoGrammarTest.java#L6954)):*
> ```Java
>         c.getDefaultVrf().getBgpProcess().getActiveNeighbors().get(Ip.parse("10.10.10.2"));
>     assertThat(neighbor.getDescription(), equalTo("direct-parent"));
>     assertThat(neighbor.getLocalIp(), equalTo(Ip.parse("10.10.10.1")));
> ```

This is the only interface address on the device. As is, this does not obviously prove you are actually inheriting from the indirect parent. Please make a more convincing test that does not rely on intricacies of automatic local ip selection algorithm.

PR #8302, opened 2022-05-11
BDDFactory: cache BDDPairings

anothermattbrown:
Add a getPair method that computes a BDDPairing that maps variables to
variables. The BDDPairing is cached to improve cache performance of
BDD#replace(BDDPairing), BDD#replaceWith(BDDPairing), etc. Without caching,
two equivalent pairings will have different IDs, and JFactory's replacecache
will not share their work.

Caveat: BDDPairings are mutable, and mutating cached pairings is not safe.
We should improve the API to guard against that. (We currently never mutate
a pairing after creating it).

Motivation: as we build and merge `Transform` transitions in the reachability graph, 
we can create many the same BDDPairing many times.


dhalperi (commented):
> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8302)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #8307, opened 2022-05-13
IspModeling: enable not-configured local IP for BgpPeerInfo

dhalperi:
During ISP modeling, we need to generate correct ISP-side BGP configuration.
This requires, in particular, specifying the neighbor on that side -- aka,
the local IP on the node in the snapshot. This is usually configured via
an explicit update source or local address command, but may not always be.

To support ISP modeling for these cases, we:

1. provide a way to force local IP in the BgpPeerInfo
2. improve default local IP inference to include local
connected routes (same VRF) for eBGP singlehop sessions.

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/e2e/isp/IspModelingTest.java
+++ b/projects/allinone/src/test/java/org/batfish/e2e/isp/IspModelingTest.java
@@ -61,6 +61,15 @@ public class IspModelingTest {
   }
 
   @Test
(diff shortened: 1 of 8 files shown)


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8307)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8319, opened 2022-05-18
Efficient conversion from ImmutableBDDInteger constraints to LongSpace

anothermattbrown:
Uses a new BDDTraversal to walk the BDD.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/ImmutableBDDInteger.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/ImmutableBDDInteger.java
@@ -4,12 +4,12 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static org.batfish.common.bdd.BDDUtils.bitvector;
 
 import com.google.common.collect.DiscreteDomain;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableRangeSet;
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8319)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8319#-N2Nu8cr30ERzZLrXJy5:-N2Nu8cr30ERzZLrXJy6:bikmwu5) (no related file):*
after discussion:

* some of the cleverness, if we keep it, deserves comments
* the optimization to avoid projection for a BDDInteger at the "top" of the input BDD seems .. hard to get right? maybe not worth it?
* avoiding blowup on wildcards seems like a bad idea. The caller is expecting a range set, and they should probably be told that they need to do something different rather than silently get not-accurate results?
* I tried to mentally write javadoc for BDDTraversal and I had a hard time. Sort of a pre-order high-first DFS traversal. The reason I had a hard time explaining was there's no visit(var) method, instead we use the traverse_high/traverse_low methods to do the work at var. Not a problem, but if you can think of a better description/procedure I wouldn't complain. 
* just observing that we're walking all paths, which is likely fine since that's related the number of ranges in the output anyway.

___

PR #8330, opened 2022-05-24
JunOS: fix static routes - ECMP, interface vs IP, and next hop overwriting

sfraint:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @sfraint)

___
*[`projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/static-route-overwrite` line 13 at r1](https://reviewable.io/reviews/batfish/batfish/8330#-N2r-C0x5Cxjbeyg_JgO:-N2r-C0x5Cxjbeyg_JgP:b1ok2qs) ([raw file](https://github.com/batfish/batfish/blob/c6bafdaf1276c9c568e3c6272da43c5865d888a2/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/static-route-overwrite#L13)):*
> ```
> #
> # These are the only static routes that should persist
> set routing-options static route 10.0.0.0/16 next-hop 10.0.0.1
> ```

can you not do a nhint+nhip static route? That is common on most vendors.

If so, how does it work?

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330)* status: all files reviewed, 2 unresolved discussions (waiting on @dhalperi and @sfraint)

___
*[`projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/static-route-overwrite` line 13 at r1](https://reviewable.io/reviews/batfish/batfish/8330#-N2r-C0x5Cxjbeyg_JgO:-N2r3KraAutVW6c4W2bG:bmjo8io) ([raw file](https://github.com/batfish/batfish/blob/c6bafdaf1276c9c568e3c6272da43c5865d888a2/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/static-route-overwrite#L13)):*

can you not do a nhint+nhip static route? That is common on most vendors.

If so, how does it work?
</blockquote></details>

As far as I can tell, you can't combine interface and IP for a single next hop.

On a vSRX device I was able to determine that we are definitely handling some cases wrong, where both next hop interface and next hop IP are set, e.g.
```
root# show interfaces | display set 
set interfaces ge-0/0/0 unit 0 family inet address 10.10.10.1/24
set interfaces ge-0/0/1 unit 0 family inet address 10.10.11.1/24

root# show routing-options | display set 
set routing-options static route 10.11.12.0/24 next-hop ge-0/0/1.0
set routing-options static route 10.11.12.0/24 next-hop 10.10.10.111

root# commit 
commit complete

root# run show route 10.11.12.0/24              
inet.0: 8 destinations, 8 routes (8 active, 0 holddown, 0 hidden)
+ = Active Route, - = Last Active, * = Both

10.11.12.0/24      *[Static/5] 00:00:24
                      to 10.10.10.111 via ge-0/0/0.0
                    > via ge-0/0/1.0
```

Also, as you might expect, you can set many next hops IPs and or interfaces. I've updated this PR to allow it.

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330)* status: all files reviewed, 2 unresolved discussions (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java` line 116 at r2](https://reviewable.io/reviews/batfish/batfish/8330#-N2rj_3M8BoKPc0gifaJ:-N2rksOCCwSeo87Yyqzk:b-3i6a54) ([raw file](https://github.com/batfish/batfish/blob/bc59a4a072cc8e3e3d2e9b07b1108e99d2a3884c/projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java#L116)):*

where is this called with `false`? is clearing still correct in that scenario?
</blockquote></details>

It is never called with `false`, but in the case where you would remove the `discard` with a `delete ... discard`, the behavior should still be correct. I will go ahead and add a docstring.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330)* status: all files reviewed, 2 unresolved discussions (waiting on @dhalperi and @sfraint)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java` line 116 at r2](https://reviewable.io/reviews/batfish/batfish/8330#-N2rj_3M8BoKPc0gifaJ:-N2rm3i8AeZg3s6xP_3V:b-94ujra) ([raw file](https://github.com/batfish/batfish/blob/bc59a4a072cc8e3e3d2e9b07b1108e99d2a3884c/projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java#L116)):*

It is never called with `false`, but in the case where you would remove the `discard` with a `delete ... discard`, the behavior should still be correct. I will go ahead and add a docstring.
</blockquote></details>

Suggest just removing the parameter then. (Delete is handled in pre-processing and would never make it here)

sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java` line 116 at r2](https://reviewable.io/reviews/batfish/batfish/8330#-N2rj_3M8BoKPc0gifaJ:-N2rmcUr1B_5DHjtvt0I:b-bu91xm) ([raw file](https://github.com/batfish/batfish/blob/bc59a4a072cc8e3e3d2e9b07b1108e99d2a3884c/projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRoute.java#L116)):*

Suggest just removing the parameter then. (Delete is handled in pre-processing and would never make it here)
</blockquote></details>

Updated

dhalperi (commented):
:lgtm:

> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8330#-:-N2rnk6V1D-iXuhWRcyB:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

PR #8340, opened 2022-05-26
FileBasedStorage: stripe dataplane by host

dhalperi:
DataPlane is one big ball of complexity which takes a long time to serialize
and deserialize. To improve on this, stripe it by host. This enables less work
by the Java serialization framework (which keeps all serialized objects in an
identity hashmap) and by also enables use of many cores (CPU overhead for
serializing objects) and better pipelining of CPU with I/O.

Note that Java serialization will only serialize pointed-to objects once no
matter how many times they are referenced. Striping means that each object
is now serialized many times, increasing both a) on-disk storage (after
serialization) and b) in-memory storage (after de-serialization of the same
object used in different devices).

These penalties are overcome via the followng insights:

a) in practice, the largest objects are things like routes that are not likely
to be identical across different devices (e.g., they reflect next-hop IP or AS
Path). So pointer equality is kept within a device's data plane. This reduces
blow-up of both on-disk and in-memory storage.

b) the small, numerous objects are written to many files but are interned after
deserialization preventing in-memory blowup.

I ran experiments with one BGP intensive network. I found that striping:

1. increases serialized, compressed size by about 10%, from 36 MB to 40 MB
2. decreases store time by 3x: on a 4-core 2-threads laptop from 10s to 3s
3. results in essentially the same in-memory use after cache expiry and reload
4. has a negligible effect on dataplane load time, about 3s either method

In another, larger BGP intensive network:

1. Serialized compressed size from 385 MB to 387 MB.
2. Serialization time from 2.5m to 3.5s (40x)

commit-id:d199f19c

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ForwardingAnalysis.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ForwardingAnalysis.java
@@ -1,9 +1,10 @@
 package org.batfish.datamodel;
 
+import java.io.Serializable;
 import java.util.Map;
 import javax.annotation.Nonnull;
 
-public interface ForwardingAnalysis {
+public interface ForwardingAnalysis extends Serializable {
   /** Mapping: hostname -&gt; inInterface -&gt; ipsToArpReplyTo */
   Map<String, Map<String, IpSpace>> getArpReplies();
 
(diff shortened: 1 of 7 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8340)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8340)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8340)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8341, opened 2022-05-26
BgpRouteAttributes: shrink size by 29%

dhalperi:
112 bytes to 80 bytes.

Java enums are by-reference, so 8 bytes each. Swapping to bytes
(which is safe since there are << 128 values) reduces this to 1
byte each. There were 7 bytes of padding already, so this is
in fact a full 32 byte reduction - all 4 new bytes are in that
padding.

This is not best practice as enum ordinals are not stable
if you add new values, but it seems worth it.

commit-id:e109a139

---

**Stack**:
- #8340
- #8341 ⬅


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8341)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java` line 143 at r1](https://reviewable.io/reviews/batfish/batfish/8341#-N312ufr7z_DsHYzjHjU:-N312uftC49ESrFze5oP:b1oz5pf) ([raw file](https://github.com/batfish/batfish/blob/e054ebc97c4e64ab378797840fd48ee55e752332/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java#L143)):*
> ```Java
>         h = h * 31 + _protocol.ordinal();
>         h = h * 31 + Boolean.hashCode(_receivedFromRouteReflectorClient);
>         h = h * 31 + (_srcProtocol == null ? 0 : _srcProtocol.ordinal());
> ```

was this a bug before? ordinal would presumably == 0 for some protocol?



_Code quote:_
```Java
(_srcProtocol == null ? 0 : _srcProtocol.ordinal());
```

PR #8350, opened 2022-06-02
Juniper: Convert tunnel attributes

corinaminer:
Includes parsing, extraction, and conversion for Juniper tunnel attributes such as:
```
set policy-options tunnel-attribute tunnel-attr-1 remote-end-point 1.2.3.4
set policy-options tunnel-attribute tunnel-attr-1 tunnel-type ipip
```
The resulting VI tunnel attributes are not yet used for anything.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8350)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java` line 838 at r1](https://reviewable.io/reviews/batfish/batfish/8350#-N3_IOKu28nS8Kc2-mWb:-N3_IOKu28nS8Kc2-mWc:b-1wbgq) ([raw file](https://github.com/batfish/batfish/blob/c8fa07124216115a1551b72c112d95f79e0bd078/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L838)):*
> ```Java
>   public @Nonnull Map<String, TunnelAttribute> getTunnelAttributes() {
>     return _tunnelAttributes;
>   }
> ```

How did you come up with this VI model?

I would not have expected this to be a top-level concept, but rather a property of BgpRoute, for example.

Do the other vendors that support applying BGP tunnel attributes do it in the same way as Juniper?



_Code quote:_
```Java
  @JsonProperty(PROP_TUNNEL_ATTRIBUTES)
  public @Nonnull Map<String, TunnelAttribute> getTunnelAttributes() {
    return _tunnelAttributes;
  }
```

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8350)* status: 6 of 11 files reviewed, 2 unresolved discussions (waiting on @corinaminer and @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java` line 838 at r1](https://reviewable.io/reviews/batfish/batfish/8350#-N3_IOKu28nS8Kc2-mWb:-N3b4hco-SSHyHjY__Gg:bn3t2yv) ([raw file](https://github.com/batfish/batfish/blob/c8fa07124216115a1551b72c112d95f79e0bd078/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Configuration.java#L838)):*

> SetTunnelAttribute
Yeah, seeing the other PR this needs to change. We have an existing pattern for this:

1. `SetTunnelEncapsulationAttribute(TunnelEncapsulationAttributeExpr)`
2. `LiteralTunnelAttributeExpr` (inline value)
3. `NamedTunnelAttributeExpr` (something in config)
4. ... ?

Prefer 1+2 to start, rather than 1+3. Less blowup.

> space-efficient if the same tunnel attribute is referenced in many routing policy terms.

We can intern `TunnelEncapsulationAttribute` if we need to  -- e.g., like CommunitySet.
</blockquote></details>

Ack. Reducing this PR to tunnel attribute extraction. The VS tunnel attributes will be used in #8353 when converting policy terms with `then tunnel-attribute...`.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/TunnelAttribute.java` line 15 at r1](https://reviewable.io/reviews/batfish/batfish/8350#-N3aYDXu9T-Aqegz_Y3v:-N3bM9573m5vBEqgvtOT:bivvvh7) ([raw file](https://github.com/batfish/batfish/blob/c8fa07124216115a1551b72c112d95f79e0bd078/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/TunnelAttribute.java#L15)):*

1. Looks like this should be `TunnelEncapsulationAttribute`. E.g., this first sentence of the RFC 9012:

>This document defines a BGP path attribute known as the "Tunnel Encapsulation attribute"

2. Should either move this to the `datamodel.bgp` package or name it something like `BgpX`.
</blockquote></details>

Done and done, though this class is now introduced in #8352 instead of here.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/TunnelAttribute.java` line 19 at r1](https://reviewable.io/reviews/batfish/batfish/8350#-N3aZX9V-XJKKbw400ZV:-N3b568TCHkVuwAhj_88:b-yqawjn) ([raw file](https://github.com/batfish/batfish/blob/c8fa07124216115a1551b72c112d95f79e0bd078/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/TunnelAttribute.java#L19)):*

Looking at the RFC, we know this will have to change. (We should already be using it for VXLAN/EVPN and maybe some specific  extended communities?) If we were being forward looking, we would include the type happen here (it's a 2-byte unsigned integer) and have a named constructor like `ofIpIp(Ip removeEndpoint)`.

ok as is, I guess, but we will pay soon enough.
</blockquote></details>

yeah... i figured it makes sense to do the minimum to support our use case first.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8350)* status: 5 of 11 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 5802 at r1](https://reviewable.io/reviews/batfish/batfish/8350#-N3aXw4DAaJLuZ1ceAcl:-N3faK1v844cY9N8jevr:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/c8fa07124216115a1551b72c112d95f79e0bd078/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L5802)):*

add an `assert ctx.IPIP() != null`? (ish, might not compile as-is)
</blockquote></details>

done

dhalperi (commented):
> Reviewed 5 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8350)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #8359, opened 2022-06-07
BGP routes questions: Add tunnel encapsulation attr column

corinaminer:


SLarkworthy (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8359)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

___
*[`projects/question/src/main/java/org/batfish/question/routes/RoutesAnswerer.java` line 660 at r1](https://reviewable.io/reviews/batfish/batfish/8359#-N43kCNA6LQbTaOoQMtb:-N43kCNA6LQbTaOoQMtc:bmyyze9) ([raw file](https://github.com/batfish/batfish/blob/6d317b4d203fbe929465f96dd81c49f85153e806/projects/question/src/main/java/org/batfish/question/routes/RoutesAnswerer.java#L660)):*
> ```Java
>         columnBuilder.add(
>             new ColumnMetadata(
>                 COL_BASE_PREFIX + COL_TUNNEL_ENCAPSULATION_ATTRIBUTE,
> ```

Looks like these diff values never get set (similarly, I noticed that Weight also isn't getting set in the diff version of this question). If we don't care about this attribute in the diff version for some reason, we don't need these columns. If we do want it, it would need to be added to RouteRowAttribute and then populated in populateBgpRouteAttributes.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8359)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @SLarkworthy)

___
*[`projects/question/src/main/java/org/batfish/question/routes/RoutesAnswerer.java` line 660 at r1](https://reviewable.io/reviews/batfish/batfish/8359#-N43kCNA6LQbTaOoQMtb:-N44LE1kC92wqr5OWNnI:b-45ne5u) ([raw file](https://github.com/batfish/batfish/blob/6d317b4d203fbe929465f96dd81c49f85153e806/projects/question/src/main/java/org/batfish/question/routes/RoutesAnswerer.java#L660)):*
<details><summary><i>Previously, SLarkworthy (Sarah Larkworthy) wrote…</i></summary><blockquote>

Looks like these diff values never get set (similarly, I noticed that Weight also isn't getting set in the diff version of this question). If we don't care about this attribute in the diff version for some reason, we don't need these columns. If we do want it, it would need to be added to RouteRowAttribute and then populated in populateBgpRouteAttributes.
</blockquote></details>

good call, done. Weird that weight isn't getting set, but out of scope for this PR

SLarkworthy (commented):
:lgtm:

> Reviewed 9 of 10 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8359#-:-N44OwI850nDsswe6r-L:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8361, opened 2022-06-10
PAN: support for template variables

sfraint:
Add support for address-object-like template variables (`ip-netmask` and `ip-range`).

For now, these are treated like `address` objects in the VS model. This is mostly correct but is too permissive of references, as you cannot reference template variables outside of a template on a real device.


dhalperi (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8361#-:-N4EKmAD3_lioJOX8qKj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sfraint)

PR #8363, opened 2022-06-11
Add xfail BGP additional-paths tests

arifogel:
IOS-based e2e add-path tests:
- Test that a router that sends all additional-paths will advertise multiple routes that are
  duplicate post-export, and that both will be present in the BGP RIB on the receiver
- Test that a router that sends all additional-paths will advertise routes that are not ECMP-best,
  and that both will be present in the BGP RIB on a receiver
- Test that a router with multiple routes for a prefix will advertise only one route per
  unique next hop for that given prefix

Tests will pass without xfail when add-path is implemented properly

### DIFF
--- a/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathDuplicateTest.java
+++ b/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathDuplicateTest.java
@@ -62,7 +62,6 @@ public final class BgpAddPathDuplicateTest {
   @Test
   public void testBgpRib() {
     Table<String, String, Set<Bgpv4Route>> bgpRoutes = _dp.getBgpRoutes();
-    Table<String, String, Set<Bgpv4Route>> backupRoutes = _dp.getBgpBackupRoutes();
 
     // as2border
     assertThat(bgpRoutes.get("as2border", DEFAULT_VRF_NAME), hasSize(2));
--- a/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathDuplicateTest.java
+++ b/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathDuplicateTest.java
@@ -1,0 +1,86 @@
+package org.batfish.dataplane.ibdp;
+
+import static org.batfish.datamodel.Configuration.DEFAULT_VRF_NAME;
+import static org.hamcrest.Matchers.hasSize;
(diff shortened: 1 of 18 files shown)


corinaminer (commented):
> Reviewed 17 of 18 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8363)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[`projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathUniqueNextHopTest.java` line 101 at r2](https://reviewable.io/reviews/batfish/batfish/8363#-N4Yl9ku9mW0vTdjc7rX:-N4Yl9ku9mW0vTdjc7rY:b5zeani) ([raw file](https://github.com/batfish/batfish/blob/b7f67bb167246042b85ead2d387c291277457854/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathUniqueNextHopTest.java#L101)):*
> ```Java
> 
>     // This should be 2, but this is separate issue from add-path
>     assertThat(ribs.get("as2rr", DEFAULT_VRF_NAME).getRoutes(PREFIX), hasSize(3));
> ```

is the reason it should be 2 because only one of the routes with the same next hop should go in the main RIB?

I think the test should check for 3 routes in the route reflector's BGP RIB rather than main RIB, especially since the main RIB shouldn't have 3 routes, but also on principle because the RR exports from its BGP RIB and this test cares about what routes are advertised to the leaf more than what's in the RR's main RIB.

corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8363)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

___
*[`projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathUniqueNextHopTest.java` line 101 at r2](https://reviewable.io/reviews/batfish/batfish/8363#-N4Yl9ku9mW0vTdjc7rX:-N4Yvchu6SJCDTksyzpV:bxigdv8) ([raw file](https://github.com/batfish/batfish/blob/b7f67bb167246042b85ead2d387c291277457854/projects/batfish/src/test/java/org/batfish/dataplane/ibdp/BgpAddPathUniqueNextHopTest.java#L101)):*
> testBgpRib does test that 3 routes are in as2rr's BGP RIB.

Oops missed that. Ok lgtm.

PR #8377, opened 2022-06-17
Junos: Define a structure type for BGP neighbor

ratulm:
Helps pull out the neighbor definitions in the config file. 

(If this makes sense, will do other vendors in separate PRs.)


arifogel (commented):
> Reviewed 6 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8377)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @ratulm)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureUsage.java` line 17 at r2](https://reviewable.io/reviews/batfish/batfish/8377#-N4nJ-opCvswlX8VBEif:-N4nJ-opCvswlX8VBEig:b-6h6l0x) ([raw file](https://github.com/batfish/batfish/blob/88a0fc862274a5fa7c5d493b8142113957fca9f2/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureUsage.java#L17)):*
> ```Java
>   BGP_IMPORT_POLICY("bgp import policy-statement"),
>   BGP_NEIGHBOR("bgp group neighbor"),
>   BGP_NEIGHBOR_SELF_REFERENCE("bgp neighbor"),
> ```

Please update the text to something like `bgp neighbor self ref`. As is, makes it hard to tell which is the def and which is the ref in test output (both read as "bgp neighbor" currently).

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8377)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

Issue #8388, opened 2022-06-21
FortiOS: ACLs that use IPv6 (protocol ICMP6) should be unsupported


dhalperi:
Consider this service:

```
config firewall service custom
    edit "ping6"
        set protocol ICMP6
        set icmptype 128
    next
end
config firewall policy
    edit 5
        set srcintf "all"
        set dstintf "all"
        set srcaddr "all"
        set dstaddr "all"
        set action accept
        set service "ping6"
    next
end
```

This firewall rule should be unsupported, not turn into permit all (ipProtocol=58, icmpType=128).

PR #8392, opened 2022-06-21
finalizeConfiguration: check invariants on converted ACLs

dhalperi:
We have had a number of issues (see batfish/batfish#8388 for example)
where an ACL checks a protocol-specific property but does not actually
restrict to packets of the correct protocol. This can lead to strange
behavior, such as allowing flows that have ICMP Type 128 but are not
ICMP packets - mainly in symbolic tools like reachability.

Check converted ACLs when assertions are enabled. Note that we only check ACLs,
rather than individual MatchExprs -- because the latter can be composed and
it's difficult to tell when we're at an abstraction boundary when these
invariants should be enforced.

ASA's object groups are not that well formed, so skip this validation
on ASA for now.

commit-id:aca59b81

**Issue references**: 

 - batfish/batfish#8388

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/util/AclPacketMatchValidityChecker.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/util/AclPacketMatchValidityChecker.java
@@ -1,0 +1,92 @@
+package org.batfish.common.bdd.util;
+
+import com.google.common.annotations.VisibleForTesting;
+import javax.annotation.Nonnull;
+import net.sf.javabdd.BDD;
+import org.batfish.common.bdd.BDDPacket;
+import org.batfish.common.bdd.BDDSourceManager;
+import org.batfish.common.bdd.IpAccessListToBdd;
+import org.batfish.common.bdd.IpAccessListToBddImpl;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.IpAccessList;
+import org.batfish.datamodel.IpProtocol;
+
+/** Checks ACLs for validity with respect to which IPv4 variables they test. */
+public final class AclPacketMatchValidityChecker {
+  public static AclPacketMatchValidityChecker checkerFor(Configuration c) {
+    BDDPacket pkt = new BDDPacket();
+    IpAccessListToBdd toBdd =
+        new IpAccessListToBddImpl(
+            pkt,
+            BDDSourceManager.forInterfaces(pkt, c.getAllInterfaces().keySet()),
+            c.getIpAccessLists(),
+            c.getIpSpaces());
+    return new AclPacketMatchValidityChecker(pkt, toBdd);
+  }
+
+  /**
+   * Checks that the given BDD obeys packet invariants about which fields can be tested
+   * independently vs must be tested together (e.g., ICMP Code should only be tested if this is an
+   * ICMP packet).
+   */
+  public boolean check(IpAccessList acl) {
+    BDD aclBdd = _toBdd.toBdd(acl);
+    return aclMeetsPacketMatchInvariants(aclBdd);
+  }
+
+  @VisibleForTesting
+  boolean aclMeetsPacketMatchInvariants(BDD aclBdd) {
+    // TCP Flags should only be set for TCP packets
+    BDD notTcp = aclBdd.and(_notTcp);
+    boolean tcpOk = !notTcp.testsVars(_pkt.getTcpFlagsVars());
+    notTcp.free();
+    if (!tcpOk) {
+      return false;
+    }
+
+    // ICMP type/code should only be set for ICMP packets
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8392)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java` line 77 at r1](https://reviewable.io/reviews/batfish/batfish/8392#-N9XtExN1HocWJT44vU4:-N9YIXbN3SpQLuCPfvme:bsgb7ua) ([raw file](https://github.com/batfish/batfish/blob/8c637fd4422834fb83d2c1e1bad7268b0f8a9d91/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java#L77)):*

seems like this is going to be called many times, and each call will recreate a bunch of common BDDs. likely better to factor out a class and store those common BDDs as fields. this would also allow you to make fewer `testVars` calls.
</blockquote></details>

General: this is not code that is worth optimizing, especially if it makes the code less readable. (It's run node-parallel during init snapshot only, low memory use, and it's assert only).

But I did make most of the changes.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java` line 80 at r1](https://reviewable.io/reviews/batfish/batfish/8392#-N9XsVqK2HXrJ5H0XfNs:-N9YITtX6ZbX8UpQMiXU:benxqsg) ([raw file](https://github.com/batfish/batfish/blob/8c637fd4422834fb83d2c1e1bad7268b0f8a9d91/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java#L80)):*

could use lessEq here
</blockquote></details>

Not relevant after changes.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java` line 81 at r1](https://reviewable.io/reviews/batfish/batfish/8392#-N9XryNjB1kjGZknnqGC:-N9YIRny5z-wW4juLCuv:b-5a3hx3) ([raw file](https://github.com/batfish/batfish/blob/8c637fd4422834fb83d2c1e1bad7268b0f8a9d91/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java#L81)):*

this can be done in a single call to `testsVars` -- is there a good reason not to do that?
</blockquote></details>

Add a new field in BDDPacket to handle.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8392)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java` line 77 at r1](https://reviewable.io/reviews/batfish/batfish/8392#-N9XtExN1HocWJT44vU4:-N9YJ5wQDLrZE6JZREj3:b-pld8vn) ([raw file](https://github.com/batfish/batfish/blob/8c637fd4422834fb83d2c1e1bad7268b0f8a9d91/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDUtils.java#L77)):*

General: this is not code that is worth optimizing, especially if it makes the code less readable. (It's run node-parallel during init snapshot only, low memory use, and it's assert only).

But I did make most of the changes.
</blockquote></details>

Correction - it's not all assert only.

anothermattbrown (commented):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8392)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8392)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/util/AclPacketMatchValidityChecker.java` line 34 at r3](https://reviewable.io/reviews/batfish/batfish/8392#-N9YkX3_-3GQNk5kzP3T:-N9YkX3aD6HZ83fW49Gc:bjxnh4a) ([raw file](https://github.com/batfish/batfish/blob/71c5ebef6e4b0a0163768d378a5d521ac88683ea/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/util/AclPacketMatchValidityChecker.java#L34)):*
> ```Java
>   public boolean check(IpAccessList acl) {
>     BDD aclBdd = _toBdd.toBdd(acl);
>     return aclMeetsPacketMatchInvariants(aclBdd);
> ```

FYI BDD reuse of some kind, likely related to `AclAclLine`.

PR #8398, opened 2022-06-23
JunOS: remove IPv6 options from ip_protocol

dhalperi:
ip_protocol is used for identifying IPv4 protocols in family inet
firewall-filter.

The same rule is abused for family inet6 as well as family inet, but fixing
that is a bigger journey I'm punting for now.


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8398)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8411, opened 2022-06-29
BDDInteger: make conversion methods from Ip/Prefix/value to BDD abstract

anothermattbrown:
The MutableBDDInteger has the previous implementation based on orAll and andAll. This makes sense since the bit BDDs can be arbitrarily large. The ImmutableBDDInteger implementation simply builds the result bottom up, which is more efficient since the bit BDDs are single variables and the order is guaranteed.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8411)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #8413, opened 2022-07-05
BDDFactory: introduce andLiterals operation

anothermattbrown:
Introduce a new operation that computes the logical AND of some number of literals (constraints on exactly one variable -- i.e. the variable must be true or must be false), and use it in `ImmutableBDDInteger#firstBitsEqual`. 

The `andLiterals` implementation does the same bottom-up construction we were doing in `firstBitsEqual`, but without caching in the operator cache. This avoids evicting more valuable entries (i.e. expensive operations) for these low-value ones (adding a node to the root of a BDD is very cheap).

We could consider caching the entire `andLiterals` operations in the multi-op cache, but that would be redundant with the cache in IpSpaceToBDD.


dhalperi (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8413#-:-N6FzMiy0hXB53dHDQAm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #8414, opened 2022-07-06
IOS: Enable OSPF for interfaces defined after router OSPF

ratulm:
Before: While parsing `router ospf` stanza, the set of OSPF networks were being computed based on the interfaces defined at the time. This meant that any interfaces defined after the OSPF stanza were ignored. This PR changes that behavior. 

A user reported that when they defined interface after router OSPF, as in append to the original config, OSPF is enabled on that interface. This behavior is consistent with what Batfish does for other vendors. 

No other change in this PR.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8414)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8419, opened 2022-07-07
Differential BGP RIB: Make receivedFromIp a key

corinaminer:
Makes `receivedFromIp` part of the `RouteRowSecondaryKey` for differential BGP RIB, so that if there is a similar route across two snapshots with different `receivedFromIp`, they will be considered different routes for diffing purposes. Paving the way for a received path ID column, which will also be a key.

Also added `receivedFromIp` to the comparator that sorts BGP RIB rows. The sort order is now: node, VRF, network, next hop, receivedFromIp. Once the received path ID column is added, that will go into the sort order after receivedFromIp.

Because `receivedFromIp` isn't an attribute of main RIB routes (and path ID won't be either), made `RouteRowSecondaryKey` abstract with separate implementations to represent a main RIB key and a BGP RIB key.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/BgpRoute.java
@@ -27,7 +27,6 @@ import org.batfish.datamodel.bgp.community.ExtendedCommunity;
 import org.batfish.datamodel.bgp.community.StandardCommunity;
 import org.batfish.datamodel.route.nh.NextHop;
 import org.batfish.datamodel.routing_policy.communities.CommunitySet;
-import org.batfish.datamodel.visitors.LegacyReceivedFromToIpConverter;
 
 /** A generic BGP route containing the common properties among different types of BGP routes */
 @ParametersAreNonnullByDefault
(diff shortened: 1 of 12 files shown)


corinaminer (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8419)* status: 4 of 12 files reviewed, all discussions resolved (waiting on @arifogel)

___
*[`projects/question/src/main/java/org/batfish/question/routes/RouteRowSecondaryKey.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/8419#-N6PZ3MM1reTEpCyj0-h:-N6PkF4WDlBqIqJ8xqgK:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0acfcff74172a52e2c6ff7260a2977ac62066fa1/projects/question/src/main/java/org/batfish/question/routes/RouteRowSecondaryKey.java#L22)):*

These getters should probably be `final`
</blockquote></details>

done

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8419)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @corinaminer)

PR #8421, opened 2022-07-08
Juniper: fix firewall filter VSIDs

sfraint:
In Juniper VS -> VI conversion, `VendorStructureIds` are created for each firewall filter term.

This PR just adds definitions for firewall filter terms for zone inbound filters, so the VSIDs no longer point to non-existent structure definitions.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8421)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8437, opened 2022-07-26
Clean up path ID column name

corinaminer:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8437)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @corinaminer and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8437#-N7vrPtf5w5_UEtZHNeC:-N7vrPtf5w5_UEtZHNeD:bczcql6) (no related file):*
Let's not do this; this code has been released, appears in notebooks, etc. It's also no longer consistent with `Originator_Id` (which maybe is how it was introduced?).

___

corinaminer:
Leaving as-is to prevent churn, and anyway it's consistent with `Originator_Id`.

PR #8439, opened 2022-07-27
JunOS: add app junos-rdp, remove invalid test for app junos-host

dhalperi:
Without junos-rdp in the list, it was lexing as NAME and resulting
in an undefined reference.

This fix also caught the improper use of junos-host, which is not
an application but rather a zone.

commit-id:e99ff3ec


sfraint (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8439)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel and @corinaminer)

Issue #8335, opened 2022-05-25
[Cisco ASA] The nat syntax is described outside the object network.


Katsuya414:
Config syntax can be processed if the nat syntax is described inside the object network, but not if the nat syntax is described outside the object network.

object network INSIDE_LOCAL
 host 10.0.12.1
!
object network INSIDE_GLOBAL
  host 10.0.34.1
!
object network OUTSIDE_LOCAL
 host 10.0.45.5
!
object network OUTSIDE_GLOBAL
 host 10.0.23.1
!
nat (inside,outside) source static INSIDE_LOCAL INSIDE_GLOBAL destination static OUTSIDE_GLOBAL OUTSIDE_LOCAL

Also, here is a Traceroute run on Batfish
[ASA_NAT.zip](https://github.com/batfish/batfish/files/8770977/ASA_NAT.zip)
Here is the config of the device actually used
[ASA_NAT_configs.zip](https://github.com/batfish/batfish/files/8771000/ASA_NAT_configs.zip)

dhalperi:
@anothermattbrown - do you understand / can you quickly refresh on https://github.com/batfish/batfish/issues/3005 ?

dhalperi:
cc: @corinaminer

dhalperi:
This issue is a duplicate of #3005. No PR will be considered that does not solve a substantial fraction of the issues in #3005.

Especially including correctly handling multiple NAT rules that may conflict if not properly matched against original flow.

And note that this needs to work with *multiple chained firewalls*. That is, it cannot be "original flow" in a traceroute, but only "original flow" at the current firewall.

Issue #3005, opened 2019-01-23
Support Cisco ASA NAT


mkremerbbn:
I am working on vendor-specific support for Cisco ASA NAT. However, there are some compatibility issues between ASA NATs and the Batfish vendor-independent data plane.

First, some background. ASA NATs optionally specify a real and a mapped interface, which
correspond to "inside" and "outside" interfaces. NAT rules can match both source and
destination and transform both source and destination. Only one NAT rule applies for any given
packet.

1) ASA NAT's can optionally circumvent routing for packets that match transformation. For
   example, if a packet would otherwise be routed to 'DMZ' but matches a rule which specifies
   'outside' as the mapped interface, the packet would be diverted to 'outside'. Similarly, an
   ASA could have zero non-connected routes and rely completely on NAT. "Identity NAT" does not
   transform packets but can be used to impose policy routing. Identity NAT rules which specify
   both real and mapped interfaces can optionally require route lookups anyway, which can be
   useful for management traffic.
2) Both inbound and outbound ACLs match against packets using the original source address and
   the post-NAT destination address.
3) When the NAT rule does not specify a mapped interface, the routing table is used.
   The routing lookup uses the original destination address, not the post-NAT address.
4) The NAT rules match based on the original source and destination.

Not necessarily in chronological order, the ASA packet flow when considering twice NATs:
- Receive packet with original source and original destination
- Choose a single source and destination transformation based on original source and original
  destination
- Check inbound ACL using original source and transformed destination
- Route based on original destination or chosen transformation
- Check outbound ACL using original source and transformed destination
- Transmit packet with transformed source and destination

Consider two NAT rules:
- match source=1 and destination=2 and transform into source=3 and destination=4
- match source=1 and destination=5 and transform into source=6 and destination=4
These rules do not match the same packets and therefore do not require an order of precedence.
The ASA will pick the correct transformation and route it based on the rule or original
destination.
If ASA NAT transformations were implemented first as a destination transformation and a subsequent source transformation (e.g. at ingress and egress respectively), the subsequent transformation would not be able to discern which rule applies.
There's no single good place to implement ASA NAT transformations since they are fully entangled with routing and ACLs.

In discussion with @anothermattbrown, we have the following proposal:
- For ASA, do not use the usual inbound or outbound ACLs.
- Add a "filtering" Transformation
- Route packets first and then apply egress transformations which combine inbound ACLs, NAT, and outbound ACLs.
- There are two cases for routing ASA packets
[...]

PR #8459, opened 2022-08-10
Cisco-ASA: Support for Twice NAT

NobutakaNiiya:
https://github.com/batfish/batfish/issues/8335 (https://github.com/batfish/batfish/issues/8335)
PR for the above issues.

ttps://wwwin-github.cisco.com/MoBillsPoC/batfish/blob/master/projects/batfish/src/test/resources/org/batfish/grammar/cisco_asa
/testconfigs/asa-nat-twice-static (https://wwwin-github.cisco.com/MoBillsPoC/batfish/blob/master/projects/batfish/src/test/resources/org/batfish/grammar/cisco_asa/testconfigs/asa-nattwice-static)
nat (inside,any) source static undef-source-real source-mapped-group inactive
I deleted the above config because it was rejected even by ASA config.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/Flow.java
@@ -18,6 +18,7 @@ public final class Flow implements Comparable<Flow>, Serializable {
   public static class Builder {
     private int _dscp;
     private Ip _dstIp;
+    private Ip _dstOrgIp;
     @Nullable private Integer _dstPort;
     private int _ecn;
     private int _fragmentOffset;
@@ -43,6 +44,7 @@ public final class Flow implements Comparable<Flow>, Serializable {
     private Builder() {
       _dscp = 0;
       _dstIp = Ip.ZERO;
+      _dstOrgIp = Ip.ZERO;
       _ecn = 0;
       _fragmentOffset = 0;
       _ipProtocol = null;
@@ -58,6 +60,7 @@ public final class Flow implements Comparable<Flow>, Serializable {
       _ingressVrf = flow._ingressVrf;
       _srcIp = flow._srcIp;
       _dstIp = flow._dstIp;
+      _dstOrgIp = flow._dstOrgIp;
       _srcPort = flow._srcPort;
       _dstPort = flow._dstPort;
       _ipProtocol = flow._ipProtocol;
@@ -114,6 +117,7 @@ public final class Flow implements Comparable<Flow>, Serializable {
           _ingressVrf,
           _srcIp,
           _dstIp,
+          _dstOrgIp,
           srcPort,
           dstPort,
           firstNonNull(_ipProtocol, IpProtocol.fromNumber(0)),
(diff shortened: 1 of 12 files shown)


dhalperi:
Changes to the VI model need to come with design docs explaining how they work. Can you share one? I can't understand the reasoning here.

As is, this change also looks incredibly incomplete: for example, I see no change anywhere in `FlowTracer` or in `BDDReachabilityAnalysis` to handle the new field. How do you intend to actually affect the dataflow in the network?

NobutakaNiiya:
@dhalperi 
I will create design documents, can you give me a format to use as a reference?

NobutakaNiiya:
[CiscoAsaTwiceNAT.pptx](https://github.com/batfish/batfish/files/9422840/CiscoAsaTwiceNAT.pptx)

@dhalperi 
I created a design document.
Could you please confirm?

dhalperi (commented):
Hi @nobutakaniiya - I'm having a hard time how this new field is taken up. The reason I asked about `FlowTracer` and `BDDReachabilityAnalysis` is that those are what implement Traceroute and Reachability analysis, respectively.

Can you provide a snapshot where reachability finds a result and traceroute realizes it, where it is only happening based on the new ASA specific behavior you're trying to add?

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8459#-:-NAQ-HR_8ZKLCd2kFX0V:b-1x7mo6)* status: 0 of 12 files reviewed, all discussions resolved

NobutakaNiiya:
Hi, @dhalperi - I provide a snapshot.
Please confirm.
[Snapshot.zip](https://github.com/batfish/batfish/files/9442473/Snapshot.zip)

NobutakaNiiya:
Hi @dhalperi , I provide Traceroute and Reachability run logs.
Please confirm.
[Snapshot_20220830.zip](https://github.com/batfish/batfish/files/9450442/Snapshot_20220830.zip)

dhalperi:
Hi, I can confirm that the snapshot DOES NOT handle the cases we are concerned about in #3005.

For example, one of the core cases we are worried about is this one:

> Consider two NAT rules:
> 
> 1. match source=1 and destination=2 and transform into source=3 and destination=4
> 2. match source=1 and destination=5 and transform into source=6 and destination=4
> These rules do not match the same packets and therefore do not require an order of precedence. The ASA will pick the correct transformation and route it based on the rule or original destination. If ASA NAT transformations were implemented first as a destination transformation and a subsequent source transformation (e.g. at ingress and egress respectively), the subsequent transformation would not be able to discern which rule applies. There's no single good place to implement ASA NAT transformations since they are fully entangled with routing and ACLs.

Your snapshot only has a single NAT rule on `dev3`. So this does not exhibit this bug!

dhalperi:
Please let us when you are ready with a solution to #3005 that meets the requirements in there, not just your one snapshot.

Thanks, and we look forward to this great contribution!

NobutakaNiiya:
[Snapshot_0920.zip](https://github.com/batfish/batfish/files/9703297/Snapshot_0920.zip)
I have attached the results of testing with multiple NAT rules and multiple chained firewalls.
Please check the attached file.

PR #8472, opened 2022-08-16
PrefixTrieMultiMap: reduce allocations in hotpath

dhalperi:
PrefixTrieMultiMap#longestPrefixMatch(ip, int) is used on the hotpath during dataplane,
and it immediately converts its arguments into a Prefix. That involves allocation and
locking (interning prefixes), that is essentially unused later - the two components
of the Prefix are decoupled in trie traversal.

Change the relevant internal APIs to work on the component parts, saving allocations.

commit-id:6a7eec28


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8472)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8482, opened 2022-08-18
BDD benchmarks, stress tests, and some simple optimizations

anothermattbrown:
Added new benchmarks, notably some that use real data (building the reachability graph for an input snapshot, and converting all ACLs to BDDs). Did some simple optimizations based profiling on the ACL benchmark, resulting in a 1.5x speedup on one large network.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDOps.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDOps.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.collect.Lists;
 import java.io.Serializable;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -25,7 +26,7 @@ public final class BDDOps implements Serializable {
     return _factory.andAll(conjuncts);
   }
 
-  public @Nonnull BDD and(Iterable<BDD> conjuncts) {
+  public @Nonnull BDD and(Collection<BDD> conjuncts) {
     return _factory.andAll(conjuncts);
   }
 
@@ -33,7 +34,7 @@ public final class BDDOps implements Serializable {
     return _factory.orAll(disjuncts);
   }
 
-  public @Nonnull BDD or(Iterable<BDD> disjuncts) {
+  public @Nonnull BDD or(Collection<BDD> disjuncts) {
     return _factory.orAll(disjuncts);
   }
 
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java
@@ -182,69 +182,61 @@ public final class HeaderSpaceToBDD {
         toBDD(tcpFlags.getUseAck(), tcpFlags.getTcpFlags().getAck(), _bddPacket.getTcpAck()));
   }
 
+  private static void addIfNonNull(List<BDD> bdds, @Nullable BDD bdd) {
+    if (bdd != null) {
+      bdds.add(bdd);
+    }
+  }
+
   public BDD toBDD(HeaderSpace headerSpace) {
     // HeaderSpace has null fields to mean "unconstrained". Thus we must be careful to treat these
     // as correctly "everything" when positive or "nothing" when negated.
-    List<BDD> conjuncts = new ArrayList<>();
+    List<BDD> bdds = new ArrayList<>();
 
-    BDD bdd = toBDD(headerSpace.getPacketLengths(), _bddPacket.getPacketLength());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotPacketLengths(), _bddPacket.getPacketLength()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getFragmentOffsets(), _bddPacket.getFragmentOffset());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd =
-        negateIfNonNull(toBDD(headerSpace.getNotFragmentOffsets(), _bddPacket.getFragmentOffset()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getEcns(), _bddPacket.getEcn());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotEcns(), _bddPacket.getEcn()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getDscps(), _bddPacket.getDscp());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotDscps(), _bddPacket.getDscp()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getTcpFlags());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getIcmpTypes(), _bddPacket.getIcmpType());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotIcmpTypes(), _bddPacket.getIcmpType()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getIcmpCodes(), _bddPacket.getIcmpCode());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotIcmpCodes(), _bddPacket.getIcmpCode()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getIpProtocols());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotIpProtocols()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = toBDD(headerSpace.getSrcPorts(), _bddPacket.getSrcPort());
-    if (bdd != null) conjuncts.add(bdd);
-    bdd = negateIfNonNull(toBDD(headerSpace.getNotSrcPorts(), _bddPacket.getSrcPort()));
-    if (bdd != null) conjuncts.add(bdd);
-    bdd =
+    addIfNonNull(bdds, toBDD(headerSpace.getPacketLengths(), _bddPacket.getPacketLength()));
+    addIfNonNull(
+        bdds,
+        negateIfNonNull(toBDD(headerSpace.getNotPacketLengths(), _bddPacket.getPacketLength())));
+    addIfNonNull(bdds, toBDD(headerSpace.getFragmentOffsets(), _bddPacket.getFragmentOffset()));
+
+    addIfNonNull(
+        bdds,
+        negateIfNonNull(
+            toBDD(headerSpace.getNotFragmentOffsets(), _bddPacket.getFragmentOffset())));
+    addIfNonNull(bdds, toBDD(headerSpace.getEcns(), _bddPacket.getEcn()));
+    addIfNonNull(bdds, negateIfNonNull(toBDD(headerSpace.getNotEcns(), _bddPacket.getEcn())));
+    addIfNonNull(bdds, toBDD(headerSpace.getDscps(), _bddPacket.getDscp()));
+    addIfNonNull(bdds, negateIfNonNull(toBDD(headerSpace.getNotDscps(), _bddPacket.getDscp())));
+    addIfNonNull(bdds, toBDD(headerSpace.getTcpFlags()));
+    addIfNonNull(bdds, toBDD(headerSpace.getIcmpTypes(), _bddPacket.getIcmpType()));
+    addIfNonNull(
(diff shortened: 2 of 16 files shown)


dhalperi (commented):
> Reviewed 2 of 2 files at r1, 6 of 6 files at r2, 2 of 2 files at r3, 5 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8482)* status: all files reviewed, 6 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java` line 92 at r4](https://reviewable.io/reviews/batfish/batfish/8482#-N9ilrWL1wnsTsh8PMga:-N9ilrWL1wnsTsh8PMgb:b8s630f) ([raw file](https://github.com/batfish/batfish/blob/23973e97feaf7c2dcb2e1c4da49abb48a8cec00f/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L92)):*
> ```Java
>       StateExpr dirtyState = dirtyStates.remove();
>       visitCounts.compute(dirtyState, (unused, oldCount) -> oldCount == null ? 1 : oldCount + 1);
>       LinkedList<BDD> inputs = (LinkedList<BDD>) dirtyInputs.removeAll(dirtyState);
> ```

This is not guaranteed by the API, is it?

I do not even believe the result is guaranteed to be safely modifiable.

e.g., this is what I see in the source code for Guava 31.1-jre

```
  /**
   * {@inheritDoc}
   *
   * <p>The returned list is immutable and implements {@link java.util.RandomAccess}.
   */
  @CanIgnoreReturnValue
  @Override
  public List<V> removeAll(@Nullable Object key) {
    /*
     * Safe because all we do is remove values for the key, not add them. (If we wanted to make sure
     * to call getCopy and removeAllNodes only with a true K, then we could check containsKey first.
     * But that check wouldn't eliminate the warnings.)
     */
    @SuppressWarnings({"unchecked", "nullness"})
    K castKey = (K) key;
    List<V> oldValues = getCopy(castKey);
    removeAllNodes(castKey);
    return oldValues;
  }

  private List<V> getCopy(@ParametricNullness K key) {
    return unmodifiableList(Lists.newArrayList(new ValueForKeyIterator(key)));
  }
```



_Code quote:_
```Java
(LinkedList<BDD>)
```
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 258 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifKmU0TtnF78Q4xAi:-N9ifKmU0TtnF78Q4xAj:b-214cc9) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L258)):*
> ```Java
> 
>     Path snapshotDir = Paths.get(dir);
>     checkArgument(snapshotDir.toFile().exists(), "%s does not exist.", dir);
> ```

also check isDirectory()?



_Code quote:_
```Java
snapshotDir.toFile().exists()
```
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 262 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifZe10lHCfPfNcG7e:-N9ifZe25rQR2-oyuZNH:b-kcuu9w) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L262)):*
> ```Java
>     // layer 1 topology
>     try {
>       Path l1TopologyPath = snapshotDir.resolve("batfish/layer1_topology.json");
> ```

It is usually better to resolve in two steps, not relying on `/` - which a different FS could actually treat as part of the filename.
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 267 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifSOIDT8K2NGBwjVR:-N9ifSOIDT8K2NGBwjVS:bcnro8b) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L267)):*
> ```Java
>       }
>     } catch (Exception e) {
>       e.printStackTrace();
> ```

Why not actually throw this? and below?



_Code quote:_
```Java
e.printStackTrace();
```
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java` line 191 at r3](https://reviewable.io/reviews/batfish/batfish/8482#-N9ikU0C66n8OPepNKQQ:-N9ikU0C66n8OPepNKQR:buwgdp6) ([raw file](https://github.com/batfish/batfish/blob/6d471cc6524de2045288e363d8fb45b33dd5d7a9/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java#L191)):*
> ```Java
> 
>     BDD bdd = toBDD(headerSpace.getPacketLengths(), _bddPacket.getPacketLength());
>     if (bdd != null) conjuncts.add(bdd);
> ```

This is an optimization that hurts readability quite a lot. Some suggestions:

1. Leave the code the same and just swap the stream out? It's two for loops, likely the same overhead just amortized differently (and maybe more optimizable).

2. Introduce a helper like `addIfNotNull` and use that instead of the `if (...)` that is missing braces.

3. (what we probably should do) add MatchX operators since most MatchHeaderSpace are just matching on one field. Would be easy to do this incrementally for the most important X.



_Code quote:_
```Java
    BDD bdd = toBDD(headerSpace.getPacketLengths(), _bddPacket.getPacketLength());
    if (bdd != null) conjuncts.add(bdd);
```
___
*[`projects/bdd/src/main/java/net/sf/javabdd/BDDFactory.java` line 319 at r4](https://reviewable.io/reviews/batfish/batfish/8482#-N9imgYb2XslO29oY3V-:-N9imgYb2XslO29oY3V0:b-kt9b7o) ([raw file](https://github.com/batfish/batfish/blob/23973e97feaf7c2dcb2e1c4da49abb48a8cec00f/projects/bdd/src/main/java/net/sf/javabdd/BDDFactory.java#L319)):*
> ```Java
>    */
>   public final BDD andAll(BDD... bddOperands) {
>     return andAll(Arrays.asList(bddOperands), false);
> ```

Seems a lot more sensible to make the other functions go to this one now, right? Otherwise they go right back to array.



_Code quote:_
```Java
andAll(Arrays.asList(bddOperands), false);
```

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8482)* status: 11 of 16 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java` line 92 at r4](https://reviewable.io/reviews/batfish/batfish/8482#-N9ilrWL1wnsTsh8PMga:-N9j7_SjC5K4In78FvaO:b-od5ue9) ([raw file](https://github.com/batfish/batfish/blob/23973e97feaf7c2dcb2e1c4da49abb48a8cec00f/projects/batfish/src/main/java/org/batfish/bddreachability/BDDReachabilityUtils.java#L92)):*

This is not guaranteed by the API, is it?

I do not even believe the result is guaranteed to be safely modifiable.

e.g., this is what I see in the source code for Guava 31.1-jre

```
  /**
   * {@inheritDoc}
   *
   * <p>The returned list is immutable and implements {@link java.util.RandomAccess}.
   */
  @CanIgnoreReturnValue
  @Override
  public List<V> removeAll(@Nullable Object key) {
    /*
     * Safe because all we do is remove values for the key, not add them. (If we wanted to make sure
     * to call getCopy and removeAllNodes only with a true K, then we could check containsKey first.
     * But that check wouldn't eliminate the warnings.)
     */
    @SuppressWarnings({"unchecked", "nullness"})
    K castKey = (K) key;
    List<V> oldValues = getCopy(castKey);
    removeAllNodes(castKey);
    return oldValues;
  }

  private List<V> getCopy(@ParametricNullness K key) {
    return unmodifiableList(Lists.newArrayList(new ValueForKeyIterator(key)));
  }
```
</blockquote></details>

nope, rewritten. For now, I'm making a copy. Plan to revisit soon.
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 258 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifKmU0TtnF78Q4xAi:-N9j9hK_AJTUiuvIAxh9:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L258)):*

also check isDirectory()?
</blockquote></details>

done
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 262 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifZe10lHCfPfNcG7e:-N9j9iZy-HOWOErSurob:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L262)):*

It is usually better to resolve in two steps, not relying on `/` - which a different FS could actually treat as part of the filename.
</blockquote></details>

done
___
*[`projects/batfish/src/test/java/org/batfish/main/TestrigText.java` line 267 at r2](https://reviewable.io/reviews/batfish/batfish/8482#-N9ifSOIDT8K2NGBwjVR:-N9jB5jG6krL7x4FRtdk:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/af2cc3a998ed4f89f8b72896fec6dacf9136720b/projects/batfish/src/test/java/org/batfish/main/TestrigText.java#L267)):*

Why not actually throw this? and below?
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java` line 191 at r3](https://reviewable.io/reviews/batfish/batfish/8482#-N9ikU0C66n8OPepNKQQ:-N9jB7ve86HvBWhfFpYu:bclal3j) ([raw file](https://github.com/batfish/batfish/blob/6d471cc6524de2045288e363d8fb45b33dd5d7a9/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java#L191)):*

This is an optimization that hurts readability quite a lot. Some suggestions:

1. Leave the code the same and just swap the stream out? It's two for loops, likely the same overhead just amortized differently (and maybe more optimizable).

2. Introduce a helper like `addIfNotNull` and use that instead of the `if (...)` that is missing braces.

3. (what we probably should do) add MatchX operators since most MatchHeaderSpace are just matching on one field. Would be easy to do this incrementally for the most important X.
</blockquote></details>

went with 2.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8482)* status: 11 of 16 files reviewed, 2 unresolved discussions (waiting on @dhalperi)

___
*[`projects/bdd/src/main/java/net/sf/javabdd/BDDFactory.java` line 319 at r4](https://reviewable.io/reviews/batfish/batfish/8482#-N9imgYb2XslO29oY3V-:-N9jDNwb4sc-Ohq7kvdJ:b-fxwb3e) ([raw file](https://github.com/batfish/batfish/blob/23973e97feaf7c2dcb2e1c4da49abb48a8cec00f/projects/bdd/src/main/java/net/sf/javabdd/BDDFactory.java#L319)):*

Seems a lot more sensible to make the other functions go to this one now, right? Otherwise they go right back to array.
</blockquote></details>

The variants delegate to the collection version, which doesn't require making a copy (`Arrays.asList` does not copy the array, just allocates a wrapper). If we standardized on the array version, we would have to make a copy (to `BDD[]`). We can't avoid the conversion to `int[]` that happens in the subclass.

dhalperi (commented):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8482)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java` line 240 at r6](https://reviewable.io/reviews/batfish/batfish/8482#-N9jYBQH64-sFGOC8L3H:-N9jYBQH64-sFGOC8L3I:b-llww52) ([raw file](https://github.com/batfish/batfish/blob/7486654c8574ac2f1beb6a296aca3d84906ebd88/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java#L240)):*
> ```Java
>     addIfNonNull(bdds, negateIfNonNull(toBDD(headerSpace.getNotDstIps(), _dstIpSpaceToBdd)));
>     BDD positiveSpace = _bddFactory.andAll(bdds);
>     return headerSpace.getNegate() ? positiveSpace.not() : positiveSpace;
> ```

should this be `.notEq()` now that that exists?



_Code quote:_
```Java
.not()
```

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8482)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java` line 240 at r6](https://reviewable.io/reviews/batfish/batfish/8482#-N9jYBQH64-sFGOC8L3H:-N9ltvPy93fXl5TdKdFg:b-83lwso) ([raw file](https://github.com/batfish/batfish/blob/7486654c8574ac2f1beb6a296aca3d84906ebd88/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/HeaderSpaceToBDD.java#L240)):*

should this be `.notEq()` now that that exists?
</blockquote></details>

that's coming in the next pr :smile:

PR #8490, opened 2022-08-19
andLiterals: always return a fresh BDD

anothermattbrown:
When called with a single literal, return a copy of it. Otherwise freeing the result is unsafe in general


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8490)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #8515, opened 2022-08-29
common: separate specifier/parboiled code out

dhalperi:
Since the parboiled parser does not play nice with Bazel code coverage,
pull it into its own library. Much like parser_common, I had a huge
challenge with modularity as the existing code doesn't really have
any.

Related work: removing the parser_common files from being compiled
twice, adding the missing dependency, adding missing pmd_test,
and making a "virtual :common" target to ensure that no existing
code has to be changed (even downstream).

commit-id:d4ccf516

---

**Stack**:
- #8518
- #8517
- #8515 ⬅

### DIFF
--- a/projects/batfish-common-protocol/BUILD
+++ b/projects/batfish-common-protocol/BUILD
@@ -4,46 +4,51 @@ load("@batfish//skylark:pmd_test.bzl", "pmd_test")
 
 package(default_visibility = ["//visibility:public"])
 
-filegroup(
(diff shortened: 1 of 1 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8515)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8542, opened 2022-09-13
JUNOS: Add and fix VNI/VxLan properties

jeffkala:
Update and fix vni/vxlan l2 and l3 properties.


jeffkala (commented):
Working on figuring out how/why the testIntefaceArp test is failing, doesn't appear to be anything change in this PR that would cause that to fail now.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542#-:-NBsnB6QDygjYkDWpStz:bo9jt9v)* status: 0 of 36 files reviewed, all discussions resolved

jeffkala (commented):
Got this cleaned up, forgot I removed a snippet of code during testing that I added back.  Everything else should be passing test now.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542#-:-NBtcquX3ueqex-JCnHZ:b-ouyagd)* status: 0 of 36 files reviewed, all discussions resolved (waiting on @arifogel)

dhalperi (commented):
> Reviewed 3 of 25 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 5 unresolved discussions (waiting on @arifogel and @jeffkala)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8542#-NCLrTCa6O7Lzed18DQx:-NCLrTCa6O7Lzed18DQy:bfo5imn) (no related file):*
Hi Jeff, please remove all the `*.orig` files that made it in by accident

___
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3411 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLrn0W168RzTfHm1bO:-NCLrn0W168RzTfHm1bP:b9akv8r) ([raw file](https://github.com/batfish/batfish/blob/99d821e90ec82c9f1c0f483810385a6554f29760/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3411)):*
> ```Java
>   public void enterS_switch_options(S_switch_optionsContext ctx) {
>     todo(ctx);
>   }
> ```

confirming: is everything under `s_switch_options` now fully handled, or are there some rules lower down in the hierarchy we should move this `todo(ctx)` under.



_Code quote:_
```Java
  @Override
  public void enterS_switch_options(S_switch_optionsContext ctx) {
    todo(ctx);
  }
```
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCLsVWS0kSfxWkrwDCf:b-urcb8x) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*
> ```Java
>           .setVrfTargetCommunityorAuto(
>               ExtendedCommunityOrAuto.of(
>                   ExtendedCommunity.parse(ctx.extended_community().getText())));
> ```

Usually, we would introduce a `toExtendedCommunity` rule that takes the `Extended_communityContext` and returns either an `ExtendedCommunity` or an `Optional<ExtendedCommunity>`.

We use the former (not optional) when the lexer will only allow things we know that `ExtendedCommunity#parse` handles. We use the latter (optional) when the lexer is generous, and we use `ExtendedCommunity#tryParse` instead.

Can you:
1. Figure out which alternative to use? (Is the lexer strict or generous?)
2. Introduce the helper?

Btw, if it turns out we should use the optional then we do a lot of (abbrv'ed): 

```
Optional<EC> maybeEc = toEC(ctx.ec);
if (!maybeEc.isPresent()) {
   warn(ctx, ...);
   maybe do some other stuff so inner rules don't crash.
   return;
}
EC ec = maybeEc.get();
handle for real
```



_Code quote:_
```Java
ExtendedCommunity.parse(ctx.extended_community().getText())));
```
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4244 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtLD5BvUtai6vm4TV:-NCLtLD5BvUtai6vm4TW:b-mb9yyf) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4244)):*
> ```Java
>   @Override
>   public void enterB_family(B_familyContext ctx) {
>     _currentBgpGroup.setEvpnAf(false);
> ```

This pattern looks suspicious to me -- what is happening here? What does this boolean mean?
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7085 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtmil74AQGpA2GrWb:-NCLtmil74AQGpA2GrWc:bu2bqyy) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7085)):*
> ```Java
>       Optional<Integer> vlan = toInteger(ctx, ctx.id);
>       vlan.ifPresent(_currentNamedVlan::setVlanId);
>     }
> ```

What do you do if there's a `none` instead? Do you not need to clear it?

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 5 unresolved discussions (waiting on @arifogel, @dhalperi, and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3411 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLrn0W168RzTfHm1bO:-NCQ1tusDi9m20C-gkL5:bely7sj) ([raw file](https://github.com/batfish/batfish/blob/99d821e90ec82c9f1c0f483810385a6554f29760/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3411)):*

confirming: is everything under `s_switch_options` now fully handled, or are there some rules lower down in the hierarchy we should move this `todo(ctx)` under.
</blockquote></details>

I'd say everything that has expectations to need support is covered now.  The remaining commands at that stanza level are things like mac-limits etc. Which I wouldn't see Batfish ever supporting?  I could be wrong. With that being said I'm not overly sure if that means this todo should just be removed completely?
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCQ2Ivn85xfO3usEUx2:b-ya42ao) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*

Usually, we would introduce a `toExtendedCommunity` rule that takes the `Extended_communityContext` and returns either an `ExtendedCommunity` or an `Optional<ExtendedCommunity>`.

We use the former (not optional) when the lexer will only allow things we know that `ExtendedCommunity#parse` handles. We use the latter (optional) when the lexer is generous, and we use `ExtendedCommunity#tryParse` instead.

Can you:
1. Figure out which alternative to use? (Is the lexer strict or generous?)
2. Introduce the helper?

Btw, if it turns out we should use the optional then we do a lot of (abbrv'ed): 

```
Optional<EC> maybeEc = toEC(ctx.ec);
if (!maybeEc.isPresent()) {
   warn(ctx, ...);
   maybe do some other stuff so inner rules don't crash.
   return;
}
EC ec = maybeEc.get();
handle for real
```
</blockquote></details>

Will take a look.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4244 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtLD5BvUtai6vm4TV:-NCQ4My06eSYMw2Szb5z:b-5mcwv4) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4244)):*

This pattern looks suspicious to me -- what is happening here? What does this boolean mean?
</blockquote></details>

Will get back to you on this one, this might have been leftover on my end before Ari helped me figured this out.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 5 unresolved discussions (waiting on @arifogel, @dhalperi, and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3411 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLrn0W168RzTfHm1bO:-NCRpdYI3cf1U_LYyZHv:b-pwlpg5) ([raw file](https://github.com/batfish/batfish/blob/99d821e90ec82c9f1c0f483810385a6554f29760/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3411)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

I'd say everything that has expectations to need support is covered now.  The remaining commands at that stanza level are things like mac-limits etc. Which I wouldn't see Batfish ever supporting?  I could be wrong. With that being said I'm not overly sure if that means this todo should just be removed completely?
</blockquote></details>

think this is accurate relooking at it.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCRsFs95jYBoT3HY9Oo:b-1fiij6) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will take a look.
</blockquote></details>

I implemented the ExtendedCommunityOrAuto which to me should make the lexer strict. I don't see how/what the helper would do in this case. Very well just be a misunderstanding on my part.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 5 unresolved discussions (waiting on @arifogel, @dhalperi, and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7085 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtmil74AQGpA2GrWb:-NCRvFKWAurfrs7jvYa4:b-plqte8) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7085)):*

What do you do if there's a `none` instead? Do you not need to clear it?
</blockquote></details>

grammar supports NONE, not doing anything with it ATM, don't think clearing it is needed at this time.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 4 unresolved discussions (waiting on @arifogel, @dhalperi, and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCRw1_cAcvdQpVHnFXV:b-k84xqv) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

I implemented the ExtendedCommunityOrAuto which to me should make the lexer strict. I don't see how/what the helper would do in this case. Very well just be a misunderstanding on my part.
</blockquote></details>

1. Even if `extended_community` is perfect (more on that later), we want a single `toExtendedCommunity(Extended_communityContext)` function that we use everywhere. Much like `toLong`, it dramatically reduces code reuse (of `EC.parse(ctx.getText())`). And if we have to evolve the parser rule, we don't have to hunt down all the various versions of it. (E.g., in the past we've had to change from `ctx.getText()` to a function that has special logic to handle spaces, or maybe there's a new juniper-specific alternative to `origin:foo:bar` that is not recognized by `ExtendedCommunity.parse`. Better to have it all in one place.

2. Hmm, I'm not sure the lexer/parser are strict. Here's the grammar for  `extended_community`:

```
ec_literal
:
  dec COLON dec COLON dec
;

ec_named
:
  ec_type COLON ec_administrator COLON assigned_number = dec
;

ec_type
:
  ORIGIN
  | TARGET
;

extended_community
:
  ec_literal
  | ec_named
;
```

`dec COLON dec COLON dec` means that `1122334455667788990011223344556677889900:0:0` could be treated as `ec_literal`. But that would crash `ExtendedCommunity.parse` -- too many bits.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7085 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtmil74AQGpA2GrWb:-NCRxLyNBiEd5GJBfOVN:b-uc9903) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7085)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

grammar supports NONE, not doing anything with it ATM, don't think clearing it is needed at this time.
</blockquote></details>

Remember that Batfish is designed to handle incremental changes. That is when a user has their running config at the top, and then appends their interactive CLI commands.

What is the result if a user has this config:

```
set ... vlan-id 7
```

and then types

```
set ... vlan-id none
```

?

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 4 unresolved discussions (waiting on @arifogel, @dhalperi, and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCSe2L84ULRNJqGHK_R:b-hkve9e) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*

1. Even if `extended_community` is perfect (more on that later), we want a single `toExtendedCommunity(Extended_communityContext)` function that we use everywhere. Much like `toLong`, it dramatically reduces code reuse (of `EC.parse(ctx.getText())`). And if we have to evolve the parser rule, we don't have to hunt down all the various versions of it. (E.g., in the past we've had to change from `ctx.getText()` to a function that has special logic to handle spaces, or maybe there's a new juniper-specific alternative to `origin:foo:bar` that is not recognized by `ExtendedCommunity.parse`. Better to have it all in one place.

2. Hmm, I'm not sure the lexer/parser are strict. Here's the grammar for  `extended_community`:

```
ec_literal
:
  dec COLON dec COLON dec
;

ec_named
:
  ec_type COLON ec_administrator COLON assigned_number = dec
;

ec_type
:
  ORIGIN
  | TARGET
;

extended_community
:
  ec_literal
  | ec_named
;
```

`dec COLON dec COLON dec` means that `1122334455667788990011223344556677889900:0:0` could be treated as `ec_literal`. But that would crash `ExtendedCommunity.parse` -- too many bits.
</blockquote></details>

That's fair, I think alot of this is just me still being pretty novice at java and batfish development. What you're saying does indeed make more sense to me now.  I'll give writing the helper a try.  Thanks for the information and patience.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7085 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtmil74AQGpA2GrWb:-NCSeHm78hV6BUhVHRlG:b3r6ny8) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7085)):*

Remember that Batfish is designed to handle incremental changes. That is when a user has their running config at the top, and then appends their interactive CLI commands.

What is the result if a user has this config:

```
set ... vlan-id 7
```

and then types

```
set ... vlan-id none
```

?
</blockquote></details>

That is fair, I forgot about that aspect of it, will take a look at this.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 4 unresolved discussions (waiting on @arifogel and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8542#-NCLrTCa6O7Lzed18DQx:-NCSfgEFDZuPgOcQcsxR:bb6un1j) (no related file):*

Hi Jeff, please remove all the `*.orig` files that made it in by accident
</blockquote></details>

Done

___

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 4 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3441 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLsVWS0kSfxWkrwDCe:-NCa1dHa1x2sJsisM7I3:bvbmj0j) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3441)):*
> Extended_communityContext
Gave it a shot, hopefully this is closer to what your ask was?

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 4 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4244 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtLD5BvUtai6vm4TV:-NCavu0PBEKIuiNgYyKD:b-kuko9d) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4244)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Will get back to you on this one, this might have been leftover on my end before Ari helped me figured this out.
</blockquote></details>

Defaulting the creation of the EVPN address family to False (off) to avoid the memory of building this out when evpn AF is not used in the configuration.  We then build the AF bgp process for these only when necessary.  This was setup with help from Ari previously.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 3 of 36 files reviewed, 3 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7085 at r3](https://reviewable.io/reviews/batfish/batfish/8542#-NCLtmil74AQGpA2GrWb:-NCbC86T9xckHezCVs8u:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/0e877d4d79da440ef1fafa46657cb2a86cda0ea4/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7085)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

That is fair, I forgot about that aspect of it, will take a look at this.
</blockquote></details>

done



_Code quote:_
```Java
    if (ctx.id != null) {
      Optional<Integer> vlan = toInteger(ctx, ctx.id);
      vlan.ifPresent(_currentNamedVlan::setVlanId);
    }
```

jeffkala:
anyway to push this along to get it over the finish line?

arifogel (commented):
> Reviewed 15 of 25 files at r1, 5 of 17 files at r2, 3 of 3 files at r3, 8 of 9 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: all files reviewed, 26 unresolved discussions (waiting on @dhalperi and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2041 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXc77s7bQIWjd8pLg1:-NFXc77s7bQIWjd8pLg2:b3ye2mz) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2041)):*
> ```Java
>   private static @Nonnull ExtendedCommunity toExtendedCommunity(Vt_communityContext ctx) {
>     if (ctx.getText() != null) {
>       return ExtendedCommunity.parse(ctx.getText());
> ```

I'm not convinced the grammar enforces that a `vt_community` nor `extended_community` as guaranteed to be parseable by `ExtendedCommunity.parse`, as `@dhalperi pointed out.

Either:
* update the grammar to make it so (if possible)
OR:
* 
- this should be non-static, return an `Optional<ExtendedCommunity>` via `tryParse`, and callers will need to handle the case where `Optional.empty()` is returned gracefully.

Same goes for your other converter.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2042 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXbaZo8cd89j93okDF:-NFXbaZo8cd89j93okDG:bkrvkz5) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2042)):*
> ```Java
>     if (ctx.getText() != null) {
>       return ExtendedCommunity.parse(ctx.getText());
>     } else {
> ```

This branch should never trigger, as an invariant of `ParserRuleContext`. Just get rid of the if and this branch. Same for your other converter.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3464 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXdZh40WnQyR7rNBWc:-NFXdZh40WnQyR7rNBWd:b-wtlqn4) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3464)):*
> ```Java
>   @Override
>   public void exitSovt_auto(Sovt_autoContext ctx) {
>     if (ctx.AUTO() != null) {
> ```

This is guaranteed to be true by the grammar. Ditch the `if`.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4408 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXgahe12OUNoqGWo1i:-NFXgahe12OUNoqGWo1j:bd2drth) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4408)):*
> ```Java
>   @Override
>   public void enterE_vni_options(E_vni_optionsContext ctx) {
>     _currentVni = toInt(ctx.id);
> ```

This is not safe since `id` is currently a `dec`. You should follow the converter+optional pattern used in e.g. `CiscoNxosControlPlaneExtractor.toInteger(ParserRuleContext messageCtx, Eigrp_asnContext ctx)`

See: https://github.com/batfish/batfish/blob/master/docs/extraction/README.md#validating-and-converting-parse-tree-nodes-with-variable-text
And while you're at it, couldn't hurt to check out our new parsing docs: https://github.com/batfish/batfish/blob/master/docs/parsing/README.md
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4409 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXhi4M0mDqA_2hmtT9:-NFXhi4M0mDqA_2hmtTA:b-hzehtv) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4409)):*
> ```Java
>   public void enterE_vni_options(E_vni_optionsContext ctx) {
>     _currentVni = toInt(ctx.id);
>     assert _currentLogicalSystem.getVniOptions() != null;
> ```

This assertion is unnecessary. You should instead fix the annotation. See comment in `LogicalSystem.java`
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4421 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXjH7QCBJ5I_LgHubC:-NFXjH7QCBJ5I_LgHubD:b9qz90) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4421)):*
> ```Java
>   @Override
>   public void exitEvovt_auto(Evovt_autoContext ctx) {
>     if (ctx.AUTO() != null) {
> ```

This is guaranteed to be true by the grammar. Remove the `if.`
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4424 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXjaYaB3N5KVZr0_3Z:-NFXjaYaB3N5KVZr0_3_:bxh5uek) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4424)):*
> ```Java
>       _currentLogicalSystem
>           .getVniOptions()
>           .get(_currentVni)
> ```

If you make `_currentVni` a `VniOptions` (and rename appropriately), you can simplify all instances of:
```
      _currentLogicalSystem
          .getVniOptions()
          .get(_currentVni)
```
to:
```
_currentVniOptions
```
You should then set `_currentVniOptions = null` for sanity in `exitE_vni_options` (not enterE_vni_options`).
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 6081 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXkFtZDUgOgG9xf2hK:-NFXkFtZDUgOgG9xf2hL:b-c06tpg) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L6081)):*
> ```Java
>   @Override
>   public void enterRo_route_distinguisher_id(Ro_route_distinguisher_idContext ctx) {
>     if (ctx.ip_address() != null) {
> ```

Guaranteed to be true by the grammar. Remove the `if`.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7237 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXkoTV20lvnpizKqG1:-NFXkoTV20lvnpizKqG2:b-bkhzxb) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7237)):*
> ```Java
>   public void exitBf_evpn(Bf_evpnContext ctx) {
>     _currentBgpGroup.setEvpnAf(false);
>     if (ctx.EVPN() != null) {
> ```

This is guaranteed to be true by the grammar, so this whole function is effectively:
```
_currentBgpGroup.setEvpnAf(true);
```
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/ExtendedCommunityOrAuto.java` line 10 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFX_OTZCtLKXH6FA5F0:-NFX_OTZCtLKXH6FA5F1:bc45ai4) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/ExtendedCommunityOrAuto.java#L10)):*
> ```Java
> 
> /** Either {@code auto} or an explicit {@link ExtendedCommunity}. */
> public final class ExtendedCommunityOrAuto implements Serializable {
> ```

Prefer not to use `null` as a sentinel value. I think overall this class has a confusing API, and its content is largely unnecessary.

I'd suggest just changing this from a class to an interface that extends `Serializable`.
Then I'd suggest an `ExtendedCommunityLiteral` and `ExtendedCommunityAuto` class in this package that implement that interface.
The literal class would just have a single field that is an `org.batfish.datamodel.bgp.community.ExtendedCommunity`.
The auto class would be a singleton with a private constructor,  a private static `INSTANCE` field that is  a `new ExtendedCommunityAuto()`, and a public static  `instance()` method that returns `INSTANCE`.

You'll probably need to override hashCode and equals for those classes since you did it here.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 779 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXfAt79BUXAAFk1pcy:-NFXeyba7lnhfXhhLcd-:b-hp2pg) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L779)):*
> ```Java
>       }
>       neighbor.setLocalIp(localIp);
>       if (ig.getEvpnAf() != null) {
> ```

What if it's false?
This should be exactly `Boolean.TRUE.equals(ig.getEvpnAf())`
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 784 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXl8H66MUT_yEsXLaj:-NFXl8H66MUT_yEsXLak:b4rfn7d) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L784)):*
> ```Java
>             .setPropagateUnmatched(true)
>             .setAddressFamilyCapabilities(
>                 AddressFamilyCapabilities.builder().setAllowRemoteAsOut(ALWAYS).build());
> ```

Sanity check: allow remote as out is always true for juniper? Just this address family?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 785 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXlnuQAlDSlHUx8c3R:-NFXlnuQAlDSlHUx8c3S:bsj96c4) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L785)):*
> ```Java
>             .setAddressFamilyCapabilities(
>                 AddressFamilyCapabilities.builder().setAllowRemoteAsOut(ALWAYS).build());
>         evpnAfBuilder.setL3Vnis(convertL3Evpn().build());
> ```

First, `setL3Vnis` cannot accept a null argument (you annotated return type of `convertL3Evpn` `@Nullable`.

Second, just have `convertL3Evpn` return a (`@Nonnull`) `Set`, and build on its return statement(s).
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2070 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXnPe-4qmSSbp14Njd:-NFXnPe-4qmSSbp14Nje:bp2ideq) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2070)):*
> ```Java
>   }
> 
>   private @Nullable ImmutableSet.Builder<Layer3VniConfig> convertL3Evpn() {
> ```

This function is too long, and I'm having trouble reading it. Please break it up into helper functions for the different cases.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2170 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXmRpC8atW-xrUs1H2:-NFXmRpC8atW-xrUs1H3:b4qgrkr) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2170)):*
> ```Java
>         } else {
>           // RI is used, but RD is not manually configured. Junos takes route-distinguisher-id +
>           // random number to generate unique RD.
> ```

This is far from unique. In a typical large network, you are virtually guaranteed to have collisions here.

That aside, you must not create non-deterministic data model elements, because Batfish must produce the same output for the same set of inputs.

This is going to need a more thorough design to avoid collisions and preserve determinism.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2203 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXo2uA2exoc5_Uso66:-NFXo2uA2exoc5_Uso67:b-4pubwn) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2203)):*
> ```Java
>       } else {
>         if (_c.getVrfs()
>                 .get(getInterfaceOrUnitByName(l3Interface).get().getRoutingInstance().getName())
> ```

Is it  guaranteed  that the named l3 interface actually exists? This will throw an NPE if it doesn't.
Same question for `vtepSource`, and also in `convertL2Vni`.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2214 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXoUABDqKxiaw7n2f3:-NFXoUABDqKxiaw7n2f4:b-nvoekd) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2214)):*
> ```Java
>                   .setVni(vxlan.getVniId())
>                   .setSrcVrf(
>                       getInterfaceOrUnitByName(l3Interface).get().getRoutingInstance().getName())
> ```

Please Compute `getInterfaceOrUnityByName(l3Interface)` once, and reuse the value. Apply this pattern generally.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 3927 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXpHsJ76UJ7dX1iMsK:-NFXpHsJ76UJ7dX1iMsL:b-l38le) ([raw file](https://github.com/batfish/batfish/blob/99d821e90ec82c9f1c0f483810385a6554f29760/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L3927)):*
> ```Java
> 
>     // convert vxlan.
>     convertVxlan();
> ```

Any reason the replacement functions were moved so far up?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 4377 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXpM4WD39lJlVLNgtk:-NFXpM4X7zdxSwlGSFw4:b-w5vr5e) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L4377)):*
> ```Java
>                         // set IRB VLAN ID if assigned
>                         newUnitInterface.setVlan(irbVlanIds.get(name));
> ```

Remove unrelated whitespace change
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java` line 109 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXijpC-Ez-nXqUjn6r:-NFXijpC-Ez-nXqUjn6s:b8ry0lp) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/LogicalSystem.java#L109)):*
> ```Java
>   @Nullable private SwitchOptions _switchOptions;
> 
>   @Nullable private Map<Integer, VniOptions> _vniOptions;
> ```

This is guaranteed to be non-null by the assignment in the constructor.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/VniOptions.java` line 7 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXfag2BvPlG0Hl9P0W:-NFXfag2BvPlG0Hl9P0X:b-m1h95j) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/VniOptions.java#L7)):*
> ```Java
> import org.batfish.datamodel.bgp.community.ExtendedCommunity;
> 
> public class VniOptions implements Serializable {
> ```

Please add javadocs for all added non-private classes and methods, with the exception of simple getters and setters, and `@Override` methods of subtypes.
Also nit: mark any classes that will not be extended as `final`.
Also, Make getter return types and setter argument types have the same null annotation as the corresponding fields.

For non-null parameters of a new class, can skip the annotation and instead mark the class with `@ParametersAreNonnullByDefault`. Note this default does not apply to return types.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java` line 3506 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXpZl5Bsu_5jXnfOgs:-NFXpZl5Bsu_5jXnfOgt:b-o3dhb7) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java#L3506)):*
> ```Java
>                     .build())));
> 
>     // Ensure a vlan-id can be present and that it can be cleared back to null.
> ```

How does this assertion test that vlan-id can be present?
___
*[`projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-vxlan-l3vni` line 4 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXb3OU-PdJmAvl89d3:-NFXb3OU-PdJmAvl89d4:brmt18x) ([raw file](https://github.com/batfish/batfish/blob/99d821e90ec82c9f1c0f483810385a6554f29760/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-vxlan-l3vni#L4)):*
> ```
> set system host-name juniper-vxlan-l3vni
> #
> set interfaces xe-0/0/0 description to_host-1
> ```

Can you explain the changes in this file?

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: all files reviewed, 26 unresolved discussions (waiting on @dhalperi and @jeffkala)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/VniOptions.java` line 7 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXfag2BvPlG0Hl9P0W:-NHcyAer3Xw8b3tUaJIx:b-35p29v) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/VniOptions.java#L7)):*

Please add javadocs for all added non-private classes and methods, with the exception of simple getters and setters, and `@Override` methods of subtypes.
Also nit: mark any classes that will not be extended as `final`.
Also, Make getter return types and setter argument types have the same null annotation as the corresponding fields.

For non-null parameters of a new class, can skip the annotation and instead mark the class with `@ParametersAreNonnullByDefault`. Note this default does not apply to return types.
</blockquote></details>

This guidance has since been documented [here](https://github.com/batfish/batfish/tree/master/docs/contributing#java-coding-standards). That content in that link supersedes any prior comments in this PR about what to annotate, javadoc.

jeffkala:
Thanks @arifogel, hoping to take care of all this feedback soon.

jeffkala:
Hoping to take this feedback and implement soon.

dhalperi:
Thanks @jeffkala  - let us know!

You should probably start by merging/rebasing as things have continued to change.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 30 of 38 files reviewed, 24 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2042 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXbaZo8cd89j93okDF:-NPdTHVG81aVGjwxWQ0z:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2042)):*

This branch should never trigger, as an invariant of `ParserRuleContext`. Just get rid of the if and this branch. Same for your other converter.
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4421 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXjH7QCBJ5I_LgHubC:-NPdTNje-3CKpBNs7DA6:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4421)):*

This is guaranteed to be true by the grammar. Remove the `if.`
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 7237 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXkoTV20lvnpizKqG1:-NPdTX3KB8Bj1dQqKqG_:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L7237)):*

This is guaranteed to be true by the grammar, so this whole function is effectively:
```
_currentBgpGroup.setEvpnAf(true);
```
</blockquote></details>

Done.

arifogel (commented):
> Reviewed 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: all files reviewed, 21 unresolved discussions (waiting on @dhalperi and @jeffkala)

arifogel (commented):
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: all files reviewed, 21 unresolved discussions (waiting on @dhalperi and @jeffkala)

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 25 of 39 files reviewed, 17 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4424 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXjaYaB3N5KVZr0_3Z:-NgdWeRUA_dbEvYly3GD:byqd0di) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4424)):*
<details><summary><i>Previously, jeffkala (Jeff Kala) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Tried this and _currentVniOptions was always null, so I reverted back to this "longer" pattern.  I tried a few different ways to get it to work with no luck.

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 25 of 39 files reviewed, 17 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2041 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXc77s7bQIWjd8pLg1:-NgdYwbS20UFlP2WCx2o:bysqm2t) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2041)):*

I'm not convinced the grammar enforces that a `vt_community` nor `extended_community` as guaranteed to be parseable by `ExtendedCommunity.parse`, as `@dhalperi pointed out.

Either:
* update the grammar to make it so (if possible)
OR:
* 
- this should be non-static, return an `Optional<ExtendedCommunity>` via `tryParse`, and callers will need to handle the case where `Optional.empty()` is returned gracefully.

Same goes for your other converter.
</blockquote></details>

Changed this to tryParse as suggested.

jeffkala (commented):
Did a first pass on the first set of recommendations.  Will be working on the rest soon. Would love to get this one over the finish line ASAP.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542#-:-Ngdb9OB-yOlLeAZSsTc:b9a6sdf)* status: 25 of 39 files reviewed, 17 unresolved discussions (waiting on @arifogel and @dhalperi)

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 25 of 39 files reviewed, 17 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 779 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXfAt79BUXAAFk1pcy:-NgeOwBUCngayOyL8_Zy:bnbbx03) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L779)):*

What if it's false?
This should be exactly `Boolean.TRUE.equals(ig.getEvpnAf())`
</blockquote></details>

I have the below snippet. Thought by creating new Evpn() as a "dummy" would handle it. Mind expanding on this concern?



_Code snippet:_
```Java
  @Override
  public void enterP_evpn(P_evpnContext ctx) {
    if (_currentLogicalSystem.getEvpn() == null) {
      _currentLogicalSystem.setEvpn(new Evpn());
    }
  }
  
```
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 784 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXl8H66MUT_yEsXLaj:-NgeODBB-UtdTHGpSoqf:bsjd8k7) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L784)):*

Sanity check: allow remote as out is always true for juniper? Just this address family?
</blockquote></details>

will get this fixed, looks like its disabled by default https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/advertise-peer-as-edit-protocols-bgp.html#id-13323108__d67115e166

jeffkala (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8542)* status: 25 of 39 files reviewed, 16 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 785 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXlnuQAlDSlHUx8c3R:-NgepHLD0vXK3RZfxAUM:b-o3im1p) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L785)):*

First, `setL3Vnis` cannot accept a null argument (you annotated return type of `convertL3Evpn` `@Nullable`.

Second, just have `convertL3Evpn` return a (`@Nonnull`) `Set`, and build on its return statement(s).
</blockquote></details>

reworked it
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 2070 at r5](https://reviewable.io/reviews/batfish/batfish/8542#-NFXnPe-4qmSSbp14Njd:-NgepEq38WJOAXRG2kg9:b-o3im1p) ([raw file](https://github.com/batfish/batfish/blob/4164886bf144aa8cb30c94cda1346165ffffc684/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L2070)):*

This function is too long, and I'm having trouble reading it. Please break it up into helper functions for the different cases.
</blockquote></details>

reworked it

jeffkala:
@dhalperi @arifogel I know its been awhile, but few weeks back I did some updates here.  I believe I covered the concerns, is it possible to get a re-review.

arifogel:
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2062 at r9](https://reviewable.io/reviews/batfish/batfish/8542#-NiGiBFe3kJHhCdGip4y:-NiGiBFf35DRedw-J1AK:bxv6xct) ([raw file](https://github.com/batfish/batfish/blob/fd32735a784e16f45cb756006e93ca7816e9bcc7/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2062)):*
> ```Java
>     // system are as follows:
>     //
>     //      4-octet AS number: 00000000-22222222
> ```

I'm having a hard time understanding this comment.


The text in the quoted line looks like 8 octets, not 4. Or is this a range?

arifogel:
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 2065 at r9](https://reviewable.io/reviews/batfish/batfish/8542#-NiGiVPwDSo3n-XAMFo5:-NiGiVPwDSo3n-XAMFo6:b-xrdaxd) ([raw file](https://github.com/batfish/batfish/blob/fd32735a784e16f45cb756006e93ca7816e9bcc7/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L2065)):*
> ```Java
>     //      4-octet AS number followed by 4-octet value: 00000000-22222222:11111111-44444444
>     //      16-octet value: 11111111-22222222-33333333-44444444
>     //      The first two patterns are used for standard extended communities, while the third
> ```

What is a "standard" extended community, and what is a "large" extended community?

IIUC there are three broad types of communities:
1. standard (4 octets)
2. extended (6 octets) (with many subtypes, like site-of-origin)
3. large (12 octets)

See https://datatracker.ietf.org/doc/html/rfc8092#section-1

PR #8571, opened 2022-10-22
InterfaceNameComparator: cache expensive comparisons

dhalperi:
InterfaceNameComparator does a ton of work to attempt to sort interface names
correctly. Caching that work improves performance by 4x-5x of some of the big
sorted structures.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java
@@ -8,6 +8,7 @@ import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.Comparator;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8571)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java` line 114 at r1](https://reviewable.io/reviews/batfish/batfish/8571#-NFCIImL-3_6LlL7SKb5:-NFCIImL-3_6LlL7SKb6:b-f3ykb8) ([raw file](https://github.com/batfish/batfish/blob/cfb6d28c577ee1f5b60f459629f7563993b1b91f/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java#L114)):*
> ```Java
>   }
> 
>   private final transient ConcurrentHashMap<CacheKey, Integer> _cmp = new ConcurrentHashMap<>();
> ```

This cache is unbounded, and the comparator is used across all node-interface pairs, right?  I'd expect the number of entries to be O((#interfaces per node*#nodes)^2). Just want to ensure we believe this is acceptable static memory overhead for large networks.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8571)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java` line 114 at r1](https://reviewable.io/reviews/batfish/batfish/8571#-NFCIImL-3_6LlL7SKb5:-NFCT8OdD4CGzLGYKxJH:bna0wuy) ([raw file](https://github.com/batfish/batfish/blob/cfb6d28c577ee1f5b60f459629f7563993b1b91f/projects/batfish-common-protocol/src/main/java/org/batfish/common/util/InterfaceNameComparator.java#L114)):*

This cache is unbounded, and the comparator is used across all node-interface pairs, right?  I'd expect the number of entries to be O((#interfaces per node*#nodes)^2). Just want to ensure we believe this is acceptable static memory overhead for large networks.
</blockquote></details>

Sorry that's wrong, looks like it doesn't include node. Similar comment but less urgency: O((#distinct interface names across all nodes)^2). Probably fine.

PR #8574, opened 2022-10-26
ExtendedCommunity: make internals more generic to support opaque ECs

SLarkworthy:
See https://tools.ietf.org/html/rfc4360 section 3.3 - opaque extended communities can not necessarily be split into local and global administrators in the way that other extended community types can. To enable more use cases of the 6 byte value field, change the internals to work in terms of the entire value rather than administrators, and convert the value into ga/la as needed. Also, add a method to create a new opaque community, that is decoupled from ga/la.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
@@ -22,32 +22,35 @@ import org.batfish.datamodel.Ip;
  */
 @ParametersAreNonnullByDefault
 public final class ExtendedCommunity extends Community {
+  /** The max value that can be stored in 6 bytes. */
+  private static final long VALUE_MAX = 0xFFFF_FFFF_FFFFL;
 
-  private static final Set<Byte> VALID_TYPES =
-      ImmutableSet.of(
-          (byte) 0x00,
-          (byte) 0x01,
-          (byte) 0x02,
-          (byte) 0x03,
-          (byte) 0x40,
-          (byte) 0x41,
-          (byte) 0x43);
+  private static final Set<Integer> VALID_TYPES =
+      ImmutableSet.of(0x00, 0x01, 0x02, 0x03, 0x40, 0x41, 0x43);
 
-  private final byte _type;
-  private final byte _subType;
-  private final long _globalAdministrator;
-  private final long _localAdministrator;
+  /** 1 byte. */
+  private final int _type;
+
+  /** 1 byte. */
+  private final int _subType;
+
+  /** 6 bytes. */
+  private final long _value;
 
   // Cached string representations
   @Nullable private transient String _str;
   @Nullable private transient String _regexStr;
 
-  private ExtendedCommunity(
-      byte type, byte subType, long globalAdministrator, long localAdministrator) {
+  private ExtendedCommunity(int type, int subType, long value) {
+    checkArgument(VALID_TYPES.contains(type), "Not a valid BGP extended community type: %s", type);
+    checkArgument(
+        subType >= 0 && type <= 0xFF, "Subtype not within accepted 0-0xFF range: %s", type);
+    checkArgument(
+        value >= 0 && value <= VALUE_MAX, "Value %s is not within the accepted range", value);
+
     _type = type;
     _subType = subType;
-    _globalAdministrator = globalAdministrator;
-    _localAdministrator = localAdministrator;
+    _value = value;
   }
 
   @JsonCreator
@@ -58,6 +61,8 @@ public final class ExtendedCommunity extends Community {
 
   @Nonnull
   public static ExtendedCommunity parse(String communityStr) {
+    // TODO: move vendor-specific parsing into vendor-specific context, remove redundant validation.
+
     String[] parts = communityStr.trim().split(":");
     checkArgument(parts.length == 3, "Invalid extended community string: %s", communityStr);
     long gaLong;
@@ -155,21 +160,22 @@ public final class ExtendedCommunity extends Community {
         type);
     byte typeByte = (byte) (type >> 8);
     checkArgument(
-        VALID_TYPES.contains(typeByte), "Not a valid BGP extended community type: %s", type);
-    checkArgument(
         globalAdministrator >= 0 && localAdministrator >= 0,
         "Administrator values must be positive");
+    int gaOffset;
     if (typeByte == 0x00 || typeByte == 0x40) {
       checkArgument(
           globalAdministrator <= 0xFFFFL && localAdministrator <= 0xFFFFFFFFL,
           "Extended community administrator values are not within the allowed range");
+      gaOffset = 32;
     } else {
       checkArgument(
           globalAdministrator <= 0xFFFFFFFFL && localAdministrator <= 0xFFFFL,
           "Extended community administrator values are not within the allowed range");
+      gaOffset = 16;
     }
-    return new ExtendedCommunity(
-        typeByte, (byte) (type & 0xFF), globalAdministrator, localAdministrator);
+    long value = (globalAdministrator << gaOffset) | localAdministrator;
+    return new ExtendedCommunity(typeByte, type & 0xFF, value);
   }
 
   public static ExtendedCommunity of(int type, Ip globalAdministrator, long localAdministrator) {
@@ -184,6 +190,12 @@ public final class ExtendedCommunity extends Community {
   public static ExtendedCommunity target(long asn, long value) {
     int type = asn <= 0xFFFFL ? 0x00 : 0x02;
     return of(type << 8 | 0x02, asn, value);
+  }
+
+  /** Return an opaque extended community. See https://tools.ietf.org/html/rfc4360 section 3.3 */
+  public static ExtendedCommunity opaque(boolean transitive, int subtype, long value) {
+    int type = transitive ? 0x03 : 0x43;
+    return new ExtendedCommunity(type, subtype, value);
   }
 
   @Override
@@ -217,21 +229,53 @@ public final class ExtendedCommunity extends Community {
   }
 
   /**
-   * Return the global administrator value.
+   * Returns the global administrator value, if applicable to this type of extended community. May
+   * be either 2 or 4 bytes depending on type/subtype. **
    *
-   * <p>It is the middle part of the extended community (type:globalAdmin:localAdmin)
+   * @throws UnsupportedOperationException if this extended community is not of a type with a global
+   *     administrator
    */
   public long getGlobalAdministrator() {
-    return _globalAdministrator;
+    if (!hasGlobalAndLocalAdministrator()) {
+      throw new UnsupportedOperationException(
+          String.format("Extended community does not have a global administrator: %s", this));
+    }
+    return _value >> localAdministratorBits();
   }
 
   /**
-   * Return the local administrator value.
+   * Returns the local administrator value, if applicable to this type of extended community. May be
+   * either 2 or 4 bytes depending on type/subtype. **
    *
-   * <p>It is the last part of the extended community (type:globalAdmin:localAdmin)
+   * @throws UnsupportedOperationException if this extended community is not of a type with a local
+   *     administrator
    */
   public long getLocalAdministrator() {
-    return _localAdministrator;
+    if (!hasGlobalAndLocalAdministrator()) {
+      throw new UnsupportedOperationException(
+          String.format("Extended community does not have a local administrator: %s", this));
+    }
+    return _value & (VALUE_MAX >> globalAdministratorBits());
+  }
+
+  public long getValue() {
+    return _value;
+  }
+
+  @Nonnull
+  private boolean hasGlobalAndLocalAdministrator() {
+    // return false if it is an opaque extended community.
+    return !(_type == 0x03 || _type == 0x43);
+  }
+
+  @Nonnull
+  private int globalAdministratorBits() {
+    return _type == 0x00 || _type == 0x40 ? 16 : 32;
+  }
+
+  @Nonnull
+  private int localAdministratorBits() {
+    return 48 - globalAdministratorBits();
   }
 
   @Override
@@ -243,25 +287,26 @@ public final class ExtendedCommunity extends Community {
       return false;
     }
     ExtendedCommunity that = (ExtendedCommunity) o;
-    return _type == that._type
-        && _subType == that._subType
-        && _globalAdministrator == that._globalAdministrator
-        && _localAdministrator == that._localAdministrator;
+    return _type == that._type && _subType == that._subType && _value == that._value;
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(_type, _subType, _globalAdministrator, _localAdministrator);
+    return Objects.hash(_type, _subType, _value);
   }
 
   @Nonnull
   @Override
   public String matchString() {
-    if (_regexStr == null) {
-      // Type is skipped for regex matching
-      _regexStr = _globalAdministrator + ":" + _localAdministrator;
+    if (!hasGlobalAndLocalAdministrator()) {
+      return toString();
+    } else {
+      if (_regexStr == null) {
+        // Type is skipped for regex matching
+        _regexStr = getGlobalAdministrator() + ":" + getLocalAdministrator();
+      }
+      return _regexStr;
     }
-    return _regexStr;
   }
 
   @Override
@@ -269,15 +314,19 @@ public final class ExtendedCommunity extends Community {
   @JsonValue
   public String toString() {
     if (_str == null) {
-      // To differentiate 4 vs 2-byte global admin values
-      String gaSuffix = _type == 0x00 || _type == 0x40 ? "" : "L";
-      _str =
-          (((int) _type << 8) | _subType)
-              + ":"
-              + _globalAdministrator
-              + gaSuffix
-              + ":"
-              + _localAdministrator;
+      if (!hasGlobalAndLocalAdministrator()) {
+        _str = String.format("0x%x:0x%x:0x%x", _type, _subType, _value);
+      } else {
+        // To differentiate 4 vs 2-byte global admin values
+        String gaSuffix = _type == 0x00 || _type == 0x40 ? "" : "L";
+        _str =
+            ((_type << 8) | _subType)
+                + ":"
+                + getGlobalAdministrator()
+                + gaSuffix
+                + ":"
+                + getLocalAdministrator();
+      }
     }
     return _str;
   }
@@ -285,11 +334,9 @@ public final class ExtendedCommunity extends Community {
   @Nonnull
   @Override
   protected BigInteger asBigIntImpl() {
-    int gaOffset = _type == 0x00 || _type == 0x40 ? 32 : 16;
     return BigInteger.valueOf(_type)
         .shiftLeft(56)
         .or(BigInteger.valueOf(_subType).shiftLeft(48))
-        .or(BigInteger.valueOf(_globalAdministrator).shiftLeft(gaOffset))
-        .or(BigInteger.valueOf(_localAdministrator));
+        .or(BigInteger.valueOf(_value));
   }
 }
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
@@ -230,7 +230,7 @@ public final class ExtendedCommunity extends Community {
 
   /**
    * Returns the global administrator value, if applicable to this type of extended community. May
-   * be either 2 or 4 bytes depending on type/subtype. **
+   * be either 2 or 4 bytes depending on type/subtype.
    *
    * @throws UnsupportedOperationException if this extended community is not of a type with a global
    *     administrator
@@ -245,7 +245,7 @@ public final class ExtendedCommunity extends Community {
 
   /**
    * Returns the local administrator value, if applicable to this type of extended community. May be
-   * either 2 or 4 bytes depending on type/subtype. **
+   * either 2 or 4 bytes depending on type/subtype.
    *
    * @throws UnsupportedOperationException if this extended community is not of a type with a local
    *     administrator
@@ -258,13 +258,14 @@ public final class ExtendedCommunity extends Community {
     return _value & (VALUE_MAX >> globalAdministratorBits());
   }
 
+  /** Return the 6 byte value */
   public long getValue() {
     return _value;
   }
 
+  /** Return true if the value field contains a global and local administrator */
   @Nonnull
   private boolean hasGlobalAndLocalAdministrator() {
-    // return false if it is an opaque extended community.
     return !(_type == 0x03 || _type == 0x43);
   }
 
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java
@@ -130,6 +130,20 @@ public final class ExtendedCommunityTest {
   }
 
   @Test
+  public void testGetLAInapplicable() {
+    ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
+    thrown.expect(UnsupportedOperationException.class);
+    community.getLocalAdministrator();
+  }
+
+  @Test
+  public void testGetGAInapplicable() {
+    ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
+    thrown.expect(UnsupportedOperationException.class);
+    community.getGlobalAdministrator();
+  }
+
+  @Test
   public void testIsTransitive() {
     assertTrue(ExtendedCommunity.parse("1:1:1").isTransitive());
     assertFalse(ExtendedCommunity.parse("16384:1:1").isTransitive());
@@ -139,6 +153,10 @@ public final class ExtendedCommunityTest {
   public void testMatchString() {
     ExtendedCommunity ec = ExtendedCommunity.parse("origin:65555L:1");
     assertThat(ec.matchString(), equalTo("65555:1"));
+
+    // check no local/global administrator case
+    assertThat(
+        ExtendedCommunity.opaque(true, 0x00, 65559).matchString(), equalTo("0x3:0x0:0x10017"));
   }
 
   @Test
@@ -147,6 +165,10 @@ public final class ExtendedCommunityTest {
     assertThat(ExtendedCommunity.parse("origin:1L:1").toString(), equalTo("515:1L:1"));
     assertThat(ExtendedCommunity.parse("origin:1:1").toString(), equalTo("3:1:1"));
     assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
+    assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
+
+    // check no local/global administrator case
+    assertThat(ExtendedCommunity.opaque(true, 0x00, 65559).toString(), equalTo("0x3:0x0:0x10017"));
   }
 
   @Test
@@ -176,6 +198,12 @@ public final class ExtendedCommunityTest {
         equalTo(ExtendedCommunity.of(0x02 << 8 | 0x02, 65555, 1)));
     assertThat(ExtendedCommunity.target(1, 65555).toString(), equalTo("2:1:65555"));
     assertThat(ExtendedCommunity.target(65555, 1).toString(), equalTo("514:65555L:1"));
+  }
+
+  @Test
+  public void testOpaqueCreation() {
+    ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
+    assertThat(community.getValue(), equalTo((long) 65559));
   }
 
   @Test
@@ -210,10 +238,23 @@ public final class ExtendedCommunityTest {
   @Test
   public void testGetGlobalAdmin() {
     assertThat(ExtendedCommunity.of(1, 2, 3).getGlobalAdministrator(), equalTo(2L));
+    assertThat(
+        ExtendedCommunity.of(1, 42995L, 4294967295L).getGlobalAdministrator(), equalTo(42995L));
+    assertThat(
+        ExtendedCommunity.of(0x01 << 8 | 1, 4294967295L, 1).getGlobalAdministrator(),
+        equalTo(4294967295L));
+    assertThat(
+        ExtendedCommunity.of(0x01 << 8 | 1, 1, 42995L).getGlobalAdministrator(), equalTo(1L));
   }
 
   @Test
   public void testGetLocalAdmin() {
     assertThat(ExtendedCommunity.of(1, 2, 3).getLocalAdministrator(), equalTo(3L));
+    assertThat(
+        ExtendedCommunity.of(1, 42995L, 4294967295L).getLocalAdministrator(), equalTo(4294967295L));
+    assertThat(
+        ExtendedCommunity.of(0x01 << 8 | 1, 4294967295L, 1).getLocalAdministrator(), equalTo(1L));
+    assertThat(
+        ExtendedCommunity.of(0x01 << 8 | 1, 1, 42995L).getLocalAdministrator(), equalTo(42995L));
   }
 }
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java
@@ -165,10 +165,11 @@ public final class ExtendedCommunityTest {
     assertThat(ExtendedCommunity.parse("origin:1L:1").toString(), equalTo("515:1L:1"));
     assertThat(ExtendedCommunity.parse("origin:1:1").toString(), equalTo("3:1:1"));
     assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
-    assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
 
     // check no local/global administrator case
     assertThat(ExtendedCommunity.opaque(true, 0x00, 65559).toString(), equalTo("0x3:0x0:0x10017"));
+    assertThat(
+        ExtendedCommunity.opaque(false, 0x00, 65559).toString(), equalTo("0x43:0x0:0x10017"));
   }
 
   @Test
@@ -202,8 +203,10 @@ public final class ExtendedCommunityTest {
 
   @Test
   public void testOpaqueCreation() {
-    ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
-    assertThat(community.getValue(), equalTo((long) 65559));
+    ExtendedCommunity communityTransitive = ExtendedCommunity.opaque(true, 0x00, 65559);
+    assertThat(communityTransitive.getValue(), equalTo(65559L));
+    ExtendedCommunity communityIntransitive = ExtendedCommunity.opaque(false, 0x01, 2);
+    assertThat(communityIntransitive.getValue(), equalTo(2L));
   }
 
   @Test


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 11 unresolved discussions (waiting on @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFMqNWQ3ff04cpC8VD2:bbgt6ft) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*
> ```Java
>   }
> 
>   public static ExtendedCommunity of(int type, long value) {
> ```

The `of` functions are not particularly useful to users when compared with the helpers (past sins). Consider making this `private`, or alternatively package-private with `@VisibleForTesting` annotation if you are testing it.

Also, nit: `@Nonnull` return type here and for the helper functions you added.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFMUkUJ4Hp06j_4CnES:b5yujam) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*
> ```Java
>     checkArgument(value <= VALUE_MAX, "Value is not within the allowed space");
> 
>     return new ExtendedCommunity(typeByte, (byte) (type & 0xFF), value);
> ```

Since you're continuing the incorrect pattern (which is the right thing, since fixing is out of scope), please add comment here and in above constructor call:
```TODO: fix reversed type and subtype```
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 208 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFM_pva5M4jNvhZ4sX7:-NFM_pva5M4jNvhZ4sX8:bmpo747) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L208)):*
> ```Java
>    * Return an opaque extended community. See https://tools.ietf.org/html/rfc4360 section 3.3
>    * Assumes a type of 0x03 or 0x43 and a variable input subtype. Opaque extended communities value
>    * field does not always map to a local and global administrator Leave it a generic 6 octets for
> ```

I'd leave any details about local and global administrator in the javadocs for those getters for:
* Legibility: The user needn't know about the unfortunate/confusing historical context of this class.
* Maintainability: If we mention ga/la here, then we'd be obligated to update this javadoc every time this class adds supports any other non-general EC field.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 211 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMVGS_DX5AHhhh3fDS:-NFMVGS_DX5AHhhh3fDT:bii47s) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L211)):*
> ```Java
>    * flexibility based on subtype.
>    */
>   public static ExtendedCommunity opaque(byte type, byte subtype, long value) {
> ```

We can reorganize this to be a bit more granular and strict:
* `transitiveOpaque` function, which only takes `subtype` and `value`
* `nonTransitiveOpaque` function, which only takes `subtype` and `value`

This removes the possibility of the assumption being violated.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 246 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMYwgj24cb-2mJdI7e:-NFMYwgk0wwKyAxe4hFv:b-rz5zpn) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L246)):*
> ```Java
> 
>   /**
>    * Return the global administrator value.
> ```

The javadoc should be updated since it may now throw an exception.
Consider:
```
Returns the global administrator value, if applicable to this type of extended community. May be either 2 or 4 bytes depending on type/subtype.

@throws UnsupportedOperationException if this extended community is not of a type with a global administrator
```.
Similar for local administrator.
```

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 253 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMZM_c9lQQUOY798xe:-NFMZM_c9lQQUOY798xf:b-mm396e) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L253)):*
> ```Java
>     if (!hasGlobalAdministrator()) {
>       throw new UnsupportedOperationException(
>           "Extended community does not have a global administrator");
> ```

Can you include the type and subtype in the message?
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 275 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaqng4U43z2FOvF42:-NFMaqng4U43z2FOvF43:b-iemos0) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L275)):*
> ```Java
>   }
> 
>   private boolean hasGlobalAdministrator() {
> ```

Nit: Aren't GA and LA always together?
If so, can simplify to one function: `hasGlobalAndLocalAdministator`
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 288 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaJM7Bst9vgMvKC0c:-NFMaJM7Bst9vgMvKC0d:b7zzwla) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L288)):*
> ```Java
>   }
> 
>   private int localAdministratorSize() {
> ```

Nit: Please include unit for clarity, e.g. `localAdministratorSizeBits`
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 329 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMbj7x4j1YYG8WlqN5:-NFMbj7x4j1YYG8WlqN6:bnzw302) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L329)):*
> ```Java
>     if (_str == null) {
>       if (!hasGlobalAdministrator() || !hasLocalAdministrator()) {
>         _str = String.format("%x:%x:%x", _type, _subType, _value);
> ```

Could you add `0x` before each `%` to disambiguate from the other format?
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 133 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMpAdzC4caT-b0YQ48:-NFMpAdzC4caT-b0YQ49:b-th3crw) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L133)):*
> ```Java
> 
>   @Test
>   public void testOpaqueGetLA() {
> ```

nit: Strictly speaking, it may not just be opaque ECs that may not have GA/LA. Consider instead e.g. `testGetLAInapplicable` or `testGetLAWrongType`.
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 134 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMouse11-lNZa-0bHV:-NFMouse11-lNZa-0bHW:bjar102) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L134)):*
> ```Java
>   @Test
>   public void testOpaqueGetLA() {
>     thrown.expect(UnsupportedOperationException.class);
> ```

nit: prefer to put this right before the line you expect to throw. Otherwise it's not clear which line is throwing (would pass if either line threw that exception).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 11 unresolved discussions (waiting on @arifogel and @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFPFRHe4GEEnhoZbmia:b-9a63yt) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*

The `of` functions are not particularly useful to users when compared with the helpers (past sins). Consider making this `private`, or alternatively package-private with `@VisibleForTesting` annotation if you are testing it.

Also, nit: `@Nonnull` return type here and for the helper functions you added.
</blockquote></details>

This definitely needs javadoc.

Are `type` and `subtype` mandatory parameters, or do only some `types` have subtypes?

If the former, just go ahead and change the constructor to have `type` and `subtype` arguments, and ditch this form.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFPFInA-idotf83N5ST:b-ner9ra) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*

Since you're continuing the incorrect pattern (which is the right thing, since fixing is out of scope), please add comment here and in above constructor call:
```TODO: fix reversed type and subtype```
</blockquote></details>

Say more about this? We can fix private internals easily. And we can make new APIs.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 208 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFM_pva5M4jNvhZ4sX7:-NFPFFXp0lg7WMDGZMdP:b5nlme0) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L208)):*

I'd leave any details about local and global administrator in the javadocs for those getters for:
* Legibility: The user needn't know about the unfortunate/confusing historical context of this class.
* Maintainability: If we mention ga/la here, then we'd be obligated to update this javadoc every time this class adds supports any other non-general EC field.
</blockquote></details>

Agreed. GA/LA are specific to certain types, and we don't need to talk about them everywhere.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 253 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMZM_c9lQQUOY798xe:-NFPEr2dEv3z1vOQSVMF:b-vv8dil) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L253)):*

Can you include the type and subtype in the message?
</blockquote></details>

Why not just `"%s, this` in a format string?
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 275 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaqng4U43z2FOvF42:-NFPEx5QDHiEk9MSdqAk:b32v57) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L275)):*

Nit: Aren't GA and LA always together?
If so, can simplify to one function: `hasGlobalAndLocalAdministator`
</blockquote></details>

+1
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 288 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaJM7Bst9vgMvKC0c:-NFPF-461jOwPYE_VxTW:baqg0d4) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L288)):*

Nit: Please include unit for clarity, e.g. `localAdministratorSizeBits`
</blockquote></details>

Or just `bits`.

**Review by dhalperi — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 212 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFPG4-mAmwznmFFu4Ff:-NFPG4-mAmwznmFFu4Fg:b4s1415) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L212)):*
> ```Java
>    */
>   public static ExtendedCommunity opaque(byte type, byte subtype, long value) {
>     return of((((int) type) << 8) | subtype, value);
> ```

Unless I'm missing something, I would strongly disagree with adding new invocations of this bad pattern. Why is this valuable?

It looks to me like you're calling into a new function you made with unnecessary stuff to combine fields you immediately pull apart.



_Code quote:_
```Java
of((((int) type) << 8) | subtype, value);
```

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 12 unresolved discussions (waiting on @dhalperi and @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFPNa5e4kFnFhNVSmQH:b8opg5v) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*

This definitely needs javadoc.

Are `type` and `subtype` mandatory parameters, or do only some `types` have subtypes?

If the former, just go ahead and change the constructor to have `type` and `subtype` arguments, and ditch this form.
</blockquote></details>

Ugh it seems that only some ECs have subtype according to RFC 4360, and this is determined by the first byte (type). So we should have subtype getter with logic similar to `getGa` from this PR, and internal representation should probably only have one byte for type, and 7 for value.

Or, perhaps we should just use a single long for the whole internal representation, and do bit math for every getter.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFPLBcxA5U4YeeZEQt8:bh781qs) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*

Say more about this? We can fix private internals easily. And we can make new APIs.
</blockquote></details>

The two byte and subtype bytes are consistently switched in our internal representation. We can fix here and now, but I thought it could use a separate PR / pass through usages.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 212 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFPG4-mAmwznmFFu4Ff:-NFPMe9h22sT9XoetmLS:bf20o8j) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L212)):*

Unless I'm missing something, I would strongly disagree with adding new invocations of this bad pattern. Why is this valuable?

It looks to me like you're calling into a new function you made with unnecessary stuff to combine fields you immediately pull apart.
</blockquote></details>

I think I agree, but want to make sure I understand. Are you saying this should call a creator function that takes type and subtype separately, such that no bit math is necessary since those are fields of the internal representation?
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 253 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMZM_c9lQQUOY798xe:-NFPNAWs8QW4AWjL5vaL:bb6mt9b) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L253)):*

Why not just `"%s, this` in a format string?
</blockquote></details>

:+1:

SLarkworthy (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 12 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFPE6ZD9aqDkdAovG-T:b3fcblv) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*

Ugh it seems that only some ECs have subtype according to RFC 4360, and this is determined by the first byte (type). So we should have subtype getter with logic similar to `getGa` from this PR, and internal representation should probably only have one byte for type, and 7 for value.

Or, perhaps we should just use a single long for the whole internal representation, and do bit math for every getter.
</blockquote></details>

While it is possible for an EC to have a 7-byte value, I can't find a single documented example that uses this. Okay to defer until we have an actual usecase, or should we update now? We have no accepted types that allow for 7-byte values so a hasSubtype helper function would currently trivially return true.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFPFZq3D-TgFw3fhiQZ:b-bywnza) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*

The two byte and subtype bytes are consistently switched in our internal representation. We can fix here and now, but I thought it could use a separate PR / pass through usages.
</blockquote></details>

I'm no longer convinced these are actually switched - it seems to be consistently <type><subtype>.  While in actual representations this would be <type><subtype><value> instead of what this is (<empty 0's><subtype><value>) I think it still makes sense intuitively.

I am generally updating this to minimize the number of times we combine type and subtype only to separate them again immediately after, so this may be a moot point after that revision is done.

SLarkworthy:
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFPymc87zBqe3TBuDeM:b-l91ta3) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*
<details><summary><i>Previously, SLarkworthy (Sarah Larkworthy) wrote…</i></summary><blockquote>

I'm no longer convinced these are actually switched - it seems to be consistently <type><subtype>.  While in actual representations this would be <type><subtype><value> instead of what this is (<empty 0's><subtype><value>) I think it still makes sense intuitively.

I am generally updating this to minimize the number of times we combine type and subtype only to separate them again immediately after, so this may be a moot point after that revision is done. 
</blockquote></details>

to clarify since reviewable weirdly formatted this comment - consistently our code is doing (from most to least significant) type then subtype then value (and value is ga followed by la) . In this case since we want only type and not value, it's just type then subtype.  It's hard to tell by the RFC what should be most significant (ex: it labels value as byte 0, but calls type high and subtype low) , but as long as we are consistent with the order we should be fine.

SLarkworthy (commented):
Changed type and subtype internals to be ints instead of bytes - using bytes wasn't saving any space, and an int allows for the full range of 0 - 0xFF. Moved validation to the constructor - there is some redundant validation in parse(), but I am considering updating parse to be out of scope of this PR. 

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574#-:-NFQ3TnX7pHdfS0wrCGB:bo9afnh)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @arifogel and @dhalperi)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 208 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFM_pva5M4jNvhZ4sX7:-NFPp8O81caOp-ALbcvO:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L208)):*

Agreed. GA/LA are specific to certain types, and we don't need to talk about them everywhere.
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 211 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMVGS_DX5AHhhh3fDS:-NFPpEJv5xM1ZgBqyMpq:brbl8zf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L211)):*

We can reorganize this to be a bit more granular and strict:
* `transitiveOpaque` function, which only takes `subtype` and `value`
* `nonTransitiveOpaque` function, which only takes `subtype` and `value`

This removes the possibility of the assumption being violated.
</blockquote></details>

done - made opaque accept a boolean for if it is transitive or not.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 212 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFPG4-mAmwznmFFu4Ff:-NFPrNFN98Z011-FOIg7:bvm047t) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L212)):*

I think I agree, but want to make sure I understand. Are you saying this should call a creator function that takes type and subtype separately, such that no bit math is necessary since those are fields of the internal representation?
</blockquote></details>

removed this in favor of a constructor that just accepts type and subtype separately. 
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 246 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMYwgj24cb-2mJdI7e:-NFPruSLAjICdp3XQxTr:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L246)):*

The javadoc should be updated since it may now throw an exception.
Consider:
```
Returns the global administrator value, if applicable to this type of extended community. May be either 2 or 4 bytes depending on type/subtype.

@throws UnsupportedOperationException if this extended community is not of a type with a global administrator
```.
Similar for local administrator.
```

</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 253 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMZM_c9lQQUOY798xe:-NFPsSfa3KimeQANHvBs:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L253)):*

:+1:
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 275 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaqng4U43z2FOvF42:-NFPsT_s49O_yZtzNj0o:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L275)):*

+1
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 288 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMaJM7Bst9vgMvKC0c:-NFPtCii-h_3zbYMQx9t:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L288)):*

Or just `bits`.
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 329 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMbj7x4j1YYG8WlqN5:-NFPtFwT0yHJuA40o-vR:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L329)):*

Could you add `0x` before each `%` to disambiguate from the other format?
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 133 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMpAdzC4caT-b0YQ48:-NFPtXhLBQ3_4SMrd_iA:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L133)):*

nit: Strictly speaking, it may not just be opaque ECs that may not have GA/LA. Consider instead e.g. `testGetLAInapplicable` or `testGetLAWrongType`.
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 134 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMouse11-lNZa-0bHV:-NFPtWd16n8L9V45Etxh:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L134)):*

nit: prefer to put this right before the line you expect to throw. Otherwise it's not clear which line is throwing (would pass if either line threw that exception).
</blockquote></details>

done

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @dhalperi and @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFQ7Gk443zi8fXEuOqK:b-44deo) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*
<details><summary><i>Previously, SLarkworthy (Sarah Larkworthy) wrote…</i></summary><blockquote>

While it is possible for an EC to have a 7-byte value, I can't find a single documented example that uses this. Okay to defer until we have an actual usecase, or should we update now? We have no accepted types that allow for 7-byte values so a hasSubtype helper function would currently trivially return true.
</blockquote></details>

The QoS Marking EC has a 7-byte value. The internals of that value are in draft only, but what is clear is that there is no subtype.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 196 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMUvdf3PGYt7bd2zSj:-NFQ2iEg0mD8u-MoBxlK:bct9o6p) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L196)):*
<details><summary><i>Previously, SLarkworthy (Sarah Larkworthy) wrote…</i></summary><blockquote>

to clarify since reviewable weirdly formatted this comment - consistently our code is doing (from most to least significant) type then subtype then value (and value is ga followed by la) . In this case since we want only type and not value, it's just type then subtype.  It's hard to tell by the RFC what should be most significant (ex: it labels value as byte 0, but calls type high and subtype low) , but as long as we are consistent with the order we should be fine.
</blockquote></details>

I think there are two important goals here:
1. Pack things in such a way that you get the same stuff out from the getters that was passed to the helpers
2. Pack things in such a way that you always get a different combo of internal fields for different helper creators, and for different values passed to a helper.

We should not obsess over capturing the RFC wire format.

So I would suggest never mind on the TODO, let's just achieve these goals by the end of this PR.

arifogel (commented):
> Reviewed 2 of 2 files at r7, 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 7 unresolved discussions (waiting on @dhalperi and @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFQOIoK2oedXs6aOHQn:b-o226zx) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*

The QoS Marking EC has a 7-byte value. The internals of that value are in draft only, but what is clear is that there is no subtype.
</blockquote></details>

Fine to defer until such time as we support ECs with 7-byte value.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 233 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQJ9xQEGrcBbcdrxvS:-NFQJ9xRBmu0qYHeHZhK:b1tp7r5) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L233)):*
> ```Java
>   /**
>    * Returns the global administrator value, if applicable to this type of extended community. May
>    * be either 2 or 4 bytes depending on type/subtype. **
> ```

nit: trailing `**` here and in doc for getLocalAdministrator
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 261 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQOyeu1SmKC_ouwWc7:-NFQOyeu1SmKC_ouwWc8:b3melmi) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L261)):*
> ```Java
>   }
> 
>   public long getValue() {
> ```

javadoc please saying it's 6 bytes
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 267 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQPWz3At1jgk91nY6f:-NFQPWz3At1jgk91nY6g:b-2w7r70) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L267)):*
> ```Java
>   @Nonnull
>   private boolean hasGlobalAndLocalAdministrator() {
>     // return false if it is an opaque extended community.
> ```

nit: it's the true case we care about
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 168 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQQCNB8A9xP3TyVcBx:-NFQQCNB8A9xP3TyVcBy:b-sf0vas) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L168)):*
> ```Java
>     assertThat(ExtendedCommunity.parse("origin:1:1").toString(), equalTo("3:1:1"));
>     assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
>     assertThat(ExtendedCommunity.parse("origin:0.0.0.0:1").toString(), equalTo("259:0L:1"));
> ```

identical to above line
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 206 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQQIQ5Cv1VtS-a_ZY0:-NFQQIQ5Cv1VtS-a_ZY1:beeh5ig) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L206)):*
> ```Java
>   public void testOpaqueCreation() {
>     ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
>     assertThat(community.getValue(), equalTo((long) 65559));
> ```

simpler: 65559L, no need for cast

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 6 unresolved discussions (waiting on @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 206 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQRaz0EBy7o_MIEQYa:-NFQRaz0EBy7o_MIEQYb:b-t548m1) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L206)):*
> ```Java
>   public void testOpaqueCreation() {
>     ExtendedCommunity community = ExtendedCommunity.opaque(true, 0x00, 65559);
>     assertThat(community.getValue(), equalTo((long) 65559));
> ```

Also test false case.

SLarkworthy (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: all files reviewed, 2 unresolved discussions (waiting on @arifogel)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 233 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQJ9xQEGrcBbcdrxvS:-NFQTJGMBsPVK4T4wsLO:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L233)):*

nit: trailing `**` here and in doc for getLocalAdministrator
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 261 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQOyeu1SmKC_ouwWc7:-NFQTriFCU64mSKV6dlY:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L261)):*

javadoc please saying it's 6 bytes
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 267 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQPWz3At1jgk91nY6f:-NFQTwfiA19APyy1NGtK:bvnrpcm) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L267)):*

nit: it's the true case we care about
</blockquote></details>

updated
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 168 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQQCNB8A9xP3TyVcBx:-NFQU4FDENJZeif_3Gcm:b-hfpv0l) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L168)):*

identical to above line
</blockquote></details>

removed
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 206 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQQIQ5Cv1VtS-a_ZY0:-NFQUBr391h50deRDD9D:bb7f2mf) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L206)):*

simpler: 65559L, no need for cast
</blockquote></details>

done
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java` line 206 at r8](https://reviewable.io/reviews/batfish/batfish/8574#-NFQRaz0EBy7o_MIEQYa:-NFQUv9rA0mhBLIpJv6D:bmgdv4d) ([raw file](https://github.com/batfish/batfish/blob/eb386a610cfab8b8932b1be6f59d236c0db6a0d3/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/bgp/community/ExtendedCommunityTest.java#L206)):*

Also test false case.
</blockquote></details>

done, and added a false test case for toString since that shows the difference in the types.

dhalperi (commented):
> Reviewed 6 of 6 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SLarkworthy)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 185 at r2](https://reviewable.io/reviews/batfish/batfish/8574#-NFMqNWQ3ff04cpC8VD1:-NFQ-CNcBwRFd-x1Ew3t:bogorfd) ([raw file](https://github.com/batfish/batfish/blob/fbbc65b181e0065a5f59cc9d5ddc215bda45e87a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L185)):*

Fine to defer until such time as we support ECs with 7-byte value.
</blockquote></details>

My 2c: I think it's fine to keep it separate for now. We can continue to evolve.

One thing we've learned in looking deeper at the RFC is that lots of parts of it are counter-intuitive. We might even want to eventually subclass EC into, e.g., ExtendedEC, RegularEC, 

So agreed with Ari, and Sarah.

arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8574)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SLarkworthy)

Issue #8567, opened 2022-10-19
Redistribution between different EIGRP processes does not advertise some routes


rynogawa:
**[Problem]**
Some routes are not advertised when redistributing routes learned by EIGRP.

**[Topology]**
![image](https://user-images.githubusercontent.com/115674092/196586587-5f53d70c-5ee0-41be-aeab-bf33babeacf1.png)

1. dev1,2 and dev3 is configured with EIGRP of AS 100.
2. dev3,4 and dev5 is configured with EIGRP of AS 200. 
3. dev3 assigns Loopback interfaces to AS100 and AS200 respectively.
4. In dev3's EIGRP settings, AS100 will redistribute AS200's route, and AS200 will redistribute AS100's route. 

**[Config] ※Minimum Config**
- dev1
```
!
interface GigabitEthernet0/1
 ip address 10.0.12.1 255.255.255.0
!
router eigrp 100
 network 10.0.12.0 0.0.0.255
!
```
- dev2
```
!
interface GigabitEthernet0/0
 ip address 10.0.12.2 255.255.255.0
!
interface GigabitEthernet0/1
 ip address 10.0.23.2 255.255.255.0
!
router eigrp 100
 network 10.0.12.0 0.0.0.255
 network 10.0.23.0 0.0.0.255
!
```
- dev3
```
interface Loopback100
 ip address 10.99.100.3 255.255.255.255
!
interface Loopback200
 ip address 10.99.200.3 255.255.255.255
!
interface GigabitEthernet0/0
 ip address 10.0.23.3 255.255.255.0
!
interface GigabitEthernet0/1
 ip address 10.0.34.3 255.255.255.0
!
router eigrp 100
 network 10.0.23.0 0.0.0.255
 network 10.99.100.3 0.0.0.0
 redistribute eigrp 200
!
router eigrp 200
 network 10.0.34.0 0.0.0.255
 network 10.99.200.3 0.0.0.0
 redistribute eigrp 100
!
```

- dev4
```
!
interface GigabitEthernet0/0
 ip address 10.0.34.4 255.255.255.0
!
interface GigabitEthernet0/1
 ip address 10.0.45.4 255.255.255.0
!
router eigrp 200
 network 10.0.34.0 0.0.0.255
 network 10.0.45.0 0.0.0.255
!
```

- dev5
```
!
interface GigabitEthernet0/0
 ip address 10.0.45.5 255.255.255.0
!
router eigrp 200
[...]

PR #8590, opened 2022-11-16
IOS:Advertise ConnectedRoute when redistributing between different EIGRP

NobutakaNiiya:
#8567
PR for the above issues.


NobutakaNiiya:
@dhalperi 
Thank you for confirming the contents.

By similar concept do you mean properties such as:

> set routing-options static route 192.168.10.1/32 discard
> set routing-options static route 192.168.20.1/32 discard
> 
> set policy-options policy-statement EXPORTSTATIC term 1 from protocol static
> set policy-options policy-statement EXPORTSTATIC term 1 from route-filter 192.168.0.0/19 orlonger
> set policy-options policy-statement EXPORTSTATIC term 1 then accept
> 
> set protocols bgp group internal-peers neighbor 10.0.12.2 export EXPORTSTATIC
> set protocols bgp group internal-peers neighbor 10.0.23.3


For Juniper's processing, should I use the following as a reference?
https://github.com/batfish/batfish/blob/6d2267a2ad50c2ceb6cd23a5f63a69851f0795c1/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L491-L495

NobutakaNiiya:
@dhalperi 
I uploaded the fix, please check it.

NobutakaNiiya:
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ConnectedRoute.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/8590#-NIhvOTE2jiEg0bzFdVF:-NRWRmAT1EcwMjL95CHE:bmvbp1p) ([raw file](https://github.com/batfish/batfish/blob/c030471e88f673dd66dcba32d0b541f799e61f3b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/ConnectedRoute.java#L22)):*

Changing `ConnectedRoute` to have EIGRP properties is a non-starter.

You need to make a different design to somehow enable the connected route to get these properties only when redistributed into EIGRP.

One place to look might be something like the `ConnectedRouteMetadata` and how that is used in the VI Model and Dataplane.

But I don't think that's precisely the right approach for the underlying problem -- because, e.g., Juniper has similar concepts where you can put BGP properties on static routes.

Need a more general solution.
</blockquote></details>

I thought of having a Map<String, EigrpInterfaceSettings> variable in the Configuration class, referencing the Juniper implementation, to store ASN information.
Then, for the ConnectedRoute case, I considered an approach to retrieve the ASN from that variable in the implementation of MatchProcessAsn.evaluate().

If this approach is not good, could you please suggest another way to solve the problem?

PR #8597, opened 2022-11-18
TransferBDD: per-path analysis

millstein:
Updates the symbolic route-map analysis (TransferBDD) to track information on a per-execution-path basis, rather than monolithically.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 198 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAq1MB8dLjD07BpDon:-NHAq1MB8dLjD07BpDoo:b6jkjte) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L198)):*
> ```Java
> 
>   /*
>    * Produce one TransferResult per path through the given boolean expression.
> ```

this surprised me -- why do we need to distinguish between paths within expressions? is it only that we want to distinguish between paths within an `CallExpr`? Would be useful to describe what the paths are abstractly



_Code quote:_
```Java
Produce one TransferResult per path through the given boolean expression.
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 201 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAneRB5UilubKT08wf:-NHAneRCBgmqa8zslN4B:bue0n4y) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L201)):*
> ```Java
>    * TODO: Any updates to the TransferParam in expr are lost currently
>    */
>   private Set<TransferResult> compute(BooleanExpr expr, TransferBDDState state)
> ```

would it make sense to return a list with a meaningful order? it seems possibly useful to know which path each `TransferResult` describes. 

(not asking that you do this now, or at all)



_Code quote:_
```Java
Set<TransferResult>
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 214 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAoExK-GZaF_0OFzzL:-NHAoExK-GZaF_0OFzzM:bm34knh) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L214)):*
> ```Java
>       allResults.add(result.setReturnValueBDD(_factory.one()).setReturnValueAccepted(true));
> 
>       /*
> ```

this is a lot of commented-out code! I assume this was not intentional? (if we really don't need it, better to just delete it)



_Code quote:_
```Java
/*
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHBA0J61-CAmMFaWz3S:b-21qpt9) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*
> ```Java
>                 .getTransferResult()
>                 // restore the callee state
>                 .setReturnAssignedValue(oldReturnAssigned));
> ```

do you need to call `id()` now that you're copying it into multiple `TransferResult`s? or is there an invariant that they will never be mutated?



_Code quote:_
```Java
oldReturnAssigned
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHBBmJYBv-jz2ydfeaV:b-hkke36) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*
> ```Java
> 
> /**
>  * The data produced by the symbolic route policy analysis performed in {@link TransferBDD}. It is a
> ```

What is the relationship between these three things?

e.g. one might expect that we'd encode rejects by having the BDDRoute map (part of) the BDD to zero. 

What does it mean if the BDDRoute maps the entire BDD to zero but the boolean says the path accepts?



_Code quote:_
```Java
 * The data produced by the symbolic route policy analysis performed in {@link TransferBDD}. It is a
 * triple representing the analysis results along a particular execution path through the route
 * policy:
 *
 * <p>1. A {@link BDDRoute} that represents a function from input announcements to the corresponding
 * output announcements produced by the analyzed route policy.
 *
 * <p>2. A {@link BDD} that represents the set of input announcements that take this particular
 * path.
 *
 * <p>3. A boolean indicating whether the path accepts or rejects the input announcement.
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 198 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAq1MB8dLjD07BpDon:-NHBKKnH67Ef4CkvIfk8:bytr4pb) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L198)):*

this surprised me -- why do we need to distinguish between paths within expressions? is it only that we want to distinguish between paths within an `CallExpr`? Would be useful to describe what the paths are abstractly
</blockquote></details>

Yes, because the expression could itself be a call to another route map, which has its own If statements, in general we want to distinguish these paths and keep them separate from one another.  In the common case where we don't have a CallExpr this code ends up producing exactly two paths, representing the conditions when this expression is true and the conditions when this expression is false.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 201 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAneRB5UilubKT08wf:-NHBKvciB7ok9-Xoq5m7:b4n3yeg) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L201)):*

would it make sense to return a list with a meaningful order? it seems possibly useful to know which path each `TransferResult` describes. 

(not asking that you do this now, or at all)
</blockquote></details>

Maybe, but it's unclear we need that, or even that an ordering would help much.  The statements are all being traced so we can use that to reconstruct the particular path through the route map that corresponds to each one.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHBM0vs7xyM7n2vmQV5:boz5ekm) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*

What is the relationship between these three things?

e.g. one might expect that we'd encode rejects by having the BDDRoute map (part of) the BDD to zero. 

What does it mean if the BDDRoute maps the entire BDD to zero but the boolean says the path accepts?
</blockquote></details>

We want to keep track of paths that reject as well as paths that accept, since both could be useful depending on the goals of a particular analysis.  So these things are all logically distinct:  the BDD is the path condition -- the predicate that ensures you go down this particular path; the boolean indicates whether this path is accepting or denying; and the BDDRoute describes the output route (in the case that the path is accepting) as a function of the input route when this path is taken.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 214 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAoExK-GZaF_0OFzzL:-NHBSAJV3qx8orslLIer:b-k7mx7q) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L214)):*

this is a lot of commented-out code! I assume this was not intentional? (if we really don't need it, better to just delete it)
</blockquote></details>

Deleted.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHBS6iZ7CMoOe7Sef0O:b-cwyva7) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

do you need to call `id()` now that you're copying it into multiple `TransferResult`s? or is there an invariant that they will never be mutated?
</blockquote></details>

Yes, the BDDs are all immutable.

anothermattbrown (commented):
Please add new tests for the new result-per-path stuff (e.g. test the correspondence between results and paths, etc).

I'm concerned about removing support for constructs we used to support. A high-level description of why that's necessary, what the plan is to restore support etc would help me review.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597#-:-NHHJbm0EGZm3n6MlPbM:b-s06kpd)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 198 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAq1MB8dLjD07BpDon:-NHHI6lw1nmN_D0aLMln:b-ckpq4b) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L198)):*

Yes, because the expression could itself be a call to another route map, which has its own If statements, in general we want to distinguish these paths and keep them separate from one another.  In the common case where we don't have a CallExpr this code ends up producing exactly two paths, representing the conditions when this expression is true and the conditions when this expression is false.
</blockquote></details>

makes sense -- please add similar explanation to the javadoc
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 201 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAneRB5UilubKT08wf:-NHHIIY0-LW8IhXS1huL:biwi7ux) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L201)):*

Maybe, but it's unclear we need that, or even that an ordering would help much.  The statements are all being traced so we can use that to reconstruct the particular path through the route map that corresponds to each one.
</blockquote></details>

seems reasonable -- again, would be good to add to the javadoc. another thing -- is the uniqueness provided `Set` meaningful? just trying to understand if/why `Set` is an important choice or not
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 214 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAoExK-GZaF_0OFzzL:-NHHIk7p63nifX-IGplr:b-iou23v) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L214)):*

Deleted.
</blockquote></details>

does that mean we no longer support those constructs? i.e. is it a regression?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHHIuKHCi22j0i3ES7C:b-z6s1jr) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

Yes, the BDDs are all immutable.
</blockquote></details>

would be good to comment that in the code for future readers (since mutation is sometimes used in this code)
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 641 at r3](https://reviewable.io/reviews/batfish/batfish/8597#-NHHHGR4223nZpYm34Ea:-NHHHGR4223nZpYm34Eb:b6yv025) ([raw file](https://github.com/batfish/batfish/blob/e170f9395210f7b2b1bc0d8aa1238adee0f56da9/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L641)):*
> ```Java
> 
>   @Test
>   public void testConjunction() {
> ```

Are you planning to add support for these constructs back in?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 685 at r3](https://reviewable.io/reviews/batfish/batfish/8597#-NHHH71w3kIPRvZIvPIh:-NHHH71w3kIPRvZIvPIi:bmhpyhs) ([raw file](https://github.com/batfish/batfish/blob/302ac40f2b68e300a1ef32bb984da4c6db8708ab/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L685)):*
> ```Java
>   }
> 
>   /* Not supporting conjunctions currently.
> ```

is this a regression? what's the plan for this?

millstein (commented):
I've added several tests for the per-path analysis.  I've added back support for Conjunction and Disjunction (along with their original tests plus new ones for the per-path analysis).

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597#-:-NHLF7Uu4KYFUvUTYcqU:b-kdkv0j)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 198 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAq1MB8dLjD07BpDon:-NHLGWSsDyxARC-n72sw:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L198)):*

makes sense -- please add similar explanation to the javadoc
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 201 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAneRB5UilubKT08wf:-NHLGczZ202SY2b_IGRy:b-7r1tjc) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L201)):*

seems reasonable -- again, would be good to add to the javadoc. another thing -- is the uniqueness provided `Set` meaningful? just trying to understand if/why `Set` is an important choice or not
</blockquote></details>

The use of sets is not in order to enforce that there are no duplicates.  By construction there should not be any duplicates, since each path has a unique path condition.  Rather, I'm using sets to indicate to the caller that there will not be any duplicates (which would not be obvious otherwise) and also that the collection of paths is unordered.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 214 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAoExK-GZaF_0OFzzL:-NHLIICC5fSLhm8-YS8Q:bgyuykj) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L214)):*

does that mean we no longer support those constructs? i.e. is it a regression?
</blockquote></details>

Re-added.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHLIL-s94-YO0xK51ev:b-y9f9h) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

would be good to comment that in the code for future readers (since mutation is sometimes used in this code)
</blockquote></details>

I'm confused about this point.  BDDs in this code are never mutated (and I didn't think that they could/should ever be mutated, other than being consumed through functions like andWith). 
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 641 at r3](https://reviewable.io/reviews/batfish/batfish/8597#-NHHHGR4223nZpYm34Ea:-NHLId5E5cFV_HnmpCc7:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/fae30507df6287bb5b412f182f9ab26905da692b/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L641)):*

Are you planning to add support for these constructs back in?
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 685 at r3](https://reviewable.io/reviews/batfish/batfish/8597#-NHHH71w3kIPRvZIvPIh:-NHLIeS1BTnyaLJXEsBo:bgyuykj) ([raw file](https://github.com/batfish/batfish/blob/781aa0be6b2a56e4f665b8aea6323fdb7b5066cf/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L685)):*

is this a regression? what's the plan for this?
</blockquote></details>

Re-added.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 201 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHAneRB5UilubKT08wf:-NHQcjG2Bz39JLlznwoz:b-xnk3td) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L201)):*

The use of sets is not in order to enforce that there are no duplicates.  By construction there should not be any duplicates, since each path has a unique path condition.  Rather, I'm using sets to indicate to the caller that there will not be any duplicates (which would not be obvious otherwise) and also that the collection of paths is unordered.
</blockquote></details>

ok -- might be good to clarify in the javadoc.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHQcJz_7QKgw35Euvws:bby56bb) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

I'm confused about this point.  BDDs in this code are never mutated (and I didn't think that they could/should ever be mutated, other than being consumed through functions like andWith). 
</blockquote></details>

yeah I'm talking about things like `andWith` -- it mutates the receiver and consumes the argument. since `BDDRoute` uses it I tend to worry about any aliasing :)
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 240 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHQBxk73Fs7QKXCQgvF:-NHQBxk73Fs7QKXCQgvG:b-9yotz) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L240)):*
> ```Java
>             nextResults.addAll(
>                 compute(e, toTransferBDDState(p.indent(), curr)).stream()
>                     .map(r -> r.setReturnValueBDD(r.getReturnValue().getSecond().and(currBDD)))
> ```

you can rewrite this using `forEach` to skip the extra collection 



_Code quote:_
```Java
            nextResults.addAll(
                compute(e, toTransferBDDState(p.indent(), curr)).stream()
                    .map(r -> r.setReturnValueBDD(r.getReturnValue().getSecond().and(currBDD)))
                    .collect(ImmutableSet.toImmutableSet()));
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 243 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHQB2bH1iogKrbMdbSC:-NHQB2bH1iogKrbMdbSD:b4i4ypt) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L243)):*
> ```Java
>                     .collect(ImmutableSet.toImmutableSet()));
>           } else {
>             // if the current state is not accepting then we've reached a false conjunct, so we
> ```

This is surprising, because we haven't yet done anything with `e`. Is the idea that we processed a false conjunct in one of the previous iterations? If so, what if the false conjunct is the last one processed? Can we move the check to the result of `compute` above? Can we pull it out of the loop entirely (i.e. can we identify a "false conjunct" without calling `compute`)? 

Also, does "false conjunct" mean always false, or falsifiable?



_Code quote:_
```Java
            // if the current state is not accepting then we've reached a false conjunct, so we
            // short-circuit the rest of the computation
            allResults.add(curr);
```

anothermattbrown (commented):
> Reviewed 2 of 5 files at r1, 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8597#-NHT23LE5aHobBu1pIMn:-NHT23LE5aHobBu1pIMo:b-bushlz) (no related file):*
I'm wondering if the exponential blow-up could cause problems for use cases (like SRP) where we don't need separate results per path. 

If we track "path IDs" using BDDs, then we could turn it off when we don't need it (by giving all paths the same ID). 

___
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 482 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHSyHP98NV0I7zJ4Zt3:-NHSyHP98NV0I7zJ4Zt4:b-39soph) ([raw file](https://github.com/batfish/batfish/blob/fae30507df6287bb5b412f182f9ab26905da692b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L482)):*
> ```Java
>           result =
>               result.setReturnAssignedValue(
>                   ite(unreachable(result), result.getReturnAssignedValue(), _factory.one()));
> ```

why did this change?



_Code quote:_
```Java
          result =
              result.setReturnAssignedValue(
                  ite(unreachable(result), result.getReturnAssignedValue(), _factory.one()));
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 536 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHSzjoa50hvEfxgfCJ3:-NHSzjoa50hvEfxgfCJ4:b-y98qnn) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L536)):*
> ```Java
>       }
> 
>       return ImmutableSet.copyOf(newStates);
> ```

it would be good to re-standardize the cases here, so they all either return or fall-through. preferable return, since that would get us closer to converting this huge if-else chain to a visitor.

if you choose to mix return and fall-through, it would help to have a comment explaining why 



_Code quote:_
```Java
return ImmutableSet.copyOf(newStates);
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 1101 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHT1fJbBif0R8I1SCJ4:-NHT1fJbBif0R8I1SCJ5:b-q93j5e) ([raw file](https://github.com/batfish/batfish/blob/fae30507df6287bb5b412f182f9ab26905da692b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1101)):*
> ```Java
>     TransferReturn ret = new TransferReturn(r.getReturnValue().getFirst(), b);
>     BDD exitAsgn = ite(notReached, r.getExitAssignedValue(), _factory.one());
>     return r.setReturnValue(ret).setExitAssignedValue(exitAsgn);
> ```

it's nice that all this `ite` stuff is getting simplified away!



_Code quote:_
```Java
  private TransferResult exitValue(TransferResult r, BDD val) {
    BDD notReached = unreachable(r);
    BDD b = ite(notReached, r.getReturnValue().getSecond(), val);
    TransferReturn ret = new TransferReturn(r.getReturnValue().getFirst(), b);
    BDD exitAsgn = ite(notReached, r.getExitAssignedValue(), _factory.one());
    return r.setReturnValue(ret).setExitAssignedValue(exitAsgn);
```

millstein (commented):
I will add back ConjunctionChain and FirstMatchChain support, so that there is no regression.  Currently in progress.

It turns out that your instinct about using Sets was correct -- since BDDRoutes are mutable it's not a good idea to use sets here.  So I switched to lists but still explain in the javadoc comments that the paths are logically unordered.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597#-:-NHSdpoi2eKN3qtgciPH:bp103iq)* status: 3 of 5 files reviewed, 5 unresolved discussions (waiting on @anothermattbrown)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8597#-NHT23LE5aHobBu1pIMn:-NHV61bxByoehpYtb_bH:brje4v2) (no related file):*

I'm wondering if the exponential blow-up could cause problems for use cases (like SRP) where we don't need separate results per path. 

If we track "path IDs" using BDDs, then we could turn it off when we don't need it (by giving all paths the same ID). 
</blockquote></details>

We could consider that down the road, but there's not an exponential blow-up in practice, from what I've seen.  For example, in a typical case where you have N stanzas, each of which permits or denies, then there are just N+1 paths (each of them plus not matching anything).  

___
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHR1J0KD_6g-9UojfhI:buuvlb5) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

yeah I'm talking about things like `andWith` -- it mutates the receiver and consumes the argument. since `BDDRoute` uses it I tend to worry about any aliasing :)
</blockquote></details>

got it.  BDDRoute was only using andWith in one place and I'm not sure if it is safe, so I changed them to use and. so now as far as I can tell TransferBDD never mutates BDDs anywhere.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 243 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHQB2bH1iogKrbMdbSC:-NHR07pN28XRGh_oC7Ol:bbiily9) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L243)):*

This is surprising, because we haven't yet done anything with `e`. Is the idea that we processed a false conjunct in one of the previous iterations? If so, what if the false conjunct is the last one processed? Can we move the check to the result of `compute` above? Can we pull it out of the loop entirely (i.e. can we identify a "false conjunct" without calling `compute`)? 

Also, does "false conjunct" mean always false, or falsifiable?
</blockquote></details>

I moved the check to the result of compute.  It can't be pulled out of the loop since we have to check for a false conjunct after each conjunct is evaluated.  "false conjunct" means always false -- since we are doing a per-path analysis, on each path there is a definite result for each conjunct.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 482 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHSyHP98NV0I7zJ4Zt3:-NHV6ydy6t_D3Uq5Wbm7:b-ao2vsv) ([raw file](https://github.com/batfish/batfish/blob/fae30507df6287bb5b412f182f9ab26905da692b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L482)):*

why did this change?
</blockquote></details>

We don't need the "if unreachable" everywhere any more, since that is handled in a single place now, before we start the symbolic execution of a statement. 
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 536 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHSzjoa50hvEfxgfCJ3:-NHVKUWz433-Icl94hfr:b-rjr9rm) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L536)):*

it would be good to re-standardize the cases here, so they all either return or fall-through. preferable return, since that would get us closer to converting this huge if-else chain to a visitor.

if you choose to mix return and fall-through, it would help to have a comment explaining why 
</blockquote></details>

Good point.  I updated this to return from each branch.

anothermattbrown (commented):
> Reviewed 1 of 5 files at r1, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: all files reviewed, 4 unresolved discussions (waiting on @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8597#-NHT23LE5aHobBu1pIMn:-NHVUij507YEe9uOqvae:b-pqov82) (no related file):*

We could consider that down the road, but there's not an exponential blow-up in practice, from what I've seen.  For example, in a typical case where you have N stanzas, each of which permits or denies, then there are just N+1 paths (each of them plus not matching anything).  
</blockquote></details>

makes sense

___
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 307 at r5](https://reviewable.io/reviews/batfish/batfish/8597#-NHVUsIc-MxdZjiX9HSP:-NHVUsIc-MxdZjiX9HSQ:b-u709zb) ([raw file](https://github.com/batfish/batfish/blob/11c84d42aa37d6c93665a15a923e359c75e60754/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L307)):*
> ```Java
>       for (int j = i + 1; j < _asPathRegexAtomicPredicates.length; j++) {
>         asPathConstraint =
>             asPathConstraint.and(
> ```

why did we stop using `andWith` here? seems completely safe, since everything being mutated was clearly created in this method. 



_Code quote:_
```Java
        asPathConstraint =
            asPathConstraint.and(
                _asPathRegexAtomicPredicates[i].nand(_asPathRegexAtomicPredicates[j]));
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 315 at r5](https://reviewable.io/reviews/batfish/batfish/8597#-NHVVCUt4eIqpbXJMu3H:-NHVVCUt4eIqpbXJMu3I:b-s7nar4) ([raw file](https://github.com/batfish/batfish/blob/11c84d42aa37d6c93665a15a923e359c75e60754/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L315)):*
> ```Java
>     BDD nextHopConstraint = _nextHop.range(Ip.ZERO.asLong() + 1, Ip.MAX.asLong() - 1);
> 
>     return protocolConstraint.and(prefLenConstraint).and(asPathConstraint).and(nextHopConstraint);
> ```

andWith seems safe here too (assuming protocolConstraint is freshly allocated). can also use `andAllAndFree`



_Code quote:_
```Java
.and(prefLenConstraint).and(asPathConstraint).and(nextHopConstraint);
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHVZmjcAgQIJwWkuqhE:bl4q8wx) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

got it.  BDDRoute was only using andWith in one place and I'm not sure if it is safe, so I changed them to use and. so now as far as I can tell TransferBDD never mutates BDDs anywhere.
</blockquote></details>

ok -- well, those uses looked safe to me -- maybe my concern was unwarranted. I think those changes can be reverted, but up to you
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 243 at r4](https://reviewable.io/reviews/batfish/batfish/8597#-NHQB2bH1iogKrbMdbSC:-NHVVbvU8C8I6xgxBQCF:b-dc1zo5) ([raw file](https://github.com/batfish/batfish/blob/17bbd3aa827ac23c0ce702dd79442fdf934c7cca/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L243)):*

I moved the check to the result of compute.  It can't be pulled out of the loop since we have to check for a false conjunct after each conjunct is evaluated.  "false conjunct" means always false -- since we are doing a per-path analysis, on each path there is a definite result for each conjunct.
</blockquote></details>

makes sense, thanks
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHV_hw9DRyOkCsw6NpS:b-i4hto6) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*

We want to keep track of paths that reject as well as paths that accept, since both could be useful depending on the goals of a particular analysis.  So these things are all logically distinct:  the BDD is the path condition -- the predicate that ensures you go down this particular path; the boolean indicates whether this path is accepting or denying; and the BDDRoute describes the output route (in the case that the path is accepting) as a function of the input route when this path is taken.
</blockquote></details>

I see -- so what does the BDDRoute look like when the path is rejecting? Is there an invariant that it be `null` or somehow trivial in that case? If yes, can you document/enforce that invariant?  Could also combine the BDDRoute and the boolean into an `Optional<BDDRoute>`. 
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 19 at r5](https://reviewable.io/reviews/batfish/batfish/8597#-NHVbZYs5jz7mUMHv79k:-NHVbZYs5jz7mUMHv79l:b9b76lq) ([raw file](https://github.com/batfish/batfish/blob/11c84d42aa37d6c93665a15a923e359c75e60754/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L19)):*
> ```Java
>  * <p>3. A boolean indicating whether the path accepts or rejects the input announcement.
>  */
> public class TransferReturn extends Tuple<BDDRoute, BDD> {
> ```

I don't think we're getting any value out of extending this anymore :)



_Code quote:_
```Java
Tuple<BDDRoute, BDD>
```
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 206 at r5](https://reviewable.io/reviews/batfish/batfish/8597#-NHVbylg5Hu7AKGAmQg3:-NHVbylg5Hu7AKGAmQg4:bqi3ogj) ([raw file](https://github.com/batfish/batfish/blob/11c84d42aa37d6c93665a15a923e359c75e60754/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L206)):*
> ```Java
> 
>   @Test
>   public void testCaptureAllPaths() {
> ```

I don't understand this name. Suggestion: `testComputePaths_ExitAccept`. 

And:
`testCaptureNone` --> `testCompute_ExitReject`
`testCaptureNonePaths` --> `testComputePaths_ExitReject`
`testOnePrefixRangePaths` --> `testComputePaths_If`
etc

edit: looks like elsewhere the convention is `test<StatementVariant>` when the method under test is `compute` and `test<StatementVariant>` when the method under test is `computePaths` (which makes sense to me). In that case, should this be `testExitAcceptPaths`?



_Code quote:_
```Java
testCaptureAllPaths
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 361 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBA0J61-CAmMFaWz3R:-NHWkAhv4PsvsajZ5DJK:bvth56w) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L361)):*

ok -- well, those uses looked safe to me -- maybe my concern was unwarranted. I think those changes can be reverted, but up to you
</blockquote></details>

reverted
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHWcoS02__3qevztUUW:b9foyz1) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*

I see -- so what does the BDDRoute look like when the path is rejecting? Is there an invariant that it be `null` or somehow trivial in that case? If yes, can you document/enforce that invariant?  Could also combine the BDDRoute and the boolean into an `Optional<BDDRoute>`. 
</blockquote></details>

The BDDRoute will still contain all the updates made along that path, even if it is rejecting.  For example, nothing prevents a stanza from updating the local pref and then rejecting.  The updates won't actually happen since the route is rejected, but still we record it (and that info may be useful for some clients potentially).  Notably we also record when we've hit an unsupported statement, which is useful to help clients know when the results of the analysis may not be accurate.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 206 at r5](https://reviewable.io/reviews/batfish/batfish/8597#-NHVbylg5Hu7AKGAmQg3:-NHWkF0aDXsIdORZI56U:birzovy) ([raw file](https://github.com/batfish/batfish/blob/11c84d42aa37d6c93665a15a923e359c75e60754/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L206)):*

I don't understand this name. Suggestion: `testComputePaths_ExitAccept`. 

And:
`testCaptureNone` --> `testCompute_ExitReject`
`testCaptureNonePaths` --> `testComputePaths_ExitReject`
`testOnePrefixRangePaths` --> `testComputePaths_If`
etc

edit: looks like elsewhere the convention is `test<StatementVariant>` when the method under test is `compute` and `test<StatementVariant>` when the method under test is `computePaths` (which makes sense to me). In that case, should this be `testExitAcceptPaths`?
</blockquote></details>

updated as you suggested

anothermattbrown (commented):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHz9SVG3BTmGEUK87hJ:b-gnjc5g) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*

The BDDRoute will still contain all the updates made along that path, even if it is rejecting.  For example, nothing prevents a stanza from updating the local pref and then rejecting.  The updates won't actually happen since the route is rejected, but still we record it (and that info may be useful for some clients potentially).  Notably we also record when we've hit an unsupported statement, which is useful to help clients know when the results of the analysis may not be accurate.
</blockquote></details>

makes sense -- would be helpful to include that explanation in the javadoc, and add a test of that behavior (if you haven't already)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/8597#-NHBBmJYBv-jz2ydfeaU:-NHz_cP3596cv8l2dznJ:bbd59qj) ([raw file](https://github.com/batfish/batfish/blob/a78f8e70869d03bbc9a09a009fb7ec3f017f3e6e/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferReturn.java#L7)):*

makes sense -- would be helpful to include that explanation in the javadoc, and add a test of that behavior (if you haven't already)
</blockquote></details>

Updated the javadoc and added a test for doing route updates and then rejecting.  There is already an existing test for hitting an unsupported statement and then rejecting.

anothermattbrown (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8597)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

PR #8612, opened 2022-12-02
Compare routing policies

nickgian:
This PR implements a question to perform a symbolic diff between two route-maps.

To compare RM1 and RM2 we rely on computing every path (= condition on the input route + set of output routes) through them using `TransferBDD` and comparing every path where there is some overlap between their path conditions (input routes).


projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java
```diff
@@ -0,0 +1,212 @@
+package org.batfish.minesweeper.bdd;
```
nickgian, line 1: This whole file is merely a refactoring of code that used to live in `SearchRoutePoliciesAnswerer`.

anothermattbrown (commented):
> Reviewed 13 of 17 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: 12 of 17 files reviewed, 7 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 249 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKI-Wn4aYO9rDwOOR5:-NIKI-Wn4aYO9rDwOOR6:b-9o6mq2) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L249)):*
> ```Java
>     _factory = route._factory;
> 
>     // We do not need to clone the atomic predicates.
> ```

why not? (please elaborate in the comment)



_Code quote:_
```Java
  // We do not need to clone the atomic predicates.
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java` line 8 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKF2YFEVTO8PQO6kqY:-NIKF2YFEVTO8PQO6kqZ:b-45uvnj) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java#L8)):*
> ```Java
> import java.util.Objects;
> 
> public class BDDRouteDiff {
> ```

please make this class `final` and add javadoc. 



_Code quote:_
```Java
public class BDDRouteDiff {
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java` line 49 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKEnJ2282-cfsdnWhj:-NIKEnJ2282-cfsdnWhk:b7utlqp) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java#L49)):*
> ```Java
>   }
> 
>   public List<Difference> computeDifferences(BDDRoute r1, BDDRoute r2) {
> ```

should this be static? if yes, please also add a private constructor to prevent instantiation of this class.

Also, it seems like you could just return a list of `DifferenceType`, since the caller already knows about `r1` and `r2`.



_Code quote:_
```Java
computeDifferences
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 658 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKFdk4ECL88P8qA4K0:-NIKFdk4ECL88P8qA4K1:bpgegom) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L658)):*
> ```Java
>       } else {
>         curP.indent().debug("Value: E2");
>         newValue.setValue(OspfType.E2);
> ```

nice. can you add a unit test for this? Also -- in the future, it would be good to send fixes like this in their own PRs



_Code quote:_
```Java
newValue.setValue(OspfType.E2);
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 221 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKIxA3-c-JhThM8lOz:-NIKIxA3-c-JhThM8lP-:b-3chhts) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L221)):*
> ```Java
>           } else {
>             // If both policies perform the same action, then check that their outputs match.
>             // We compute the outputs of interest, by restricting the sets of output routes to the
> ```

is this valuable when the action is reject?



_Code quote:_
```Java
            // If both policies perform the same action, then check that their outputs match.
            // We compute the outputs of interest, by restricting the sets of output routes to the
            // intersection
            // of the input routes and then comparing them.
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 277 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKKXRf40X_Rbt0RTar:-NIKKXRf40X_Rbt0RTas:bjdvsfu) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L277)):*
> ```Java
>   public AnswerElement answer(NetworkSnapshot snapshot) {
>     SpecifierContext context = _batfish.specifierContext(snapshot);
>     List<Row> rows =
> ```

it's usually important to sort rows so the output is deterministic, especially if we add a ref test or notebook for this. e.g. you could create a new class for your row type and make it comparable. Then have `comparePoliciesForNode` return a stream of those things, sort it, and then convert to `Row`.

Since no two rows will have the same route, you could probably sort based on that. But maybe there's a better ordering, e.g. (line number 1, line number 2).



_Code quote:_
```Java
List<Row> rows =
```
___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 121 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKNP968pV5CfxDHoDc:-NIKNP968pV5CfxDHoDd:bsonowo) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L121)):*
> ```Java
>    * @return a table row containing the differences of the simulation
>    */
>   public static Row diffRowResultsFor(
> ```

deferrable: would be better to move this (and the other public helpers) out of this class/package. e.g. see `FilterQuestionUtils`.

anothermattbrown (commented):
> Reviewed 4 of 17 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: all files reviewed, 8 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 171 at r3](https://reviewable.io/reviews/batfish/batfish/8612#-NIKPjBpDYdaUU94eNVw:-NIKPjBpDYdaUU94eNVx:b32ge68) ([raw file](https://github.com/batfish/batfish/blob/3cf1990ba93e999624ac09132ffa3555d5e3942d/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L171)):*
> ```Java
>             }
>           }
>       }
> ```

please add a default case and `throw` :)

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 277 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKKXRf40X_Rbt0RTar:-NIUZcCQ2sedG78t1C6p:b-35jag9) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L277)):*

it's usually important to sort rows so the output is deterministic, especially if we add a ref test or notebook for this. e.g. you could create a new class for your row type and make it comparable. Then have `comparePoliciesForNode` return a stream of those things, sort it, and then convert to `Row`.

Since no two rows will have the same route, you could probably sort based on that. But maybe there's a better ordering, e.g. (line number 1, line number 2).
</blockquote></details>

I am a bit puzzled as to how approach this.
It seems that `TableAnswerElement` uses the class `Rows` which explicitly uses an unordered structure (`Multiset`) to store `Row` objects.
How do other questions handle determinism? Should this be a larger refactoring where `Multiset` is replaced by `List` and `TableAnswerElement` implements a `sort` method?

Alternatively, I can try to sort the answers as they are generated before they are converted into table rows. This assumes that iterating the `Multiset` is deterministic - which it is for the implementation used (`LinkedHashSet`), but otherwise not guaranteed by the Interface of `Multiset`. I think the latter is outside the scope of this PR though.

projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
```diff
@@ -972,6 +976,10 @@ BDDRoute ite(BDD guard, BDDRoute r1, BDDRoute r2) {
 
     ret.setUnsupported(ite(guard, r1.getUnsupported(), r2.getUnsupported()));
 
+    MutableBDDInteger i =
```
nickgian, line 979: This was missing for setOspfMetric to work properly (other cases too).
I believe this is not on the code path for `CompareRoutingPolicies`, only SRP, hence why my unit tests for comparison did not uncover the bug.
The unit test added for `TransferBDD` caught this.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: 12 of 17 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 221 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKIxA3-c-JhThM8lOz:-NIYii2m9Hs_6WU1FlsQ:b-ub5be7) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L221)):*

is this valuable when the action is reject?
</blockquote></details>

Adapted this to only do the comparison if the paths were permitted.

anothermattbrown (commented):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: all files reviewed, 3 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 277 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKKXRf40X_Rbt0RTar:-NI_8z0SBJM1nKhyyNue:b-wv4922) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L277)):*

I am a bit puzzled as to how approach this.
It seems that `TableAnswerElement` uses the class `Rows` which explicitly uses an unordered structure (`Multiset`) to store `Row` objects.
How do other questions handle determinism? Should this be a larger refactoring where `Multiset` is replaced by `List` and `TableAnswerElement` implements a `sort` method?

Alternatively, I can try to sort the answers as they are generated before they are converted into table rows. This assumes that iterating the `Multiset` is deterministic - which it is for the implementation used (`LinkedHashSet`), but otherwise not guaranteed by the Interface of `Multiset`. I think the latter is outside the scope of this PR though.
</blockquote></details>

Yes the multiset implementation used (`LinkedHashMultiset`) uses insertion order. I agree that modifying `TableAnswerElement` is out of scope. The best approach is to sort your list before calling `postProcessAnswer` -- that will determine the row order.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: 11 of 17 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 277 at r2](https://reviewable.io/reviews/batfish/batfish/8612#-NIKKXRf40X_Rbt0RTar:-NIdjXLJ6EyK9AIK4dN2:b-t4jmu4) ([raw file](https://github.com/batfish/batfish/blob/e843bb20fa70e135f558157c309ee0b0e9efb8c2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L277)):*

Yes the multiset implementation used (`LinkedHashMultiset`) uses insertion order. I agree that modifying `TableAnswerElement` is out of scope. The best approach is to sort your list before calling `postProcessAnswer` -- that will determine the row order. 
</blockquote></details>

The results are sorted by node, policy and then based on the prefix of the input route. Ideally they would be sorted based on the full input BGP route but comparison for BgpRoute is not defined.
Still, the result should be deterministic assuming SRP is deterministic (I believe a valid assumption).

anothermattbrown (commented):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)

nickgian:
> > Reviewed 9 of 9 files at r6, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)_ status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)
> 
> _[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 442 at r6](https://reviewable.io/reviews/batfish/batfish/8612#-NImiIWwBp8KP66l45pm:-NImiIWwBp8KP66l45pn:bgbst7s) ([raw file](https://github.com/batfish/batfish/blob/dea7e09191127d3db32d77ccd2d24ab7dca012b2/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L442)):_
> 
> > ```java
> >   }
> > 
> >   private static @Nullable Row toCompareRow(Result currentResult, Result proposedResult) {
> > ```
> 
> we have two conventions for terminology around snapshot comparison: base/delta (the old convention) and current/reference (the new convention). Reference means past, last-known-good, etc. It would be good to follow the current/reference convention
> 
> _Code quote:_
> 
> ```java
> Result currentResult, Result proposedResult
> ```

With the latest commit, at least for customer-facing names that is the case; `reference` is last-known-good, `snapshot` is the one you are trying out now (it was the other way around, Ratul brought it up).

I will adjust the variable names to conform to this convention too.

anothermattbrown (commented):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8612)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickgian)

PR #8615, opened 2022-12-06
TransferBDD: better handling for extended/large communities

millstein:
Updates support for extended/large community literals in the symbolic route-map analysis:
* separate handling of these communities from the handling of standard communities/regexes, so they don't get confused and so we match their treatment by the Batfish routing simulation
* support for matching on opaque extended communities


anothermattbrown (commented):
> Reviewed 1 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8615)* status: 1 of 12 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java` line 74 at r1](https://reviewable.io/reviews/batfish/batfish/8615#-NIcQp6D3Cg7rh6kM1hZ:-NIcQp6D3Cg7rh6kM1h_:b-k2ei1f) ([raw file](https://github.com/batfish/batfish/blob/83a2eeab9427587b6930f205c5d1fc0651714b68/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java#L74)):*
> ```Java
>     Predicate<CommunityVar> isStandardCommunity =
>         cvar ->
>             cvar.getType() == CommunityVar.Type.REGEX
> ```

Why does type `REGEX` imply it's a standard community? Are regexes not supported for extended communities?



_Code quote:_
```Java
   cvar.getType() == CommunityVar.Type.REGEX
```

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8615)* status: 1 of 12 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java` line 74 at r1](https://reviewable.io/reviews/batfish/batfish/8615#-NIcQp6D3Cg7rh6kM1hZ:-NIcXvUs62Zi_CC7olnL:br81zna) ([raw file](https://github.com/batfish/batfish/blob/83a2eeab9427587b6930f205c5d1fc0651714b68/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java#L74)):*

Batfish in general does not currently support regexes for extended communities, as far as I understand, and we are following that here as well.  I can add a comment to this effect.
</blockquote></details>

thanks, that would help. something to think about: what will happen here if batfish starts supporting regexes for extended communities? can we easily check that the assumption (that it does not) holds?

anothermattbrown (commented):
> Reviewed 11 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8615)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java` line 195 at r1](https://reviewable.io/reviews/batfish/batfish/8615#-NId5vwvETDJDfMpANkg:-NId5vww1ylfvb6sQdXm:b-r0vyks) ([raw file](https://github.com/batfish/batfish/blob/83a2eeab9427587b6930f205c5d1fc0651714b68/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java#L195)):*
> ```Java
>               assert c.getLiteralValue() != null;
>               ExtendedCommunity ec = (ExtendedCommunity) c.getLiteralValue();
>               return ec.isOpaque()
> ```

can you comment this code? are there remaining unsupported features? 



_Code quote:_
```Java
              return ec.isOpaque()
                  && ec.isTransitive() == opaqueExtendedCommunities.getIsTransitive()
                  && ec.getSubtype() == opaqueExtendedCommunities.getSubtype();
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8615)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java` line 74 at r1](https://reviewable.io/reviews/batfish/batfish/8615#-NIcQp6D3Cg7rh6kM1hZ:-NId8LO968sSbeiJdli2:bb0our0) ([raw file](https://github.com/batfish/batfish/blob/83a2eeab9427587b6930f205c5d1fc0651714b68/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java#L74)):*

thanks, that would help. something to think about: what will happen here if batfish starts supporting regexes for extended communities? can we easily check that the assumption (that it does not) holds?
</blockquote></details>

I added a comment.  In terms of your question, I think that's even much broader than extended communities -- how do we keep this code in sync with the route simulation as there are new bug fixes, features, etc. there?  One idea is to have a set of common tests and check that they are consistent with one another.  In principle every TRP test can also be an SRP test, where we use the same route map and enforce constraints on the input route to be equal to a specific concrete route.

In this particular case I'm not sure what would easily check changes to extended communities, particularly because right now the Batfish behavior is more "whatever it happens to do" rather than some specific intended behavior, as far as I can tell (e.g., there are no tests for extended communities matching regexes).
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java` line 195 at r1](https://reviewable.io/reviews/batfish/batfish/8615#-NId5vwvETDJDfMpANkg:-NId9VSg5AJ0R24mD7oJ:b-9nokgf) ([raw file](https://github.com/batfish/batfish/blob/83a2eeab9427587b6930f205c5d1fc0651714b68/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java#L195)):*

can you comment this code? are there remaining unsupported features? 
</blockquote></details>

This code is complete.  It's exactly analogous to the concrete version in CommunityMatchExprEvaluator::visitOpaqueExtendedCommunities.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8615)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

PR #8618, opened 2022-12-08
TransferBDD: Ignore directives about intermediate BGP attributes

millstein:
Updates the symbolic route analysis to ignore directives about intermediate BGP attributes.  These directives are not well document/tested, but from discussions their main purpose is to properly handle the semantics of read-write operations like additive community sets and AS-path prepends (which implicitly both read and write the current state).  We are already properly accounting for this semantics so do not need to include the notion of intermediate attributes here.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8618)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8622, opened 2022-12-16
SearchRoutePolicies: Support per-path analysis

millstein:
Updates the SearchRoutePolicies question to allow getting (up to) one result per execution path, rather than one result in total.  This is done by migrating to the new computePaths API of TransferBDD analysis; the old API has been removed and the analysis also simplified a bit as a result.


nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8622)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 124 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKRFmpG1bzmaB9uU0je:-NKRFmpG1bzmaB9uU0jf:b-n9n2uu) ([raw file](https://github.com/batfish/batfish/blob/56fc446c01d62b90beaa7751d99984dba0cecac2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L124)):*
> ```Java
>   // whether the analysis encountered an unsupported route-policy
>   // statement/expression
>   private boolean _unsupported;
> ```

I think eventually (but likely not as part of this PR), this should be a list of unsupported statements so we can do better error reporting and not completely discard paths with unsupported features.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 316 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKRIIV6DHIOM8W7IdZX:-NKRIIV6DHIOM8W7IdZY:b-7ltpve) ([raw file](https://github.com/batfish/batfish/blob/56fc446c01d62b90beaa7751d99984dba0cecac2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L316)):*
> ```Java
>             // only search for models on paths where no unsupported route-policy feature was
>             // encountered
>             .filter(p -> !p.getFirst().getUnsupported())
> ```

We should probably avoid filtering here, at least not like this.
There are two choices:
1. We consider paths with unsupported features. This might mean our analysis is not always precise/sound. On the other hand, you do get some "coverage" over such paths and the features that are not supported might not be of interest (especially if this is some set statement for a BGP feature we do not care about).
2. We completely drop paths with unsupported features. I am not sure this buys us anything. If my SRP query returns no results because of unsupported features this is more misleading than getting some path that might not be 100% correct, i.e., ignoring paths doesn't make the analysis sound, it just gives a false sense of security.

I would suggest we do not filter here, and we do a better job of reporting unsupported features during the analysis by incorporating the information in the results (which is not super trivial because there's also TRP).
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 338 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKRMuP94has94U6rJ7l:-NKRMuPA5wDb1tSMMmBu:br52kwi) ([raw file](https://github.com/batfish/batfish/blob/56fc446c01d62b90beaa7751d99984dba0cecac2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L338)):*
> ```Java
>         if (!_perPath) {
>           // return the first result we find
>           break;
> ```

you can also do `return builder.build();`. Not sure which one is better.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8622)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 316 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKRIIV6DHIOM8W7IdZX:-NKTcsHtE3PixQ9l7hM9:byndovc) ([raw file](https://github.com/batfish/batfish/blob/56fc446c01d62b90beaa7751d99984dba0cecac2/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L316)):*

We should probably avoid filtering here, at least not like this.
There are two choices:
1. We consider paths with unsupported features. This might mean our analysis is not always precise/sound. On the other hand, you do get some "coverage" over such paths and the features that are not supported might not be of interest (especially if this is some set statement for a BGP feature we do not care about).
2. We completely drop paths with unsupported features. I am not sure this buys us anything. If my SRP query returns no results because of unsupported features this is more misleading than getting some path that might not be 100% correct, i.e., ignoring paths doesn't make the analysis sound, it just gives a false sense of security.

I would suggest we do not filter here, and we do a better job of reporting unsupported features during the analysis by incorporating the information in the results (which is not super trivial because there's also TRP).
</blockquote></details>

Right now what SRP does is #2 above, plus we also log unsupported features as they are encountered during the analysis.  This is not a new behavior for this PR.  We have discussed alternatives, for example including all results but adding an extra column to the output that lists any encountered unsupported features.  I think that's similar to what you're suggesting, and I agree it is worth revisiting.  But probably that should be done as a separate PR since it's not related to the per-path change.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8622)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 314 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKtq1l70pA_64Bn_Qlj:-NKtq1l70pA_64Bn_Qlk:b-u9g4q7) ([raw file](https://github.com/batfish/batfish/blob/6d2267a2ad50c2ceb6cd23a5f63a69851f0795c1/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L314)):*
> ```Java
>       BDD outConstraints = routeConstraintsToBDD(_outputConstraints, outputRoute, true, configAPs);
>       intersection = acceptedAnnouncements.and(inConstraints).and(outConstraints);
>     } else {
> ```

The code no longer negates `acceptedAnnouncements` when `_action` is not `PERMIT`. Can you explain that change? Is it because `path.getSecond()` now returns the denied announcements when action is `DENY`? In that case, the name `acceptedAnnouncements` is a bit confusing, because I assume `accepted` means `permitted`. Perhaps `pathAnnouncements` would be a better name?



_Code quote:_
```Java
    } else {
      intersection = acceptedAnnouncements.not().and(inConstraints);
    }
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8622)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 314 at r1](https://reviewable.io/reviews/batfish/batfish/8622#-NKtq1l70pA_64Bn_Qlj:-NL-ILHn-CajF7YNNgpY:b-pf0vs2) ([raw file](https://github.com/batfish/batfish/blob/6d2267a2ad50c2ceb6cd23a5f63a69851f0795c1/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L314)):*

The code no longer negates `acceptedAnnouncements` when `_action` is not `PERMIT`. Can you explain that change? Is it because `path.getSecond()` now returns the denied announcements when action is `DENY`? In that case, the name `acceptedAnnouncements` is a bit confusing, because I assume `accepted` means `permitted`. Perhaps `pathAnnouncements` would be a better name?
</blockquote></details>

Good point.  Yes, the BDD now always represents the set of announcements that go down the current execution path, regardless of whether the path accepts or denies.  I have updated the name as you suggested.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8622)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8625, opened 2022-12-27
SearchRoutePolicies - only build AtomicPredicates for relevant policies.

nickgian:
Updates the SearchRoutePolicies/CompareRoutePolicies questions to only compute the AtomicPredicates for the set of policies that are used in the question instead of the whole file. This is done for performance reasons; computing APs in configs with multiple regex based AS-path/community definitions can be very expensive.

Updates community/AS visitors to recursively visit called policies. This assumes routing policies do not include mutual recursion (a valid assumption AFAIK), otherwise the visitors would loop forever.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java
@@ -1,8 +1,11 @@
 package org.batfish.minesweeper;
 
 import com.google.common.collect.ImmutableSet;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
@@ -15,6 +18,7 @@ import org.batfish.datamodel.routing_policy.RoutingPolicy;
 import org.batfish.datamodel.routing_policy.statement.Statement;
 import org.batfish.minesweeper.aspath.RoutePolicyStatementAsPathCollector;
 import org.batfish.minesweeper.communities.RoutePolicyStatementVarCollector;
+import org.batfish.minesweeper.utils.Tuple;
 
 /**
  * This class computes the community-regex and AS-path-regex atomic predicates for a single router
@@ -58,16 +62,20 @@ public class ConfigAtomicPredicates {
    * @param router the name of the router whose configuration is being analyzed
    * @param communities additional community regexes to track, from user-defined constraints
    * @param asPathRegexes additional as-path regexes to track, from user-defined constraints
+   * @param policies the set of policies to create AtomicPredicates for
    */
   public ConfigAtomicPredicates(
       IBatfish batfish,
       NetworkSnapshot snapshot,
       String router,
       @Nullable Set<CommunityVar> communities,
-      @Nullable Set<String> asPathRegexes) {
+      @Nullable Set<String> asPathRegexes,
+      @Nullable Collection<RoutingPolicy> policies) {
     _configuration = batfish.loadConfigurations(snapshot).get(router);
+    Collection<RoutingPolicy> usedPolicies =
+        policies == null ? _configuration.getRoutingPolicies().values() : policies;
 
-    Set<CommunityVar> allCommunities = findAllCommunities(communities);
+    Set<CommunityVar> allCommunities = findAllCommunities(communities, usedPolicies);
 
     // currently we only support regex matching for standard communities
     Predicate<CommunityVar> isStandardCommunity =
@@ -90,19 +98,37 @@ public class ConfigAtomicPredicates {
     for (int i = 0; i < nonStandardCommunityVars.length; i++) {
       _nonStandardCommunityLiterals.put(i + numAPs, nonStandardCommunityVars[i]);
     }
-
     _asPathRegexAtomicPredicates =
         new RegexAtomicPredicates<>(
-            findAllAsPathRegexes(asPathRegexes), SymbolicAsPathRegex.ALL_AS_PATHS);
+            findAllAsPathRegexes(asPathRegexes, usedPolicies), SymbolicAsPathRegex.ALL_AS_PATHS);
   }
 
   /**
-   * Identifies all of the community literals and regexes in the given configurations. An optional
+   * Compute atomic predicates for the given router's configuration.
+   *
+   * @param batfish the batfish object
+   * @param snapshot the current snapshot
+   * @param router the name of the router whose configuration is being analyzed
+   * @param communities additional community regexes to track, from user-defined constraints
+   * @param asPathRegexes additional as-path regexes to track, from user-defined constraints
+   */
+  public ConfigAtomicPredicates(
+      IBatfish batfish,
+      NetworkSnapshot snapshot,
+      String router,
+      @Nullable Set<CommunityVar> communities,
+      @Nullable Set<String> asPathRegexes) {
+    this(batfish, snapshot, router, communities, asPathRegexes, null);
+  }
+
+  /**
+   * Identifies all of the community literals and regexes in the given routing policies. An optional
    * set of additional community literals and regexes is also included, which is used to support
    * user-specified community constraints for symbolic analysis.
    */
-  private Set<CommunityVar> findAllCommunities(@Nullable Set<CommunityVar> communities) {
-    Set<CommunityVar> allCommunities = findAllCommunities();
+  private Set<CommunityVar> findAllCommunities(
+      @Nullable Set<CommunityVar> communities, Collection<RoutingPolicy> policies) {
+    Set<CommunityVar> allCommunities = findAllCommunities(policies);
     if (communities != null) {
       allCommunities.addAll(communities);
     }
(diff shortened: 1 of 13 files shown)


millstein (commented):
> Reviewed 7 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 286 at r1](https://reviewable.io/reviews/batfish/batfish/8625#-NKKDYN46cSXYNOn6tzv:-NKKDYN46cSXYNOn6tzw:b-fngc8j) ([raw file](https://github.com/batfish/batfish/blob/dc6382f5a1b84e9f07c4b7323ccc95a09b5bb856/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L286)):*
> ```Java
>             _asPathRegexes,
>             allPolicies);
> ```

An alternative to consider is to compute APs separately for each policy (or I guess in this context each pair of policies of the same name).  If there are many policies, then this may be more efficient than computing one set of APs across all policies.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java` line 343 at r1](https://reviewable.io/reviews/batfish/batfish/8625#-NKKDyy20VXAPoQNpNRn:-NKKDyy20VXAPoQNpNRo:b7y66j2) ([raw file](https://github.com/batfish/batfish/blob/dc6382f5a1b84e9f07c4b7323ccc95a09b5bb856/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswerer.java#L343)):*
> ```Java
>             _asPathRegexes,
>             policies);
> ```

An alternative to consider is to compute APs separately for each policy.  If there are many policies, then this may be more efficient than computing one set of APs across all policies.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 286 at r1](https://reviewable.io/reviews/batfish/batfish/8625#-NKKDYN46cSXYNOn6tzv:-NKLMyuLDPoP0n5jLx9L:ba5nke5) ([raw file](https://github.com/batfish/batfish/blob/dc6382f5a1b84e9f07c4b7323ccc95a09b5bb856/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L286)):*

An alternative to consider is to compute APs separately for each policy (or I guess in this context each pair of policies of the same name).  If there are many policies, then this may be more efficient than computing one set of APs across all policies.
</blockquote></details>

(Applies to the comment below about SRP too)

It seems like there is a trade-off here; if your set of policies use different communities/as-path lists then what you suggest makes sense.
If the policies under consideration all use the same (or, largely the same?) communities/as-path lists then it probably makes sense to compute APs only once.

Optimizing for the common use-case it seems the latter might be more prevalent?

Another alternative is to do per-policy APs, but also cache APs over communities/AS regexes such that we only reconstruct APs if we haven't already done so for the given collection of communities/ASes.

millstein (commented):
> Reviewed 2 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java` line 286 at r1](https://reviewable.io/reviews/batfish/batfish/8625#-NKKDYN46cSXYNOn6tzv:-NKLzIMR2HqHIQ5bzJcd:b-ft257) ([raw file](https://github.com/batfish/batfish/blob/dc6382f5a1b84e9f07c4b7323ccc95a09b5bb856/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareRoutePolicies/CompareRoutePoliciesAnswerer.java#L286)):*

(Applies to the comment below about SRP too)

It seems like there is a trade-off here; if your set of policies use different communities/as-path lists then what you suggest makes sense.
If the policies under consideration all use the same (or, largely the same?) communities/as-path lists then it probably makes sense to compute APs only once.

Optimizing for the common use-case it seems the latter might be more prevalent?

Another alternative is to do per-policy APs, but also cache APs over communities/AS regexes such that we only reconstruct APs if we haven't already done so for the given collection of communities/ASes.
</blockquote></details>

Makes sense; fine to keep as is, and we can always revisit later.

anothermattbrown (commented):
> Reviewed 1 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8625#-NL1op2bE9pPptlDJFOx:-NL1op2bE9pPptlDJFOy:b-dbvbvc) (no related file):*
ACLs can also potentially be mutually recursive. We avoid infinite loops in that case using `NonRecursiveSupplier` to build a lazy cache, and then compute the top-level ones we care about. The supplier will throw if we try to get the value while it's being computed. The infinite loop will throw stack overflow eventually too, but `NonRecursiveSupplier` allows us to give a better error message. 

https://github.com/batfish/batfish/blob/master/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/IpAccessListToBdd.java#L92-L104

Another approach that may be simpler and throw less often would be to split this into two phases: first collect the routing policies to be included, then extract the regexes from each one. The first phase is just transitive closure of the dependency graph. The second phase is basically the existing code. 

___

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: 1 of 13 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8625#-NL1op2bE9pPptlDJFOx:-NM0vqKVAcdLajr7zl8l:b-c3u03z) (no related file):*

ACLs can also potentially be mutually recursive. We avoid infinite loops in that case using `NonRecursiveSupplier` to build a lazy cache, and then compute the top-level ones we care about. The supplier will throw if we try to get the value while it's being computed. The infinite loop will throw stack overflow eventually too, but `NonRecursiveSupplier` allows us to give a better error message. 

https://github.com/batfish/batfish/blob/master/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/IpAccessListToBdd.java#L92-L104

Another approach that may be simpler and throw less often would be to split this into two phases: first collect the routing policies to be included, then extract the regexes from each one. The first phase is just transitive closure of the dependency graph. The second phase is basically the existing code. 
</blockquote></details>

I failed to use the memoization supplier, so I took the following approach instead:
When walking through a policy, the visitors that collect communities/as-regexes (`RoutePolicyStatementAsPathCollector` and `RoutePolicyStatementVarCollector`) keep track of the policies they've visited while walking through a given policy. On a Call statement/expression, if they have visited the called policy then they ignore it, otherwise they recursively walk through it.
This has the benefit that we will never throw a stackoverflow error, even for cyclic policies.

The downside is that we might walk through some policies multiple times, e.g.
if policy1 calls policy2, when we walk through policy1 we will walk through policy2. If we then separately walk through policy2 there is no memoization, we will do everything again.
However, the performance downside is limited: policies are not that big in size so walking over them multiple times is not terrible, and we are merely collecting strings/communities.

___

anothermattbrown (commented):
> Reviewed 1 of 9 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 74 at r2](https://reviewable.io/reviews/batfish/batfish/8625#-NM132X67w7uAtU3vwmi:-NM132X67w7uAtU3vwmj:b19yg7k) ([raw file](https://github.com/batfish/batfish/blob/48f839c874948597907c8f47c53aed16f0db2119/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L74)):*
> ```Java
>     // Otherwise update the set of seen policies and recurse.
>     Set<String> newSeen = new HashSet<>(arg.getFirst());
>     newSeen.add(callExpr.getCalledPolicyName());
> ```

can we just add to `arg.getFirst()` (and not remove it after the recursive call)? that way if we see a policy twice (not in a cycle), we still only walk it once.



_Code quote:_
```Java
    // Otherwise update the set of seen policies and recurse.
    Set<String> newSeen = new HashSet<>(arg.getFirst());
    newSeen.add(callExpr.getCalledPolicyName());
```

nickgian:
> > Reviewed 1 of 9 files at r1, 12 of 12 files at r2, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)_ status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)
> 
> _[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 74 at r2](https://reviewable.io/reviews/batfish/batfish/8625#-NM132X67w7uAtU3vwmi:-NM132X67w7uAtU3vwmj:b19yg7k) ([raw file](https://github.com/batfish/batfish/blob/48f839c874948597907c8f47c53aed16f0db2119/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L74)):_
> 
> > ```java
> >     // Otherwise update the set of seen policies and recurse.
> >     Set<String> newSeen = new HashSet<>(arg.getFirst());
> >     newSeen.add(callExpr.getCalledPolicyName());
> > ```
> 
> can we just add to `arg.getFirst()` (and not remove it after the recursive call)? that way if we see a policy twice (not in a cycle), we still only walk it once.
> 
> _Code quote:_
> 
> ```java
>     // Otherwise update the set of seen policies and recurse.
>     Set<String> newSeen = new HashSet<>(arg.getFirst());
>     newSeen.add(callExpr.getCalledPolicyName());
> ```

that is sensible I think.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8625)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 74 at r2](https://reviewable.io/reviews/batfish/batfish/8625#-NM132X67w7uAtU3vwmi:-NM1GIyACwJ-yWIAj3yu:bt2vr1b) ([raw file](https://github.com/batfish/batfish/blob/48f839c874948597907c8f47c53aed16f0db2119/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L74)):*

that is sensible I think.
</blockquote></details>

Fixed.

PR #8628, opened 2022-12-29
Docs: Add parser docs for _null rules

SLarkworthy:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8628)* status: all files reviewed, 2 unresolved discussions (waiting on @SLarkworthy)

___
*[`docs/parsing/README.md` line 654 at r1](https://reviewable.io/reviews/batfish/batfish/8628#-NKTvyypCSfHz6tFyVRS:-NKTvyypCSfHz6tFyVRT:bnbrol4) ([raw file](https://github.com/batfish/batfish/blob/da2d7c2e4d1d46f4ffd6c3377e74dee358e22cc8/docs/parsing/README.md#L654)):*
> ```Markdown
> This makes the parser more fragile and can break recovery in the event that a child rule is only partially recognized.
> 
> #### Ending rules in `_null`
> ```

Hmm. My goal was to make sure our documentation (somewhere) explains our warning philosophy – aka, the thing we drew on the whiteboard a few months ago.

IIUC, there are 5 states of config support and accompanying warnings.

1. Not in the grammar at all (unrecognized)
2. In the grammar, not implemented yet, but known to be wrong if used (we warn, with things like `todo(...)`).
3. In the grammar, but never needed to be extracted: silently ignored, but we add `_null` suffix to indicate it.
4. In the grammar and extracted, but depending on how it's used may not be supported correctly. In that case, we warn during conversion (`Warnings#redFlag` typically) if we can tell that it's not supported. If we can't tell, we warn unconditionally (and try to come up with a better system).
5. Fully implemented. No warnings.

Do you think there's a good way to put this in the docs?
___
*[`docs/parsing/README.md` line 676 at r1](https://reviewable.io/reviews/batfish/batfish/8628#-NKTtz3k5dQRMVOFwT7u:-NKTtz3k5dQRMVOFwT7v:b-v1rd4y) ([raw file](https://github.com/batfish/batfish/blob/da2d7c2e4d1d46f4ffd6c3377e74dee358e22cc8/docs/parsing/README.md#L676)):*
> ```Markdown
> ```
> 
> Note that the `_null` indicator should always be added to the leaf rule.
> ```

.. For example, if `log syslog` was to be ignored, but `log access-list ACL` was to implemented (in order to track the use of `ACL`) we would have something like:

```
s_log:
  log_access_list
  log_syslog_null
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8628)* status: all files reviewed, 1 unresolved discussion (waiting on @SLarkworthy)

PR #8630, opened 2023-01-06
SearchRoutePolicies: AS-path output constraints

millstein:
Updates SearchRoutePolicies to support user-specified constraints on the AS-path of the output route (e.g., require that a specific AS appear in the AS-path of the output route).  This ability was disabled when support for AS-path prepending was added recently.  This PR re-enables the ability to constrain the AS-path of the output route and does so correctly in the presence of AS-path prepending.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8630)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/AsPathRegexAtomicPredicates.java` line 37 at r1](https://reviewable.io/reviews/batfish/batfish/8630#-NM0vwPn92FXGS7ihnzY:-NM0vwPo0gRXbPEgKTP5:b-id8omp) ([raw file](https://github.com/batfish/batfish/blob/43992770ce2833e1582cc8e8ab4acb6e07865d3a/projects/minesweeper/src/main/java/org/batfish/minesweeper/AsPathRegexAtomicPredicates.java#L37)):*
> ```Java
>       return;
>     }
>     Map<Integer, Automaton> apAutomata = this.getAtomicPredicateAutomata();
> ```

non-blocking: seems dangerous to mutate internal state we got from the public API. would be safer to make the public API return immutable data and use a protected method to mutate



_Code quote:_
```Java
this.getAtomicPredicateAutomata();
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java` line 204 at r1](https://reviewable.io/reviews/batfish/batfish/8630#-NM0zBFOAKIUdwPXVHqg:-NM0zBFOAKIUdwPXVHqh:bk1xfp5) ([raw file](https://github.com/batfish/batfish/blob/43992770ce2833e1582cc8e8ab4acb6e07865d3a/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java#L204)):*
> ```Java
> 
>   @Override
>   public int hashCode() {
> ```

What are equals/hashCode used for? It's not safe to hash based on mutable data, so if they're just for testing, may be better to find another way



_Code quote:_
```Java
  public int hashCode() {
    return Objects.hash(
        _configuration,
        _standardCommunityAtomicPredicates,
        _nonStandardCommunityLiterals,
        _asPathRegexAtomicPredicates);
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8630)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/AsPathRegexAtomicPredicates.java` line 37 at r1](https://reviewable.io/reviews/batfish/batfish/8630#-NM0vwPn92FXGS7ihnzY:-NML3MP_2unSY_QWYdXT:b-2ib03x) ([raw file](https://github.com/batfish/batfish/blob/43992770ce2833e1582cc8e8ab4acb6e07865d3a/projects/minesweeper/src/main/java/org/batfish/minesweeper/AsPathRegexAtomicPredicates.java#L37)):*

non-blocking: seems dangerous to mutate internal state we got from the public API. would be safer to make the public API return immutable data and use a protected method to mutate
</blockquote></details>

Updated so that the data is immutable but there is a protected method to set the atomic predicate automata.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java` line 204 at r1](https://reviewable.io/reviews/batfish/batfish/8630#-NM0zBFOAKIUdwPXVHqg:-NMKv8WBEu3jTmBv9AJa:b-k8xzzx) ([raw file](https://github.com/batfish/batfish/blob/43992770ce2833e1582cc8e8ab4acb6e07865d3a/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java#L204)):*

What are equals/hashCode used for? It's not safe to hash based on mutable data, so if they're just for testing, may be better to find another way
</blockquote></details>

Yes, just for testing.  I removed them here and also in RegexAtomicPredicates for the same reason, and updated the tests.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8630)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/RegexAtomicPredicates.java` line 138 at r2](https://reviewable.io/reviews/batfish/batfish/8630#-NMlXQGX1js-lhGy64gu:-NMlXQGY-RkJWLjtJx1m:bjix9a6) ([raw file](https://github.com/batfish/batfish/blob/6be9248a760acc174ea48a061931c28156c45deb/projects/minesweeper/src/main/java/org/batfish/minesweeper/RegexAtomicPredicates.java#L138)):*
> ```Java
>   protected void setAtomicPredicateAutomata(Map<Integer, Automaton> apAutomata) {
>     _atomicPredicateAutomata =
>         ImmutableMap.<Integer, Automaton>builder().putAll(apAutomata).build();
> ```

(optional,fyi) you can use `ImmutableMap.copyOf` here



_Code quote:_
```Java
ImmutableMap.<Integer, Automaton>builder().putAll(apAutomata).build();
```

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8630)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickgian)

PR #8642, opened 2023-01-24
SRP: try to use default values for model.

nickgian:
Try to use the default value for route fields rather than the smallest possible value.
Only applied to local-pref because the rest of the fields (weight, tag, metric) seem to have a zero default value anyway.

Also fixes a bug where we didn't set the weight during model generation.


millstein (commented):
Better to make a separate PR for the bug fix if it's not too onerous, and also add a test case for that.

Maybe even more than local pref, the thing that has a bad value currently is the prefix, which will typically be something like 0.0.0.0/32.  That's the default route and so looks like a very special case.  It doesn't have to be in this PR (though it could) but worth thinking about how to address that issue.  I believe other Batfish questions have dealt with something similar.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642#-:-NMWabMw-GdwWWGsyySI:b-v268i0)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswererTest.java` line 855 at r1](https://reviewable.io/reviews/batfish/batfish/8642#-NMWalnJCqZbg6ExsbAP:-NMWalnJCqZbg6ExsbAQ:b-ra1j9z) ([raw file](https://github.com/batfish/batfish/blob/5cef1d0b535452820ca78e7e3566791df56d2d33/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswererTest.java#L855)):*
> ```Java
>             .setNextHopIp(Ip.parse("0.0.0.1"))
>             .setLocalPreference(Bgpv4Route.DEFAULT_LOCAL_PREFERENCE)
>             .setLocalPreference(Bgpv4Route.DEFAULT_LOCAL_PREFERENCE)
> ```

Presumably this redundancy was an accident.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 230 at r1](https://reviewable.io/reviews/batfish/batfish/8642#-NMWbXQy149CPnL18-mn:-NMWbXQy149CPnL18-mo:bq4vdy5) ([raw file](https://github.com/batfish/batfish/blob/5cef1d0b535452820ca78e7e3566791df56d2d33/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L230)):*
> ```Java
>             // the next-hop since it is constrained to not be 0.0.0.0
>             .setNextHopIp(Ip.parse("0.0.0.1"))
>             .setLocalPreference(Bgpv4Route.DEFAULT_LOCAL_PREFERENCE)
> ```

Are there any tests that exhibit the case when the default local pref cannot be used?  I guess that would happen if we match on the local pref or have an input constraint on the local pref.  If we have no such tests it would be useful to add one.

nickgian (commented):
That's a good callout, other batfish questions seem to deal with this by trying some well-known IP addresses.
I tried to do something similar, falling back to a wide range of prefixes if some predefined ones are not compatible with the model.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642#-:-NM_r9ZEA7J7-e1ytyVB:b-6ucvjs)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswererTest.java` line 855 at r1](https://reviewable.io/reviews/batfish/batfish/8642#-NMWalnJCqZbg6ExsbAP:-NM_qm6r524sclyc6RKK:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/5cef1d0b535452820ca78e7e3566791df56d2d33/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswererTest.java#L855)):*

Presumably this redundancy was an accident.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 230 at r1](https://reviewable.io/reviews/batfish/batfish/8642#-NMWbXQy149CPnL18-mn:-NM_qnof7ZBTaVJzdd-N:bwasjee) ([raw file](https://github.com/batfish/batfish/blob/5cef1d0b535452820ca78e7e3566791df56d2d33/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L230)):*

Are there any tests that exhibit the case when the default local pref cannot be used?  I guess that would happen if we match on the local pref or have an input constraint on the local pref.  If we have no such tests it would be useful to add one.
</blockquote></details>

I don't think TransferBDD supports matches on local-pref. Introduced a test where the local-pref of the input route is constrained.

progwriter (commented):
> Reviewed 1 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

millstein (commented):
Looks good to me.  One note: You will also need to do a PR for pybatfish because this change will break tests there.

> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642#-:-NN46qgc9KGoGQiqwyy2:brpwuxj)* status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 2112 at r2](https://reviewable.io/reviews/batfish/batfish/8642#-NN46eZZ00mT3LgH9xYR:-NN46eZZ00mT3LgH9xYS:b-461om6) ([raw file](https://github.com/batfish/batfish/blob/ece281862f8392508117634b8f65f640677ba64f/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L2112)):*
> ```Java
>     BgpRoute inputRoute =
>         BgpRoute.builder()
>             .setNetwork(Prefix.parse("128.0.0.0/32"))
> ```

I'm confused why this starts with 128 rather than 20, given the input constraint on the prefix?

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)* status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 2112 at r2](https://reviewable.io/reviews/batfish/batfish/8642#-NN46eZZ00mT3LgH9xYR:-NN84HAV7rsUgEgkSRo1:bw4o5qm) ([raw file](https://github.com/batfish/batfish/blob/ece281862f8392508117634b8f65f640677ba64f/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L2112)):*

I'm confused why this starts with 128 rather than 20, given the input constraint on the prefix?
</blockquote></details>

I haven't done the bit-level analysis but I am guessing it's a combination of the constraints generated and how they are converted to ints.

`20.0.0.0/0:28-32` I believe means any prefix with length 28/32. So the result is not wrong, maybe less expected.
Added a test where the prefix is `20.0.0.0/31:28-32` and then you get `20.0.0.0/32` as you would expect.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @dhalperi, @millstein, and @progwriter)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 2112 at r2](https://reviewable.io/reviews/batfish/batfish/8642#-NN46eZZ00mT3LgH9xYR:-NNrWqF3Crd6JF0LO9zK:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/ece281862f8392508117634b8f65f640677ba64f/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L2112)):*

Because this is `20.0.0.0/0`, emphasis on the `/0`, the `20.` is meaningless. Make it `/8`? Or just change it back to `0.0.0.0`? Agree that it's confusing as-is.
</blockquote></details>

Done.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @millstein, and @progwriter)

dhalperi (commented):
> Dismissed @millstein from a discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @millstein, and @progwriter)

nickgian:
> Dismissed @millstein from a discussion.
> _[Reviewable](https://reviewable.io/reviews/batfish/batfish/8642)_ status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @millstein, and @progwriter)

thanks!

PR #8649, opened 2023-02-01
TestRoutePolicies: fix invalid assertion

dhalperi:
And add test

---

**Stack**:
- #8652
- #8651
- #8650
- #8649 ⬅

### DIFF
--- a/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java
+++ b/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java
@@ -406,7 +406,15 @@ public final class TestRoutePoliciesAnswerer extends Answerer {
         .build();
   }
 
-  private static @Nullable Row toDiffRow(Result snapshotResult, Result referenceResult) {
+  /**
+   * Returns a {@link Row} describing the difference between the two results, in the case where the
+   * actual outcome of the policy is different, or {@code null} otherwise.
+   *
+   * <p>Note that no difference in the outcome includes the case when, e.g., only the name of a
+   * policy changed but the action and output attributes are identical across the two snapshots.
+   */
+  @VisibleForTesting
+  static @Nullable Row toDiffRow(Result snapshotResult, Result referenceResult) {
     assert snapshotResult.getKey().equals(referenceResult.getKey());
 
     if (snapshotResult.equals(referenceResult)) {
@@ -420,7 +428,10 @@ public final class TestRoutePoliciesAnswerer extends Answerer {
 
     boolean equalAction = snapshotResult.getAction() == referenceResult.getAction();
     boolean equalOutputRoutes = Objects.equals(snapshotOutputRoute, referenceOutputRoute);
-    assert !(equalAction && equalOutputRoutes);
+    if (equalAction && equalOutputRoutes) {
+      // This can happen if the trace is different.
+      return null;
+    }
 
     BgpRouteDiffs routeDiffs =
         new BgpRouteDiffs(routeDiffs(referenceOutputRoute, snapshotOutputRoute));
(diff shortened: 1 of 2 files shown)


nickgian (commented):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8649)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 426 at r1](https://reviewable.io/reviews/batfish/batfish/8649#-NNDOZe96_TXeIXtuCOp:-NNDOZeA0B3lZT2OJPnQ:bky4an1) ([raw file](https://github.com/batfish/batfish/blob/a69da53e68d999c34d0a687d00441549ca4668c9/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L426)):*
> ```Java
>     if (equalAction && equalOutputRoutes) {
>       // This can happen if the trace is different.
>       return null;
> ```

oh, interesting case.
Should we be using the more type-safe alternative of returning an `Optional` instead of `null`? Clients of this code would need to change, but it should be an easy change.

Regardless, would also be useful to have javadoc comment to document what this method returns and when to except a `null`/`None` value.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8649)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 426 at r1](https://reviewable.io/reviews/batfish/batfish/8649#-NNDOZe96_TXeIXtuCOp:-NNDTYAM-jL_6Oh3uN0x:b-73evpr) ([raw file](https://github.com/batfish/batfish/blob/a69da53e68d999c34d0a687d00441549ca4668c9/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L426)):*

oh, interesting case.
Should we be using the more type-safe alternative of returning an `Optional` instead of `null`? Clients of this code would need to change, but it should be an easy change.

Regardless, would also be useful to have javadoc comment to document what this method returns and when to except a `null`/`None` value.
</blockquote></details>

The IDE and tools do a good job using `@Nullable` and type inference. We go back and forth, and we do different things in different places. However my general feeling has been that in `Optional` has turned out to be a fairly big JVM performance hit, code complexity hit, and actually makes the nullity analysis work worse in practice.

(I wish I knew how to alleviate these pragmatic issues).

Did update the javadoc.

nickgian (commented):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8649#-:-NNDXYrU4pTI6M_4lfaO:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 426 at r1](https://reviewable.io/reviews/batfish/batfish/8649#-NNDOZe96_TXeIXtuCOp:-NNDVuLV3dkMDhO_IN19:b-2w456r) ([raw file](https://github.com/batfish/batfish/blob/a69da53e68d999c34d0a687d00441549ca4668c9/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L426)):*

The IDE and tools do a good job using `@Nullable` and type inference. We go back and forth, and we do different things in different places. However my general feeling has been that in `Optional` has turned out to be a fairly big JVM performance hit, code complexity hit, and actually makes the nullity analysis work worse in practice.

(I wish I knew how to alleviate these pragmatic issues).

Did update the javadoc.
</blockquote></details>

Looks good then! I figured that your convention is leaning towards `@Nullable`.
I'd be curious to see if performance issues are still there with newer JVMs, but that's obviously outside the scope of this PR.

PR #8657, opened 2023-02-14
Junos: buggy test of replace failing

dhalperi:


arifogel:
Integrated into batfish/batfish#8667

PR #8663, opened 2023-02-20
SearchRoutePolicies: Change the default handling of unsupported features

millstein:
Previously SearchRoutePolicies only reported results along execution paths that do not encounter any statements that are currently unsupported by the symbolic route analysis.  This choice meant that if a route map has unsupported features, even ones that are unrelated to the properties of interest to a user, SRP will tend to give no results.  The user then must manually remove the unsupported features from their route maps and try again (or worse, they believe that there are no results when really the desired behavior does exist).  

In the new approach, we simply treat unsupported statements as no-ops.  So SRP will now produce results along paths containing unsupported statements.  The downside is that it's now possible for these results to be false positives (i.e., to not exhibit the behaviors that the user requested).  But this seems like a better tradeoff than producing false negatives as before.

Unsupported features are logged and emitted as warnings, as previously.  In the future we could use this information to provide additional context for users about the SRP results.


nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663#-:-NOmniteB9UmIhvfzoTK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NOu95ea1gbty5Ei2pAY:b5q0n5b) (no related file):*
Are they non-deterministic no-ops, or just no-ops?

Aka, if I have `match unsupported`, will SRP explore just the true path, or both?

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NOufo0i9OO9YQzFRIEI:b-36oi4v) (no related file):*

Are they non-deterministic no-ops, or just no-ops?

Aka, if I have `match unsupported`, will SRP explore just the true path, or both?
</blockquote></details>

Currently the semantics is that the nearest enclosing statement that an unsupported thing is part of is treated as a no-op.  So if the unsupported thing is in an "if" guard then the whole if statement is skipped.  This could do something reasonable like skip a single stanza (if the stanza is represented with an if-then statement) or something unreasonable like skip large swaths of the route map (if we have an if-then-else statement where the "else" contains the entire remainder of the route map).

We could switch to the semantics that I think you're suggesting, where if an "if" guard has an unsupported statement then we treat the guard as if it is a non-deterministic choice and so explore both branches.  I think the issue there is that we are likely to get false positives.  Suppose the analysis finds that taking the "then" branch leads to the behavior that the user requested.  But when we solve for a concrete example, because we aren't modeling the guard it's very unlikely that we will actually get one that happens to satisfy it.

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NOulgBK53PgSxNy1cC4:b10p94d) (no related file):*

Currently the semantics is that the nearest enclosing statement that an unsupported thing is part of is treated as a no-op.  So if the unsupported thing is in an "if" guard then the whole if statement is skipped.  This could do something reasonable like skip a single stanza (if the stanza is represented with an if-then statement) or something unreasonable like skip large swaths of the route map (if we have an if-then-else statement where the "else" contains the entire remainder of the route map).

We could switch to the semantics that I think you're suggesting, where if an "if" guard has an unsupported statement then we treat the guard as if it is a non-deterministic choice and so explore both branches.  I think the issue there is that we are likely to get false positives.  Suppose the analysis finds that taking the "then" branch leads to the behavior that the user requested.  But when we solve for a concrete example, because we aren't modeling the guard it's very unlikely that we will actually get one that happens to satisfy it.
</blockquote></details>

Maybe a middle ground to consider is to just analyze the "else" branch and ignore the "then" branch.  That way we still analyze some of the "if" but have less likelihood of generating a false positive.

___

dhalperi:

Maybe a middle ground to consider is to just analyze the "else" branch and ignore the "then" branch.  That way we still analyze some of the "if" but have less likelihood of generating a false positive.
</blockquote></details>

I guess I don't understand why taking the true branch is somehow more or less likely to introduce a false positive than taking the false branch.

dhalperi:
> if the unsupported thing is in an "if" guard then the whole if statement is skipped.

One more clarifying Q. Is this true as written? Aka, `if (unsupported & false)` we still skip the `if`?

nickgian:
> > if the unsupported thing is in an "if" guard then the whole if statement is skipped.
> 
> One more clarifying Q. Is this true as written? Aka, `if (unsupported & false)` we still skip the `if`?

Yes, we still skip `if`. A similar problem can happen with set statements.

This probably can/should be improved, just to ensure that we are not skipping commands we support because of some potentially irrelevant command. It's unclear how much effort we should put into that vs. ensuring that we support all features, but it's definitely worth investigating how to make the results more trustworthy.

millstein (commented):
By "unsupported  feature" here we mean something that Batfish parses and represents in the VI model but which SRP does not handle.  There are no such things in TRP.  

Separately there is the class of features that Batfish does not include in the VI model.  In those cases implicitly both SRP and TRP treat them as no-ops.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663#-:-NOvCXFP7A_JV0-Q3nkS:b-8xwphy)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NOvBuW624a5QAvYNxiB:b-htr9gr) (no related file):*

> if the unsupported thing is in an "if" guard then the whole if statement is skipped.

One more clarifying Q. Is this true as written? Aka, `if (unsupported & false)` we still skip the `if`?
</blockquote></details>

My intuition is that the guard in a stanza is typically asking for some relatively specific thing, which we are unlikely to accidentally solve for when finding a model (particularly since we default to things like 0s for the values of attributes).  So it's more likely that a concrete example we produce will falsify a guard than satisfy it, particularly if the guard contains some kind of expression that we are not modeling currently.

___

nickgian:
> There are no such things in TRP.

I am not sure I buy that, I vaguely recall seeing things missing even from the type of routes computed.
But, I might be wrong.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NOzYWRmDSJ2v0PAx7f9:b-yujbke) (no related file):*

My intuition is that the guard in a stanza is typically asking for some relatively specific thing, which we are unlikely to accidentally solve for when finding a model (particularly since we default to things like 0s for the values of attributes).  So it's more likely that a concrete example we produce will falsify a guard than satisfy it, particularly if the guard contains some kind of expression that we are not modeling currently.
</blockquote></details>

Results are now per-path, right? Can we add a column indicating whether the path includes unsupported features?

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NP-6dD3CwvgHSNNb5Bw:b-1lw81u) (no related file):*

Results are now per-path, right? Can we add a column indicating whether the path includes unsupported features?
</blockquote></details>

There's an option to provide one result per path, as opposed to one result total.  But the default is just one result.  Either way we know what path the analysis went down for any result that we provide, and know if we encountered unsupported features.  So we could do what you suggest.  

The thing I'd also want to think through is how to be clear to users about what it means if the "unsupported" box is checked, and what to do as a result.  Roughly I think they should 1) See if the given example in fact meets their desired behavior.  2) If it does and they just asked for one result, they should be happy.  If it does and they asked for all results (perPath) then it's possible they are missing results for some paths.  3) If it does not meet their desired behavior, then it's a false positive, and they also might be missing real results.

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NOu95ea1gbty5Ei2pAX:-NPoImSL9O0BkxsLh-EX:bhwkbfs) (no related file):*

There's an option to provide one result per path, as opposed to one result total.  But the default is just one result.  Either way we know what path the analysis went down for any result that we provide, and know if we encountered unsupported features.  So we could do what you suggest.  

The thing I'd also want to think through is how to be clear to users about what it means if the "unsupported" box is checked, and what to do as a result.  Roughly I think they should 1) See if the given example in fact meets their desired behavior.  2) If it does and they just asked for one result, they should be happy.  If it does and they asked for all results (perPath) then it's possible they are missing results for some paths.  3) If it does not meet their desired behavior, then it's a false positive, and they also might be missing real results.
</blockquote></details>

After discussing with @nickgian we'd prefer to not add such a column right now.  I think we need more experience to understand the failure modes and how best to convey them.  For now our main goal is to make it easier to experiment with symbolic route analysis in the presence of unsupported features

___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/8663#-NPoJ8R_44mP8T2PVGti:-NPoJ8R_44mP8T2PVGtj:bw39m4q) (no related file):*
Based on an earlier comment from @dhalperi and some discussion with @nickgian, I updated the handling of unsupported expressions inside conjunctions and disjunctions.  Now they are treated like no-ops but the rest of the conjunction/disjunction is still analyzed as usual.  For example, if(A /\ B /\ C) ... is treated as if(A /\ C) in the case where B is an unsupported expression.  This enables us to still accurately analyze many scenarios (e.g., the case when either A or C is false) and also capture potentially useful behaviors even when B's status is relevant.

___

anothermattbrown (commented):
I don't know of VI constructs that are unsupported by TRP. Seems unlikely that we would add them to the VI model without interpreting them. More likely that there are VS constructs that are not converted to VI. But @dhalperi or @arifogel would know more.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663#-:-NQQnNTl3wlih4mtkSgP:b9pzl14)* status: all files reviewed, 2 unresolved discussions (waiting on @dhalperi and @nickgian)

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8663#-:-NSEmBZw24TiGVUeoF98:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @dhalperi, and @nickgian)

PR #8676, opened 2023-03-06
fix_java_format.sh: fix missing command detection

dhalperi:
Without quotes, the if seems to return true even if the command does not exist


arifogel (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8676)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8685, opened 2023-03-16
BgpRib: sort lazily

dhalperi:
1. getRoutes was sorting unnecessarily, can simply keep the maximum
2. multipath code to update best routes was pre-emptively sorting
even though it might not be necessary.

Changing it to sort lazily cut bestPathComparator time by more than
a factor of 10 on one large network.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java
@@ -18,6 +18,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


dhalperi:
*[`projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java` line 295 at r1](https://reviewable.io/reviews/batfish/batfish/8685#-NQd9wY38DpxkKe9ekJj:-NQd9wY4A5IJhRpYRInW:b9w4n4l) ([raw file](https://github.com/batfish/batfish/blob/9a8dd4d0149a62c7cf6e5ed997829af510533f9b/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java#L295)):*
> ```Java
>                   if (!routesForNh.isEmpty()) {
>                     // the removed route was best, so promote the new best
>                     builder.from(action).add(routesForNh.last());
> ```

FYI this `.from(action)` is redundant with the one 3 lines up.



_Code quote:_
```Java
  builder.from(action).add(routesForNh.last());
```

PR #8690, opened 2023-03-17
TransferBDD: MatchClusterListLength

millstein:
Adds support for the MatchClusterListLength boolean expression in the symbolic route map analysis.


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8690)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java` line 328 at r1](https://reviewable.io/reviews/batfish/batfish/8690#-NR-AxEq4Mqh5WaQuqtm:-NR-ucQMBLEKTfRpxpf2:b-s0pt7p) ([raw file](https://github.com/batfish/batfish/blob/ecd75d37c06fee0483f562bd7eee6177a51a3d48/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/questions/BgpRoute.java#L328)):*

is that actually a set in BGP implementations?
</blockquote></details>

Here I'm following how org.batfish.datamodel.BgpRoute represents the cluster list, which is why I used a set.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 210 at r1](https://reviewable.io/reviews/batfish/batfish/8690#-NR-CYsU79J54WDzzem4:-NR-wQ8_BddqwNVndXgx:bkxc0xt) ([raw file](https://github.com/batfish/batfish/blob/ecd75d37c06fee0483f562bd7eee6177a51a3d48/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L210)):*

I wonder whether using 32-bits for this is an overkill, but I couldn't find any specs on the max-length.
</blockquote></details>

Yeah, I was also looking but couldn't find anything about the max length so chose 32.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 197 at r1](https://reviewable.io/reviews/batfish/batfish/8690#-NR-DekbCHC099DrH_6p:-NR-xAnICIQfMiz61rnO:bk748mt) ([raw file](https://github.com/batfish/batfish/blob/ecd75d37c06fee0483f562bd7eee6177a51a3d48/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L197)):*

This makes me wonder whether we should really just keep track of the length/present the length to the user.
Since the match happens on the length, presenting a random cluster-list might be more confusing, rather than telling them this happens when the length is K.
</blockquote></details>

I agree that the produced list is very artificial.  But since a route contains a cluster list, and Batfish already models it as having such a list, I felt that giving a concrete list was still better than just a length.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8690)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 197 at r1](https://reviewable.io/reviews/batfish/batfish/8690#-NR-DekbCHC099DrH_6p:-NR4FOU1Dr9kSGPzqC-j:bfy7dhm) ([raw file](https://github.com/batfish/batfish/blob/ecd75d37c06fee0483f562bd7eee6177a51a3d48/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L197)):*

I agree that the produced list is very artificial.  But since a route contains a cluster list, and Batfish already models it as having such a list, I felt that giving a concrete list was still better than just a length.
</blockquote></details>

If the cluster list length is being determined by a cluster list input from the user, it would be good to retain that value. That wouldn't need to happen here though. 
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 2216 at r1](https://reviewable.io/reviews/batfish/batfish/8690#-NR4GASY0bVHBadzVqPP:-NR4GASY0bVHBadzVqPQ:bxu5vd) ([raw file](https://github.com/batfish/batfish/blob/ecd75d37c06fee0483f562bd7eee6177a51a3d48/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L2216)):*
> ```Java
>             .setNextHopIp(Ip.parse("0.0.0.1"))
>             .setLocalPreference(Bgpv4Route.DEFAULT_LOCAL_PREFERENCE)
>             .setClusterList(ImmutableSet.of(0L, 1L))
> ```

would be good to have a test where the cluster list does not happen to match what `ModelGeneration` produces, e.g. `1, 2`. 



_Code quote:_
```Java
setClusterList(ImmutableSet.of(0L, 1L))
```

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8690#-:-NS8_7SpCptWuQH_aaQ-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8718, opened 2023-05-13
Move route reflection logic to a shared method

SLarkworthy:
Route reflection criteria was applied in two places, but one location was incomplete. Specifically, we have a check `isReflectable` that was used for advertising BGP routes that are not installed in the main RIB if they are a result of route reflection. Moved this logic into a shared method used for both the RIB failure spot and the BgpProtocolHelper spot. 

Functionally, this changes a single case: in the event of RIB failure, client-to-client and non-client-to-client cases were already covered, and this CR adds client-to-non-client.

Does not effect anything eBGP-related.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8718)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @arifogel)

PR #8724, opened 2023-05-25
TransferBDD: Support some community set regexes

millstein:
The symbolic route analysis handles regexes on individual communities but not regexes on sets of communities, because the semantics of the latter is complex and not easy to model symbolically.  This PR conservatively identifies community set regexes that can be treated as regexes on individual communities, and hence can be handled by the analysis.  It leverages Batfish's translation of the _ character to a Java regex, so this PR also pulls out that translation as a constant that is shared by many vendor-specific conversions.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @millstein and @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8724#-NWIX8cR68n-a8q9fBtN:-NWIX8cS0DhO3eVIliMV:b-omrpea) (no related file):*
1. This is really cool!
2. I was just thinking about this the other day. Would it be possible / better to do this in VS land so that the slow / non-symbolic ways of checking communities via interpreting the policy are also sped up?

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8724#-NWIX8cR68n-a8q9fBtN:-NWIbxdc8SiobIuAeXVB:b-kbn0rd) (no related file):*

1. This is really cool!
2. I was just thinking about this the other day. Would it be possible / better to do this in VS land so that the slow / non-symbolic ways of checking communities via interpreting the policy are also sped up?
</blockquote></details>

Doing it separately for each vendor seems painful and not easily maintainable.  An alternative would be to have a visitor pass right after creation of the VI model to walk the model and replace instances of CommunitySetMatchRegex with HasCommunityIn(CommunityMatchRegex) where possible.  Not sure if we already have passes like that (similar to optimization passes in a compiler) or where that would fit?

In terms of the motivation, I'm surprised that for non-symbolic stuff that this really matters in terms of performance -- it's still just a regex match (but now maybe preceded by sorting the route's communities).  

___

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 0 of 20 files reviewed, 1 unresolved discussion (waiting on @millstein and @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8724#-NWIX8cR68n-a8q9fBtN:-NWKTqKP-JhNzCy2Xjov:b-nwrumx) (no related file):*

Doing it separately for each vendor seems painful and not easily maintainable.  An alternative would be to have a visitor pass right after creation of the VI model to walk the model and replace instances of CommunitySetMatchRegex with HasCommunityIn(CommunityMatchRegex) where possible.  Not sure if we already have passes like that (similar to optimization passes in a compiler) or where that would fit?

In terms of the motivation, I'm surprised that for non-symbolic stuff that this really matters in terms of performance -- it's still just a regex match (but now maybe preceded by sorting the route's communities).  
</blockquote></details>

Maybe the lowest order fruit would be a new static factory that is something like `public static MatchCommunitySet matchRegex(String regex);` Then it could internally convert into `MatchCommunitySet(HasCommunity(...))` or `MatchCommunitySet(CSMR(regex))` or whatever based on use.

I can see a lot of value here, actually. If someone uses `_1_` and we can convert that into `HasCommunity(Literal(1))` or `_.*:5_` and we turn that into `HasCommunity(HasLower16(5)` that would be pretty cool. Primitive ops instead of regexes.

Of course, more than you've already done is incremental value to come later.

___

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @dhalperi and @nickgian)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8724#-NWIX8cR68n-a8q9fBtN:-NWO0AMx8VFLcV9Oa8K9:b-dicg4y) (no related file):*

Maybe the lowest order fruit would be a new static factory that is something like `public static MatchCommunitySet matchRegex(String regex);` Then it could internally convert into `MatchCommunitySet(HasCommunity(...))` or `MatchCommunitySet(CSMR(regex))` or whatever based on use.

I can see a lot of value here, actually. If someone uses `_1_` and we can convert that into `HasCommunity(Literal(1))` or `_.*:5_` and we turn that into `HasCommunity(HasLower16(5)` that would be pretty cool. Primitive ops instead of regexes.

Of course, more than you've already done is incremental value to come later.
</blockquote></details>

I did a version of that -- a static factory for creating community-set regexes.

___

nickgian (commented):
> Reviewed 7 of 20 files at r1, 13 of 20 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 20 of 28 files reviewed, 2 unresolved discussions (waiting on @dhalperi and @millstein)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java` line 59 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO67vs8lWuItA1jQtN:-NWO67vv1CKpyRF-pHWR:b-20a1o3) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L59)):*
> ```Java
>     // community in the set: the regex optionally starts with _, optionally ends with _, and in
>     // between only accepts strings containing digits and colons
>     String underscore = StringEscapeUtils.unescapeJava(DEFAULT_UNDERSCORE_REPLACEMENT);
> ```

Could we generalize this code further, by also checking for other symbols like `$` instead of `_`?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 108 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO7Hzs8NkLg_Y1SeBS:-NWO7Hzs8NkLg_Y1SeBT:b-k838le) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L108)):*
> ```Java
>     // We've already ensured in CommunitySetMatchExprVarCollector that there are no community-set
>     // regexes, so this code should be unreachable.
>     checkArgument(false, "Unexpected community set match regex");
> ```

what does checkArgument do? Does it throw?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 20 of 28 files reviewed, 2 unresolved discussions (waiting on @dhalperi and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java` line 59 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO67vs8lWuItA1jQtN:-NWO85CEDxyHH8f6ryqU:byb3lw4) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L59)):*

Could we generalize this code further, by also checking for other symbols like `$` instead of `_`?
</blockquote></details>

There may be ways to generalize it, but in general uses of `$` and `^` are not ok.  For example, something like "_20:30$" requires that the community list end with the community 20:30, so the order still matters -- it's not sufficient for 20:30 to be anywhere in the list, which is what we need.  
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 108 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO7Hzs8NkLg_Y1SeBS:-NWO8Tr05oPdh_n5r5ps:btf6o5r) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L108)):*

what does checkArgument do? Does it throw?
</blockquote></details>

Yes it throws.  If there's a better, more standard way to signal unreachable code, I'm happy to do that instead.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 20 of 28 files reviewed, 3 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java` line 37 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO8lue4a7DoZAkO8fo:-NWO8lue4a7DoZAkO8fp:b-as2dca) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L37)):*
> ```Java
> 
> /** Utility functions for generating {@link RoutingPolicy routing policies}. */
> public final class Common {
> ```

This is pretty much exactly what I was thinking. Just one request - these proliferation of generic `Common` with lots of interspersed unrelated functionality becomes hard to maintain over time.

We have a dedicated place for this for Communities, inside the `.communities` package: `org.batfish.datamodel.routing_policy.communities.CommunitySetExprs`. Can you move it there?

Then you can also simplify the name `communitySet` is implicit in the path/classname.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 20 of 28 files reviewed, 2 unresolved discussions (waiting on @dhalperi and @millstein)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java` line 59 at r3](https://reviewable.io/reviews/batfish/batfish/8724#-NWO67vs8lWuItA1jQtN:-NWOSq1p8K0-0i5eqi-E:b-v73j0x) ([raw file](https://github.com/batfish/batfish/blob/adfe5004dce624ef47d60711793c932c9a235955/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/Common.java#L59)):*

There may be ways to generalize it, but in general uses of `$` and `^` are not ok.  For example, something like "_20:30$" requires that the community list end with the community 20:30, so the order still matters -- it's not sufficient for 20:30 to be anywhere in the list, which is what we need.  
</blockquote></details>

that makes sense.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724#-:-NWOSyw15f4i9grbsbcq:bnfp4nl)* status: 20 of 28 files reviewed, 2 unresolved discussions (waiting on @dhalperi and @millstein)

dhalperi (commented):
> Reviewed 3 of 11 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: 14 of 29 files reviewed, all discussions resolved (waiting on @nickgian)

dhalperi (commented):
> Reviewed 1 of 20 files at r1, 6 of 20 files at r3, 10 of 10 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 108 at r4](https://reviewable.io/reviews/batfish/batfish/8724#-NWOiuwKAuAGGfCNoFXT:-NWOiuwKAuAGGfCNoFXU:b45h) ([raw file](https://github.com/batfish/batfish/blob/9953aa72f851fb275d1a1f86e6b167c4dfb111e2/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L108)):*
> ```Java
>     // We've already ensured in CommunitySetMatchExprVarCollector that there are no community-set
>     // regexes, so this code should be unreachable.
>     checkArgument(false, "Unexpected community set match regex");
> ```





_Suggestion:_
```Java
   throw new IllegalStateException(String.format("Unexpected community set match regex %s", communitySetMatchRegex));
```

dhalperi (commented):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8724)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8732, opened 2023-06-09
TransferBDD: FirstMatchChain

millstein:
Updates the symbolic route analysis to support a form of Juniper policy chains, which executes a sequence of policies until one of them makes a decision (accept or reject).

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java
@@ -25,6 +25,7 @@ import javax.annotation.Nullable;
 import org.batfish.common.plugin.TracerouteEngine;
 import org.batfish.common.topology.IpOwners;
 import org.batfish.common.topology.L3Adjacencies;
+import org.batfish.common.traceroute.TraceDag;
 import org.batfish.datamodel.BgpActivePeerConfig;
 import org.batfish.datamodel.BgpPassivePeerConfig;
 import org.batfish.datamodel.BgpPeerConfig;
@@ -51,6 +52,7 @@ import org.batfish.datamodel.NamedPort;
 import org.batfish.datamodel.NetworkConfigurations;
 import org.batfish.datamodel.Vrf;
 import org.batfish.datamodel.collections.NodeInterfacePair;
+import org.batfish.datamodel.flow.FirewallSessionTraceInfo;
 import org.batfish.datamodel.flow.Hop;
 import org.batfish.datamodel.flow.Trace;
 import org.batfish.datamodel.flow.TraceAndReverseFlow;
@@ -316,15 +318,15 @@ public final class BgpTopologyUtils {
                     feasibleLocalIpsForPeeringWithCandidate.forEach(
                         ip -> addEdges(neighbor, neighborId, ip, candidateId, graph, nc));
                   } else {
-                    initiateBgpSessions(
-                            neighborId,
-                            candidateId,
-                            neighbor,
-                            feasibleLocalIpsForPeeringWithCandidate,
-                            tracerouteEngine)
-                        .stream()
-                        .filter(BgpSessionInitiationResult::isSuccessful)
-                        .map(initiationResult -> initiationResult.getFlow().getSrcIp())
+                    feasibleLocalIpsForPeeringWithCandidate.stream()
+                        .filter(
+                            initiatorLocalIp ->
+                                canEstablishBgpSession(
+                                    neighborId,
+                                    candidateId,
+                                    neighbor,
+                                    initiatorLocalIp,
+                                    tracerouteEngine))
                         .forEach(
                             srcIp -> addEdges(neighbor, neighborId, srcIp, candidateId, graph, nc));
                   }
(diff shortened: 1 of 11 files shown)


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8732)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 379 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXQAe1BCEwcDclu2Am:-NXXQAe1BCEwcDclu2An:bid38ms) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L379)):*
> ```Java
>                   r -> {
>                     TransferResult updated =
>                         r.setReturnValueBDD(r.getReturnValue().getSecond().and(currBDD));
> ```

can you comment this logic? 
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 389 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXcTUo1KrZxRb10XYe:-NXXcTUo1KrZxRb10XYf:b-5bbdr6) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L389)):*
> ```Java
>         currResults = nextResults;
>       }
>       if (!currResults.isEmpty()) {
> ```

why did you move this check here? is there a case when the old check would have passed and this one fail? if so, is this error message accurate?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 2999 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXd6SWCvBewYfT-E54:-NXXd6SWCvBewYfT-E55:b75b6h7) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L2999)):*
> ```Java
> 
>   @Test
>   public void testFirstMatchChainStateful() {
> ```

can you add a stateful test where both route maps set the same field (to different values)?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8732)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 379 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXQAe1BCEwcDclu2Am:-NX_mn4KDqRGfcPKpr7R:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L379)):*

can you comment this logic? 
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 389 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXcTUo1KrZxRb10XYe:-NX_mpSY32MB4dQKEd1z:bhrzvob) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L389)):*

why did you move this check here? is there a case when the old check would have passed and this one fail? if so, is this error message accurate?
</blockquote></details>

This check will catch more things than the other check -- it fails if we ever reach the end of the policy chain but the path is still saying it wants to fall through to the next policy.  I updated the error message.  This logic is analogous to what happens in FirstMatchChain::evaluate, where we  complain about a missing default policy if we end up falling through to it.  However,  here we will also complain if the default policy itself has fallthroughs -- I'm not sure of the semantics of that so it seemed safer to reject it and notify the user.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 2999 at r1](https://reviewable.io/reviews/batfish/batfish/8732#-NXXd6SWCvBewYfT-E54:-NX_nkAYCt3LiWBgRg0K:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/628ffd80836883878140f49bdaa0bda6606d4964/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L2999)):*

can you add a stateful test where both route maps set the same field (to different values)? 
</blockquote></details>

Done.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8732#-:-NXaB_ceC-RmoYpHdb-I:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown)

PR #8740, opened 2023-07-02
Symbolic Route Analysis: Better support for next-hops

millstein:
Adds support to the symbolic route analysis and associated questions (SearchRoutePolicies and CompareRoutePolicies) for several next-hop-related route-map constructs:

* Matching on the interface name of the next hop
* Setting the next hop to the local IP address
* Setting the next hop to the remote IP address


nickgian (commented):
> Reviewed 33 of 33 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: all files reviewed, 5 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddress.java` line 9 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_655L0EPG84_B1Hmgu:-N_655L0EPG84_B1Hmgv:b-d053wk) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddress.java#L9)):*
> ```Java
> 
> /**
>  * Represents the fact that the next-hop IP of a BGP route is set to the address of the BGP peer by
> ```

Can you amend the comment to explain when objects of this class are used by SRP?
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 10 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_64mooAcAL6S0rh-fD:-N_64mopEqLMIlz8gg2h:b2nreh6) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L10)):*
> ```Java
> /**
>  * Represents the fact that the next-hop IP of a BGP route is set to the local address of the BGP
>  * session by a particular route policy. Objects of this class can be used in results produced by
> ```

Same as above.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 109 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_66wQv4q1h4Tw4SQ-s:-N_66wQv4q1h4Tw4SQ-t:b-ao225z) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L109)):*
> ```Java
>   }
> 
>   private NextHopType _nextHopType;
> ```

This is not a BDD because it does not depend on the input, right?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 193 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6928oDR0XTndmi3NP:-N_6928oDR0XTndmi3NQ:blnpfnu) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L193)):*
> ```Java
>     checkState(
>         nextHopInterfaces.size() <= 1,
>         "Error in symbolic route analysis: at most one source VRF can be in the environment");
> ```

"at most one NextHop Interface can be set"?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 554 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_69mVu-Kk-dRxBnTyO:-N_69mVu-Kk-dRxBnTyP:b-7ddhsz) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L554)):*
> ```Java
>       if (_useOutputAttributes) {
>         // we don't yet properly model the situation where a modified next-hop can be later matched
>         // upon, so we don't allow matching if that is possible
> ```

can this check be stricter to check whether the next-hop has actually been modified?
Can you look at `TransferBDDState` and only panic if `_nextHopSet` is true?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java` line 18 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6LRL403kyWRmteMt4:-N_6LRL403kyWRmteMt5:b3tct2v) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java#L18)):*
> ```Java
>   public void testCopyConstructorAndEquality() {
>     BDDFactory factory = JFactory.init(100, 100);
>     BDDRoute r1 = new BDDRoute(factory, 3, 4, 5, 6, 2);
> ```

is 6 the number of interfaces? Is it meaningful somehow or just random?
___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 417 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6OnMu2PFNaq8kabv8:-N_6OnMu2PFNaq8kabv9:b-6d92sg) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L417)):*
> ```Java
>    * @return a version of the result suitable for output from this analysis
>    */
>   public static Result<BgpRoute> toQuestionResult(Result<Bgpv4Route> result) {
> ```

Is there a fundamental reason that we need two formats or is it trying to maintain backwards compatibility?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: all files reviewed, 5 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddress.java` line 9 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_655L0EPG84_B1Hmgu:-N_9pQM0D-QNpMzWHO93:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddress.java#L9)):*

Can you amend the comment to explain when objects of this class are used by SRP?
</blockquote></details>

Done.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 10 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_64mooAcAL6S0rh-fD:-N_9pRQjAcxkLvjnKFKy:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L10)):*

Same as above.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 109 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_66wQv4q1h4Tw4SQ-s:-N_9pSRu-bgpKEWx3QgG:btqqu5l) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L109)):*

This is not a BDD because it does not depend on the input, right?
</blockquote></details>

Correct.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 193 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6928oDR0XTndmi3NP:-N_9pUEC5KdCDGXwl33m:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L193)):*

"at most one NextHop Interface can be set"?
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 554 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_69mVu-Kk-dRxBnTyO:-N_9pVSO2zZaR45N11rf:bz7ngai) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L554)):*

can this check be stricter to check whether the next-hop has actually been modified?
Can you look at `TransferBDDState` and only panic if `_nextHopSet` is true?
</blockquote></details>

Good call.  Done.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java` line 18 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6LRL403kyWRmteMt4:-N_9pXDC22YmySUhCzKb:b-2qces) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java#L18)):*

is 6 the number of interfaces? Is it meaningful somehow or just random?
</blockquote></details>

Just random, for testing purposes.
___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 417 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6OnMu2PFNaq8kabv8:-N_9pa1-5Z83eVz2YOp4:bqhi607) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L417)):*

Is there a fundamental reason that we need two formats or is it trying to maintain backwards compatibility?
</blockquote></details>

Good question.  I don't know the history there ( @anothermattbrown may) and so have just kept things as is.  Logically I can see how it makes sense to separate the internal representation of routes used for the route simulation from the presentation of routes to users as part of question results, as the needs may be different and they may evolve differently over time.  It is true that currently they are very similar though.

anothermattbrown (commented):
> Reviewed 12 of 33 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 31 of 33 files reviewed, 9 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8740#-N_A1g4Z0Wqiuj_XYK-e:-N_A0tuy6mQw0QpOaU6S:bmgc1yb) ([raw file](https://github.com/batfish/batfish/blob/066ddd9ff51c035905ac378c2e6c099c475c0246/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java#L25)):*
> ```Java
>   @JsonSubTypes.Type(value = NextHopVtep.class, name = "vtep")
> })
> public interface NextHop extends NextHopResult, Serializable {
> ```

would be better not to modify this internal class. IIRC the idea is that `NextHopResult` is just for questions, and will include all `NextHop` variants plus a few more. In that case it's cleaner to either: create two separate class hierarchies and convert between them, or add a variant to `NextHopResult` that just wraps a `NextHop`. 
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopResult.java` line 19 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6MvUy-Jm72CcHCzWu:-N_6MvUzB-8l4rGbt6HI:b-9subuw) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopResult.java#L19)):*
> ```Java
>   @JsonSubTypes.Type(value = NextHopBgpPeerAddress.class)
> })
> public interface NextHopResult extends Serializable {}
> ```

if this class hierarchy is only for Questions, should probably live in the `org.batfish.datamodel.answers` package.

does it need to be `Serializable`? 
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6Nnkz7I5G3szvMrrx:-N_6Nnkz7I5G3szvMrry:bn6x1wf) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L29)):*
> ```Java
>   @Override
>   public int hashCode() {
>     return 569605681; // randomly generated
> ```

I'm not familiar with this pattern. Why prefer this over the default implementation?
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddressTest.java` line 15 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6OMbb876AgWB1r49n:-N_6OMbb876AgWB1r49o:bnfiepn) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/route/nh/NextHopBgpPeerAddressTest.java#L15)):*
> ```Java
> 
>   @Test
>   public void testJavaSerialiation() {
> ```

do we need java serialization?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 31 of 33 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8740#-N_A1g4Z0Wqiuj_XYK-e:-N_Ai5x7CH8WNmGL-5pY:b5kza4i) ([raw file](https://github.com/batfish/batfish/blob/066ddd9ff51c035905ac378c2e6c099c475c0246/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java#L25)):*

would be better not to modify this internal class. IIRC the idea is that `NextHopResult` is just for questions, and will include all `NextHop` variants plus a few more. In that case it's cleaner to either: create two separate class hierarchies and convert between them, or add a variant to `NextHopResult` that just wraps a `NextHop`. 
</blockquote></details>

Is `NextHop` really internal?  It seems that it's already being used as part of the output of other questions, for example maybe being provided in results about flows from traceroute?  And it's part of the pybatfish API as well.  So the distinction seems to be less about internal/external than about a "real" next-hop vs. a possibly symbolic one.

Let me know your thoughts.  I had originally gone with the separate hierarchies approach, but that seemed like too much code duplication.  If you think they should still be separated I can try the wrapper approach.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 31 of 33 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 417 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6OnMu2PFNaq8kabv8:-N_BcBW_4BuyrUaYUnuh:bi39n10) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L417)):*

Good question.  I don't know the history there ( @anothermattbrown may) and so have just kept things as is.  Logically I can see how it makes sense to separate the internal representation of routes used for the route simulation from the presentation of routes to users as part of question results, as the needs may be different and they may evolve differently over time.  It is true that currently they are very similar though.
</blockquote></details>

I think Todd is getting at the principle, tho I'd be stronger.

The "engine" in these questions should not be coupled - input, algorithms, or output - to the format of input spec or output presentation.

We're not great at this, but we've seen a pattern over and over and over again that we need to move the "engine" logic out of an Answerer so we can use it in other places. ;)

--

Another likely reason would be a consumer of this public API that's not open source.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 31 of 33 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8740#-N_A1g4Z0Wqiuj_XYK-e:-N_G9bGy1wMlt942WqCV:b-8na3ap) ([raw file](https://github.com/batfish/batfish/blob/066ddd9ff51c035905ac378c2e6c099c475c0246/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java#L25)):*

Is `NextHop` really internal?  It seems that it's already being used as part of the output of other questions, for example maybe being provided in results about flows from traceroute?  And it's part of the pybatfish API as well.  So the distinction seems to be less about internal/external than about a "real" next-hop vs. a possibly symbolic one.

Let me know your thoughts.  I had originally gone with the separate hierarchies approach, but that seemed like too much code duplication.  If you think they should still be separated I can try the wrapper approach.
</blockquote></details>

@dhalperi - this issue is related to the other one that you commented on in this PR, and Matt and I are interested in your opinion on how to structure this one.  Here's the situation.  Currently `NextHop` is being used both internally and externally in question results.  With this PR, now question results need to include a few more kinds of "symbolic" next-hops that don't correspond to "real" ones.  What I did (after considering some other options) is to create a new superinterface `NextHopResult` of `NextHop`, and add new symbolic kinds of next-hops as classes that extend `NextHopResult`.  So the internal stuff can continue to use `NextHop` but the external facing results use `NextHopResult`, which can be either a "real" next-hop or a symbolic one.  

The question is whether that is reasonable or whether there should be a fuller separation between the two; see above discussion.  Thanks.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 31 of 33 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8740#-N_A1g4Z0Wqiuj_XYK-e:-N_KglVz6xutR4rGQAWt:bjoclfc) ([raw file](https://github.com/batfish/batfish/blob/066ddd9ff51c035905ac378c2e6c099c475c0246/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java#L25)):*

@dhalperi - this issue is related to the other one that you commented on in this PR, and Matt and I are interested in your opinion on how to structure this one.  Here's the situation.  Currently `NextHop` is being used both internally and externally in question results.  With this PR, now question results need to include a few more kinds of "symbolic" next-hops that don't correspond to "real" ones.  What I did (after considering some other options) is to create a new superinterface `NextHopResult` of `NextHop`, and add new symbolic kinds of next-hops as classes that extend `NextHopResult`.  So the internal stuff can continue to use `NextHop` but the external facing results use `NextHopResult`, which can be either a "real" next-hop or a symbolic one.  

The question is whether that is reasonable or whether there should be a fuller separation between the two; see above discussion.  Thanks.
</blockquote></details>

I'm with Matt here. I'd introduce a question-specific next-hop implementation that's decoupled and evolvable separately from this one.

I'd also link the "Prefer composition over inheritance" guidance from Effective Java. If we want to reuse work to have the new abstaction internally be "a concrete next_hop, or one of these new ones", the new class could still be a union of this and some new values.

FWIW, every time we actually do expose a functional class as a presentation class, we have a lot of pain. In retrospect, I think it was a mistake to ever apply `@Json` annotations to anything used by Engine code :).

Protobuf, Smithy, and other API spec languages fix this and that's (part of) why all things we've wrapped around Batfish use those.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: 16 of 35 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java` line 417 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6OnMu2PFNaq8kabv8:-N_P-8fv6NpG3F2XQrcp:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/question/src/main/java/org/batfish/question/testroutepolicies/TestRoutePoliciesAnswerer.java#L417)):*

I think Todd is getting at the principle, tho I'd be stronger.

The "engine" in these questions should not be coupled - input, algorithms, or output - to the format of input spec or output presentation.

We're not great at this, but we've seen a pattern over and over and over again that we need to move the "engine" logic out of an Answerer so we can use it in other places. ;)

--

Another likely reason would be a consumer of this public API that's not open source.
</blockquote></details>

Done.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/8740#-N_A1g4Z0Wqiuj_XYK-e:-N_OzXoKAd_kiAgBHxxE:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/066ddd9ff51c035905ac378c2e6c099c475c0246/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHop.java#L25)):*

I'm with Matt here. I'd introduce a question-specific next-hop implementation that's decoupled and evolvable separately from this one.

I'd also link the "Prefer composition over inheritance" guidance from Effective Java. If we want to reuse work to have the new abstaction internally be "a concrete next_hop, or one of these new ones", the new class could still be a union of this and some new values.

FWIW, every time we actually do expose a functional class as a presentation class, we have a lot of pain. In retrospect, I think it was a mistake to ever apply `@Json` annotations to anything used by Engine code :).

Protobuf, Smithy, and other API spec languages fix this and that's (part of) why all things we've wrapped around Batfish use those.
</blockquote></details>

Done.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopResult.java` line 19 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6MvUy-Jm72CcHCzWu:-N_OzZEP8niaZYBHZCVX:b-25kisz) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopResult.java#L19)):*

if this class hierarchy is only for Questions, should probably live in the `org.batfish.datamodel.answers` package.

does it need to be `Serializable`? 
</blockquote></details>

Moved the hierarchy as suggested, and got rid of serializability.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6Nnkz7I5G3szvMrrx:-N_OzdxrBAfck9OwU3r2:b-c8dlws) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L29)):*

I'm not familiar with this pattern. Why prefer this over the default implementation?
</blockquote></details>

This style was copied from similar classes, like NextHopDiscard.  The contract of hashCode is that if two objects are equal then their hashes should be the same.  Here any two objects of the class NextHopSelf are considered equal to one another, but that then means that we can't use the default hashCode implementation or else they will have different hashes.

anothermattbrown (commented):
> Reviewed 7 of 33 files at r1, 2 of 4 files at r2, 11 of 17 files at r3, 11 of 11 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: all files reviewed, 7 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/question/BUILD.bazel` line 28 at r4](https://reviewable.io/reviews/batfish/batfish/8740#-N_jZjCHDTezZUlHvGvW:-N_jZjCHDTezZUlHvGvX:bpyw6tx) ([raw file](https://github.com/batfish/batfish/blob/508db52930d34c8ece0b32e2640a96578549835f/projects/question/BUILD.bazel#L28)):*
> ```Starlark
>         "@maven//:com_google_guava_guava",
>         "@maven//:org_apache_commons_commons_lang3",
>         "@maven//:org_parboiled_parboiled_core",
> ```

is this really part of this change, or did it sneak in somehow?
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6Nnkz7I5G3szvMrrx:-N_jXcfM7eXOkML_-LJC:bsrq0r9) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L29)):*

This style was copied from similar classes, like NextHopDiscard.  The contract of hashCode is that if two objects are equal then their hashes should be the same.  Here any two objects of the class NextHopSelf are considered equal to one another, but that then means that we can't use the default hashCode implementation or else they will have different hashes.
</blockquote></details>

ok -- sounds like we can't ensure this is a singleton class

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/question/BUILD.bazel` line 28 at r4](https://reviewable.io/reviews/batfish/batfish/8740#-N_jZjCHDTezZUlHvGvW:-N_kLedW7glGhI7633H1:b-kp1qbh) ([raw file](https://github.com/batfish/batfish/blob/508db52930d34c8ece0b32e2640a96578549835f/projects/question/BUILD.bazel#L28)):*

is this really part of this change, or did it sneak in somehow?
</blockquote></details>

This is part of the change -- it's needed for the Preconditions.checkArgument method.
___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/8740#-N_6Nnkz7I5G3szvMrrx:-N_kLTh_6FKtkCSEA0sa:b-cn4tbq) ([raw file](https://github.com/batfish/batfish/blob/05f10f74ef3c7975dae2d9106f5c71bf7c4bbfbf/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/route/nh/NextHopSelf.java#L29)):*

ok -- sounds like we can't ensure this is a singleton class
</blockquote></details>

It's a singleton from the outside, but nothing prevents (accidentally) creating multiple instances from within the class definition itself.

anothermattbrown (commented):
> Reviewed 1 of 33 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/question/BUILD.bazel` line 28 at r4](https://reviewable.io/reviews/batfish/batfish/8740#-N_jZjCHDTezZUlHvGvW:-Na7fAbs3kJubof42Lwu:b-a08mj5) ([raw file](https://github.com/batfish/batfish/blob/508db52930d34c8ece0b32e2640a96578549835f/projects/question/BUILD.bazel#L28)):*

This is part of the change -- it's needed for the Preconditions.checkArgument method.
</blockquote></details>

we should be using guava's `checkArgument`, not parboiled's

dhalperi:
*[`projects/question/BUILD.bazel` line 28 at r4](https://reviewable.io/reviews/batfish/batfish/8740#-N_jZjCHDTezZUlHvGvW:-Na7gWcWCoolJyWMpzr-:b-1rdgmw) ([raw file](https://github.com/batfish/batfish/blob/508db52930d34c8ece0b32e2640a96578549835f/projects/question/BUILD.bazel#L28)):*

we should be using guava's `checkArgument`, not parboiled's
</blockquote></details>

Note to self: wtf? I thought this was disallowed by code analysis tools

anothermattbrown (commented):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

anothermattbrown (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8740#-:-Na7mXEF8A8uj-HoVvn5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8763, opened 2023-07-28
Bugfix: Consider using the output/intermediate route when matching on the metric

millstein:
Check for usage of output/intermediate route attributes when matching on the metric, similar to how matching is done for other attributes; reorder the existing checks when matching on the tag to be consistent with how it's done elsewhere.


progwriter (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8763#-:-NbGCGel3AUFCeYjoXQt:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8768, opened 2023-08-07
Validation for TransferBDD tests

millstein:
Adds validation for each TransferBDD test against Batfish's concrete route simulation -- for each symbolic path produced by TransferBDD for a route map, we solve for an input route that takes that path and compare the expected result to what the concrete route simulation produces.

This PR also fixes two modeling errors in TransferBDD that were identified via this validation:
* When matching on the next hop we need to check the prefix length in addition to the IP address.
* The SetReadIntermediateBgpAttributes directive should be used to determine whether to read from the original input route attributes or the current ones when matching/reading attributes.


nickgian (commented):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8768)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 745 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh30_VW76N_0ksvtQSc:-Nh30_VXEyogf6E72A0F:buxrqtk) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L745)):*
> ```Java
>            * <p>This code, {@link TransferBDD}, properly handles two common cases: 1) platforms that
>            * never use these directives, so that the original route attributes are always read; and
>            * 2) platforms that always use the SetRead... and SetWrite... directives up front, so
> ```

is there any easy way to detect a violation of this assumption in our code and throw an unsupported feature error? e.g., can we set a flag if some other state mutating statement has been processed before these?
I suppose we can at least easily check if you try to turn it off.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 1344 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh37Xc_4S_at3nQ_nPu:-Nh37XcaBbHP36ZijJuH:b2utol) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1344)):*
> ```Java
>   // Returns the appropriate route to use for matching on attributes.
>   private BDDRoute routeForMatching(TransferParam p) {
>     return _useOutputAttributes || p.getReadIntermediateBgpAtttributes()
> ```

I've missed this, where does `useOutputAttributes` come from, when is it true? It's a different mechanism for `ReadIntermediateBgpAttributes` right?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 239 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh3Bp0o6aEn-0WnHq_D:-Nh3Bp0o6aEn-0WnHq_E:b7yyjgp) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L239)):*
> ```Java
>    * @return a boolean indicating whether the check succeeded
>    */
>   private boolean validatePaths(
> ```

Just throwing it out there, an alternative signature to consider for this method:
`private List<TransferReturn> paths validatePaths(TransferBDD tbdd)`
computing the paths and asserting they are equal to TRP's model, then returning them for any further assertions.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8768)* status: all files reviewed, 3 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 745 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh30_VW76N_0ksvtQSc:-Nh3USxiBeqVl8pSlkN1:b-7m9r9y) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L745)):*

is there any easy way to detect a violation of this assumption in our code and throw an unsupported feature error? e.g., can we set a flag if some other state mutating statement has been processed before these?
I suppose we can at least easily check if you try to turn it off.
</blockquote></details>

Good question!  I removed the (no-op) handling of `UnsetWriteIntermediateBgpAttributes`, so now we'll throw the unsupported exception if that is used, to handle the case when writing to intermediate attributes is turned off.  For the other case, I realized that things are a bit more subtle than I thought -- for example, there are allowed to be writes before we turn on this feature, as long as those writes are never read.  So I don't think it's easy to check.  I updated the comment there to be more clear about what the expectations are, however.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 1344 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh37Xc_4S_at3nQ_nPu:-Nh3HywkB2BkPGkopAiu:b-9pjig) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1344)):*

I've missed this, where does `useOutputAttributes` come from, when is it true? It's a different mechanism for `ReadIntermediateBgpAttributes` right?
</blockquote></details>

Yes, it's a different mechanism.  It's based solely on the configuration format (Cisco, Juniper, etc.) and is determined by the method `Environment.useOutputAttributesFor`.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 239 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh3Bp0o6aEn-0WnHq_D:-Nh3IVqs8HYHCe6ReiVt:b-c3ff4t) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L239)):*

Just throwing it out there, an alternative signature to consider for this method:
`private List<TransferReturn> paths validatePaths(TransferBDD tbdd)`
computing the paths and asserting they are equal to TRP's model, then returning them for any further assertions.
</blockquote></details>

Logically I like keeping the generation of paths separate from their validation.  This also makes sense because in some cases (e.g., where we have unsupported features) we want to skip validation and just do the generation so we can check other properties of the paths.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8768#-:-NhmEmODElDoSHDiY6up:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 745 at r2](https://reviewable.io/reviews/batfish/batfish/8768#-Nh30_VW76N_0ksvtQSc:-NhmEWd67kC8AwhQXF96:b-e349s5) ([raw file](https://github.com/batfish/batfish/blob/d18cbc8c7e383e2c5119285778d8decb04920e0b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L745)):*

Good question!  I removed the (no-op) handling of `UnsetWriteIntermediateBgpAttributes`, so now we'll throw the unsupported exception if that is used, to handle the case when writing to intermediate attributes is turned off.  For the other case, I realized that things are a bit more subtle than I thought -- for example, there are allowed to be writes before we turn on this feature, as long as those writes are never read.  So I don't think it's easy to check.  I updated the comment there to be more clear about what the expectations are, however.
</blockquote></details>

can you create an issue with some details about what the semantics are missing so we can keep track of it in the future?

PR #8777, opened 2023-08-17
BgpSessionStatus: use established session instead of configured values if present

dhalperi:
Mainly affects local/remote IP.

The new test fails without this change (Local_IP is null).


dhalperi:
I want to mess with it some more.

PR #8787, opened 2023-08-25
SemDiff question.

nickgian:
This PR introduces a new differential question which uses CompareRoutePolicies to semantically compare routing policies that were changed. The key idea is for each peer group, do a lightweight syntactic comparison between the old and new snapshot, and if the result says they differ then check the exact differences using CRP.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java
@@ -1,0 +1,48 @@
+package org.batfish.minesweeper.collectors;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchAny;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchExpr;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchExprReference;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchExprVisitor;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchRegex;
+import org.batfish.datamodel.routing_policy.as_path.AsSetsMatchingRanges;
+import org.batfish.datamodel.routing_policy.as_path.HasAsPathLength;
+
+/** Collect all AS-path list names in a {@link AsPathMatchExpr}. */
+@ParametersAreNonnullByDefault
+public class AsPathNameAsPathMatchExprCollector
+    implements AsPathMatchExprVisitor<Set<String>, Configuration> {
+
+  @Override
+  public Set<String> visitAsPathMatchAny(AsPathMatchAny asPathMatchAny, Configuration arg) {
+    return asPathMatchAny.getDisjuncts().stream()
+        .flatMap(expr -> expr.accept(this, arg).stream())
+        .collect(ImmutableSet.toImmutableSet());
+  }
+
+  @Override
+  public Set<String> visitAsPathMatchExprReference(
+      AsPathMatchExprReference asPathMatchExprReference, Configuration arg) {
+    return ImmutableSet.of(asPathMatchExprReference.getName());
+  }
+
+  @Override
+  public Set<String> visitAsPathMatchRegex(AsPathMatchRegex asPathMatchRegex, Configuration arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitAsSetsMatchingRanges(
+      AsSetsMatchingRanges asSetsMatchingRanges, Configuration arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitHasAsPathLength(HasAsPathLength hasAsPathLength, Configuration arg) {
+    return ImmutableSet.of();
+  }
+}
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameBooleanExprCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameBooleanExprCollector.java
@@ -1,0 +1,36 @@
+package org.batfish.minesweeper.collectors;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.as_path.AsPathMatchExpr;
+import org.batfish.datamodel.routing_policy.as_path.MatchAsPath;
+import org.batfish.datamodel.routing_policy.expr.AsPathSetExpr;
+import org.batfish.datamodel.routing_policy.expr.BooleanExpr;
+import org.batfish.datamodel.routing_policy.expr.LegacyMatchAsPath;
+import org.batfish.datamodel.routing_policy.expr.NamedAsPathSet;
+import org.batfish.minesweeper.aspath.BooleanExprMatchCollector;
+import org.batfish.minesweeper.utils.Tuple;
+
+/** Collect all AS-path list names in a {@link BooleanExpr}. */
+@ParametersAreNonnullByDefault
+public class AsPathNameBooleanExprCollector extends BooleanExprMatchCollector<String> {
+
+  @Override
+  public Set<String> visitMatchAsPath(
+      MatchAsPath matchAsPath, Tuple<Set<String>, Configuration> arg) {
+    AsPathMatchExpr matchExpr = matchAsPath.getAsPathMatchExpr();
+    return matchExpr.accept(new AsPathNameAsPathMatchExprCollector(), arg.getSecond());
+  }
+
+  @Override
+  public Set<String> visitMatchLegacyAsPath(
+      LegacyMatchAsPath legacyMatchAsPath, Tuple<Set<String>, Configuration> arg) {
+    AsPathSetExpr expr = legacyMatchAsPath.getExpr();
+    if (expr instanceof NamedAsPathSet) {
+      return ImmutableSet.of(((NamedAsPathSet) expr).getName());
+    }
+    return ImmutableSet.of();
+  }
+}
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java
@@ -1,0 +1,64 @@
+package org.batfish.minesweeper.collectors;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.communities.MatchCommunities;
+import org.batfish.datamodel.routing_policy.expr.BooleanExpr;
+import org.batfish.datamodel.routing_policy.expr.CallExpr;
+import org.batfish.datamodel.routing_policy.expr.WithEnvironmentExpr;
+import org.batfish.minesweeper.aspath.BooleanExprMatchCollector;
+import org.batfish.minesweeper.utils.Tuple;
+
+/** Collect all community-list names in a {@link BooleanExpr}. */
+@ParametersAreNonnullByDefault
+public class CommunityBooleanExprCollector extends BooleanExprMatchCollector<String> {
+
+  @Override
+  public Set<String> visitMatchCommunities(
+      MatchCommunities matchCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.<String>builder()
+        .addAll(matchCommunities.getCommunitySetExpr().accept(new CommunitySetExprCollector(), arg))
+        .addAll(
+            matchCommunities
+                .getCommunitySetMatchExpr()
+                .accept(new CommunitySetMatchExprCollector(), arg))
+        .build();
+  }
+
+  @Override
+  public Set<String> visitCallExpr(CallExpr callExpr, Tuple<Set<String>, Configuration> arg) {
+    if (arg.getFirst().contains(callExpr.getCalledPolicyName())) {
+      // If we have already visited this policy then don't visit again
+      return ImmutableSet.of();
+    }
+    // Otherwise update the set of seen policies and recurse.
+    arg.getFirst().add(callExpr.getCalledPolicyName());
+
+    return new RoutingPolicyStatementCommunityCollector()
+        .visitAll(
+            arg.getSecond()
+                .getRoutingPolicies()
+                .get(callExpr.getCalledPolicyName())
+                .getStatements(),
+            arg);
+  }
+
+  @Override
+  public Set<String> visitWithEnvironmentExpr(
+      WithEnvironmentExpr withEnvironmentExpr, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.<String>builder()
+        .addAll(withEnvironmentExpr.getExpr().accept(this, arg))
+        .addAll(
+            new RoutingPolicyStatementCommunityCollector()
+                .visitAll(withEnvironmentExpr.getPreStatements(), arg))
+        .addAll(
+            new RoutingPolicyStatementCommunityCollector()
+                .visitAll(withEnvironmentExpr.getPostStatements(), arg))
+        .addAll(
+            new RoutingPolicyStatementCommunityCollector()
+                .visitAll(withEnvironmentExpr.getPostTrueStatements(), arg))
+        .build();
+  }
+}
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityMatchExprCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityMatchExprCollector.java
@@ -1,0 +1,214 @@
+package org.batfish.minesweeper.collectors;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.batfish.common.BatfishException;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.communities.AllExtendedCommunities;
+import org.batfish.datamodel.routing_policy.communities.AllLargeCommunities;
+import org.batfish.datamodel.routing_policy.communities.AllStandardCommunities;
+import org.batfish.datamodel.routing_policy.communities.CommunityAcl;
+import org.batfish.datamodel.routing_policy.communities.CommunityAclLine;
+import org.batfish.datamodel.routing_policy.communities.CommunityIn;
+import org.batfish.datamodel.routing_policy.communities.CommunityIs;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchAll;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchAny;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchExpr;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchExprReference;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchExprVisitor;
+import org.batfish.datamodel.routing_policy.communities.CommunityMatchRegex;
+import org.batfish.datamodel.routing_policy.communities.CommunityNot;
+import org.batfish.datamodel.routing_policy.communities.ExtendedCommunityGlobalAdministratorHighMatch;
+import org.batfish.datamodel.routing_policy.communities.ExtendedCommunityGlobalAdministratorLowMatch;
+import org.batfish.datamodel.routing_policy.communities.ExtendedCommunityGlobalAdministratorMatch;
+import org.batfish.datamodel.routing_policy.communities.ExtendedCommunityLocalAdministratorMatch;
+import org.batfish.datamodel.routing_policy.communities.OpaqueExtendedCommunities;
+import org.batfish.datamodel.routing_policy.communities.RouteTargetExtendedCommunities;
+import org.batfish.datamodel.routing_policy.communities.SiteOfOriginExtendedCommunities;
+import org.batfish.datamodel.routing_policy.communities.StandardCommunityHighMatch;
+import org.batfish.datamodel.routing_policy.communities.StandardCommunityLowMatch;
+import org.batfish.datamodel.routing_policy.communities.VpnDistinguisherExtendedCommunities;
+import org.batfish.minesweeper.utils.Tuple;
+
+/**
+ * Collect all community-list names in a {@link CommunityMatchExpr}. The visitor takes as argument a
+ * set of seen community names (used to break out of potential infinite loops in a malformed
+ * snapshot) and a Batfish {@link Configuration} used to dereference some community expressions.
+ */
+@ParametersAreNonnullByDefault
+public class CommunityMatchExprCollector
+    implements CommunityMatchExprVisitor<Set<String>, Tuple<Set<String>, Configuration>> {
+
+  private static final Logger LOGGER = LogManager.getLogger(CommunityMatchExprCollector.class);
+
+  @Override
+  public Set<String> visitAllExtendedCommunities(
+      AllExtendedCommunities allExtendedCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitAllLargeCommunities(
+      AllLargeCommunities allLargeCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitAllStandardCommunities(
+      AllStandardCommunities allStandardCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitCommunityAcl(
+      CommunityAcl communityAcl, Tuple<Set<String>, Configuration> arg) {
+    return visitAll(
+        communityAcl.getLines().stream()
+            .map(CommunityAclLine::getCommunityMatchExpr)
+            .collect(ImmutableList.toImmutableList()),
+        arg);
+  }
+
+  @Override
+  public Set<String> visitCommunityIn(
+      CommunityIn communityIn, Tuple<Set<String>, Configuration> arg) {
+    return communityIn.getCommunitySetExpr().accept(new CommunitySetExprCollector(), arg);
+  }
+
+  @Override
+  public Set<String> visitCommunityIs(
+      CommunityIs communityIs, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitCommunityMatchAll(
+      CommunityMatchAll communityMatchAll, Tuple<Set<String>, Configuration> arg) {
+    return visitAll(communityMatchAll.getExprs(), arg);
+  }
+
+  @Override
+  public Set<String> visitCommunityMatchAny(
+      CommunityMatchAny communityMatchAny, Tuple<Set<String>, Configuration> arg) {
+    return visitAll(communityMatchAny.getExprs(), arg);
+  }
+
+  @Override
+  public Set<String> visitCommunityMatchExprReference(
+      CommunityMatchExprReference communityMatchExprReference,
+      Tuple<Set<String>, Configuration> arg) {
+    String name = communityMatchExprReference.getName();
+    // In case we have already dereferenced this community in this visit, issue a warning and return
+    // an empty set.
+    if (arg.getFirst().contains(name)) {
+      LOGGER.warn("Cycle detected in communities - this denotes a malformed snapshot.");
+      return ImmutableSet.of();
+    }
+
+    CommunityMatchExpr matchExpr = arg.getSecond().getCommunityMatchExprs().get(name);
+    if (matchExpr == null) {
+      throw new BatfishException("Cannot find community match expression: " + name);
+    }
+
+    // Add this name to set of seen references.
+    Set<String> newSeen = new HashSet<>(arg.getFirst());
+    newSeen.add(name);
+    return ImmutableSet.<String>builder()
+        .add(name)
+        .addAll(matchExpr.accept(this, new Tuple<>(newSeen, arg.getSecond())))
+        .build();
+  }
+
+  @Override
+  public Set<String> visitCommunityMatchRegex(
+      CommunityMatchRegex communityMatchRegex, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitCommunityNot(
+      CommunityNot communityNot, Tuple<Set<String>, Configuration> arg) {
+    return communityNot.getExpr().accept(this, arg);
+  }
+
+  @Override
+  public Set<String> visitExtendedCommunityGlobalAdministratorHighMatch(
+      ExtendedCommunityGlobalAdministratorHighMatch extendedCommunityGlobalAdministratorHighMatch,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitExtendedCommunityGlobalAdministratorLowMatch(
+      ExtendedCommunityGlobalAdministratorLowMatch extendedCommunityGlobalAdministratorLowMatch,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitExtendedCommunityGlobalAdministratorMatch(
+      ExtendedCommunityGlobalAdministratorMatch extendedCommunityGlobalAdministratorMatch,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitExtendedCommunityLocalAdministratorMatch(
+      ExtendedCommunityLocalAdministratorMatch extendedCommunityLocalAdministratorMatch,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitOpaqueExtendedCommunities(
+      OpaqueExtendedCommunities opaqueExtendedCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitRouteTargetExtendedCommunities(
+      RouteTargetExtendedCommunities routeTargetExtendedCommunities,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitSiteOfOriginExtendedCommunities(
+      SiteOfOriginExtendedCommunities siteOfOriginExtendedCommunities,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitStandardCommunityHighMatch(
+      StandardCommunityHighMatch standardCommunityHighMatch,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitStandardCommunityLowMatch(
+      StandardCommunityLowMatch standardCommunityLowMatch, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitVpnDistinguisherExtendedCommunities(
+      VpnDistinguisherExtendedCommunities vpnDistinguisherExtendedCommunities,
+      Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  private Set<String> visitAll(
+      Collection<CommunityMatchExpr> exprs, Tuple<Set<String>, Configuration> arg) {
+    return exprs.stream()
+        .flatMap(expr -> expr.accept(this, arg).stream())
+        .collect(ImmutableSet.toImmutableSet());
+  }
+}
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunitySetExprCollector.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunitySetExprCollector.java
@@ -1,0 +1,96 @@
+package org.batfish.minesweeper.collectors;
+
+import com.google.common.collect.ImmutableSet;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.batfish.common.BatfishException;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.communities.CommunityExprsSet;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetDifference;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetExpr;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetExprReference;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetExprVisitor;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetReference;
+import org.batfish.datamodel.routing_policy.communities.CommunitySetUnion;
+import org.batfish.datamodel.routing_policy.communities.InputCommunities;
+import org.batfish.datamodel.routing_policy.communities.LiteralCommunitySet;
+import org.batfish.minesweeper.utils.Tuple;
+
+/** Collect all community-lists referenced in a {@link CommunitySetExpr}. */
+@ParametersAreNonnullByDefault
+public class CommunitySetExprCollector
+    implements CommunitySetExprVisitor<Set<String>, Tuple<Set<String>, Configuration>> {
+
+  private static final Logger LOGGER = LogManager.getLogger(CommunitySetExprCollector.class);
+
+  @Override
+  public Set<String> visitCommunityExprsSet(
+      CommunityExprsSet communityExprsSet, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitCommunitySetDifference(
+      CommunitySetDifference communitySetDifference, Tuple<Set<String>, Configuration> arg) {
+    Set<String> s1 = communitySetDifference.getInitial().accept(this, arg);
+    Set<String> s2 =
+        communitySetDifference.getRemovalCriterion().accept(new CommunityMatchExprCollector(), arg);
+    return ImmutableSet.<String>builder().addAll(s1).addAll(s2).build();
+  }
+
+  @Override
+  public Set<String> visitCommunitySetExprReference(
+      CommunitySetExprReference communitySetExprReference, Tuple<Set<String>, Configuration> arg) {
+    String name = communitySetExprReference.getName();
+    // In case we have already dereferenced this community in this visit, issue a warning and return
+    // an empty set.
+    if (arg.getFirst().contains(name)) {
+      LOGGER.warn("Cycle detected in communities - this denotes a malformed snapshot.");
+      return ImmutableSet.of();
+    }
+
+    CommunitySetExpr setExpr = arg.getSecond().getCommunitySetExprs().get(name);
+    if (setExpr == null) {
+      throw new BatfishException("Cannot find community set expression: " + name);
+    }
+
+    // Add this name to set of seen references.
+    Set<String> newSeen = new HashSet<>(arg.getFirst());
+    newSeen.add(name);
+
+    return ImmutableSet.<String>builder()
+        .add(name)
+        .addAll(setExpr.accept(this, new Tuple<>(newSeen, arg.getSecond())))
+        .build();
+  }
+
+  @Override
+  public Set<String> visitCommunitySetReference(
+      CommunitySetReference communitySetReference, Tuple<Set<String>, Configuration> arg) {
+    String name = communitySetReference.getName();
+    return ImmutableSet.of(name);
+  }
+
+  @Override
+  public Set<String> visitCommunitySetUnion(
+      CommunitySetUnion communitySetUnion, Tuple<Set<String>, Configuration> arg) {
+    return communitySetUnion.getExprs().stream()
+        .flatMap(e -> e.accept(this, arg).stream())
+        .collect(ImmutableSet.toImmutableSet());
+  }
+
+  @Override
+  public Set<String> visitInputCommunities(
+      InputCommunities inputCommunities, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public Set<String> visitLiteralCommunitySet(
+      LiteralCommunitySet literalCommunitySet, Tuple<Set<String>, Configuration> arg) {
+    return ImmutableSet.of();
+  }
+}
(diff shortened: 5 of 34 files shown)


dhalperi (commented):
> Reviewed 1 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 1 of 26 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`questions/experimental/semdiff.json` line 1 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ncx6SY32q0mfkNAN-bH:-Ncx6SY32q0mfkNAN-bI:b-ft30ss) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/questions/experimental/semdiff.json#L1)):*
> ```JSON
> {
> ```

From a user perspective, can we pick a clearer name that better fits with other Batfish names? One suggestion might be: `compareBgpPolicies`, but do your own thinking.

Thoughts:
* "Semantic difference" feels more academic rather than what it does for users
* It's BGP specific
* we have `compareFilters` to which this feels pretty similar.

millstein (commented):
> Reviewed 10 of 26 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 11 of 26 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java` line 45 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgXDLR6C_wJJV6xeGk:-NdgXDLR6C_wJJV6xeGl:b-zhvfft) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java#L45)):*
> ```Java
> public class CommunityBooleanExprCollector
>     implements BooleanExprVisitor<Set<String>, Tuple<Set<String>, Configuration>> {
> ```

Can't this be simplified a lot by inheriting from `BooleanExprMatchCollector`?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java` line 66 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgYQca6D_E0rzGIt2b:-NdgYQca6D_E0rzGIt2c:bryiy7n) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java#L66)):*
> ```Java
>     // Otherwise update the set of seen policies and recurse.
>     arg.getFirst().add(callExpr.getCalledPolicyName());
> ```

The mutation of the set worries me a bit, particularly since you detect cycles later in CommunityMatchExprCollector.  It seems like the use of mutation here could lead to false positives there, since by mutating the set you're updating it for the rest of the traversal, not just this path.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java` line 46 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgZMq475rsnQ6Kv_rv:-NdgZMq475rsnQ6Kv_rw:b2wkbkq) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java#L46)):*
> ```Java
> public class RouteFilterBooleanExprCollector
>     implements BooleanExprVisitor<Set<String>, Tuple<Set<String>, Configuration>> {
> ```

Again it seems like you could inherit from `BooleanExprMatchCollector` here.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java` line 67 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgZauS6ILuCp1mkcnZ:-NdgZauS6ILuCp1mkcn_:b-4c7gmo) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java#L67)):*
> ```Java
>     // Otherwise update the set of seen policies and recurse.
>     arg.getFirst().add(callExpr.getCalledPolicyName());
> ```

Mutation here as well.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementCommunityCollector.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgZluZ0LwXcq-SARrU:-NdgZluZ0LwXcq-SARrV:bodoesh) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementCommunityCollector.java#L57)):*
> ```Java
>     // Otherwise update the set of seen policies and continue.
>     arg.getFirst().add(callStatement.getCalledPolicyName());
> ```

And here.  I guess it's ok if you're not trying to detect cycles, but still the fact that we sometimes copy and sometimes don't is confusing and potentially error prone.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java` line 42 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndg_S9s8bGJW-UFp6-g:-Ndg_S9s8bGJW-UFp6-h:b-pgyxfx) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java#L42)):*
> ```Java
> public class RoutingPolicyStatementRouteFilterCollector
>     implements StatementVisitor<Set<String>, Tuple<Set<String>, Configuration>> {
>   @Override
> ```

You might be able to inherit from `RoutePolicyStatementMatchCollector` which would be a lot simpler.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndg_2x21wA_vSWxrK3I:-Ndg_2x21wA_vSWxrK3J:b-28gv60) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java#L57)):*
> ```Java
>     // Otherwise update the set of seen policies and continue.
>     arg.getFirst().add(callStatement.getCalledPolicyName());
> ```

Another one.  Just marking them for completeness.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 11 of 26 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`questions/experimental/semdiff.json` line 1 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ncx6SY32q0mfkNAN-bH:-Ndg_qej91BUvrcZcVlx:b32v57) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/questions/experimental/semdiff.json#L1)):*

From a user perspective, can we pick a clearer name that better fits with other Batfish names? One suggestion might be: `compareBgpPolicies`, but do your own thinking.

Thoughts:
* "Semantic difference" feels more academic rather than what it does for users
* It's BGP specific
* we have `compareFilters` to which this feels pretty similar.
</blockquote></details>

+1

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 11 of 26 files reviewed, 11 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-Ndgcdlx8ghKnvNBIMLW:b-qicw2a) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*
> ```Java
> 
>   /** Pretty prints a route-map trace. Assumes there are no route-map calls. */
>   private static String prettyPrintTrace(List<TraceTree> trace) {
> ```

Are we not handling the case when there are route-map calls, or is that handled in some other way?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 64 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndge4Kv0tn6AqvloKFS:-Ndge4Kv0tn6AqvloKFT:b-m1etyl) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L64)):*
> ```Java
>   /**
>    * This method only looks through the names used by the currentConfig (and the currentPolicy).
>    * This is because it assumes that the referencePolicy and currentPolicy are syntactically equal,
> ```

So this difference is only done when the two policies are syntactically equal?  What is the intuition there?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 142 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgf9tw1KEDGyayeJnX:-Ndgf9tx4fJYgcuyxWQ9:b8n4sf3) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L142)):*
> ```Java
>     // TODO: For syntactic equality we should be doing a context-aware equality such that routing
>     // policy calls are dereferenced.
>     // That said, this works correctly for most purposes.
> ```

What is the issue?  Is it that two policies will be seen to be equal even though they are not, if the only differences are in other policies that they call?

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 7 of 26 files reviewed, 11 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java` line 45 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgXDLR6C_wJJV6xeGk:-Ndwne089B5J2t1h_iDL:bazqdvw) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java#L45)):*

Can't this be simplified a lot by inheriting from `BooleanExprMatchCollector`?
</blockquote></details>

Yes, but it's not as clear. Community expressions also appear in Set statements, while this class is only designed to go through Matches. The problem arises in methods like `visitCallExpr` which in `BooleanExprMatchCollector` calls `RoutePolicyStatementMatchCollector` rather than `RoutingPolicyStatementCommunityCollector` which also handles `Set` statements.
We can reuse if we override these two, but it's not as nice...
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java` line 66 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgYQca6D_E0rzGIt2b:-Ndwnvk96qWk0S0e5wIW:bxm2d9z) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/CommunityBooleanExprCollector.java#L66)):*

The mutation of the set worries me a bit, particularly since you detect cycles later in CommunityMatchExprCollector.  It seems like the use of mutation here could lead to false positives there, since by mutating the set you're updating it for the rest of the traversal, not just this path.
</blockquote></details>

While this code is gone now that I extend the right classes, it is still present in those classes (I think I was the one to put this de-duplication there :P).
The goal is to mutate it for all paths. It suffices for one path to explore the called policy in callExpr and all others should avoid doing so, because the visitor is idempotent as it's building a Set.

CommunityMatchExprCollector is doing a different kind of deduplication/loop detection. It detects cycles in CommunityReferenceExpr which hint at a problematic snapshot (Matt suggested that).
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java` line 46 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgZMq475rsnQ6Kv_rv:-NdwpO-Z1YQpG90dDOfQ:b-7iq8wi) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RouteFilterBooleanExprCollector.java#L46)):*

Again it seems like you could inherit from `BooleanExprMatchCollector` here.
</blockquote></details>

true!
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementCommunityCollector.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-NdgZluZ0LwXcq-SARrU:-NdwpYji9VLeaaORJk85:b31860) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementCommunityCollector.java#L57)):*

And here.  I guess it's ok if you're not trying to detect cycles, but still the fact that we sometimes copy and sometimes don't is confusing and potentially error prone.
</blockquote></details>

indeed not trying to detect cycles, just avoid duplicate work.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java` line 42 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndg_S9s8bGJW-UFp6-g:-NdwpeoQ90r68peVSlX9:bb7sa0c) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java#L42)):*

You might be able to inherit from `RoutePolicyStatementMatchCollector` which would be a lot simpler.
</blockquote></details>

yes.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndg_2x21wA_vSWxrK3I:-NdwplGz2wd4wqZth32A:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutingPolicyStatementRouteFilterCollector.java#L57)):*

Another one.  Just marking them for completeness.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-NdwqZ2v9r7qD4_-xHbx:b-kqytka) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

Are we not handling the case when there are route-map calls, or is that handled in some other way?
</blockquote></details>

printing currently does not handle route-map calls, though it should be fixable. I will look into it, it wasn't high priority for the use case I had in mind.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 64 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndge4Kv0tn6AqvloKFS:-NdwqjqK1g5tp2bj9CVM:b-u42ihr) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L64)):*

So this difference is only done when the two policies are syntactically equal?  What is the intuition there?
</blockquote></details>

That's right. If the two policies differ syntactically then we already know we need to compare them semantically. If they syntactically equal we compare their contexts (list definitions) to see if a semantic comparison is necessary.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 142 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgf9tw1KEDGyayeJnX:-NdwrAWmEX_0Ak8Mp5BD:bebhih9) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L142)):*

What is the issue?  Is it that two policies will be seen to be equal even though they are not, if the only differences are in other policies that they call?
</blockquote></details>

yes, we need a custom visitor to do the comparison properly.
___
*[`questions/experimental/semdiff.json` line 1 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ncx6SY32q0mfkNAN-bH:-NdwrHAb13aTe0VFIhc7:b-ac4zx6) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/questions/experimental/semdiff.json#L1)):*

+1
</blockquote></details>

We also have `CompareRoutePolicies`. `CompareChangedRoutingPolicies` could be an option, albeit confusing.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 7 of 26 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-NdzNolw-ZZ68r0rK8qy:b8snwvp) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

printing currently does not handle route-map calls, though it should be fixable. I will look into it, it wasn't high priority for the use case I had in mind.
</blockquote></details>

What will happen if the trace contains route-map calls?  I feel like calls from one route map to another are common, so just trying to understand the consequence of this restriction/assumption.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 142 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgf9tw1KEDGyayeJnX:-NdzODp0872LtsCnT8tl:b-1az6vg) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L142)):*

yes, we need a custom visitor to do the comparison properly.
</blockquote></details>

I see.  But even if the policies are erroneously seen to be syntactically equal, your other code that does more semantic checks will still catch the differences, is that right?

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 7 of 26 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-Ne6eGwg3TPzcGarV8Q-:bkjlpvl) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

What will happen if the trace contains route-map calls?  I feel like calls from one route map to another are common, so just trying to understand the consequence of this restriction/assumption.
</blockquote></details>

The trace will use the wrong route-map name for called route-maps. The assumption is used to simplify printing; it should be relatively easy to do a better job at this to be honest.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 142 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgf9tw1KEDGyayeJnX:-Ne6dvW_-H19eS9BeuGO:b-hjj6t0) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L142)):*

I see.  But even if the policies are erroneously seen to be syntactically equal, your other code that does more semantic checks will still catch the differences, is that right?
</blockquote></details>

Yes, the idea is to have some efficient checks to rule out policies that are definitely semantically equivalent (i.e., syntactically equal and same context), and for the rest to use CRP to determine whether if/how they differ.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 7 of 26 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-Ne8ve8w0D5u-vWag_zG:bnziv70) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

The trace will use the wrong route-map name for called route-maps. The assumption is used to simplify printing; it should be relatively easy to do a better job at this to be honest.
</blockquote></details>

Seems worth fixing if it's easy to do, but I'll leave that to your judgment.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 7 of 26 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-NfIKKOn1WqavGqsFMw_:b-ophtxg) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

Seems worth fixing if it's easy to do, but I'll leave that to your judgment.
</blockquote></details>

Agreed, I think it's worth fixing.

anothermattbrown (commented):
> Reviewed 7 of 26 files at r1, 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 13 of 26 files reviewed, 5 unresolved discussions (waiting on @dhalperi, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java` line 17 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS-4d4-M5WIyEcnlrP:-NfS-4d4-M5WIyEcnlrQ:b87uqxw) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java#L17)):*
> ```Java
> /** Collect all AS-path list names in a {@link AsPathMatchExpr}. */
> @ParametersAreNonnullByDefault
> public class AsPathNameAsPathMatchExprCollector
> ```

this can do a lot of intermediate copies, which could meaningfully impact performance in some cases. you could consider rewriting to avoid that. e.g. pass in a consumer or a set (builder) to add into.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/OutputRouteDifference.java` line 22 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS0Ea09clPr-D2997x:-NfS0Ea09clPr-D2997y:br964b3) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/OutputRouteDifference.java#L22)):*
> ```Java
>   }
> 
>   public static ActionDifference compute_action_difference(String current, String reference) {
> ```

can you use Action enums instead of strings here?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 105 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS21vXEqfSqbGKDNik:-NfS21vXEqfSqbGKDNil:b-k9w3dm) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L105)):*
> ```Java
>       answer.append(diff.getKey()).append("\n");
>     }
>     return new StringAnswerElement(answer.toString());
> ```

The establish paradigm is to return a table answer containing structured data, and move the rending/pretty-printing logic to the client (pybatfish). That allows clients to access the details of the data programmatically. We also don't have questions that produce markdown output.

anothermattbrown (commented):
> Reviewed 1 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 14 of 26 files reviewed, 9 unresolved discussions (waiting on @dhalperi, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 18 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4JhT4XMDSHUv4kHH:-NfS4JhT4XMDSHUv4kHI:b-hb3l86) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L18)):*
> ```Java
> /**
>  * This class describes the diff of the context (prefix-lists, community-lists, as-path lists) under
>  * which a RoutingPolicy is interpreted.
> ```

can you be more precise about the kind of diff here? it seems like it's syntactic, but includes differences to (indirect) dependencies?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 20 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS5Nix5M3b1hcTwkyn:-NfS3Nlf-wJOLBDVvEez:bahz0yb) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L20)):*
> ```Java
>  * which a RoutingPolicy is interpreted.
>  */
> public class RoutingPolicyContextDiff {
> ```

please add a test for this class, including indirect references etc
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 65 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4hDjC6e5zjxQvJHg:-NfS4hDjC6e5zjxQvJHh:bojjxpc) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L65)):*
> ```Java
>    * This method only looks through the names used by the currentConfig (and the currentPolicy).
>    * This is because it assumes that the referencePolicy and currentPolicy are syntactically equal,
>    * hence there is no point also looking for references used through the referencePolicy.
> ```

what about indirect references? e.g. if the input policy calls another policy and that gets changed



_Code quote:_
```Java
there is no point also looking for references used through the referencePolicy.
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 105 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS5DcM-kVLaX2yDSrZ:-NfS5DcM-kVLaX2yDSr_:b-4c3mmh) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L105)):*
> ```Java
>       return true;
>     }
>     if (o == null || getClass() != o.getClass()) {
> ```

if you make this class final you can simplify this to `!(o instanceof RoutingPolicyContextDiff)`



_Code quote:_
```Java
o == null || getClass() != o.getClass()
```

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 9 of 33 files reviewed, 4 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java` line 17 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS-4d4-M5WIyEcnlrP:-NfhY2iz5nJidWZngqxo:bm8p5qq) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/AsPathNameAsPathMatchExprCollector.java#L17)):*

this can do a lot of intermediate copies, which could meaningfully impact performance in some cases. you could consider rewriting to avoid that. e.g. pass in a consumer or a set (builder) to add into.
</blockquote></details>

will punt on this for now to make progress into a functional artifact, but can revisit if performance is an issue.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java` line 105 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS21vXEqfSqbGKDNik:-Nfhm8inBHius4-juwGC:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/SemDiffAnswerer.java#L105)):*

The establish paradigm is to return a table answer containing structured data, and move the rending/pretty-printing logic to the client (pybatfish). That allows clients to access the details of the data programmatically. We also don't have questions that produce markdown output.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 18 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4JhT4XMDSHUv4kHH:-NfhbJcz3euPLlAb6pJN:b-axwj8r) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L18)):*

can you be more precise about the kind of diff here? it seems like it's syntactic, but includes differences to (indirect) dependencies?
</blockquote></details>

Not sure what you mean by indirect dependencies. This diff is purely syntactic between the list definitions in two configurations and the method `differ(policy)` essentially projects this difference on a given routing-policy. If there are no differences in the definitions used by `policy` it will return false, even if there are differences in other definitions that do not appear in `policy`. It does not perform a recursive exploration of policies called from `policy` (`SyntacticCompare` handles that).
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 20 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS5Nix5M3b1hcTwkyn:-Nfhc509Em1CxD41REFO:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L20)):*

please add a test for this class, including indirect references etc
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 65 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4hDjC6e5zjxQvJHg:-NfhcBEs3v_D8ZceI2s6:bt2jmhp) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L65)):*

what about indirect references? e.g. if the input policy calls another policy and that gets changed
</blockquote></details>

See `SyntacticCompare.areEqual` - it is handled there.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 105 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS5DcM-kVLaX2yDSrZ:-NfhcKEA7nid1hTdon31:boyexb3) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L105)):*

if you make this class final you can simplify this to `!(o instanceof RoutingPolicyContextDiff)`
</blockquote></details>

Done. Thanks!

anothermattbrown (commented):
> Reviewed 7 of 26 files at r1, 16 of 17 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: all files reviewed, 8 unresolved discussions (waiting on @dhalperi, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyBooleanExprCallCollector.java` line 10 at r4](https://reviewable.io/reviews/batfish/batfish/8787#-NflChEdCINIx07ESDvb:-NflChEe6GXEVvcLgf0P:bf2sf17) ([raw file](https://github.com/batfish/batfish/blob/d99a4370c4b4cdd3669da0a69ff1d0b89712381c/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyBooleanExprCallCollector.java#L10)):*
> ```Java
> import org.batfish.minesweeper.utils.Tuple;
> 
> public class RoutePolicyBooleanExprCallCollector extends BooleanExprMatchCollector<String> {
> ```

please make this final and add javadoc and tests
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyStatementCallCollector.java` line 17 at r4](https://reviewable.io/reviews/batfish/batfish/8787#-NflD1CQ35HzBNXUCxhn:-NflD1CQ35HzBNXUCxho:b-b1wfme) ([raw file](https://github.com/batfish/batfish/blob/d99a4370c4b4cdd3669da0a69ff1d0b89712381c/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyStatementCallCollector.java#L17)):*
> ```Java
>  * callees.
>  */
> public class RoutePolicyStatementCallCollector extends RoutePolicyStatementMatchCollector<String> {
> ```

please make this final and add a test.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswerer.java` line 47 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ3zIY9eHSUn3Ly6dC:-NgQ3zIZ3khxtN0JuQ4F:b-miemwn) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswerer.java#L47)):*
> ```Java
>   }
> 
>   public Stream<Tuple<Result<BgpRoute>, Result<BgpRoute>>> answerDiffHelper(
> ```

1. better to move utility helper functions out of the answerer into a utils class.
2. please add javadoc
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java` line 396 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ6RsqBjZCsuca4eyS:-NgQ6RsrBXJvPrme8yMe:b-nmxm0o) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java#L396)):*
> ```Java
>    * with route-maps in r1 of the new snapshot.
>    *
>    * <p>This method returns a structured representation suitable for use by other questions, rather
> ```

would be helpful to describe how to interpret this representation
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java` line 399 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ6M0u6rDgj93__eBE:-NgQ6M0u6rDgj93__eBF:b-zgssx4) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java#L399)):*
> ```Java
>    * than a TableAnswer.
>    */
>   public Stream<Tuple<Result<BgpRoute>, Result<BgpRoute>>> answerDiffHelper(
> ```

please move this to a utility class too
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 18 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4JhT4XMDSHUv4kHH:-NgQ4aS_6HP0WVshxq-v:b-y829m7) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L18)):*

Not sure what you mean by indirect dependencies. This diff is purely syntactic between the list definitions in two configurations and the method `differ(policy)` essentially projects this difference on a given routing-policy. If there are no differences in the definitions used by `policy` it will return false, even if there are differences in other definitions that do not appear in `policy`. It does not perform a recursive exploration of policies called from `policy` (`SyntacticCompare` handles that).
</blockquote></details>

yeah the non-recursive part is what I meant. would be helpful to make that clear in the javadoc
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 65 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4hDjC6e5zjxQvJHg:-NgQ4rtP42D9mivcp4se:b2uxxcv) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L65)):*

See `SyntacticCompare.areEqual` - it is handled there.
</blockquote></details>

might be helpful to point to that in the comment and draw the distinction

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 29 of 39 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyBooleanExprCallCollector.java` line 10 at r4](https://reviewable.io/reviews/batfish/batfish/8787#-NflChEdCINIx07ESDvb:-NglirP63uI6AOWW0Bgy:bffhyg6) ([raw file](https://github.com/batfish/batfish/blob/d99a4370c4b4cdd3669da0a69ff1d0b89712381c/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyBooleanExprCallCollector.java#L10)):*

please make this final and add javadoc and tests
</blockquote></details>

Done. Excellent callout on the missing tests, there was a bug there.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyStatementCallCollector.java` line 17 at r4](https://reviewable.io/reviews/batfish/batfish/8787#-NflD1CQ35HzBNXUCxhn:-Nglj0aR5p9vq9tOG4AI:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/d99a4370c4b4cdd3669da0a69ff1d0b89712381c/projects/minesweeper/src/main/java/org/batfish/minesweeper/collectors/RoutePolicyStatementCallCollector.java#L17)):*

please make this final and add a test.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswerer.java` line 47 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ3zIY9eHSUn3Ly6dC:-Nglj5xQBlPb5SNKuxBU:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswerer.java#L47)):*

1. better to move utility helper functions out of the answerer into a utils class.
2. please add javadoc
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java` line 396 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ6RsqBjZCsuca4eyS:-NgljQeP0qWJgshM8AG9:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java#L396)):*

would be helpful to describe how to interpret this representation
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java` line 399 at r5](https://reviewable.io/reviews/batfish/batfish/8787#-NgQ6M0u6rDgj93__eBE:-NgljT50AISym1mfD_NU:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/bcf4d554a5ed02e0d967c1444eecfd78ba1a51ef/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesAnswerer.java#L399)):*

please move this to a utility class too
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java` line 155 at r1](https://reviewable.io/reviews/batfish/batfish/8787#-Ndgcdlx8ghKnvNBIMLV:-NgljfFMAVlnOV5E3gP1:b3kl21r) ([raw file](https://github.com/batfish/batfish/blob/f41817327b9229f37b97b1d8dd416c6dc50b9df5/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/DifferenceAnswer.java#L155)):*

Agreed, I think it's worth fixing.
</blockquote></details>

This code is gone from this PR.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/OutputRouteDifference.java` line 22 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS0Ea09clPr-D2997x:-NgljoQ67pZMWIiw9T76:b3kl21r) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/OutputRouteDifference.java#L22)):*

can you use Action enums instead of strings here?
</blockquote></details>

This code is gone from this PR.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java` line 18 at r3](https://reviewable.io/reviews/batfish/batfish/8787#-NfS4JhT4XMDSHUv4kHH:-NgljDBf82yujRlul36s:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f63e7ddcd072296008a6e2dde55d5e5468b13db9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/semdiff/RoutingPolicyContextDiff.java#L18)):*

yeah the non-recursive part is what I meant. would be helpful to make that clear in the javadoc
</blockquote></details>

Done.

anothermattbrown (commented):
> Reviewed 6 of 10 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

anothermattbrown (commented):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 320 at r7](https://reviewable.io/reviews/batfish/batfish/8787#-NiMmWg81yMi6kx2B6oo:-NiMmWg81yMi6kx2B6op:b-7gnk8b) ([raw file](https://github.com/batfish/batfish/blob/73a5c73a10341988731048bdc5193751a7bef9ca/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L320)):*
> ```Java
>           }
>           break;
>         case AS_PATH:
> ```

is this change intended?

nickgian (commented):
> Reviewed 12 of 26 files at r1, 3 of 5 files at r2, 1 of 1 files at r3, 11 of 17 files at r4, 2 of 4 files at r5, 6 of 10 files at r6, 2 of 4 files at r7, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: 38 of 39 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 320 at r7](https://reviewable.io/reviews/batfish/batfish/8787#-NiMmWg81yMi6kx2B6oo:-NiVoj-g2QD_53cSN067:b-yklt9s) ([raw file](https://github.com/batfish/batfish/blob/73a5c73a10341988731048bdc5193751a7bef9ca/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L320)):*

is this change intended?
</blockquote></details>

Yes, it is due to a rebase from one of Todd's PR: https://github.com/batfish/batfish/commit/80f621cc950b164c1773ca96caccce00c78b606f
As part of this change I moved the definitions to a different file and then when I rebased I had to incorporate Todd's changes.

He added a comment too, which I missed when I rebased (added just now): 
"Note that AS-path prepends are  
\*     recorded concretely rather than symbolically in {@link BDDRoute}s, so their differences are  
\*     also ignored here."

nickgian (commented):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

anothermattbrown (commented):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8787)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickgian)

PR #8789, opened 2023-08-28
TransferBDD: optimize constraints on atomic predicates

millstein:
Introduces a few optimizations to avoid building large BDDs, which can happen when there are a large number of community or as-path atomic predicates. First, when solving for concrete routes that meet the constraints, we now do so from a partial model rather than a full model. Second, we avoid a few places where previously we created a constraint over all community atomic predicates.


nickgian (commented):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java` line 146 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLLgAO6l_W1HX0N5KE:-NdLLgAO6l_W1HX0N5KF:b-5h8hsk) ([raw file](https://github.com/batfish/batfish/blob/e860efeee73d476d9562457dccd8214fe156f7d7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java#L146)):*
> ```Java
>     // to negate a predicate on a single community, we diff it from a predicate representing
>     // any community. simply negating toBeNegated is not sufficient because it would allow a model
>     // where all atomic predicates are false, which doesn't correspond to any concrete communities.
> ```

why is the old comment no longer valid?
related, is there a semantic meaning to all atomic predicates being false? (no communities attached?)
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 186 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLOTMv74TGxz63dVCS:-NdLOTMv74TGxz63dVCT:b-16vl9z) ([raw file](https://github.com/batfish/batfish/blob/2a1e2c8f7f29dfcae653dc39ef2fe38f7d0b9c47/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L186)):*
> ```Java
>                 });
> 
>     /**
> ```

use an implementation-comment rather than a javadoc comment `/*`.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 197 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLQv9L-7mm7_mgfFeo:-NdLQv9L-7mm7_mgfFep:b-owoui7) ([raw file](https://github.com/batfish/batfish/blob/2a1e2c8f7f29dfcae653dc39ef2fe38f7d0b9c47/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L197)):*
> ```Java
>   }
> 
>   // Checks whether all variables in the given variable assignment are negated.
> ```

and maybe a javadoc here if you wish that.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 63 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLTC3M8HV8-QDFknFh:-NdLTC3M8HV8-QDFknFi:b-6v17jh) ([raw file](https://github.com/batfish/batfish/blob/2a1e2c8f7f29dfcae653dc39ef2fe38f7d0b9c47/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L63)):*
> ```Java
>    */
>   static boolean mustBeTrueInModel(BDD b, BDD model) {
>     return !model.diffSat(b);
> ```

Did you switch to `diffSat` because the model is partial?

nickgian (commented):
Mostly curious, any metrics on the impact of these optimizations?

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789#-:-NdLTsdjDkEDO8blbT40:bmcchg7)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

millstein (commented):
The main issue that occurs is that if there are a lot of community atomic predicates then the BDD library can have a stack overflow when doing things like ORing them all together, and other operations like that.  By eliminating logical formulas that talk about all atomic predicates, we have made these kinds of errors much less likely -- we now only OR together APs that are actually relevant for a particular snippet of config.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789#-:-NdahNaeCXv2nqWda5rN:bvwynaz)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java` line 146 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLLgAO6l_W1HX0N5KE:-NdaieTH00D6bs-bM9IF:bg8vrsf) ([raw file](https://github.com/batfish/batfish/blob/e860efeee73d476d9562457dccd8214fe156f7d7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java#L146)):*

why is the old comment no longer valid?
related, is there a semantic meaning to all atomic predicates being false? (no communities attached?)
</blockquote></details>

Constraints on a single community can now be arbitrary formulas.  The method CommunitySetMatchExprToBDD::toCommunitySetConstraint then does the work of converting this formula into a formula on community sets.  A constraint on a single community is providing constraints on what that community should look like -- hence, a constraint that requires all atomic predicates to be false is simply not satisfiable, as there is no community that satisfies the constraint.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 63 at r1](https://reviewable.io/reviews/batfish/batfish/8789#-NdLTC3M8HV8-QDFknFh:-NdamY2j8iJsUz4Q4AQz:b-1xemap) ([raw file](https://github.com/batfish/batfish/blob/2a1e2c8f7f29dfcae653dc39ef2fe38f7d0b9c47/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L63)):*

Did you switch to `diffSat` because the model is partial?
</blockquote></details>

Yes, exactly.  This checks whether something must be true in the model, not just whether it is allowed to be true.  In a full model there is no difference between the two, but now it matters.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789#-:-Ndav8hSDD6kJUPp3tmm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

anothermattbrown (commented):
> Reviewed 3 of 10 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-Ndd5LWCDf3QHpMfoX5x:b4h68u9) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*
> ```Java
>    * @return a community set constraint
>    */
>   public static BDD toCommunitySetConstraint(BDD commConstraint, Arg arg) {
> ```

I'm struggling to understand this. 

Since we're converting community constraints to community-set constraints, we are allowing multiple APs to be true simultaneously, right? 

So a positive AP constraint "ap1" means "the set contains some community in ap1", and a negative AP constraint "!ap2" means "the set does not contain any community in ap2".

If this is right, it's not safe to throw away negative constraints until the very end, i.e. picking a concrete route. Otherwise, we're losing information. 

For example, suppose we add a constraint `(ap1 /\ !ap2)` and then `(ap2 /\ !ap1)` later. If we discard the negative constraints, we end up with `ap1 /\ ap2`, which is wrong.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-NdeknZlCSoEsApFLANh:b-71c8cv) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*

I'm struggling to understand this. 

Since we're converting community constraints to community-set constraints, we are allowing multiple APs to be true simultaneously, right? 

So a positive AP constraint "ap1" means "the set contains some community in ap1", and a negative AP constraint "!ap2" means "the set does not contain any community in ap2".

If this is right, it's not safe to throw away negative constraints until the very end, i.e. picking a concrete route. Otherwise, we're losing information. 

For example, suppose we add a constraint `(ap1 /\ !ap2)` and then `(ap2 /\ !ap1)` later. If we discard the negative constraints, we end up with `ap1 /\ ap2`, which is wrong.
</blockquote></details>

No, that's not quite right.  A community constraint is a constraint about a single community.  So the constraint `(ap1 /\ !ap2)` is only constraining a single community -- we require there to be a community that satisfies this constraint.  And similarly for `(ap2 /\ !ap1)`.  So in fact, if C1 satisfies ap1 and C2 satisfies ap2, then the set {C1, C2} does satisfy both of the single-community constraints -- there exists a community satisfying each of them.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-NdephAk0jWwp-UEz35X:b-sy7326) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*

No, that's not quite right.  A community constraint is a constraint about a single community.  So the constraint `(ap1 /\ !ap2)` is only constraining a single community -- we require there to be a community that satisfies this constraint.  And similarly for `(ap2 /\ !ap1)`.  So in fact, if C1 satisfies ap1 and C2 satisfies ap2, then the set {C1, C2} does satisfy both of the single-community constraints -- there exists a community satisfying each of them.
</blockquote></details>

This is the difference between `HasCommunity(CommunityNot(FOO))` and `CommunitySetNot(HasCommunity(FOO))`.  The former requires just that we have a community in our set that does not match FOO, and that's what we're talking about here, while the latter requires that there not exist any community in the set satisfying FOO.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-Ndi5BXBCkjd6UBzdZen:b-pc7le7) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*

This is the difference between `HasCommunity(CommunityNot(FOO))` and `CommunitySetNot(HasCommunity(FOO))`.  The former requires just that we have a community in our set that does not match FOO, and that's what we're talking about here, while the latter requires that there not exist any community in the set satisfying FOO.
</blockquote></details>

Is the input `commConstraint` over the AP variables? If so, do you allow `ap1 /\ ap2`? What does that mean? Is it an invalid community constraint but a valid community set constraint?

Why can't you just do `commConstraint.andSat(originalAPs[i])`? I think I'll understand all of this better once I understand that.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-Ndk38269CoM3uY0wvla:bgjnt2a) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*

Is the input `commConstraint` over the AP variables? If so, do you allow `ap1 /\ ap2`? What does that mean? Is it an invalid community constraint but a valid community set constraint?

Why can't you just do `commConstraint.andSat(originalAPs[i])`? I think I'll understand all of this better once I understand that.
</blockquote></details>

Yes, the input is over the AP variables.  Any formula is allowed.  The formula `ap1 /\ ap2` should be considered unsatisfiable as a single-community constraint, since no single community can satisfy both.  That's exactly why you can't just do something like what you suggest -- `ap1 /\ ap2` is satisfiable as a boolean formula, but we need to take into account the fact that APs are disjoint.  That's what this code is doing -- making sure that we find only the APs that are consistent with this formula and satisfy it on their own.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 164 at r2](https://reviewable.io/reviews/batfish/batfish/8789#-Ndd5LWCDf3QHpMfoX5w:-Ndk3uiE10T6NfAw153I:b4i8qou) ([raw file](https://github.com/batfish/batfish/blob/3f7e0845a4816bb7acab57cf447d4d863f0c2c4d/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L164)):*

Yes, the input is over the AP variables.  Any formula is allowed.  The formula `ap1 /\ ap2` should be considered unsatisfiable as a single-community constraint, since no single community can satisfy both.  That's exactly why you can't just do something like what you suggest -- `ap1 /\ ap2` is satisfiable as a boolean formula, but we need to take into account the fact that APs are disjoint.  That's what this code is doing -- making sure that we find only the APs that are consistent with this formula and satisfy it on their own.
</blockquote></details>

Specifically, if the formula is `ap1 /\ ap2` then this code will properly determine that there are no APs that satisfy the formula.

anothermattbrown (commented):
> Reviewed 3 of 10 files at r1, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8789)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8790, opened 2023-08-29
TransferBDD: Support AsSetsMatchingRanges

millstein:
Updates the symbolic route analysis to support the AsSetsMatchingRanges class, which is used to represent certain common kinds of as-path regexes.  Also introduces an optimized approach to producing atomic predicates for disjunctions of as-path matches, which occur as the result of Juniper (and maybe other) as-path groups.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 96 at r2](https://reviewable.io/reviews/batfish/batfish/8790#-Nd2Zs655nTobnG_daRx:-Nd2Zs655nTobnG_daRy:b-806t1o) ([raw file](https://github.com/batfish/batfish/blob/c009fbe5d707f7ae14c041bd92eed26157bf95c8/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L96)):*
> ```Java
>         "Unexpected unbounded ASN range in an AS-path expression");
>     ContiguousSet<Long> cs = ContiguousSet.create(r, DiscreteDomain.longs());
>     String setAsString = cs.stream().map(l -> "(" + l + ")").collect(Collectors.joining("|"));
> ```

FYI this implementation is what caused us to switch away from this implementation: it runs for multiple minutes.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 96 at r2](https://reviewable.io/reviews/batfish/batfish/8790#-Nd2Zs655nTobnG_daRx:-NdCmISSAQl4197ezkXs:bi4ok9c) ([raw file](https://github.com/batfish/batfish/blob/c009fbe5d707f7ae14c041bd92eed26157bf95c8/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L96)):*

FYI this implementation is what caused us to switch away from this implementation: it runs for multiple minutes.
</blockquote></details>

The newest commit optimizes this, leveraging existing support for reasoning about numeric ranges symbolically.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: 0 of 12 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 61 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGyklk2Qsf5OvS4DmQ:-NdGyklk2Qsf5OvS4DmR:b-qk3rmj) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L61)):*
> ```Java
>     */
>     if (disjunction.getDisjuncts().stream().allMatch(d -> d instanceof MatchAsPath)) {
>       List<SymbolicAsPathRegex> disjuncts_list = new ArrayList<>(disjuncts);
> ```

What value is the list adding? Union takes a collection, set is one...
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 64 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGyaGsAUH3K1Ok-QUC:-NdGyaGsAUH3K1Ok-QUD:b-a7nga2) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L64)):*
> ```Java
>       if (disjuncts_list.isEmpty()) {
>         return ImmutableSet.of();
>       }
> ```

We can check this up front, right? Why wait for the all-match.

dhalperi (commented):
> Reviewed 1 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: 1 of 12 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java` line 93 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGzdeo-FaPRDDlH_uY:-NdGzdeo-FaPRDDlH_uZ:bnw54ol) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java#L93)):*
> ```Java
>         equalTo(
>             "((<0-2147483647>)|(2<147483648-999999999>)|(3<000000000-999999999>)|(4<000000000-294967295>))"));
>   }
> ```

The string tests look reasonable, but are only correct when the library works a specific way.

In a trust but verify sense, can we add tests of the outer API? I'm sure you can pick a good "Stressful" relevant range (or a few) but a bad stab is something like

`2134567890-4223467890L`

And then we should have an array of tests making sure that the right boundary values are accepted or not... something like

```
long[] rejected = new long[] {1, 2, 2134567800, 2134567889, 42234567891}; // maybe more stuff baed on the way we compress range regexes
long[] accepted = new long[] {2134567890, 2555557890, 421...};
for (long l : new long[] {1, 40, 49, 419999999L, ...}) {
    assertThat(an As Path of only this is accepted by the library)
```


I'd also do something like Range(50,155) and then just actually check 1-255. to be correct.

nickgian (commented):
> Reviewed 6 of 11 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: all files reviewed, 11 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 111 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdQy0iS8BOUwaPmzQUm:-NdQy0iT9D_zGcfE_Du4:b-ujdx2c) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L111)):*
> ```Java
>       return toRegexRange(String.valueOf(lower), String.valueOf(upper));
>     } else {
>       // handle the case when the given interval is outside the range of Java integers
> ```

I am kind of confused about the special treatment of `Integer.MAX_VALUE`. Why do we need to handle intervals outside the range of java integers in a different way? 
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java` line 64 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGyaGsAUH3K1Ok-QUC:-NdR5K_T7fkGPQ-ZmsET:b-mb9kt1) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollector.java#L64)):*

We can check this up front, right? Why wait for the all-match.
</blockquote></details>

+1, similar if condition to the one used in TransferBDD should work.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 294 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdR1YhU1T_ty77984ZU:-NdR1YhV5LcMvzlfopDV:b-btg181) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L294)):*
> ```Java
>       predicates, for performance reasons. See BooleanExprAsPathCollector::visitDisjunction. We
>       have to check for that special case here in order to access the correct atomic predicates.
>       */
> ```

This computation will be repeated on each path it appears on, right?
Depending on how expensive it is we can consider caching it so that we won’t reconstruct the regexp and the BDD.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java` line 87 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdR4-LUDfkb9lhIPIfW:-NdR4-LY4xgfdgfpz6ta:b-sfweeo) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java#L87)):*
> ```Java
> 
>     assertThat(r1, equalTo("<0-2147483647>"));
>     assertThat(r2, equalTo("((<500-2147483647>)|(2<147483648-147483667>))"));
> ```

Similar question, why is `<500-2147483667>` is not valid?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java` line 118 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdR4g4i6X0lP5SoA7fo:-NdR4g4i6X0lP5SoA7fp:byd9kq5) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java#L118)):*
> ```Java
>         new Disjunction(
>             ImmutableList.of(
>                 new Disjunction(
> ```

Ideally we should still be able to collect all the MatchAsPath expressions, but I am guessing it doesn’t really matter if we don’t construct Disjunction between fields of different types (does that ever happen in practice?).

dhalperi (commented):
> Reviewed 2 of 11 files at r1, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: all files reviewed, 13 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java` line 93 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGzdeo-FaPRDDlH_uY:-Nda1cRs7KKIMkl7nGfl:beq6khk) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java#L93)):*

The string tests look reasonable, but are only correct when the library works a specific way.

In a trust but verify sense, can we add tests of the outer API? I'm sure you can pick a good "Stressful" relevant range (or a few) but a bad stab is something like

`2134567890-4223467890L`

And then we should have an array of tests making sure that the right boundary values are accepted or not... something like

```
long[] rejected = new long[] {1, 2, 2134567800, 2134567889, 42234567891}; // maybe more stuff baed on the way we compress range regexes
long[] accepted = new long[] {2134567890, 2555557890, 421...};
for (long l : new long[] {1, 40, 49, 419999999L, ...}) {
    assertThat(an As Path of only this is accepted by the library)
```


I'd also do something like Range(50,155) and then just actually check 1-255. to be correct.
</blockquote></details>

The above suggestion was incomplete: I meant to abstract out the upper bound and then we could test it with an upper bound of say 100 (rather than Integer.MAX_VALUE) and the above command.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java` line 113 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdH0kpqBzh36widlZVs:-NdH0kpqBzh36widlZVt:bxrrakx) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java#L113)):*
> ```Java
> 
>   @Test
>   public void testVisitDisjunction2() {
> ```

`VisitDisjunction<OfX>` instead of 2?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 865 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdH0w9y0Xs8t-BX5KDq:-NdH0w9y0Xs8t-BX5KDr:bothusn) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L865)):*
> ```Java
> 
>   @Test
>   public void testAsSetsMatchingRanges2() {
> ```

similarly, what's different about this test?

also, is there anything worth factoring out to get simpler tests?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: 8 of 12 files reviewed, 5 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 99 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGcrK58TumgRcRvvXQ:-NdaoR3z2jJiGjDD-tO4:b-aw84i2) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L99)):*

prefer to use the normalization logic to make this braindead.

`DiscreteDomain.longs().minValue(r)` / minValue should do the right thing. (and DD.longs() is a singleton so no need to worry about caching etc.) https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/DiscreteDomain.html#maxValue()
</blockquote></details>

I don't see a minValue(r) method, but I did it a different way that I think satisfies your point.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 111 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdQy0iS8BOUwaPmzQUm:-NdaqLpC2M81WwVGmu3P:b-n18dyf) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L111)):*

I am kind of confused about the special treatment of `Integer.MAX_VALUE`. Why do we need to handle intervals outside the range of java integers in a different way?
</blockquote></details>

The automaton library supports a syntax `<lower-upper>` for integer ranges, but those numbers must be Java ints.  So we have to treat numbers outside of that range specially.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java` line 87 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdR4-LUDfkb9lhIPIfW:-NdatF2ME4RoiDuCm7Si:b-mczl91) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java#L87)):*

Similar question, why is `<500-2147483667>` is not valid?
</blockquote></details>

The Automaton library signals a syntax error if you try to use a number outside the Java int range.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java` line 93 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdGzdeo-FaPRDDlH_uY:-Ndb6LBAEf4QP1uz2mms:b62q560) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/SymbolicAsPathRegexTest.java#L93)):*

The above suggestion was incomplete: I meant to abstract out the upper bound and then we could test it with an upper bound of say 100 (rather than Integer.MAX_VALUE) and the above command.
</blockquote></details>

Good idea; added.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java` line 118 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdR4g4i6X0lP5SoA7fo:-NdaueKO3hgoWAwhRNtC:b66pdub) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/aspath/BooleanExprAsPathCollectorTest.java#L118)):*

Ideally we should still be able to collect all the MatchAsPath expressions, but I am guessing it doesn’t really matter if we don’t construct Disjunction between fields of different types (does that ever happen in practice?).
</blockquote></details>

True, we could still combine the as-path regexes into one, even if some disjuncts are not as-path regexes.  As you said, unclear if we need that and it will make things a bit more complex.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 865 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdH0w9y0Xs8t-BX5KDq:-NdauxbT0mOAHK8xWING:b-frod4a) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L865)):*

similarly, what's different about this test?

also, is there anything worth factoring out to get simpler tests?
</blockquote></details>

this test ensures that we do in fact obtain concrete routes with AS-paths containing ASNs beyond the range of Java ints.  the tests you suggested above will do a more thorough job of testing that.

dhalperi (commented):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 865 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdH0w9y0Xs8t-BX5KDq:-NdbAxe5A2iOjzlIJsca:bvwygaj) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L865)):*

this test ensures that we do in fact obtain concrete routes with AS-paths containing ASNs beyond the range of Java ints.  the tests you suggested above will do a more thorough job of testing that.
</blockquote></details>

Good to have an e2e test of it if there's missing coverage elsewhere.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8790#-:-NdvQeU_9FqV6Zelp6et:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 111 at r3](https://reviewable.io/reviews/batfish/batfish/8790#-NdQy0iS8BOUwaPmzQUm:-NdvQO8--RipdvAWRLDR:b-ozr4xo) ([raw file](https://github.com/batfish/batfish/blob/b5644a870eca72c97fa54cb4c4b83050640f3adb/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L111)):*

The automaton library supports a syntax `<lower-upper>` for integer ranges, but those numbers must be Java ints.  So we have to treat numbers outside of that range specially.
</blockquote></details>

got it, that's a pain...

PR #8797, opened 2023-09-16
TransferBDD: Precise Number Regexes

millstein:
This PR updates the symbolic route analysis to properly account for the ranges of numbers that can appear in communities and in AS-paths.  Otherwise the analysis can provide spurious results, for example a community that is ill-formed because one half is a number that is out of range.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
@@ -43,6 +43,11 @@ public final class CommunityVar extends SymbolicRegex implements Comparable<Comm
   @Nonnull private final Type _type;
   @Nullable private final Community _literalValue;
 
+  /**
+   * A regex representing numbers that can be part of communities. The first conjunct of the regex
+   * ensures there are no leading zeros. The second conjunct of the regex ensures that the number is
+   * at most 16 bits, using the numeric interval syntax of the automaton library.
+   */
   @Nonnull private static final String NUM_REGEX = "((0|[1-9][0-9]*)&<0-65535>)";
 
   // a regex that represents the syntax of standard community literals supported by Batfish
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
@@ -43,7 +43,7 @@ public final class CommunityVar extends SymbolicRegex implements Comparable<Comm
   @Nonnull private final Type _type;
   @Nullable private final Community _literalValue;
 
-  @Nonnull private static final String NUM_REGEX = "(0|[1-9][0-9]*)";
+  @Nonnull private static final String NUM_REGEX = "((0|[1-9][0-9]*)&<0-65535>)";
 
   // a regex that represents the syntax of standard community literals supported by Batfish
   // see StandardCommunity::matchString()
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java
@@ -3,6 +3,7 @@ package org.batfish.minesweeper;
 import static org.batfish.minesweeper.CommunityVar.Type.EXACT;
 import static org.batfish.minesweeper.CommunityVar.Type.REGEX;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import dk.brics.automaton.Automaton;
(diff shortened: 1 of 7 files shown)


nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8797)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java` line 51 at r1](https://reviewable.io/reviews/batfish/batfish/8797#-NeWFXV1Baaz2AjRMELR:-NeWFXV1Baaz2AjRMELS:bce5bn5) ([raw file](https://github.com/batfish/batfish/blob/999d3f3fafa61fca72e708b3f4b837b822fe02a8/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java#L51)):*
> ```Java
>    * at most 16 bits, using the numeric interval syntax of the automaton library.
>    */
>   @Nonnull private static final String NUM_REGEX = "((0|[1-9][0-9]*)&<0-65535>)";
> ```

why do we need the first conjunct? Isn’t the interval implying the first conjunct?

From the library docs I see:
 >Numerical intervals are specified by non-negative decimal integers and include both end points, and if `_n_` and `_m_` have the same number of digits, then the conforming strings must have that length (i.e. prefixed by 0's).

But I don’t quite grasp what they mean.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java` line 35 at r1](https://reviewable.io/reviews/batfish/batfish/8797#-NeWR_rVELqeV6x8Ar1i:-NeWR_rW34l_iSa89emo:bu33n8w) ([raw file](https://github.com/batfish/batfish/blob/999d3f3fafa61fca72e708b3f4b837b822fe02a8/projects/minesweeper/src/main/java/org/batfish/minesweeper/SymbolicAsPathRegex.java#L35)):*
> ```Java
>   @VisibleForTesting @Nonnull
>   static final String AS_NUM_REGEX =
>       "((0|[1-9][0-9]*)&(<0-2147483647>|2<147483648-999999999>|3<000000000-999999999>|4<000000000-294967295>))";
> ```

I recall you doing a similar trick for 32-bit ints in another PR/file. Might be worth seeing if you can factor out this as INT_32_REGEX or something, although I don’t recall how exactly you used it in that other file.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/CommunityVarTest.java` line 43 at r1](https://reviewable.io/reviews/batfish/batfish/8797#-NeWW4X-70StyExGUGW8:-NeWW4X-70StyExGUGW9:b-44fyfd) ([raw file](https://github.com/batfish/batfish/blob/999d3f3fafa61fca72e708b3f4b837b822fe02a8/projects/minesweeper/src/test/java/org/batfish/minesweeper/CommunityVarTest.java#L43)):*
> ```Java
>     assertThat(cv3.toAutomaton(), equalTo(CommunityVar.COMMUNITY_FSM));
>     assertThat(
>         cv4.toAutomaton(), equalTo(new RegExp("^40:((0|[1-9][0-9]*)&<0-65535>)$").toAutomaton()));
> ```

nit: use `NUM_REGEX`? you do it for the as-path case.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8797)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java` line 51 at r1](https://reviewable.io/reviews/batfish/batfish/8797#-NeWFXV1Baaz2AjRMELR:-NeYg4Z70ph8tCtu8vzO:b-646amh) ([raw file](https://github.com/batfish/batfish/blob/999d3f3fafa61fca72e708b3f4b837b822fe02a8/projects/minesweeper/src/main/java/org/batfish/minesweeper/CommunityVar.java#L51)):*

why do we need the first conjunct? Isn’t the interval implying the first conjunct?

From the library docs I see:
 >Numerical intervals are specified by non-negative decimal integers and include both end points, and if `_n_` and `_m_` have the same number of digits, then the conforming strings must have that length (i.e. prefixed by 0's).

But I don’t quite grasp what they mean.
</blockquote></details>

It turns out that if I just do the second conjunct then leading zeros are allowed, so you can get numbers like `0012`.  The first conjunct could be slightly simplified but I still need something like it.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8797#-:-NeYx8Q_3Uo93-VMDgQ3:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8820, opened 2023-09-28
Validation for SearchRoutePolicies and CompareRoutePolicies

millstein:
Adds validation for answers from SearchRoutePolicies and CompareRoutePolicies, by comparing them against the results of doing a concrete route simulation with Batfish.  Also fixes a few small bugs found through this validation.


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/common/bdd/MutableBDDIntegerTest.java` line 87 at r2](https://reviewable.io/reviews/batfish/batfish/8820#-Nfky7LyCHWbTEiCPYIb:-NfnKOdJEKfSgtAnk6yl:bvfxxc1) ([raw file](https://github.com/batfish/batfish/blob/1a3d5bd24b5d495b2797a2b1f93ca5489bf37ebe/projects/batfish-common-protocol/src/test/java/org/batfish/common/bdd/MutableBDDIntegerTest.java#L87)):*

can we add 2-3 more cases (maybe testing overflow, near overflow, what happens you add `xPlusx`). Just one test does not provide very strong coverage (I am surprised there was a bug we missed from other tests tbh).
</blockquote></details>

Done. The reason the bug was never found is that we always passed this code a full model. Once you start passing in partial models it's actually not a very subtle bug -- almost anything interesting is broken.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 80 at r2](https://reviewable.io/reviews/batfish/batfish/8820#-Nfkz_ezExPU1mXyB0ld:-NfnK_C73rQbm9vSxUgb:b-m2297k) ([raw file](https://github.com/batfish/batfish/blob/1a3d5bd24b5d495b2797a2b1f93ca5489bf37ebe/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L80)):*

nice! Should we do the same for fields like `_nextHopInterfaces` and `sourceVrfs` too?
</blockquote></details>

We could but it's probably not a big win unless there are lots of them.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 153 at r2](https://reviewable.io/reviews/batfish/batfish/8820#-Nfl-qBV5aBgn1Y4X9uz:-NfnKlIw7H2UuUw2hiUk:b-sdyfb1) ([raw file](https://github.com/batfish/batfish/blob/c9a6f33e8e27456fd64e57c559834f86e9569313/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L153)):*

is this intersection redundant because it is performed in `initAtomicPredicates`?
</blockquote></details>

Now that there is exactly one as-path AP that is true, we can just directly get its automaton.  Before we were looping over all as-path APs to see which were true and then intersecting them all.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 228 at r2](https://reviewable.io/reviews/batfish/batfish/8820#-Nfl0c2q3CggD6AWM03X:-NfnL0IL06bUa3SKHMeU:b-mcpu7u) ([raw file](https://github.com/batfish/batfish/blob/1a3d5bd24b5d495b2797a2b1f93ca5489bf37ebe/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L228)):*

Is this called on test-code only or production code too? I am wary for two reasons, performance overhead, and assertion failures when it might not be a critical error. NBD for now, something to keep in mind if any of these become an issue.
</blockquote></details>

It's only called by code that is in an `assert` so it will only run when assertions are enabled (i.e., during testing rather than production).

nickgian (commented):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

anothermattbrown (commented):
> Reviewed 6 of 9 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java` line 99 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvUahOBPajrQeRcSIi:-NfvUahOBPajrQeRcSIj:b-ico54g) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java#L99)):*
> ```Java
>     // are part of the support of this MutableBDDInteger. this is necessary to properly get models
>     // in the face of mutation, where this object represents a function of the original BDD
>     // variables.
> ```

I need help understanding this. Why can't we set all variables to false unless they are assigned true in the assignment? i.e. why can't we use `fullSatOne`?



_Code quote:_
```Java
    // explicitly treat as false any variables that do not appear in the given SAT assignment but
    // are part of the support of this MutableBDDInteger. this is necessary to properly get models
    // in the face of mutation, where this object represents a function of the original BDD
    // variables.
```
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NfvYMsvDdjxyGcOjT6o:b619n4e) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*
> ```Java
>     BDD prefLenConstraint = _prefixLength.leq(32);
>     // at most one source VRF should be in the environment
>     BDD sourceVrfConstraint = atMostOneOf(_sourceVrfs);
> ```

can we also use BDDDomain for sourceVrfs and nextHopInterfaces?



_Code quote:_
```Java
    // at most one source VRF should be in the environment
    BDD sourceVrfConstraint = atMostOneOf(_sourceVrfs);
```

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java` line 99 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvUahOBPajrQeRcSIi:-NgQGKzY79V1-bcxDyyd:b-njaltl) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java#L99)):*

I need help understanding this. Why can't we set all variables to false unless they are assigned true in the assignment? i.e. why can't we use `fullSatOne`?
</blockquote></details>

We could use `fullSatOne`.  This is an optimization that has the same effect but only sets to false the variables that are mentioned by this MutableBDDInteger, to avoid the need to create a potentially large BDD that contains a valuation for every single variable.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NgQG_yxBWQ1SJwK8GHF:bd3iu8f) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*

can we also use BDDDomain for sourceVrfs and nextHopInterfaces?
</blockquote></details>

Yes we could do that.  Not a big win unless there are lots of these things, so I'm not sure how much it would help in practice.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NgQHJWP1t8vorzkMjib:bsm7cjg) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*

Yes we could do that.  Not a big win unless there are lots of these things, so I'm not sure how much it would help in practice.
</blockquote></details>

IMO the value is that you capture the semantics of these fields through their type rather than introducing relying on the general formulation and adding extra constraints. But I don't think this needs to be part of this PR.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NgQJjwS1BrN76sT10Zz:bijt8js) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*

IMO the value is that you capture the semantics of these fields through their type rather than introducing relying on the general formulation and adding extra constraints. But I don't think this needs to be part of this PR.
</blockquote></details>

Good point. Ok, I'm sold.  Probably best as a separate PR.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NgQLF2h97ZPUFVot_-O:b-ida0mi) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*

Good point. Ok, I'm sold.  Probably best as a separate PR.
</blockquote></details>

Though actually there is a difference which is that for AS-paths we always need to get a value, while these things are optional -- we want at most one rather than exactly one.  I guess that can be handled by adding an extra dummy value to mean "none" and make that the 0 value so that it's chosen by default.  Is there a better way to do this?

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java` line 99 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvUahOBPajrQeRcSIi:-NgtMrAiBt6bVs-Pu3_N:bvh9181) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java#L99)):*

We could use `fullSatOne`.  This is an optimization that has the same effect but only sets to false the variables that are mentioned by this MutableBDDInteger, to avoid the need to create a potentially large BDD that contains a valuation for every single variable.
</blockquote></details>

I'm confused by this comment then. What is "this is necessary" referring to?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 443 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvYMsvDdjxyGcOjT6n:-NgtNK2L6bhaFv5Zci_s:b-54mnj9) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L443)):*

Though actually there is a difference which is that for AS-paths we always need to get a value, while these things are optional -- we want at most one rather than exactly one.  I guess that can be handled by adding an extra dummy value to mean "none" and make that the 0 value so that it's chosen by default.  Is there a better way to do this?
</blockquote></details>

I think having an explicit value for none makes sense

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java` line 99 at r3](https://reviewable.io/reviews/batfish/batfish/8820#-NfvUahOBPajrQeRcSIi:-NgtQtG90hMiivbIBUXj:b9neshp) ([raw file](https://github.com/batfish/batfish/blob/e9d6830a13cb4cb6db11ec6ae8d241f15ca42581/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java#L99)):*

I'm confused by this comment then. What is "this is necessary" referring to?
</blockquote></details>

We're given a partial assignment.  We have to "complete" it in some way, or else this code does not work properly.  So some kind of completion is necessary, which is what the comment is trying to convey.  You could do it with fullSatOne but this is an optimized way to do that instead. I updated the comment to try to make this clearer.

anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8820)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8827, opened 2023-10-03
Fix 'forwarding-options dhcp-security' parse warning for Juniper

saidvandeklundert:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8827)* status: all files reviewed, 2 unresolved discussions

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4` line 164 at r1](https://reviewable.io/reviews/batfish/batfish/8827#-Nfqdh5E8adlLXlR8Hdg:-Nfqdh5E8adlLXlR8Hdh:b-l4dtg3) ([raw file](https://github.com/batfish/batfish/blob/fd008e73c8f2f83d76b119cb535d880c1e93d140/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4#L164)):*
I'm assuming this config is valid at top-level as well as within a VLAN, which is why you added it in both places? If so:

1/ We have an existing `s_forwarding_options`: add this as an alternative there instead. I think it would go in `fo_null`. This keeps the parser LL(1): 

> the parser may look ahead an arbitrary number of tokens to make this decision, and backtrack if needed. Backtracking is expensive and can sometimes cause issues with Batfish's parser error recovery infrastructure. So to minimize the potential for backtracking, we try to write all parser rules in such a way that a decision among alternatives can be made by looking ahead only a single token (LL( 1)). That is, when you recurse down into all referenced rules until you hit the first token reference, all such token references should be distinct.

Here, the top level of config is already used, so adding it into `s_null` will add a second place to look once the `set forwarding-options` shows up.

2/ please add this top level variant to your config test.

(If not true, just revert this change?)
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4` line 203 at r1](https://reviewable.io/reviews/batfish/batfish/8827#-Nfqfibp0O6ZBg_KFh6y:-Nfqfibp0O6ZBg_KFh6z:bckns15) ([raw file](https://github.com/batfish/batfish/blob/fd008e73c8f2f83d76b119cb535d880c1e93d140/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperParser.g4#L203)):*
> ```ANTLR
>     | vlt_vlan_id_list
>     | vlt_vni_id
>     | sl_vlans_null
> ```

I'd put this as `vlt_forwarding_options`

We have strong customers around prefix: the prefix (everything til first `_`) should indicate exactly where a thing is valid.

In this case, I had to go into git history to understand why it's `vlt`, but the answer seems to be that it came from "vlan tail", aka, things valid after the `vlan` command: https://github.com/batfish/batfish/pull/3014

So:

```
| vlt_filter
| vlt_forwarding_options  # in alphabetical order unless we can't for some reason, for readability
| ...
```

Then

```
vlt_filter: ...

# in alphabetiacl order with vlts, for readability
vlt_forwarding_options: FORWARDING_OPTIONS vltfo_null;
vltfo_null: DHCP_SECURITY

...
```

where `vltfo` is the natural prefix combining the existing `vlt` with the new `forwarding-options` :)

PR #8829, opened 2023-10-05
AbstractRoute: raise AD to maximum 32 bit integer

dhalperi:
Found a real network where AD is bigger than 255. It was an accident, but it's out there. We should
handle this more gracefully than crashing.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8829)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute.java` line 51 at r1](https://reviewable.io/reviews/batfish/batfish/8829#-NfxTY-J14g93EJMu7Md:-NfxTY-J14g93EJMu7Me:bjdokqf) ([raw file](https://github.com/batfish/batfish/blob/5095c8ed4007d13880c5f9a20b4f0d5dbabc2432/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/AbstractRoute.java#L51)):*
> ```Java
>   /** Helper to check admin distance validity. */
>   @SuppressWarnings("ComparisonOutOfRange")
>   static void checkAdmin(int admin, int lower) {
> ```

This is a little hoopla to work around the fact that with `(any int) <= Integer.MAX_VALUE` is true and we were getting a compiler error.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 203 at r1](https://reviewable.io/reviews/batfish/batfish/8829#-NfxTjrL39TLmLElgBBF:-NfxTjrL39TLmLElgBBG:b-mlp30z) ([raw file](https://github.com/batfish/batfish/blob/5095c8ed4007d13880c5f9a20b4f0d5dbabc2432/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L203)):*
> ```Java
>     int bitsToRepresentAdmin = IntMath.log2(AbstractRoute.MAX_ADMIN_DISTANCE, RoundingMode.CEILING);
>     // or else we need to do tricks in the BDDInteger.
>     assert LongMath.isPowerOfTwo(1L + AbstractRoute.MAX_ADMIN_DISTANCE);
> ```

@nickgian @millstein -- thoughts on making this code zero-maintenance (as long as max AD is an Integer)?

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8829)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian and @SLarkworthy)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8829)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian and @SLarkworthy)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 203 at r1](https://reviewable.io/reviews/batfish/batfish/8829#-NfxTjrL39TLmLElgBBF:-Ng-52HZAUJXg0jtfCVE:b-g6ztmw) ([raw file](https://github.com/batfish/batfish/blob/5095c8ed4007d13880c5f9a20b4f0d5dbabc2432/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L203)):*

@nickgian @millstein -- thoughts on making this code zero-maintenance (as long as max AD is an Integer)?
</blockquote></details>

One thing that could help is that `numNeeded` need not be exact -- it can be an over-approximation because it's just determining the number of variables that the factory needs.  So we can do something simple and overapproximate here (which then maybe works for any max distance and avoids the assert).  But then below when we actually allocate the `MutableBDDInteger` we'd need to have another constructor that can take a max value and internally compute the number of bits required, similar to what `BDDFiniteDomain` does.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8829)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickgian and @SLarkworthy)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 203 at r1](https://reviewable.io/reviews/batfish/batfish/8829#-NfxTjrL39TLmLElgBBF:-Ng-Mxkb0lQt2S5WeLZh:b-5b2f5h) ([raw file](https://github.com/batfish/batfish/blob/5095c8ed4007d13880c5f9a20b4f0d5dbabc2432/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L203)):*

One thing that could help is that `numNeeded` need not be exact -- it can be an over-approximation because it's just determining the number of variables that the factory needs.  So we can do something simple and overapproximate here (which then maybe works for any max distance and avoids the assert).  But then below when we actually allocate the `MutableBDDInteger` we'd need to have another constructor that can take a max value and internally compute the number of bits required, similar to what `BDDFiniteDomain` does.
</blockquote></details>

Yeah, I think I asked this question poorly. Suppose we need to change it to 2^m-1 where m <= 32, I think this will work.

Agree on the right solution if later for some reason we have to do this differently, but for now I *think* this is a correct check under the above constraints.

Of course if we change AD to a u32 we'll be changing types, and that's a bigger change anyway.

PR #8837, opened 2023-10-07
TransferBDD: Never crash due to an unsupported feature

millstein:
Updates the symbolic route analysis so that it never crashes due to an unsupported feature. Instead it ignores the feature, warns about it, and moves on. The analysis was already doing that for the most part, but there were a few places where exceptions were being thrown and not caught due to unsupported features.  This PR addresses those things.


nickgian (commented):
> Reviewed 6 of 8 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8837)* status: all files reviewed, 4 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java` line 58 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH7tnJDtqo8wen4_Uj:-NgH7tnJDtqo8wen4_Uk:b-fb3u3y) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java#L58)):*
> ```Java
>   public Set<SymbolicAsPathRegex> visitHasAsPathLength(
>       HasAsPathLength hasAsPathLength, CommunitySetMatchExprToBDD.Arg arg) {
>     throw new UnsupportedOperationException(hasAsPathLength.toString());
> ```

is there a dedicated construct Juniper or Cisco construct that will generate `HasAsPathLength`? Should we be supporting this since we support the regex syntax?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 130 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH8nq306HCT9LJEDoi:-NgH8nq306HCT9LJEDoj:by22bty) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L130)):*
> ```Java
>   @Override
>   public BDD visitHasSize(HasSize hasSize, Arg arg) {
>     throw new UnsupportedOperationException(hasSize.toString());
> ```

same comment as above (though, I am not necessarily in favor of making everything slower, just brainstorming)
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 547 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH9kBM9yz3x4A1Wvyz:-NgH9kBM9yz3x4A1Wvz-:bikm55x) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L547)):*
> ```Java
>           break;
>         default:
>           throw new IllegalArgumentException("Unexpected boolean expression " + b.getType());
> ```

why `IllegalArgumentException` instead of `UnsupportedOperation`?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 1017 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgHAoA4APFGi2EyM1xV:-NgHAoA4APFGi2EyM1xW:b-jsjexq) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1017)):*
> ```Java
>     RoutingPolicy policy =
>         _policyBuilder
>             .addStatement(new SetWeight(new VarInt("var")))
> ```

what router config constructs generate `VarInt`?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8837)* status: all files reviewed, 4 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java` line 58 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH7tnJDtqo8wen4_Uj:-NgK3m_r8xgiloVJeAj_:b6ozij2) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java#L58)):*

is there a dedicated construct Juniper or Cisco construct that will generate `HasAsPathLength`? Should we be supporting this since we support the regex syntax?
</blockquote></details>

It looks like Batfish uses this for Cisco XR.  I'll add it to the TODO list.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java` line 130 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH8nq306HCT9LJEDoi:-NgK449S33e4hRtxjV_4:bfs6yuv) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunitySetMatchExprToBDD.java#L130)):*

same comment as above (though, I am not necessarily in favor of making everything slower, just brainstorming)
</blockquote></details>

`HasSize` is about the number of communities -- we don't handle this currently.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 547 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH9kBM9yz3x4A1Wvyz:-NgK5AM08ncqlltQfMJI:b2409m3) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L547)):*

why `IllegalArgumentException` instead of `UnsupportedOperation`?
</blockquote></details>

This is not a situation where we are not handling a certain feature.  Rather, we are handling all elements of the StaticExpressionType enum.  So if there's a failure here it's because we later get an unexpected value, rather than because there's something we are not currently handling.  This is how other case statements on enums are handled elsewhere.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 1017 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgHAoA4APFGi2EyM1xV:-NgK5bs09c9pVusWo_d5:b-u6825o) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L1017)):*

what router config constructs generate `VarInt`?
</blockquote></details>

It seems to be used by Cisco XR.  Here I'm just using it as an example of an unsupported feature, to test that we do the right thing in that case.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8837)* status: all files reviewed, 4 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java` line 58 at r2](https://reviewable.io/reviews/batfish/batfish/8837#-NgH7tnJDtqo8wen4_Uj:-NgK8dyz2KKh1H6oOlQv:bb70g84) ([raw file](https://github.com/batfish/batfish/blob/9e9472942a63bb2ba919eb63f1f4018c36a50f40/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/AsPathMatchExprToRegexes.java#L58)):*

It looks like Batfish uses this for Cisco XR.  I'll add it to the TODO list.
</blockquote></details>

It's also used in Juniper.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8837#-:-Ng_nRo37f73kFNS5_XJ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8845, opened 2023-10-18
TransferBDD: Model "at-most-one" route attributes using BDDDomain

millstein:
Updates the symbolic route analysis to represent "at-most-one" route attributes, namely the next-hop interface name and the source VRF, using a BDDDomain.  This has the effect of requiring only a logarithmic number of BDD variables in the number of options, whereas before we were using a linear number of BDD variables. It also simplifies some of the code since the "at-most-one" constraint is now enforced implicitly through the representation.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDDomain.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDDomain.java
@@ -20,7 +20,7 @@ public class BDDDomain<T> {
   private MutableBDDInteger _integer;
 
   public BDDDomain(BDDFactory factory, List<T> values, int index) {
-    int bits = numBits(values);
+    int bits = numBits(values.size());
     _factory = factory;
     _values = values;
     _integer = MutableBDDInteger.makeFromIndex(_factory, bits, index, false);
@@ -42,8 +42,13 @@ public class BDDDomain<T> {
     _integer = other.getInteger().and(pred);
   }
 
-  private int numBits(List<T> values) {
-    int size = values.size();
+  /**
+   * Returns the number of bits used to represent a domain of the given size.
+   *
+   * @param size the number of elements in the domain
+   * @return the number of bits required
+   */
+  public static int numBits(int size) {
     double log = Math.log((double) size);
     double base = Math.log((double) 2);
     if (size == 0) {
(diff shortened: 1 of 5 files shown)


nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8845)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 596 at r1](https://reviewable.io/reviews/batfish/batfish/8845#-NhXxuVBBmmxgEZ1f9eQ:-NhXxuVBBmmxgEZ1f9eR:b-njltlw) ([raw file](https://github.com/batfish/batfish/blob/75dad27ca8c7ffc7eaaab9099dba1e59ec0e2921/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L596)):*
> ```Java
>     } else if (expr instanceof MatchSourceVrf) {
>       MatchSourceVrf msv = (MatchSourceVrf) expr;
>       // we add 1 to the index since 0 in the BDDDomain is used to represent the absence of a value
> ```

would things be easier (i.e., we wouldn't have to add 1 to the index) if none is represented by the last index rather than 0? Or is it that model generation would become more complicated?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8845)* status: all files reviewed, 1 unresolved discussion (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 596 at r1](https://reviewable.io/reviews/batfish/batfish/8845#-NhXxuVBBmmxgEZ1f9eQ:-NhYTLYL4AfUA57PjtP8:b-jg0szv) ([raw file](https://github.com/batfish/batfish/blob/75dad27ca8c7ffc7eaaab9099dba1e59ec0e2921/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L596)):*

would things be easier (i.e., we wouldn't have to add 1 to the index) if none is represented by the last index rather than 0? Or is it that model generation would become more complicated?
</blockquote></details>

Right, model generation is easier if none is 0, since we get the "least" satisfying model from things like `satOne`, so that will naturally give us none unless we need to have something.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8845#-:-NhmDpyb--MJOMxyFBN6:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #8850, opened 2023-10-26
Fix BGP route resolution tracker losing routes

arifogel:
* Fix bug where a dependent route is removed from the resolution tracker if it is determined not to be resolvable during updateActiveRoutes
* With fix, such a dependent route can now be (re-)activated when a suitable main RIB route is added in the delta supplied o a subsequent call to updateActiveRoutes


SLarkworthy (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8850#-:-Nhi-jsB7KJItergKWqb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8855, opened 2023-10-28
checkstyle: one more nullity pattern to catch

dhalperi:
The autoformatter will put @Nullable on its own line if it starts the line, but
we can require it to be moved if there's a modifier it will stay after.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8855)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8875, opened 2023-12-11
fix batfish crash due to junOS interface filter direction and name missing in config command 

pranavbj-amzn:
Update exit functions to handle missing filter direction and name. These are optional in the junOS grammer.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8875)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pranavbj-amzn)

PR #8884, opened 2023-12-18
Improve deduplication of syntactic differences across devices for ComparePeerGroupPolicies

nickgian:
This change groups together syntactic differences (current snapshot policy, reference snapshot policy) from different devices allowing us to reduce the number of calls to CRP.

Syntactic differences across devices are considered the same if the definitions (lists) of the two devices are equal, the policies in current-snapshot are syntactically identical, and the policies in the reference-snapshot are syntactically identical.


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8884)* status: all files reviewed, 2 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 95 at r1](https://reviewable.io/reviews/batfish/batfish/8884#-NlyiukX8_FYYj1C-gug:-NlyiukX8_FYYj1C-guh:b-h96o2o) ([raw file](https://github.com/batfish/batfish/blob/73a51f3ec96ec1f76d5861a2ea0aa03339da4db7/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L95)):*
> ```Java
>    *     contexts. If all of these are equal, the two differences are equal. Otherwise, we simply
>    *     order the differences by name and hostname. Note that we consider the full routing context
>    *     of the device, even if some definitions are not used in this routing policy. This might
> ```

It seems like you already have a method in RoutingPolicyContextDiff (called differ) that can scope the difference to a particular routing policy, so why not use that to make this comparison more precise?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifferenceTest.java` line 31 at r1](https://reviewable.io/reviews/batfish/batfish/8884#-Nlyk2TN9oZnuALHmEC2:-Nlyk2TN9oZnuALHmEC3:bnllehj) ([raw file](https://github.com/batfish/batfish/blob/73a51f3ec96ec1f76d5861a2ea0aa03339da4db7/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifferenceTest.java#L31)):*
> ```Java
> import projects.minesweeper.src.main.java.org.batfish.minesweeper.question.comparepeergrouppolicies.SyntacticDifference;
> 
> public class SyntacticDifferenceTest {
> ```

A few other tests seem useful:  1) testing that routingPolicySyntaxEq properly tests equality of all the callees (finding when they are equal and also when they are not); 2) testing that compareTo orders the policies in the right way when it does not return 0.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8884)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 95 at r1](https://reviewable.io/reviews/batfish/batfish/8884#-NlyiukX8_FYYj1C-gug:-NmP2TYo6TtoGaUU0OOv:b-imfli7) ([raw file](https://github.com/batfish/batfish/blob/73a51f3ec96ec1f76d5861a2ea0aa03339da4db7/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L95)):*

It seems like you already have a method in RoutingPolicyContextDiff (called differ) that can scope the difference to a particular routing policy, so why not use that to make this comparison more precise?
</blockquote></details>

We probably can. It felt a bit more work at first, and potentially a bit expensive computationally as you have to construct a SyntacticCompare object for every comparison (which involves, walking the routing policy and tracking the definitions it uses).
I think I should be able to prototype this quickly, so let me give it a shot and report back.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifferenceTest.java` line 31 at r1](https://reviewable.io/reviews/batfish/batfish/8884#-Nlyk2TN9oZnuALHmEC2:-NmP4EpwCgYJgMc2lFik:bjxoa94) ([raw file](https://github.com/batfish/batfish/blob/73a51f3ec96ec1f76d5861a2ea0aa03339da4db7/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifferenceTest.java#L31)):*

A few other tests seem useful:  1) testing that routingPolicySyntaxEq properly tests equality of all the callees (finding when they are equal and also when they are not); 2) testing that compareTo orders the policies in the right way when it does not return 0.
</blockquote></details>

Added tests for 1, not sure about 2, we don't really have any ordering in mind, other than that they should not be considered equal. I suppose, I can strengthen a test that shows lexicographical ordering of policy-name/hostname "wins".

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8884)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 95 at r1](https://reviewable.io/reviews/batfish/batfish/8884#-NlyiukX8_FYYj1C-gug:-NmPcfkwEhT_boFpw55L:bjmhcqd) ([raw file](https://github.com/batfish/batfish/blob/73a51f3ec96ec1f76d5861a2ea0aa03339da4db7/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L95)):*

We probably can. It felt a bit more work at first, and potentially a bit expensive computationally as you have to construct a SyntacticCompare object for every comparison (which involves, walking the routing policy and tracking the definitions it uses).
I think I should be able to prototype this quickly, so let me give it a shot and report back.
</blockquote></details>

actually, the problem is SyntacticCompare works on top of configurations that concern the same device. We would have to redesign this class to apply `stripStatements` to every routing policy it touches. It's not infeasible, but the benefits to doing this refactoring in an efficient way seem small to me at this point. I suggest we go ahead with this slightly imprecise solution and refactor if performance becomes an issue in the future.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8884)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickgian)

PR #8897, opened 2023-12-20
Junos: refactor reference tracking into concrete and abstract structures

dhalperi:
This adopts the A10StructureType model of classifying all structures into abstract or concrete
automatically, and always marking all usage types. This appears to have caught at least 2
structures that weren't being marked at all: as-path-groups and bgp-rib-groups.

The Juniper structure model has been rewritten enough it's finally safe to do this.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8897)* status: 0 of 5 files reviewed, all discussions resolved

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureType.java` line 67 at r1](https://reviewable.io/reviews/batfish/batfish/8897#-Nm3mc1BDlGDCV4Z-GNg:-Nm3mc1CCQpVraBLiV5X:b-1vfoeq) ([raw file](https://github.com/batfish/batfish/blob/cc588d74221945af6fe10f104b520854b7a3f1a1/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureType.java#L67)):*
> ```Java
>   }
> 
>   public static final Multimap<JuniperStructureType, JuniperStructureType> ABSTRACT_STRUCTURES =
> ```

FYI this becomes a `MultiMap` which is a `Map<T, Collection<T>>` that maps an abstract type to all the concrete types it could be. I did it by copying all the abstract references that used to be in the `JuniperConfiguration`.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureType.java` line 75 at r1](https://reviewable.io/reviews/batfish/batfish/8897#-Nm3mnhY8FwJTSDC5XZ0:-Nm3mnhY8FwJTSDC5XZ1:bkgbszx) ([raw file](https://github.com/batfish/batfish/blob/cc588d74221945af6fe10f104b520854b7a3f1a1/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperStructureType.java#L75)):*
> ```Java
>   public static final Set<JuniperStructureType> CONCRETE_STRUCTURES =
>       ImmutableSet.copyOf(
>           Sets.difference(ImmutableSet.copyOf(values()), ABSTRACT_STRUCTURES.keySet()));
> ```

FYI this is using `values()` to model "all members of this enum", so that anything that is not ABSTRACT is CONCRETE. Since there's no explicit list that's redundant with the enum itself, all enum members are guaranteed to be in exactly one of these two sets.

This found at least two concrete structures that were not being marked, explaining the ref differences.

zergling-aws (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8897)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #8939, opened 2024-01-30
TracerouteEngineImpl: limit number of traces materialized from TraceDag.

anothermattbrown:
TraceDags have been observed to contain huge numbers of traces, too many to materialize into a list. Limiting to 10k for now. Rewrote TraceDagImpl to use iterators internally, as limit() wasn't working on the previous implementation based on recursive streams flattened with flatMap. Something about the stream implementation materializes many more traces than the limit, which wastes compute time and risks OOM. Rewriting using the same style of code but based on iterators solved the problem.


dhalperi (commented):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8939)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

PR #8944, opened 2024-02-15
Optimizations for CompareRoutePolicies

millstein:
A few performance optimizations for CompareRoutePolicies:
* When identifying differences between two BDDRoutes, ignore the input constraints at first; only impose them later on specific differences that are found (rather than on the entire BDDRoute as done previously)
* Don't compute the intersection between two path conditions unless it is actually needed

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
@@ -3,12 +3,14 @@ package org.batfish.common.bdd;
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.batfish.common.bdd.BDDUtils.bitvector;
 
+import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import net.sf.javabdd.BDD;
 import net.sf.javabdd.BDDFactory;
 import org.batfish.datamodel.Ip;
@@ -218,6 +220,13 @@ public final class MutableBDDInteger extends BDDInteger {
       val._bitvec[i] = pred.and(_bitvec[i]);
     }
     return val;
+  }
+
+  public BDD xor(BDDInteger other) {
+    return _factory.orAll(
+        IntStream.range(0, _bitvec.length)
+            .mapToObj(i -> _bitvec[i].xor(other._bitvec[i]))
+            .collect(ImmutableList.toImmutableList()));
   }
 
   /*
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
@@ -224,12 +224,14 @@ public final class MutableBDDInteger extends BDDInteger {
 
   /**
    * Produces a BDD whose models represent all possible differences between the two BDDIntegers --
-   * valuations of the BDD variables that cause the two BDDIntegers to have different values.
+   * valuations of the BDD variables that cause the two BDDIntegers to have different values. The
+   * two BDDIntegers are assumed to have the same length.
    *
    * @param other the second BDDInteger
    * @return a predicate represented as a BDD
    */
   public BDD allDifferences(BDDInteger other) {
+    assert _bitvec.length == other._bitvec.length;
     return _factory.orAll(
         IntStream.range(0, _bitvec.length)
             .mapToObj(i -> _bitvec[i].xor(other._bitvec[i]))
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/MutableBDDInteger.java
@@ -222,7 +222,14 @@ public final class MutableBDDInteger extends BDDInteger {
     return val;
   }
 
-  public BDD xor(BDDInteger other) {
+  /**
+   * Produces a BDD whose models represent all possible differences between the two BDDIntegers --
+   * valuations of the BDD variables that cause the two BDDIntegers to have different values.
+   *
+   * @param other the second BDDInteger
+   * @return a predicate represented as a BDD
+   */
+  public BDD allDifferences(BDDInteger other) {
     return _factory.orAll(
         IntStream.range(0, _bitvec.length)
             .mapToObj(i -> _bitvec[i].xor(other._bitvec[i]))
--- a/projects/batfish-common-protocol/src/test/java/org/batfish/common/bdd/MutableBDDIntegerTest.java
+++ b/projects/batfish-common-protocol/src/test/java/org/batfish/common/bdd/MutableBDDIntegerTest.java
@@ -110,4 +110,18 @@ public class MutableBDDIntegerTest {
     assertEquals(x.geq(30), xPlus1.value(31)); // x >= 31 ==> x+1 == 31
     assertEquals(x.geq(15), xPlus16.value(31)); // x >= 15 ==> x+16 == 31
   }
+
+  @Test
+  public void testAllDifferences() {
+    BDDFactory factory = BDDUtils.bddFactory(10);
+    MutableBDDInteger x = MutableBDDInteger.makeFromIndex(factory, 5, 0, false);
+    MutableBDDInteger constant1 = MutableBDDInteger.makeFromValue(factory, 5, 1);
+    MutableBDDInteger xPlus1 = x.add(constant1);
+
+    BDD allDiffs1 = x.allDifferences(constant1);
+    BDD allDiffs2 = x.allDifferences(xPlus1);
+
+    assertEquals(allDiffs1, x.value(0).or(x.geq(2)));
+    assertEquals(allDiffs2, factory.one());
+  }
 }
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java
@@ -440,11 +440,19 @@ public final class CompareRoutePoliciesUtils {
       for (TransferReturn otherPath : otherPaths) {
         BDD inputRoutes = path.getSecond();
         BDD inputRoutesOther = otherPath.getSecond();
-        BDD intersection = inputRoutesOther.and(inputRoutes).and(wf);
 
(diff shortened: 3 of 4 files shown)


nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8944)* status: all files reviewed, 4 unresolved discussions (waiting on @anothermattbrown and @millstein)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8944#-Nqfe2Jl4YnWskcY9k3l:-Nqfe2Jl4YnWskcY9k3m:b-v1q26q) (no related file):*
how do we evaluate these changes? We should test them against a few route-maps to ensure they help.

___
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 326 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqfRUds0_yL_0SGOfUa:-NqfRUds0_yL_0SGOfUb:b-85xwty) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L326)):*
> ```Java
>           break;
>         case ADMIN_DIST:
>           result = r1.getAdminDist().allDifferences(r2.getAdminDist());
> ```

what changed in all of these cases and we need the extra constraint? Was the old implementation wrong or is something done differently now?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 423 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqffTzNEYK3cuYlVGxm:-NqffTzNEYK3cuYlVGxn:b-3zsbxq) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L423)):*
> ```Java
>     List<Tuple<Result<BgpRoute>, Result<BgpRoute>>> differences = new ArrayList<>();
> 
>     BDDFactory factory = JFactory.init(100000, 10000);
> ```

I’ve seen cases where the size of node-table was doubled 11 times, should we pick a larger number here (e.g., 1M)?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 435 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-Nqf_6DK6M-b_20OieDM:-Nqf_6DK6M-b_20OieDN:bpnveg1) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L435)):*
> ```Java
>     List<TransferReturn> otherPaths = computePaths(otherTBDD);
> 
>     // Potential optimization: if a set of input routes between the two paths is the same then we
> ```

Not related to the changes in this CR, just trying to think what would be a more useful variant of this optimization in the comment.

Could we also build the union of every `inputRoutesOther` we found to be part of a difference, and exit the inner loop early when that union is equal to `inputRoutes`?
i.e.:

```
  for (path : paths) {
    BDD union = BDD.ZERO;
    for (otherPath: otherPaths) {
         …
         // after andSat code
         union = union.or(inputRoutesOther);

         …. // after computing differences at the end of the loop
         if union.equals(inputRoutes) { break; }   
    }
```

That might save you from comparing with all other paths, e.g., in case you add an extra condition/stanza somewhere in your route-map. It won’t work if you remove a stanza…
Anyway, just sharing in case it sparks a better idea for you.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 480 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-Nqff5ywDPhXN3UMVS_k:-Nqff5ywDPhXN3UMVS_l:bykuokx) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L480)):*
> ```Java
>                 if (!outputConstraints.isZero()) {
>                   behaviorDiff = true;
>                   finalConstraints = intersection.and(outputConstraints);
> ```

you can have `counterExampleOutputConstraints` return the intersection directly (or `Zero`) to slightly simplify this.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8944)* status: all files reviewed, 3 unresolved discussions (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 326 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqfRUds0_yL_0SGOfUa:-Nqnl2NB-IGj_G3wxH_l:b-90iv1y) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L326)):*

what changed in all of these cases and we need the extra constraint? Was the old implementation wrong or is something done differently now?
</blockquote></details>

The old implementation was wrong.  For example, suppose one route map sets the administrative distance to the value 0 and accepts, while the other route map simply accepts.  We would previously say there's a difference but the example we provide wouldn't show a difference, since it will still use the default admin distance of 0 in the input route.  The above constraint forces it to choose a different value like 1, so that the difference is exposed.  I added a test case.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 423 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqffTzNEYK3cuYlVGxm:-NqnmB3j1Y381Zcpd034:brykiyy) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L423)):*

I’ve seen cases where the size of node-table was doubled 11 times, should we pick a larger number here (e.g., 1M)?
</blockquote></details>

I don't have a good sense.  100,000 is the largest being used currently in the project, as far as I can tell.  Maybe @anothermattbrown has a thought?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 435 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-Nqf_6DK6M-b_20OieDM:-NqnmaD13ZQxEwopO1dl:b-jpc0cd) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L435)):*

Not related to the changes in this CR, just trying to think what would be a more useful variant of this optimization in the comment.

Could we also build the union of every `inputRoutesOther` we found to be part of a difference, and exit the inner loop early when that union is equal to `inputRoutes`?
i.e.:

```
  for (path : paths) {
    BDD union = BDD.ZERO;
    for (otherPath: otherPaths) {
         …
         // after andSat code
         union = union.or(inputRoutesOther);

         …. // after computing differences at the end of the loop
         if union.equals(inputRoutes) { break; }   
    }
```

That might save you from comparing with all other paths, e.g., in case you add an extra condition/stanza somewhere in your route-map. It won’t work if you remove a stanza…
Anyway, just sharing in case it sparks a better idea for you.
</blockquote></details>

Yes, that's a good thought.  I tried something somewhat related, which is that if you swap the order of the for loops and otherPath implies path, then you don't need to compare otherPath against any more paths in paths since they must be disjoint, so you can break out of the inner loop.  But at least in the cases I was looking at, that didn't help much -- you still need to do O(n^2) intersections (and I think the same is true of the above optimization).  Further, it turned out that a much bigger bottleneck was the cost of handling two paths that do intersect, rather than avoiding unnecessary path intersections, so I ended up focusing on the former.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8944)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 326 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqfRUds0_yL_0SGOfUa:-NqsZ_Ge6pd6WxqiNLFu:b-l4g8u2) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L326)):*

The old implementation was wrong.  For example, suppose one route map sets the administrative distance to the value 0 and accepts, while the other route map simply accepts.  We would previously say there's a difference but the example we provide wouldn't show a difference, since it will still use the default admin distance of 0 in the input route.  The above constraint forces it to choose a different value like 1, so that the difference is exposed.  I added a test case.
</blockquote></details>

thanks for the explanation and adding a test.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8944)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 423 at r1](https://reviewable.io/reviews/batfish/batfish/8944#-NqffTzNEYK3cuYlVGxm:-Nr1EIni8fBwokiE53P9:b-59tj1o) ([raw file](https://github.com/batfish/batfish/blob/48b0f53990bd3657ce785ec6194cd86ea6b24afb/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L423)):*

I don't have a good sense.  100,000 is the largest being used currently in the project, as far as I can tell.  Maybe @anothermattbrown has a thought?
</blockquote></details>

It doesn't really matter. In `BDDPacket` we use 1M:
https://github.com/batfish/batfish/blob/6709c01d541dd2d3f805d41113a8ff4bec88a8d7/projects/batfish-common-protocol/src/main/java/org/batfish/common/bdd/BDDPacket.java#L41

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8944)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

PR #8972, opened 2024-03-15
Junos: parse and warn on empty interface description

zergling-aws:
The prevents crashing when interface description is empty. Junos will ignore an interface statement with an empty description (ie the statement has no effect).

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_common.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_common.g4
@@ -79,7 +79,7 @@ dec
 
 description
 :
-  DESCRIPTION text = M_Description_DESCRIPTION?
+  DESCRIPTION text = M_Description_DESCRIPTION
 ;
 
 ec_administrator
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8972)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @zergling-aws)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4929 at r1](https://reviewable.io/reviews/batfish/batfish/8972#-Nt2tTSD1GDPtrlf5Pau:-Nt2tTSD1GDPtrlf5Pav:b-giaxft) ([raw file](https://github.com/batfish/batfish/blob/cae14396f86e4173af272a1de504b492249a8ede/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4929)):*
> ```Java
>     } catch (NullPointerException e) {
>       warn(ctx, "Batfish does not support an empty description");
>     }
> ```

The better way to do this would be to check for `null`.

```
if (ctx.description() == null) {
   warn...
} else {
    setDesc...
}
```

But I wonder -- why accept description missing vs just not parse it?



_Code quote:_
```Java
    try {
      String text = toString(ctx.description());
      _currentInterfaceOrRange.setDescription(text);
    } catch (NullPointerException e) {
      warn(ctx, "Batfish does not support an empty description");
    }
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8972)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @zergling-aws)

PR #8980, opened 2024-03-28
move logicalArrivalTime from rib to route inorder to decrease the memory usage.

swb2183713389xjtu:


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8980)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @swb2183713389xjtu)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8980#-Nu29dzJ2HSb0nd8n8lf:-Nu29dzJ2HSb0nd8n8lg:b-hqysdh) (no related file):*
Hi, thanks for this submission. Unfortunately, we can't move the logical time into the route itself -- that will break interning of routes. This really is a property of the RIB, or at least some sort of device-local received route instance.

I do think the underlying problem you highlight is real. What other fixes can we consider?

___

swb2183713389xjtu:
Thanks for your reply, I understand your meaning. 
In my experiments on large FatTree dataset, I found that a route only exists in one specific device, so I think this modification reduces memory usage. But considering the reality, it is true that storing the data in the rib is closer to reality.

dhalperi:
I wonder if this:
```
diff --git a/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java b/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java
index 4089d0c91f..2c4dcb40d8 100644
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/BgpRib.java
@@ -338,6 +338,11 @@ public abstract class BgpRib<R extends BgpRoute<?, ?>> extends AbstractRib<R> {
         }
       }
     }
+    if (_tieBreaker == BgpTieBreaker.ARRIVAL_ORDER) {
+      // Might be redundant with above, but handles backup routes.
+      // See: https://github.com/batfish/batfish/issues/8971
+      _logicalArrivalTime.remove(route);
+    }
     return delta;
   }
```

will do the trick. Tests seems to pass -- can you try it out?

PR #8989, opened 2024-04-12
Hierarchy: optimize apply-groups application by tracking separate types

dhalperi:
This makes a big difference (minutes to tends of seconds) for large files with very large apply groups.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
@@ -10,6 +10,7 @@ import com.google.common.collect.ImmutableList.Builder;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -144,7 +145,7 @@ final class Hierarchy {
     }
 
     // clear children and inherit from groups
-    inheritorNode._children = new LinkedHashMap<>();
+    inheritorNode.resetChildren();
     for (String group : Lists.reverse(prioritizedGroups)) {
       HierarchyNode groupNode = findApplicableGroupNode(inheritorNodePath, group, exceptGroups);
       if (groupNode == null) {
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8989)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @arifogel)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/8989#-NvJZ8gu5mgAcfVyTEDW:-NvJZ8gu5mgAcfVyTEDX:b-ryoyyb) (no related file):*
Should review commit by commit, since the first is just a refactoring (+addition of extra bookkeeping) to enable the second.

___

arifogel (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8989)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java` line 563 at r1](https://reviewable.io/reviews/batfish/batfish/8989#-NvJjci_Co09y3P0JDo0:-NvJjci_Co09y3P0JDo1:b-mq9xgs) ([raw file](https://github.com/batfish/batfish/blob/3a5e930c5e3d37d4d4aa8a38c2712793553f3a25/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java#L563)):*
> ```Java
>           // weird case where the removed node was of a different type than the added node.
>           // let's remove it from the wrong list.
>           removeConsistently(replaced);
> ```

Doesn't this ensure neither `node` nor `replaced` are in the wildcard or literal children lists?

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8989)* status: all files reviewed, 1 unresolved discussion (waiting on @arifogel)

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java` line 563 at r1](https://reviewable.io/reviews/batfish/batfish/8989#-NvJjci_Co09y3P0JDo0:-NvKMJVP00bdSCkCvtxl:b-phce9c) ([raw file](https://github.com/batfish/batfish/blob/3a5e930c5e3d37d4d4aa8a38c2712793553f3a25/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java#L563)):*

Doesn't this ensure neither `node` nor `replaced` are in the wildcard or literal children lists?
</blockquote></details>

No. This code is only invoked if `node` is a different class from `replaced`, and then only `replaced ` will get removed from the list that is not of the type of `node`. So node will stay in.

PR #8997, opened 2024-04-25
Junos: remove convError

dhalperi:
In all but two cases, the else was fraudulent: couldn't happen in grammar.

In the last two, there were what I believe are likely fraudulent applys (from
2015). It doesn't make sense to expand apply-groups there IMO, tho I could of
course be wrong.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/8997)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @progwriter)

PR #9005, opened 2024-04-28
ComparePeerGroupPolicies: deduplicate context differences.

nickgian:
The goal of this CR is for ComparePeerGroupPolicies to dedup differences across devices even if the devices have different context (e.g., different prefix-lists or community-lists). Initially, the thinking was that completeness (do not miss a difference in any device) was very important, but using the question I realize that this approach produces a lot of redundant information, for example when two devices have slightly different definitions of a prefix-list or a community-list. 
Perhaps the best approach to finding erroneous differences is another question to do a direct comparison of those definitions instead.

Changes:
1. `SyntacticDifference` the class that determines whether two pairs of current/reference routing policy are the same, no longer compares the list definitions (context). It also does not recursively compare sub-calls, rather only compares the top-level which contains the route-map name (sub-calls contain statements with IP addresses/communities which often differ across devices).
2. Fixes a bug with `SyntacticDifference::compareTo` where it was a) not consistent with `equalsTo` and b) not deterministic due to the way the comparison was done in the non-equals case (i.e., it was possible that `a == b`, `b < c` and `a > c`).

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesUtils.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesUtils.java
@@ -80,8 +80,7 @@ public final class ComparePeerGroupPoliciesUtils {
       SyntacticDifference d =
           new SyntacticDifference(
               syntacticCompare.getCurrentConfig().getRoutingPolicies().get(currentPolicy),
-              syntacticCompare.getReferenceConfig().getRoutingPolicies().get(referencePolicy),
-              syntacticCompare.getContextDiff());
+              syntacticCompare.getReferenceConfig().getRoutingPolicies().get(referencePolicy));
       SortedSet<String> routers = differences.computeIfAbsent(d, k -> new TreeSet<>());
       routers.add(router);
     }
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/RoutingPolicyContextDiff.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/RoutingPolicyContextDiff.java
@@ -1,11 +1,17 @@
 package projects.minesweeper.src.main.java.org.batfish.minesweeper.question.comparepeergrouppolicies;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Comparators;
+import com.google.common.collect.Ordering;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import javax.annotation.Nonnull;
 import org.batfish.datamodel.Configuration;
 import org.batfish.datamodel.routing_policy.RoutingPolicy;
 import org.batfish.minesweeper.aspath.RoutePolicyStatementMatchCollector;
@@ -27,17 +33,17 @@ import org.batfish.minesweeper.utils.Tuple;
  * exploration of policy calls. If you want a full syntactic difference, including the context and
  * statements of a routing policy, you should use {@link SyntacticCompare}.
  */
-public final class RoutingPolicyContextDiff {
+public final class RoutingPolicyContextDiff implements Comparable<RoutingPolicyContextDiff> {
   private final Configuration _currentConfig;
 
   /** The set of community lists that differ between the two configurations. */
-  private final Set<String> _communityListsDiff;
+  private final SortedSet<String> _communityListsDiff;
 
   /** The set of as-path access lists that differ between the two configurations. */
-  private final Set<String> _asPathAccessListsDiff;
+  private final SortedSet<String> _asPathAccessListsDiff;
 
   /** The set of prefix lists that differ between the two configurations. */
-  private final Set<String> _routeFilterListsDiff;
+  private final SortedSet<String> _routeFilterListsDiff;
 
   public RoutingPolicyContextDiff(Configuration currentConfig, Configuration referenceConfig) {
     this._currentConfig = currentConfig;
@@ -61,9 +67,9 @@ public final class RoutingPolicyContextDiff {
    * @return the set of keys for which the value V differs between the two maps (including keys
    *     missing from one map).
    */
-  private <K, V> Set<K> mapDiffKeys(Map<K, V> first, Map<K, V> second) {
-    Set<K> diff = new HashSet<>();
-    Set<K> keys = new HashSet<>(first.keySet());
+  private <K extends Comparable<K>, V> SortedSet<K> mapDiffKeys(Map<K, V> first, Map<K, V> second) {
+    SortedSet<K> diff = new TreeSet<>();
+    SortedSet<K> keys = new TreeSet<>(first.keySet());
     keys.addAll(second.keySet());
     for (K key : keys) {
       V fv = first.get(key);
@@ -118,17 +124,17 @@ public final class RoutingPolicyContextDiff {
   }
 
   @VisibleForTesting
-  public Set<String> getCommunityListsDiff() {
+  public SortedSet<String> getCommunityListsDiff() {
     return _communityListsDiff;
   }
 
   @VisibleForTesting
-  public Set<String> getAsPathAccessListsDiff() {
+  public SortedSet<String> getAsPathAccessListsDiff() {
     return _asPathAccessListsDiff;
   }
 
   @VisibleForTesting
-  public Set<String> getRouteFilterListsDiff() {
+  public SortedSet<String> getRouteFilterListsDiff() {
     return _routeFilterListsDiff;
   }
 
@@ -158,5 +164,21 @@ public final class RoutingPolicyContextDiff {
     result = 31 * result + _asPathAccessListsDiff.hashCode();
     result = 31 * result + _routeFilterListsDiff.hashCode();
     return result;
+  }
+
+  private static final Comparator<RoutingPolicyContextDiff> COMPARATOR =
+      Comparator.comparing(
+              RoutingPolicyContextDiff::getCommunityListsDiff,
+              Comparators.lexicographical(Ordering.natural()))
+          .thenComparing(
+              RoutingPolicyContextDiff::getRouteFilterListsDiff,
+              Comparators.lexicographical(Ordering.natural()))
+          .thenComparing(
+              RoutingPolicyContextDiff::getAsPathAccessListsDiff,
+              Comparators.lexicographical(Ordering.natural()));
+
+  @Override
+  public int compareTo(@Nonnull RoutingPolicyContextDiff that) {
+    return COMPARATOR.compare(this, that);
   }
 }
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java
@@ -2,20 +2,21 @@ package projects.minesweeper.src.main.java.org.batfish.minesweeper.question.comp
 
 import static org.batfish.question.TracingHintsStripper.TRACING_HINTS_STRIPPER;
 
+import com.google.common.collect.Comparators;
+import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.ImmutableList;
 import java.util.Comparator;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
+import java.util.Optional;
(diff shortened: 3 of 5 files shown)


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: all files reviewed, 3 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 28 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa73UZ599sboNQAqZH:-Nwa73UZ599sboNQAqZI:b-2282m4) ([raw file](https://github.com/batfish/batfish/blob/eb8c824429494b07af6585cd8d7ac7b2956b4d5b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L28)):*
> ```Java
> 
>   /** The (potential) context-diff under which the two policies differ */
>   private final RoutingPolicyContextDiff _context;
> ```

Do you think it would be useful to optionally keep the context difference, allowing the client of `SyntacticDifference` to decide whether they want to include it or not?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa8ViHDbdtJZ_sUjSO:-Nwa8ViHDbdtJZ_sUjSP:b-575khl) ([raw file](https://github.com/batfish/batfish/blob/a0f57dba93943024d4652dc7b5558d333383d7e9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L57)):*
> ```Java
>    *     be small differences in prefix-lists or community-lists.
>    */
>   private int routingPolicySyntaxComparison(RoutingPolicy p1, RoutingPolicy p2) {
> ```

Make this method static?
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswererTest.java` line 682 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa98BQ03ZvITqudIU8:-Nwa98BQ03ZvITqudIU9:b-8fforp) ([raw file](https://github.com/batfish/batfish/blob/a0f57dba93943024d4652dc7b5558d333383d7e9/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswererTest.java#L682)):*
> ```Java
>             allOf(
>                 hasColumn(COL_NODE, equalTo(new Node(HOSTNAME)), Schema.NODE),
>                 hasColumn(COL_POLICY_NAME, equalTo("RM2"), Schema.STRING),
> ```

Why did you switch the order of the two expected answers?  Is that important for some reason?

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 28 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa73UZ599sboNQAqZH:-Nwbg7PQ-1Kd7w7woSMw:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/eb8c824429494b07af6585cd8d7ac7b2956b4d5b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L28)):*

Do you think it would be useful to optionally keep the context difference, allowing the client of `SyntacticDifference` to decide whether they want to include it or not?
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 57 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa8ViHDbdtJZ_sUjSO:-NwbM7dS0ejrdpXa4Fow:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a0f57dba93943024d4652dc7b5558d333383d7e9/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L57)):*

Make this method static?
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswererTest.java` line 682 at r1](https://reviewable.io/reviews/batfish/batfish/9005#-Nwa98BQ03ZvITqudIU8:-Nwbg8dA6oDT49pdHTCg:b-sdmlco) ([raw file](https://github.com/batfish/batfish/blob/a0f57dba93943024d4652dc7b5558d333383d7e9/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesAnswererTest.java#L682)):*

Why did you switch the order of the two expected answers?  Is that important for some reason?
</blockquote></details>

Not important, it's just because the `compareTo` method changed, it happens that with the new method these two had to swap.

dhalperi (commented):
> Fixes a bug with `SyntacticDifference::compareTo` where it was a) not consistent with `equalsTo` and b) not deterministic due to the way the comparison was done in the non-equals case (i.e., it was possible that `a == b`, `b < c` and `a > c`).

(No action needed)

My favorite way to test equals looks something like this:

```
L = List in sorted order
assertThat(sort(L), equalTo(L));
assertThat(sort(reverse(L)), equalTo(L));
```

See:

```

  @Test
  public void testSorting() {
    List<RoutingProtocolInstance> entries =
        ImmutableList.of(
            RoutingProtocolInstance.bgp(5),
            RoutingProtocolInstance.bgp(6),
            RoutingProtocolInstance.eigrp(6),
            RoutingProtocolInstance.eigrp(7),
            RoutingProtocolInstance.ospf());
    assertThat(ImmutableList.sortedCopyOf(entries), equalTo(entries));
    assertThat(ImmutableList.sortedCopyOf(Lists.reverse(entries)), equalTo(entries));
  }
```

In a random place.

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005#-:-NweS1XF05v_1ZD0eJ-z:b-jja1f1)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @millstein)

dhalperi (commented):
s/test equals/test comparator/

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005#-:-NweSd81AH95Teof2AEH:bxmbg6m)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @millstein)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: all files reviewed, 3 unresolved discussions (waiting on @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 37 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjIfGg16X0-SuhTQ1a:-NwjIfGg16X0-SuhTQ1b:b-n7praz) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L37)):*
> ```Java
>       RoutingPolicy currentPolicy,
>       RoutingPolicy referencePolicy,
>       RoutingPolicyContextDiff context) {
> ```

Do you expect these to be non-null?  It'd be useful to document that if so, either with Nonnull annotations or a general annotation on the class that ParametersAreNonnullByDefault.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 86 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjJdY6C1HUJPbgKM43:-NwjJdY6C1HUJPbgKM44:bowo7mt) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L86)):*
> ```Java
>    */
>   @Override
>   public int compareTo(SyntacticDifference o) {
> ```

You can probably simplify this by writing a comparator, in the same style as done above for RoutingPolicyContextDiff.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 99 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjJq8S9Ok4FtNCjEVv:-NwjJq8S9Ok4FtNCjEVw:b-xl0k0t) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L99)):*
> ```Java
>         return comparatorOpt.compare(this._context, o._context);
>       }
>     }
> ```

Since you have compareTo, do you also need a dedicated equals (and hashCode) method?  Otherwise it seems that you have a situation where compareTo can return 0 but equals returns false, which could cause inconsistencies.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: all files reviewed, 3 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 99 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjJq8S9Ok4FtNCjEVv:-Nwl7lDb4thm4hdPO3Da:b-id532n) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L99)):*

Since you have compareTo, do you also need a dedicated equals (and hashCode) method?  Otherwise it seems that you have a situation where compareTo can return 0 but equals returns false, which could cause inconsistencies.
</blockquote></details>

there is one, that uses `compareTo` for `equals`, but I need to fix `hashCode` to use the optional context too.

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 37 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjIfGg16X0-SuhTQ1a:-NwzrRhCEZZ318shahWo:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L37)):*

Do you expect these to be non-null?  It'd be useful to document that if so, either with Nonnull annotations or a general annotation on the class that ParametersAreNonnullByDefault.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 86 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjJdY6C1HUJPbgKM43:-NwzrR393DK6x9FwI7dR:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L86)):*

You can probably simplify this by writing a comparator, in the same style as done above for RoutingPolicyContextDiff.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java` line 99 at r2](https://reviewable.io/reviews/batfish/batfish/9005#-NwjJq8S9Ok4FtNCjEVv:-NwziI--2Q74psw94i_D:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/fab1098d47c4e1f8a7d6889cc1a2f9a5052e1414/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/SyntacticDifference.java#L99)):*

there is one, that uses `compareTo` for `equals`, but I need to fix `hashCode` to use the optional context too.
</blockquote></details>

Done.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9005)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickgian)

PR #9008, opened 2024-05-03
Junos: convert last F_DecByte to F_Uint8

dhalperi:
Juniper allows leading 0s in numbers, generally, including IP addresses. We already handle
this in F_Uint8, but somehow we had a nearly-equivalent F_DecByte that was only used
in IP address that did not allow zeros. Fix this last case, add a test, and delete
the now unused fragment.


anothermattbrown (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9008)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9018, opened 2024-05-25
BgpRoutingProcess: optimize away full RIB copy when add-path is not on

dhalperi:
We only use this if add-path is on, so let's save that RAM and CPU if not used.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -223,6 +223,9 @@ final class BgpRoutingProcess implements RoutingProcess<BgpTopology, BgpRoute<?,
   private @Nonnull RibDelta<Bgpv4Route> _bgpv4DeltaPrevBestPath = RibDelta.empty();
 
   // copy of RIBs from prev round, for new links in the current round.
+  // note that we only populate the full Prev (not PrevBestPath) versions when
+  // some session has additional paths and we need to be able to send them.
+  private boolean _anySessionHasAdditionalPaths = true;
   private @Nonnull Set<Bgpv4Route> _ebgpv4Prev;
   private @Nonnull Set<Bgpv4Route> _ebgpv4PrevBestPath;
   private @Nonnull Set<Bgpv4Route> _bgpv4Prev;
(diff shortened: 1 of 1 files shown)


dhalperi:
FYI this doesn't actually save much work yet, because we call these functions in stupid things like iteration statistics or hash code.

progwriter:
*[`projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java` line 2459 at r1](https://reviewable.io/reviews/batfish/batfish/9018#-Nz-G38s0q13Vll3ZhN3:-Nz-G38s0q13Vll3ZhN4:b-osalwn) ([raw file](https://github.com/batfish/batfish/blob/f4edba67879ad95e04688d92b509f22319e9a996/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L2459)):*
> ```Java
>    */
>   private boolean computeAnySessionHasAdditionalPaths() {
>     for (EdgeId edge : _bgpv4Edges) {
> ```

Is it clear how evpn and add-path interact? I would have assumed addpath for evpn is also possible (even if for type 5 only). Some nxos docs [hint at it](https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/102x/configuration/vxlan/cisco-nexus-9000-series-nx-os-vxlan-configuration-guide-release-102x/m_configuring_proportional_multipath_for_vnf.html). sorry very rudimentary search.

that said, this feels like uncharted feature territory to me, so not a blocker.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9018)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @progwriter)

___
*[`projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java` line 2459 at r1](https://reviewable.io/reviews/batfish/batfish/9018#-Nz-G38s0q13Vll3ZhN3:-Nz-IMS03eW6kJ_6V4lB:b-2secdu) ([raw file](https://github.com/batfish/batfish/blob/f4edba67879ad95e04688d92b509f22319e9a996/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java#L2459)):*

Is it clear how evpn and add-path interact? I would have assumed addpath for evpn is also possible (even if for type 5 only). Some nxos docs [hint at it](https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/102x/configuration/vxlan/cisco-nexus-9000-series-nx-os-vxlan-configuration-guide-release-102x/m_configuring_proportional_multipath_for_vnf.html). sorry very rudimentary search.

that said, this feels like uncharted feature territory to me, so not a blocker.
</blockquote></details>

I mean you're definitely right, but we don't support add-path for EVPN at all.

PR #9022, opened 2024-05-28
FileBasedStorage: skip string when writing pojo topology

dhalperi:
User reports that on a very large snapshot this causes a 2GiB limit to be hit.


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9022)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

Issue #9013, opened 2024-05-18
Palo Alto Interface missing bandwidth, cannot compute OSPF cost


netops501:
**Describe the bug and expected behavior**

Palo Alto firewall doesn't support "bandwidth" under the interface. However, it has metric keyword for OSPF interface instead. For example,

set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 metric 5001

**Runnable example**
```py
from pybatfish.client.session import Session
TXT = """
!RANCID-CONTENT-TYPE: paloalto
set network interface tunnel units tunnel.11 comment "tun11 VPN hq"
set network interface tunnel units tunnel.11 ip 169.254.11.2/30
set network interface tunnel units tunnel.11 mtu 1427
set network virtual-router default protocol bgp enable no
set network virtual-router default protocol bgp routing-options graceful-restart enable yes
set network virtual-router default protocol rip enable no
set network virtual-router default protocol ospf enable yes
set network virtual-router default protocol ospf router-id 10.10.10.1
set network virtual-router default protocol ospf reject-default-route yes
set network virtual-router default protocol ospf allow-redist-default-route yes
set network virtual-router default protocol ospf area 0.0.0.0 type normal
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 enable yes
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 passive no
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 authentication ospf_md5
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 metric 5001
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 priority 1
set network virtual-router default protocol ospf area 0.0.0.0 interface tunnel.11 link-type p2p
set network virtual-router default ecmp algorithm ip-modulo
set network virtual-router default interface [ tunnel.11 ]
set network virtual-router default routing-table ip static-route vpn-hq-0 interface tunnel.11
set network virtual-router default routing-table ip static-route vpn-hq-0 metric 10
set network virtual-router default routing-table ip static-route vpn-hq-0 destination 10.0.1.0/24
set network virtual-router default routing-table ip static-route vpn-hq-0 route-table unicast
set zone VPN network layer3 tunnel.11
set import network interface [ tunnel.11 ]
[edit]
"""
bf = Session()
bf.set_network("github-bug-report")
bf.init_snapshot_from_text(TXT)
[...]

PR #9024, opened 2024-05-29
PAN: fix OSPF metric grammar and values

dhalperi:
There are two types of OSPF metrics with different bounds. Separate them
grammatically and make the interface metric one correctly u16. Improve testing
for new behavior including both 16 bit support and using the PAN default when
metric is not configured in configs.

Fix batfish/batfish#9013, thanks @netops501!


ratulm (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9024)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9028, opened 2024-05-30
GenericRibReadOnly: rename functions for clarity

dhalperi:
Rather than calling out (some of the) typed functions, call out the untyped ones.
Note that the old getRoutes(Prefix) was typed, but not named so.


progwriter (commented):
> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9028)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9051, opened 2024-06-04
Junos: parse and ignore vlan "default"

anothermattbrown:
From [junos docs](https://www.juniper.net/documentation/us/en/software/junos/multicast-l2/topics/topic-map/bridging-and-vlans.html#:~:text=A%20Default%20VLAN%20Is%20Configured%20on%20Most%20Switches,-Some%20switches%20running&text=On%20these%20switches%2C%20each%20interface,not%20support%20a%20default%20VLAN):

> Some switches running Junos OS that do not support the ELS configuration style are preconfigured with a VLAN named default that does not tag packets and operates only with untagged packets. On these switches, each interface already belongs to the VLAN named default and all traffic uses this VLAN until you configure more VLANs and assign traffic to those VLANs.

Parse and ignore the `default` vlan, and don't count it as a reference to a vlan variable.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -465,6 +465,8 @@ DEAD_PEER_DETECTION: 'dead-peer-detection';
 
 DECAPSULATE: 'decapsulate';
 
+DEFAULT: 'default';
+
 DEFAULT_ACTION: 'default-action';
 
 DEFAULT_ADDRESS_SELECTION: 'default-address-selection';
@@ -4335,6 +4337,8 @@ M_Version_WS
 mode M_VlanMembers;
 
 M_VlanMembers_ALL: 'all' -> type(ALL), popMode;
+
+M_VlanMembers_DEFAULT: 'default' -> type(DEFAULT), popMode;
 
 M_VlanMembers_WS: F_WhitespaceChar+ -> skip;
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4
@@ -453,6 +453,7 @@ ife_vlan
    VLAN MEMBERS
    (
       ALL
+      | DEFAULT
       | range
       | name = junos_name
    )
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -5032,6 +5032,8 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
       _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(new VlanReference(name));
     } else if (ctx.ALL() != null) {
       _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(AllVlans.instance());
+    } else if (ctx.DEFAULT() != null) {
+      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().clear();
     } else {
       todo(ctx);
     }
--- a/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
@@ -3561,7 +3561,10 @@ public final class FlatJuniperGrammarTest {
                     .build())));
 
     // Vlan "default" is ignored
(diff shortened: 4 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9051)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @arifogel)

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4` line 451 at r1](https://reviewable.io/reviews/batfish/batfish/9051#-NzZtCAEAt-4yG-NrSg5:-NzZtCAEAt-4yG-NrSg6:bhpyi2) ([raw file](https://github.com/batfish/batfish/blob/a9a1c19193465ae4274aed0089a0d3a82ebb35f7/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4#L451)):*
> ```ANTLR
> ;
> 
> ife_vlan
> ```

I think we need to update the implementation of `ife_vlan` too. Two things:

1. As written, this will log a `todo`. I assume we don't want that:

```
  @Override
  public void exitIfe_vlan(Ife_vlanContext ctx) {
    if (!_currentInterfaceOrRange.getName().endsWith(".0")) {
      warn(ctx, "Vlan members can only be configured on unit 0");
      return;
    }
    if (ctx.range() != null) {
      IntegerSpace range = IntegerSpace.unionOf(toRange(ctx.range()).toArray(new SubRange[] {}));
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(new VlanRange(range));
    } else if (ctx.name != null) {
      String name = toString(ctx.name);
      _configuration.referenceStructure(VLAN, name, INTERFACE_VLAN, getLine(ctx.name.getStart()));
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(new VlanReference(name));
    } else if (ctx.ALL() != null) {
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(AllVlans.instance());
    } else {
      // DAN will hit here, right?
      todo(ctx);
    }
  }```

2. I think this should result in a mutating change, right?

E.g.,

```
set interfaces ge-0/10/0 unit 0 family ethernet-switching vlan members 5
set interfaces ge-0/10/0 unit 0 family ethernet-switching vlan members default
```

Should clear the `5` right?
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9051)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @arifogel)

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4` line 451 at r1](https://reviewable.io/reviews/batfish/batfish/9051#-NzZtCAEAt-4yG-NrSg5:-NzZtyMR2C3RXKn4HVr1:bpuhdmx) ([raw file](https://github.com/batfish/batfish/blob/a9a1c19193465ae4274aed0089a0d3a82ebb35f7/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_interfaces.g4#L451)):*

I think we need to update the implementation of `ife_vlan` too. Two things:

1. As written, this will log a `todo`. I assume we don't want that:

```
  @Override
  public void exitIfe_vlan(Ife_vlanContext ctx) {
    if (!_currentInterfaceOrRange.getName().endsWith(".0")) {
      warn(ctx, "Vlan members can only be configured on unit 0");
      return;
    }
    if (ctx.range() != null) {
      IntegerSpace range = IntegerSpace.unionOf(toRange(ctx.range()).toArray(new SubRange[] {}));
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(new VlanRange(range));
    } else if (ctx.name != null) {
      String name = toString(ctx.name);
      _configuration.referenceStructure(VLAN, name, INTERFACE_VLAN, getLine(ctx.name.getStart()));
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(new VlanReference(name));
    } else if (ctx.ALL() != null) {
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().add(AllVlans.instance());
    } else {
      // DAN will hit here, right?
      todo(ctx);
    }
  }```

2. I think this should result in a mutating change, right?

E.g.,

```
set interfaces ge-0/10/0 unit 0 family ethernet-switching vlan members 5
set interfaces ge-0/10/0 unit 0 family ethernet-switching vlan members default
```

Should clear the `5` right?
```

</blockquote></details>

I think this is likely just another `else if`:

```
    } else if (ctx.DEFAULT() != null ) {
      _currentInterfaceOrRange.getEthernetSwitching().getVlanMembers().clear();
```

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9051)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @arifogel)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9051)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @arifogel)

PR #9069, opened 2024-06-07
minesweeper: handwavy implementation of CommunitySet HasSize

dhalperi:
Filtering on community set size is an extremely fragile capability. The primary
way we've seen it used is as a safeguard against, say, 1000 communities on a route.
In practical networks, we rarely see more than a dozen communities.

Make minesweeper approximate HasSize in a way that lines up with that use case,
which may need adaptation if we learn of others. This improves behavior
in a common case.


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9069)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9081, opened 2024-06-20
Minesweeper: add support for TunnelEncapsulationAttribute

dhalperi:
Create a BDD Domain for all distinct attribute values seen,
and set it.

---

**Stack**:
- #9082
- #9081 ⬅


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: 0 of 16 files reviewed, 3 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`BUILD.bazel` line 14 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r63S442uoj72o0c_N:-O-r63S442uoj72o0c_O:b-jmls0z) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/BUILD.bazel#L14)):*
> ```Starlark
> )
> 
> java_plugin(
> ```

@nickgian @anothermattbrown thought you might like to see this machinery to automake dataclasses for you. I'm a bit torn as to how valuable it is, but it saves SOME boilerplate.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 329 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r6MBECOf9_H00uyUw:-O-r6MBECOf9_H00uyUx:b-cw27sc) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L329)):*
> ```Java
>     len = _ospfMetric.getInteger().size();
>     addBitNames("ospfMetric", len, idx, false);
>     idx += len;
> ```

FYI this increment is unnecessary, but when I originally added tunnel attribute below this it failed in really weird ways because index wasn't incremented. Seems better to just increment it after every field and then avoid unused for the last increment at the bottom.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r6aQmBZ7fWkZrooiM:-O-r6aQmBZ7fWkZrooiN:b-7ou1iu) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java#L22)):*
> ```Java
>     TAG,
>     ADMIN_DIST,
>     TUNNEL_ENCAPSULATION_ATTRIBUTE,
> ```

FYI no clue how this enum is ordered if at all, but I was looking at "everywhere weight is used" so I put it here :shrug: 
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 216 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r6kjdDQcs4qJTrLhs:-O-r6kjdDQcs4qJTrLht:b-70ennz) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L216)):*
> ```Java
>       return value.value();
>     }
>     // BDDRoute can't handle unknown, so make one up.
> ```

I mean BGPRoute.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 420 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r6so2CIcJaww-hADj:-O-r6so2CIcJaww-hADk:bpwgq57) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L420)):*
> ```Java
>    * @return the unique item consistent with the model, or null if there is none
>    */
>   private static <T> @Nullable T optionalSatisfyingItem(
> ```

revert? I think this is unnecessary and dates back to an earlier approach before I realized I also needed to model unknown.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 341 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r7-yJ60k0OYnVd6y1:-O-r7-yJ60k0OYnVd6y2:byx74ux) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L341)):*
> ```Java
>           result = factory.one();
>           break;
>         case TUNNEL_ENCAPSULATION_ATTRIBUTE: /* TODO */
> ```

@nickgian @millstein opinions on what this should do? I can't tell when we choose to model things as one() or not.

nickgian (commented):
> Reviewed 8 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: 8 of 16 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java` line 30 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01QwzA692FX_fGkscl:-O01QwzA692FX_fGkscm:ba8pplw) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java#L30)):*
> ```Java
>     TunnelEncapsulationAttributeExpr expr =
>         setTunnelAttribute.getTunnelEncapsulationAttributeExpr();
>     assert expr instanceof LiteralTunnelEncapsulationAttribute;
> ```

what do we do in other collectors? 
Wondering if this should be an assertion or should we return an empty set instead (and log an error/warning) and expect that the command will be ignored or the path will be deemed unsupported?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 723 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01SE3iCNiuRtc0ExMr:-O01SE3iCNiuRtc0ExMs:b-wszwx5) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L723)):*
> ```Java
>         && Objects.equals(_sourceVrfs, other._sourceVrfs)
>         && Arrays.equals(_tracks, other._tracks)
>         && _tunnelEncapsulationAttribute.equals(other._tunnelEncapsulationAttribute)
> ```

curious, why not use `Objects.equals`? (or why do we use it in all others :P)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: 8 of 16 files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 723 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01SE3iCNiuRtc0ExMr:-O02XvBXD4ov1MnVXk2m:b-d70o5x) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L723)):*

curious, why not use `Objects.equals`? (or why do we use it in all others :P)
</blockquote></details>

The main difference between `Objects.equals(a, b)` and `a.equals(b)` is that the first one won't crash if `a` is `null`. But since `_tunnelEncapsulationAttribute` can't be null, not needed.

millstein (commented):
> Reviewed 12 of 16 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: all files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java` line 30 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01QwzA692FX_fGkscl:-O0A1QDl53CGJKBH_eV-:bp8r0o3) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java#L30)):*

what do we do in other collectors? 
Wondering if this should be an assertion or should we return an empty set instead (and log an error/warning) and expect that the command will be ignored or the path will be deemed unsupported?
</blockquote></details>

Yes we've usually opted to not fail here so that we can still do the analysis and log a warning at that point (in TransferBDD)
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 223 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A-2U8Chn_H_LlrLac:-O0A-2U8Chn_H_LlrLad:b89tplr) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L223)):*
> ```Java
>             + numBits(numSourceVrfs + 1)
>             + numTracks
>             + numBits(tunnelEncapsulationAttributes.size() + 2)
> ```

Why + 2?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r6aQmBZ7fWkZrooiM:-O0A19wp7cDCW_DFUifX:b-vc7bia) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteDiff.java#L22)):*

FYI no clue how this enum is ordered if at all, but I was looking at "everywhere weight is used" so I put it here :shrug:
</blockquote></details>

probably fine as is, but @nickgian should confirm
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A2JgE0xIuprKNegWO:-O0A2JgE0xIuprKNegWP:b-etvhvz) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java#L25)):*
> ```Java
>     enum Type {
>       /** There is no Tunnel Encapsulation Attribute on the advertisement. */
>       ABSENT,
> ```

We have two other uses of BDDDomain that also need an absent value but it's just hacked in.  So I wonder if this would be cleaner if we first made a class that supports the optional emptiness, which could be used by all three cases, and then build your BDDTunnelEncapsulationAttribute as a wrapper on that?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java` line 63 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A1q2WBiaTgRm-Y8uQ:-O0A1q2WBiaTgRm-Y8uR:b-eg2c0b) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java#L63)):*
> ```Java
>       BDDFactory factory, int index, List<TunnelEncapsulationAttribute> knownValues) {
>     ImmutableList.Builder<Value> domainValues =
>         ImmutableList.<Value>builderWithExpectedSize(knownValues.size() + 2).add(Value.ABSENT);
> ```

You're adding 2 more values so don't you just need one extra bit instead of 2?  Or is there some edge case to consider?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 208 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A3eaCAQH7ROC7WF2C:-O0A3eaD3wkKMf4hlrAN:bsgn27r) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L208)):*
> ```Java
>    * with the assignment.
>    */
>   static @Nullable TunnelEncapsulationAttribute satAssignmentToTunnelEncapsulationAttribute(
> ```

Presumably we'd like to use ABSENT whenever possible, since it makes the model simpler.  And we only want to use UNKNOWN when necessary.  Will this do both of those things?  I guess maybe it will due to the careful way that you ordered the values when you created the BddTunnelEncapsulationAttribute?  If so, worth a comment there to that effect (and some tests to make sure it's not changed later).  If not, then it could be done here in the same way that certain IP addresses are preferred as defaults.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 884 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A4ssg5E-3QC3PLaE0:-O0A4ssg5E-3QC3PLaE1:bwzkofd) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L884)):*
> ```Java
>         SetTunnelEncapsulationAttribute st = (SetTunnelEncapsulationAttribute) stmt;
>         LiteralTunnelEncapsulationAttribute expr =
>             (LiteralTunnelEncapsulationAttribute) st.getTunnelEncapsulationAttributeExpr();
> ```

Why not check for this and use UNKNOWN otherwise?  It seems like you're not using UNKNOWN at all currently.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 330 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A6c6222HMI_R4nFJn:-O0A6c6222HMI_R4nFJo:bjqxdlh) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L330)):*
> ```Java
>           result = r1.getAdminDist().allDifferences(r2.getAdminDist());
>           break;
>         case TUNNEL_ENCAPSULATION_ATTRIBUTE:
> ```

This means that differences in tunnel encapsulation attributes will now be considered by CRP (as I understand it), so probably worth adding a test to make sure that it does the right thing.  cc @nickgian 
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttributeTest.java` line 35 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A5VMf-RB4d4mZOZW6:-O0A5VMf-RB4d4mZOZW7:b-d59g43) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttributeTest.java#L35)):*
> ```Java
>             BDDTunnelEncapsulationAttribute.create(factory, 0, ImmutableList.of(b)),
>             // Document that no values are *not* considered: different size but same numBits.
>             BDDTunnelEncapsulationAttribute.create(factory, 0, ImmutableList.of(a, b)))
> ```

I don't understand this -- why would we want something with only 1 possible value to be considered equal to something with 2 possible values?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: all files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 341 at r1](https://reviewable.io/reviews/batfish/batfish/9081#-O-r7-yJ60k0OYnVd6y1:-O0A7n_iBtLedro1s94D:b-p2oi5c) ([raw file](https://github.com/batfish/batfish/blob/043470107027e17662dc98910ab51178c8c96890/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L341)):*

@nickgian @millstein opinions on what this should do? I can't tell when we choose to model things as one() or not.
</blockquote></details>

The way you did it is correct.  The reason we need this is because let's say one route map removes the tunnel attribute and the other does not.  Without this, we will find the difference but might generate an example that doesn't show the difference, because the input route will start with an absent tunnel attribute.  This extra constraint will force us to provide an input route with a non-absent tunnel attribute, so that the output routes will differ.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: all files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java` line 30 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01QwzA692FX_fGkscl:-O0A9Q0f-9wURAWkhvs4:b1pyra3) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java#L30)):*

Yes we've usually opted to not fail here so that we can still do the analysis and log a warning at that point (in TransferBDD)
</blockquote></details>

I think that was a pragmatic choice for starting out given the large amount of old features.

But now that this functionality exists, I actually think we should crash when adding new features, so we have to add their support here.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 223 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A-2U8Chn_H_LlrLac:-O0A9jKr1Aqf5kU9nr44:b-e12gq5) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L223)):*

Why + 2?
</blockquote></details>

There are two values in the domain that are not in the input set (ABSENT and UNKNOWN). I added a useful helper that covers this.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A2JgE0xIuprKNegWO:-O0AA0wf7kcjqtcQiZ1G:b75pqru) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java#L25)):*

We have two other uses of BDDDomain that also need an absent value but it's just hacked in.  So I wonder if this would be cleaner if we first made a class that supports the optional emptiness, which could be used by all three cases, and then build your BDDTunnelEncapsulationAttribute as a wrapper on that?
</blockquote></details>

Yeah; considered it but I think it's better as follow on work. I think it goes with the thinking about domains with 0 or 1 values, because optional would turn into 1 or 2 values.

This one is weird since there's ABSENT and UNKNOWN, and it's not clear how general UNKNOWN is.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java` line 63 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A1q2WBiaTgRm-Y8uQ:-O0A9qmG1OYDiCyTp55_:b-oxok5a) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java#L63)):*

You're adding 2 more values so don't you just need one extra bit instead of 2?  Or is there some edge case to consider?
</blockquote></details>

Same as in BDDRoute, ABSENT and UNKNOWN are different things, so there's 2 extra values than are in the input set.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 208 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A3eaCAQH7ROC7WF2C:-O0ABkmd3Yv2K3iLxJsi:bk9hdwv) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L208)):*

Presumably we'd like to use ABSENT whenever possible, since it makes the model simpler.  And we only want to use UNKNOWN when necessary.  Will this do both of those things?  I guess maybe it will due to the careful way that you ordered the values when you created the BddTunnelEncapsulationAttribute?  If so, worth a comment there to that effect (and some tests to make sure it's not changed later).  If not, then it could be done here in the same way that certain IP addresses are preferred as defaults.
</blockquote></details>


Yeah this is done via the order. Added some comments, and added a comment to + fleshed out the existing test.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 884 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A4ssg5E-3QC3PLaE0:-O0AFs5S3ofK5MXVN7n0:b-882zn2) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L884)):*

Why not check for this and use UNKNOWN otherwise?  It seems like you're not using UNKNOWN at all currently.
</blockquote></details>

UNKNOWN has semantic meaning though, as in "a concrete value that is different than these" rather than "tell the user we don't know what's up here".

I'd rather just be forced to implement the functionality when there's new features added.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 330 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A6c6222HMI_R4nFJn:-O0A_1eqBSj2H_8qDDwr:bb2a177) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L330)):*

This means that differences in tunnel encapsulation attributes will now be considered by CRP (as I understand it), so probably worth adding a test to make sure that it does the right thing.  cc @nickgian
</blockquote></details>

Couldn't find any tests of this file, but I did add a test of `BDDTunnelAttributes#allDifferences`.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttributeTest.java` line 35 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A5VMf-RB4d4mZOZW6:-O0AGHYM7NI7-xi0eqxz:bprb1w6) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttributeTest.java#L35)):*

I don't understand this -- why would we want something with only 1 possible value to be considered equal to something with 2 possible values?
</blockquote></details>

This just lines up with the BDDDomain implementation:

```
    // Values is ignored here, because we should only be checking equality for the same input
    // domain and bit assignment.
    return _integer.equals(other._integer);
```

We can change this if we want, but (I think) the class doesn't include the values because it would be a bug to be comparing them in the first place if the domain values are different.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: 13 of 16 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java` line 25 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A2JgE0xIuprKNegWO:-O0AbXNo4kDpePA2BHEp:bi9pbca) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDTunnelEncapsulationAttribute.java#L25)):*

Yeah; considered it but I think it's better as follow on work. I think it goes with the thinking about domains with 0 or 1 values, because optional would turn into 1 or 2 values.

This one is weird since there's ABSENT and UNKNOWN, and it's not clear how general UNKNOWN is.
</blockquote></details>

(The other ones all just need ABSENT)

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: 13 of 18 files reviewed, 8 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 330 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A6c6222HMI_R4nFJn:-O0Ajkdp7K8Jig2fjf0Q:btdxu50) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L330)):*

Couldn't find any tests of this file, but I did add a test of `BDDTunnelAttributes#allDifferences`.
</blockquote></details>

I also started tests of `ModelGeneration` and tested the intersting cases for tunnel attributes there. I think that covers the code that CRP should be using, in the core libraries.

millstein (commented):
> Reviewed 1 of 16 files at r1, 3 of 3 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java` line 30 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01QwzA692FX_fGkscl:-O0BLqf6-jK7i1PW5zkw:bo84kb5) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java#L30)):*

I think that was a pragmatic choice for starting out given the large amount of old features.

But now that this functionality exists, I actually think we should crash when adding new features, so we have to add their support here.
</blockquote></details>

Seems reasonable.  But then what role does the UNKNOWN attribute play?  It seems like it will never be used because we are only allowing the analysis to proceed if all tunnel attributes in the config are ones we are able to collect.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java` line 330 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O0A6c6222HMI_R4nFJn:-O0BP_ioCqm9uE6Gyyuo:bk8913p) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/compareroutepolicies/CompareRoutePoliciesUtils.java#L330)):*

I also started tests of `ModelGeneration` and tested the intersting cases for tunnel attributes there. I think that covers the code that CRP should be using, in the core libraries.
</blockquote></details>

Thanks for doing that!  This is definitely useful, versus what we have been doing (adding tests into CRPAnswererTest).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java` line 30 at r2](https://reviewable.io/reviews/batfish/batfish/9081#-O01QwzA692FX_fGkscl:-O0BQpy19Lc7yHevAGEd:b-hls35e) ([raw file](https://github.com/batfish/batfish/blob/704eb5bbcb5e2907a7f259e8903314ec500ca707/projects/minesweeper/src/main/java/org/batfish/minesweeper/aspath/TunnelEncapsulationAttributeCollector.java#L30)):*
>But then what role does the UNKNOWN attribute play?

`RemoveTunnelEncapsulationAttribute` statement doesn't constrain the input. The intent of this value is to capture that. I don't know how to do it without creating the "UNKNOWN" value.

Would it be better if I called it `OTHER`?

dhalperi (commented):
> Reviewed 10 of 16 files at r1, 1 of 3 files at r2, 1 of 3 files at r3, 3 of 3 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9081)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #9084, opened 2024-06-24
CompareSameName: stop using JSONAssert

dhalperi:
They have released some breaking changes, and it's only used here. Do we need it?


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9084)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureEquivalenceSets.java` line 34 at r1](https://reviewable.io/reviews/batfish/batfish/9084#-O09tVF67gYcopgsF4RA:-O09tVF67gYcopgsF4RB:b-rpleas) ([raw file](https://github.com/batfish/batfish/blob/5a6d654cef71b222562d72bc8705399d2a46a0cd/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureEquivalenceSets.java#L34)):*
> ```Java
> 
>     private static boolean checkJsonStringEquals(String lhs, String rhs) {
>       // TODO: when is this not good enough?
> ```

@millstein -- all the tests pass. What tests are we missing that make the complicated JSONAssert logic needed?

millstein (commented):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9084)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @ratulm)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureEquivalenceSets.java` line 34 at r1](https://reviewable.io/reviews/batfish/batfish/9084#-O09tVF67gYcopgsF4RA:-O0GkP2KBn0MEBSElzeW:b-k648eb) ([raw file](https://github.com/batfish/batfish/blob/5a6d654cef71b222562d72bc8705399d2a46a0cd/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/collections/NamedStructureEquivalenceSets.java#L34)):*

@millstein -- all the tests pass. What tests are we missing that make the complicated JSONAssert logic needed?
</blockquote></details>

From what I can tell the use of `false` in the third argument in `assertEquals` means that it will use `LENIENT` mode, which for example ignores extra fields that one object has and the other lacks.  But I don't see why we need that leniency for our use case.

PR #9094, opened 2024-07-02
TransferBDD: clear unreachable paths earlier

dhalperi:
This is a no-op in the long run, because answerers make sure to present
reachable paths with examples. But it will improve efficiency, especially
if we later add richer input constraints, to filter during path
computation.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
@@ -630,6 +630,9 @@ public class TransferBDD {
       throw new UnsupportedOperationException(expr.toString());
     }
 
+    // Clear any paths that are not reachable from the inputs.
+    finalResults.removeIf(tr -> tr.getReturnValue().getInputConstraints().isZero());
+
     // in most cases above we have only provided the path corresponding to the predicate being true
     // so lastly, we add the path corresponding to the predicate being false
 


millstein (commented):
> Reviewed 1 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9094)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 635 at r1](https://reviewable.io/reviews/batfish/batfish/9094#-O0uV_zH4wOFwqL6v1kh:-O0uV_zH4wOFwqL6v1ki:b-pbzd8) ([raw file](https://github.com/batfish/batfish/blob/56f5587d43edabdee55bde8cc8515e88a84170dd/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L635)):*
> ```Java
> 
>     // Clear any paths that are not reachable from the inputs.
>     finalResults.removeIf(tr -> tr.getReturnValue().getInputConstraints().isZero());
> ```

I'm not sure if this is correct.  For example, suppose the guard is a disjunction of `False` and something else.  Then after processing the False predicate we will have no results due to this line of code (from the recursive call to `compute` for `False`).  So then we might just skip the processing of the second disjunct.

Note that the infeasible paths are removed in the handling of an `If` guard, which is where the above function is called.  So we are already filtering them early; this optimization would only help in cases where there are large boolean predicates that are matched on, and we reach infeasibility early in the processing of those predicates.  That seems like a pretty rare use case.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9094)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 635 at r1](https://reviewable.io/reviews/batfish/batfish/9094#-O0uV_zH4wOFwqL6v1kh:-O0uWv7K2XrumGWZfWnI:b-ul2qi5) ([raw file](https://github.com/batfish/batfish/blob/56f5587d43edabdee55bde8cc8515e88a84170dd/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L635)):*

I'm not sure if this is correct.  For example, suppose the guard is a disjunction of `False` and something else.  Then after processing the False predicate we will have no results due to this line of code (from the recursive call to `compute` for `False`).  So then we might just skip the processing of the second disjunct.

Note that the infeasible paths are removed in the handling of an `If` guard, which is where the above function is called.  So we are already filtering them early; this optimization would only help in cases where there are large boolean predicates that are matched on, and we reach infeasibility early in the processing of those predicates.  That seems like a pretty rare use case.
</blockquote></details>

Hmm, I guess the adding of the "else" result below solves the correctness issue that I was worried about.  So it's probably fine.  But I haven't completely convinced myself.

dhalperi (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9094)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 635 at r1](https://reviewable.io/reviews/batfish/batfish/9094#-O0uV_zH4wOFwqL6v1kh:-O0ueCwt4_QMuDRWjm4N:bjua40f) ([raw file](https://github.com/batfish/batfish/blob/56f5587d43edabdee55bde8cc8515e88a84170dd/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L635)):*

Hmm, I guess the adding of the "else" result below solves the correctness issue that I was worried about.  So it's probably fine.  But I haven't completely convinced myself.
</blockquote></details>

Yeah. I don't think your correctness concern holds but if there's a test you'd like .

My meta concern is that there are 6 recursive calls in here plus multiple nested loops, all of which can create unreachable `TransferReturn`s that may still be worked on later. My concern may also be unjustified tho, I should try to construct a bad case too.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9094)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 635 at r1](https://reviewable.io/reviews/batfish/batfish/9094#-O0uV_zH4wOFwqL6v1kh:-O0uuBrEC5pWoxR9uXuk:bd6zj6l) ([raw file](https://github.com/batfish/batfish/blob/56f5587d43edabdee55bde8cc8515e88a84170dd/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L635)):*

Yeah. I don't think your correctness concern holds but if there's a test you'd like .

My meta concern is that there are 6 recursive calls in here plus multiple nested loops, all of which can create unreachable `TransferReturn`s that may still be worked on later. My concern may also be unjustified tho, I should try to construct a bad case too.
</blockquote></details>

I think it's very unlikely that this optimization would ever be used in real configs.  It requires a single match predicate on its own to be logically false.  Note that we're not taking into account any prior match predicates earlier in the path.  So I don't see the motivation for it.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9094)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 635 at r1](https://reviewable.io/reviews/batfish/batfish/9094#-O0uV_zH4wOFwqL6v1kh:-O12lChB3s6M5U6cc04d:b2urdfp) ([raw file](https://github.com/batfish/batfish/blob/56f5587d43edabdee55bde8cc8515e88a84170dd/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L635)):*

I think it's very unlikely that this optimization would ever be used in real configs.  It requires a single match predicate on its own to be logically false.  Note that we're not taking into account any prior match predicates earlier in the path.  So I don't see the motivation for it.
</blockquote></details>

When I put in `assert !inputConstraints.isZero`  I found that zero TransferReturns are being created all over the place. So I started looking into when and why it was happening, and it was all over this function.

dhalperi:
I think this should not be necesary.

PR #9147, opened 2024-08-08
TransferBDD: optionally combine paths with equal results on the fly

dhalperi:
**Stack**:
- #9148
- #9147 ⬅

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java
@@ -5,9 +5,11 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -15,6 +17,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import net.sf.javabdd.BDD;
@@ -274,14 +277,20 @@ public class TransferBDD {
   }
 
   /**
-   * Produces one TransferResult per path through the given boolean expression. For most
+   * Produce {@link TransferResult} for all paths through the given boolean expression. For most
    * expressions, for example matching on a prefix or community, this analysis will yield exactly
    * two paths, respectively representing the case when the expression evaluates to true and false.
-   * Some expressions, such as CallExpr, Conjunction, and Disjunction, implicitly or explicitly
-   * contain branches and so can yield more than two paths. TODO: Any updates to the TransferParam
-   * in expr are lost currently
+   * Some expressions, such as {@link CallExpr}, {@link Conjunction}, and {@link Disjunction},
+   * implicitly or explicitly contain branches and so can yield more than two paths.
+   *
+   * <p>If {@code retainAllPaths} is {@code false}, then paths with the same output behavior (aka,
+   * {@link TransferResult} is equivalent except for the input conditions under which the path is
+   * reached) are combined (by unioning those input conditions).
+   *
+   * <p>TODO: Any updates to the TransferParam in expr are lost currently
    */
-  private List<TransferResult> compute(BooleanExpr expr, TransferBDDState state, Context context)
+  private List<TransferResult> compute(
+      BooleanExpr expr, TransferBDDState state, Context context, boolean retainAllPaths)
       throws UnsupportedOperationException {
     TransferParam p = state.getTransferParam();
     TransferResult result = state.getTransferResult();
@@ -296,7 +305,7 @@ public class TransferBDD {
     } else if (expr instanceof Not) {
       p.debug("mkNot");
       Not n = (Not) expr;
-      List<TransferResult> results = compute(n.getExpr(), state, context);
+      List<TransferResult> results = compute(n.getExpr(), state, context, retainAllPaths);
       for (TransferResult res : results) {
         TransferResult newRes = res.setReturnValueAccepted(!res.getReturnValue().getAccepted());
         finalResults.add(newRes);
@@ -312,7 +321,7 @@ public class TransferBDD {
         try {
           for (TransferResult curr : currResults) {
             BDD currBDD = curr.getReturnValue().getInputConstraints();
-            compute(e, toTransferBDDState(p.indent(), curr), context)
+            compute(e, toTransferBDDState(p.indent(), curr), context, retainAllPaths)
                 .forEach(
                     r -> {
                       TransferResult updated =
@@ -327,7 +336,7 @@ public class TransferBDD {
                       }
                     });
           }
-          currResults = nextResults;
+          currResults = retainAllPaths ? nextResults : combineTransferResults(nextResults);
         } catch (UnsupportedOperationException ufe) {
           // BooleanExpr e is not supported; ignore it but record the fact that we encountered it
           currResults.forEach(
@@ -381,7 +390,7 @@ public class TransferBDD {
           try {
             for (TransferResult curr : currResults) {
               BDD currBDD = curr.getReturnValue().getInputConstraints();
-              compute(e, toTransferBDDState(p.indent(), curr), context)
+              compute(e, toTransferBDDState(p.indent(), curr), context, retainAllPaths)
                   .forEach(
                       r -> {
                         TransferResult updated =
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dhalperi)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/9147#-O3jP-4z9INROQGKnqKn:-O3jP-4z9INROQGKnqKo:bfs2fe5) (no related file):*
Still needs tests.

___

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147)* status: all files reviewed, 3 unresolved discussions (waiting on @dhalperi and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 293 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3lVH9g9wtAkQD_CVO1:-O3lVH9g9wtAkQD_CVO2:b-vilzus) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L293)):*
> ```Java
>    */
>   private List<TransferResult> compute(
>       BooleanExpr expr, TransferBDDState state, Context context, boolean retainAllPaths)
> ```

Do you have any intuition about how different it would be if you combined paths only in the "main" `compute` function (the one that takes a list of statements)?

For correctness I assume it is the same, for performance it is more difficult to say. It depends on the VI structure, which is not fresh in my memory right now. I have a suspicion it would be a small number of paths you combine in the expr/statement `compute` functions, and the majority of reduction comes from the main `compute` - but I might be totally wrong :)

1. Doing it only in one `compute` function would reduce complexity (though no strong feelings about it), and might even end up being faster if we are wasting a lot of effort trying to combine elsewhere. Would be useful to have some measurements on how much this helps and at which point paths get combined.
2. Side-q: It is always safe to miss/forget to combine paths in one of these branches, right?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 1515 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3mCnz9CHV9WhI_OF7t:-O3mCnz9CHV9WhI_OF7u:bf9ar9t) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1515)):*
> ```Java
>     grouped.forEach(
>         (state, inputConstraints) ->
>             ret.add(replaceInputConstraint.apply(factory.orAll(inputConstraints), state)));
> ```

`orAllAndFree` works I think.
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java` line 3791 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3mEPQtEgKPJuZwkzfN:-O3mEPQtEgKPJuZwkzfO:b-xss0fc) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/TransferBDDTest.java#L3791)):*
> ```Java
>     initial =
>         new TransferBDDState(new TransferParam(routeParam, false), new TransferResult(routeParam));
>     assertThat(tbdd.computePaths(initial, manyMatchesSameResult, context, false), hasSize(2));
> ```

nice :D

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 293 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3lVH9g9wtAkQD_CVO1:-O3mSfwM7DPFizCE18M3:b-5hsbh4) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L293)):*

Do you have any intuition about how different it would be if you combined paths only in the "main" `compute` function (the one that takes a list of statements)?

For correctness I assume it is the same, for performance it is more difficult to say. It depends on the VI structure, which is not fresh in my memory right now. I have a suspicion it would be a small number of paths you combine in the expr/statement `compute` functions, and the majority of reduction comes from the main `compute` - but I might be totally wrong :)

1. Doing it only in one `compute` function would reduce complexity (though no strong feelings about it), and might even end up being faster if we are wasting a lot of effort trying to combine elsewhere. Would be useful to have some measurements on how much this helps and at which point paths get combined.
2. Side-q: It is always safe to miss/forget to combine paths in one of these branches, right?
</blockquote></details>

2 is easier to answer: I claim it's always safe to miss/forget, as long as there is only 1 main entry point to this code that has it on all exit paths. (I think this is true, a'la your question 1).

1 is a bit nuanced. Given that the way we implement `continue` is with many-deep `CallExpr` chains, I'm concerned we need this for `BooleanExpr` too. (Tho `CallExpr` does go back to the `List<Statements>` version I guess)

But, thinking about it, there were other confounding factors when I wrote this PR. Specifically, the filtering out of unreachables in `Conjunction`/`Disjunction` (#9148) turns out to have been really important. Had I realized that first, I might have seen different intermediate results on the way.

For context, your intuition is right. In my "hard profiling test case" that analyzes ~1k devices with hundreds of thousands of policies in a single go, much of the time is now being spent in combining. The tradeoff may have shifted to where these inner combines become less useful.

[That case went from (>24h) to <8m with these changes so far.]

I'd vote for looking at it again later, as this is a 2-way door.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 1515 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3mCnz9CHV9WhI_OF7t:-O3mUYEkC89E6-UnFUnv:bu9tsaf) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L1515)):*

`orAllAndFree` works I think.
</blockquote></details>

I'm really unsure of the ownership invariants around BDDs in this library. We often replace one field but leave the input the same, for example. That said, I think you're right.

I want to try to establish those invariants (and put in `assert` statements) before freeing BDDs outside the code context (function/line/etc) where they are created. Follow-on work.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147#-:-O3mcdUX0wbQxB3p8ZO0:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi and @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 293 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3lVH9g9wtAkQD_CVO1:-O3mcIHl9Xtj22z7C5DY:b-36550t) ([raw file](https://github.com/batfish/batfish/blob/c482f28b57c9dcc18b1bdcb72ddfd19812c6b2a7/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L293)):*

2 is easier to answer: I claim it's always safe to miss/forget, as long as there is only 1 main entry point to this code that has it on all exit paths. (I think this is true, a'la your question 1).

1 is a bit nuanced. Given that the way we implement `continue` is with many-deep `CallExpr` chains, I'm concerned we need this for `BooleanExpr` too. (Tho `CallExpr` does go back to the `List<Statements>` version I guess)

But, thinking about it, there were other confounding factors when I wrote this PR. Specifically, the filtering out of unreachables in `Conjunction`/`Disjunction` (#9148) turns out to have been really important. Had I realized that first, I might have seen different intermediate results on the way.

For context, your intuition is right. In my "hard profiling test case" that analyzes ~1k devices with hundreds of thousands of policies in a single go, much of the time is now being spent in combining. The tradeoff may have shifted to where these inner combines become less useful.

[That case went from (>24h) to <8m with these changes so far.]

I'd vote for looking at it again later, as this is a 2-way door.
</blockquote></details>

Sounds good, agreed this is not a blocker, we can always revisit.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147)* status: all files reviewed, 1 unresolved discussion (waiting on @millstein and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java` line 293 at r2](https://reviewable.io/reviews/batfish/batfish/9147#-O3lVH9g9wtAkQD_CVO1:-O3ntn5i3KMd_cFEI-rD:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/f3796d53b1ea2299ba72cb205f1fef3e28907cbf/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/TransferBDD.java#L293)):*

Sounds good, agreed this is not a blocker, we can always revisit.
</blockquote></details>

Done.

dhalperi (commented):
> Dismissed @nickgian from a discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9147)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #9164, opened 2024-08-14
Junos: minor tech debt improvement

dhalperi:
We added a layer of indirection before parboiled so that we could short-circuit some
prefixes and so we could keep the parboiled parsers in their own library. (Because
parboiled breaks the jacoco code coverage library).

We have an unused intermediate class from this refactor that only exists to be called by tests.
Move the tests to the right place, then delete this class.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9164)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9167, opened 2024-08-15
Junos: warn if a standard community regex is used that can match longer strings

dhalperi:
Junos community regexes are not implicitly truncated so can match longer
strings in many cases. This is risky, as unintended community matches can lead
to unexpected bad routing behaviors. Add a warning for these regexes.

Do not warn in common cases that do work even though the regex is not
truncated: as of modern Junos releases, the regex is not matched against
non-standard communities so 12345:3$, for example, is fine as no standard
community has 6 digits in the first half.

---

**Stack**:
- #9168
- #9167 ⬅

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -5594,6 +5594,16 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
       if (literalCommunity != null) {
         return new LiteralCommunityMember(literalCommunity);
       }
+      List<String> unintendedMatches = RegexCommunityMember.getUnintendedCommunityMatches(text);
+      if (!unintendedMatches.isEmpty()) {
+        warn(
+            ctx,
+            "RISK: Community regex "
+                + text
+                + " allows longer matches such as "
+                + String.join(" and ", unintendedMatches));
+      }
+      ;
       return new RegexCommunityMember(text);
     } else {
       assert ctx.sc_named() != null;
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/BUILD.bazel
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/BUILD.bazel
@@ -15,6 +15,7 @@ java_library(
         "@maven//:com_github_ben_manes_caffeine_caffeine",
         "@maven//:com_google_code_findbugs_jsr305",
         "@maven//:com_google_guava_guava",
+        "@maven//:dk_brics_automaton",
         "@maven//:org_apache_commons_commons_lang3",
     ],
 )
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
@@ -391,13 +391,6 @@ public final class JuniperConfiguration extends VendorConfiguration {
 
   private static final String FIRST_LOOPBACK_INTERFACE_NAME = "lo0";
 
-  private static String communityRegexToJavaRegex(String regex) {
-    String out = regex;
-    out = out.replace(":*", ":.*");
-    out = out.replaceFirst("^\\*", ".*");
-    return out;
-  }
-
   Configuration _c;
 
   /** Map of policy name to routing instances referenced in the policy, in the order they appear */
@@ -1062,8 +1055,7 @@ public final class JuniperConfiguration extends VendorConfiguration {
     public CommunityMatchExpr visitRegexCommunityMember(RegexCommunityMember regexCommunityMember) {
       // TODO: verify regex semantics and rendering
       return new CommunityMatchRegex(
-          ColonSeparatedRendering.instance(),
-          communityRegexToJavaRegex(regexCommunityMember.getRegex()));
+          ColonSeparatedRendering.instance(), regexCommunityMember.getJavaRegex());
     }
 
     private static final CommunityMemberToCommunityMatchExpr INSTANCE =
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java
@@ -1,5 +1,10 @@
 package org.batfish.representation.juniper;
 
+import com.google.common.collect.ImmutableList;
+import dk.brics.automaton.Automaton;
+import dk.brics.automaton.RegExp;
+import java.util.LinkedList;
+import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 import org.batfish.datamodel.bgp.community.Community;
@@ -7,6 +12,88 @@ import org.batfish.datamodel.bgp.community.Community;
 /** A {@link CommunityMember} representing a regex match condition for a {@link Community}. */
 @ParametersAreNonnullByDefault
 public final class RegexCommunityMember implements CommunityMember {
+  /**
+   * In Junos regexes, you can use {@code *} as a shortcut for {@code .*}, if it's alone. That is,
+   * {@code *:5$} is the same as {@code .*:5$} or {@code [0-9]*:5$}.
+   *
+   * <p>Handle this by converting standalone {@code *} literals to {@code .*} for use in Java
+   * regexes.
+   *
+   * <p>TODO: this likely needs more work for regexes over extended communities or malformed
+   * regexes.
+   *
+   * <p>See <a
+   * href="https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-bgp-communities-extended-communities-evaluation-in-routing-policy-match-conditions.html">Junos
+   * documentation</a>}
+   */
+  private static String communityRegexToJavaRegex(String regex) {
+    String out = regex;
+    out = out.replace(":*", ":.*");
+    out = out.replaceFirst("^\\*", ".*");
+    return out;
+  }
+
+  /**
+   * Returns a list of {@link Community} literals that may be unintended matches for the given
+   * regex. The definition of "unintended" is if it matches the given regex, but does not match if
+   * the regex is limited with {@code ^} and {@code $}.
+   *
+   * <p>Note that in modern Junos releases, you do not need to truncate "longest possible" matches
+   * like {@code 11111:*} because there is no standard-community that starts with 6 digits. So we
+   * must constrain to regex-type-specific regexes.
+   *
+   * <p>TODO: this function currently handles only standard community regexes.
+   *
+   * <p>See <a
+   * href="https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-bgp-communities-extended-communities-evaluation-in-routing-policy-match-conditions.html">Junos
+   * documentation</a>}
+   */
+  public static List<String> getUnintendedCommunityMatches(String junosRegex) {
+    if (junosRegex.split(":", -1).length != 2) {
+      return ImmutableList.of();
+    }
+    String javaRegex = communityRegexToJavaRegex(junosRegex);
+
+    // In Junos, a valid standard community regex is u16:u16.
+    //
+    // Note: this assumes a user doesn't do something like [community FOO members "no-ex.ort"]
+    // to turn a well-known community into a regex. It may not work anyway, but we haven't tested
+    // it. There's no good reason for this to happen, as community matches can mix regexes and
+    // literals, unlike many Cisco-style platforms.
+    String u16 = "((0|[1-9][0-9]*)&<0-65535>)";
+    // The user's regex may already have ^$, but we will unconditionally append them. Therefore, we
+    // must permit ^^ and $$. The automaton library does not implement ^ and $ as start/stop.
+    Automaton valid = new RegExp("^^" + u16 + ":" + u16 + "$$").toAutomaton();
+
+    // This replacement is necessary because otherwise something like .....:..... will be treated as
+    // allowing shorter communities such as 1:2 by matching, e.g., xx^^1:2$$xx. Make the . only
+    // match numbers to disallow this. This stems from the lack of ^$ support.
+    String regex = javaRegex.replaceAll("\\.", "[0-9]");
+
+    RegExp r = new RegExp(".*" + regex + ".*");
+    RegExp rClipped = new RegExp("^?^" + regex + "$$?");
+    Automaton validR = r.toAutomaton().intersection(valid);
+    Automaton validClipped = rClipped.toAutomaton().intersection(valid);
+    if (validR.equals(validClipped)) {
+      return ImmutableList.of();
+    }
+    Automaton exampleExtendingFront =
+        new RegExp(".*" + regex + "$.*").toAutomaton().intersection(valid).minus(validClipped);
+    Automaton exampleExtendingEnd =
+        new RegExp(".*^" + regex + ".*").toAutomaton().intersection(valid).minus(validClipped);
+    List<String> examples = new LinkedList<>();
+    if (!exampleExtendingFront.minus(validClipped).isEmpty()) {
+      String example = exampleExtendingFront.getShortestExample(true);
+      assert example.startsWith("^^") && example.endsWith("$$");
+      examples.add(example.substring(2, example.length() - 2));
+    }
+    if (!exampleExtendingEnd.minus(validClipped).isEmpty()) {
+      String example = exampleExtendingEnd.getShortestExample(true);
+      assert example.startsWith("^^") && example.endsWith("$$");
+      examples.add(example.substring(2, example.length() - 2));
+    }
+    return examples;
+  }
 
   public RegexCommunityMember(String regex) {
     _regex = regex;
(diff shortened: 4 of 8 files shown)


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: all files reviewed, 4 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @nickgian)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4Ntd8s8IiG30odiWFN:-O4Ntd8s8IiG30odiWFO:b-dw8dki) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L22)):*
> ```Java
>   }
> 
>   public static Optional<String> isRiskyCommunityRegex(String junosRegex) {
> ```

Add a comment.  Also the name sounds like it will return a boolean so consider renaming.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 24 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4NdvwU0MzQT5GsJhuZ:-O4NdvwU0MzQT5GsJhu_:b-tz1c17) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L24)):*
> ```Java
>   public static Optional<String> isRiskyCommunityRegex(String junosRegex) {
>     String javaRegex = communityRegexToJavaRegex(junosRegex);
>     String u16 = "((0|[1-9][0-9]*)&<0-65535>)";
> ```

This string is also defined in `CommunityVar` so maybe it can be shared?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 25 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4Ne0Kf5A3AmDdlft1k:-O4Ne0Kf5A3AmDdlft1l:bp3ah8z) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L25)):*
> ```Java
>     String javaRegex = communityRegexToJavaRegex(junosRegex);
>     String u16 = "((0|[1-9][0-9]*)&<0-65535>)";
>     String regex = javaRegex.replaceAll("\\.", "[0-9]");
> ```

Why this replacement?  I see that below you are intersecting (the automaton for) this regex with one that represents all (and only) valid communities, so won't that intersection have this same effect (and handle any other similar kinds of things)?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4NtnggDFZLNdvlxR9e:-O4NtnggDFZLNdvlxR9f:b-mulzff) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L50)):*
> ```Java
>       examples.add(example.substring(2, example.length() - 2));
>     }
>     return Optional.of(String.join(" and ", examples));
> ```

Why return a string rather than just the list of examples (and then let callers do what they want with them)?

nickgian (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown, @dhalperi, and @millstein)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 18 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OvppW8G5fySQUMvyL:-O4OvppX7Su7ezq2m3Yd:brm216t) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L18)):*
> ```Java
>     String out = regex;
>     out = out.replace(":*", ":.*");
>     out = out.replaceFirst("^\\*", ".*");
> ```

Since you are at it, would be useful to add a comment to this one too
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 27 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OyY_d4LVvtE_Gw9uL:-O4OyY_d4LVvtE_Gw9uM:b-z7ucnk) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L27)):*
> ```Java
>     String regex = javaRegex.replaceAll("\\.", "[0-9]");
> 
>     Automaton valid = new RegExp("^^" + u16 + ":" + u16 + "$$").toAutomaton();
> ```

I tried looking [here](https://www.brics.dk/automaton/doc/dk/brics/automaton/RegExp.html) but couldn't figure it out; is the meaning of `^^` the usual `^`?
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 41 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4P2Vrk3FwdSV9IuzyE:-O4P2Vrk3FwdSV9IuzyF:blzqzs6) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L41)):*
> ```Java
>     List<String> examples = new LinkedList<>();
>     if (!exampleExtendingFront.minus(validClipped).isEmpty()) {
>       String example = exampleExtendingFront.getShortestExample(true);
> ```

you can pull this out in another method since it is the same code as below.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4NtnggDFZLNdvlxR9e:-O4P9Dvv03zhk6irhwII:b1hizvn) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L50)):*

Why return a string rather than just the list of examples (and then let callers do what they want with them)?
</blockquote></details>

+1, especially if we end up with other types of violations too.
___
*[`projects/batfish/src/test/java/org/batfish/representation/juniper/RegexCommunityMemberTest.java` line 21 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4PPWrTDKi5C5pXKK7W:-O4PPWrTDKi5C5pXKK7X:b-upra0i) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/test/java/org/batfish/representation/juniper/RegexCommunityMemberTest.java#L21)):*
> ```Java
>       assertThat(maybeRisky, equalTo("not risky"));
>     }
>     for (String risky : new String[] {"123:4", "^123:4", "123:4$", "12345:5[12][34]"}) {
> ```

Is something like
`community lst members 123:4` a case we will warn on? This will not be parsed as a regular expression community by Batfish, right?

(It's fine for this code to mark this as risky)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 27 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OyY_d4LVvtE_Gw9uL:-O4QZnXj5ev4rcX_Bjxk:b-8uc5rk) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L27)):*

I tried looking [here](https://www.brics.dk/automaton/doc/dk/brics/automaton/RegExp.html) but couldn't figure it out; is the meaning of `^^` the usual `^`?
</blockquote></details>

No, `^` is treated as just a regular character by the automaton library.  This took me a bit to understand as well and is worth a comment. Dan is using two of them so that he can do the check that he wants to do.  In particular this allows him to distinguish `^30:` from `30:` because after intersecting with the automaton for valid communities, they will effectively become `^^30:` and `^.*30:` [not exactly for that second one, but close enough], so the difference is clear.  If you didn't have the double carets then they would end up being the same automaton after the intersection.

By the way we already use double carets elsewhere, in `SymbolicAsPathRegex`, for a somewhat related reason.

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4Ntd8s8IiG30odiWFN:-O4Qagj5-yJ8RYEef5y0:bni1cux) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L22)):*

Add a comment.  Also the name sounds like it will return a boolean so consider renaming.
</blockquote></details>

(Though I see why an option is like a boolean, so that's ok if you keep the return type as is.  But see later comment on that point.)

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: all files reviewed, 7 unresolved discussions (waiting on @anothermattbrown and @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 27 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OyY_d4LVvtE_Gw9uL:-O4QbocF9nNNvlhG3cfF:bfufkyn) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L27)):*

No, `^` is treated as just a regular character by the automaton library.  This took me a bit to understand as well and is worth a comment. Dan is using two of them so that he can do the check that he wants to do.  In particular this allows him to distinguish `^30:` from `30:` because after intersecting with the automaton for valid communities, they will effectively become `^^30:` and `^.*30:` [not exactly for that second one, but close enough], so the difference is clear.  If you didn't have the double carets then they would end up being the same automaton after the intersection.

By the way we already use double carets elsewhere, in `SymbolicAsPathRegex`, for a somewhat related reason.
</blockquote></details>

Whoops the last regex above should be `^^.*30:`

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: 2 of 8 files reviewed, 5 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 18 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OvppW8G5fySQUMvyL:-O4Qrqnu7XHIHRASS4e4:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L18)):*

Since you are at it, would be useful to add a comment to this one too
</blockquote></details>

Done.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 22 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4Ntd8s8IiG30odiWFN:-O4Qrrw34EliJUZ4o-H8:b2m5xif) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L22)):*

(Though I see why an option is like a boolean, so that's ok if you keep the return type as is.  But see later comment on that point.)
</blockquote></details>

Yeah moved it to `List`.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 24 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4NdvwU0MzQT5GsJhuZ:-O4O2VWG6uzAjU0yUibf:b-4zu2m9) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L24)):*

This string is also defined in `CommunityVar` so maybe it can be shared?
</blockquote></details>

That's more or less a coincidence in that this function is specific to how Juniper processes regexes and is "before" we get into VI land. Don't want to couple them as I expect them to diverge when we get deeper into Junos bugs.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 25 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4Ne0Kf5A3AmDdlft1k:-O4O2g4zC98iShUrZyQ1:bgnyz7i) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L25)):*

Why this replacement?  I see that below you are intersecting (the automaton for) this regex with one that represents all (and only) valid communities, so won't that intersection have this same effect (and handle any other similar kinds of things)?
</blockquote></details>

We have to do this change because the automaton library doesn't support `^$` correctly. Letting it stay `.` introduces collisions with the `^$`. This adds false positives because the user's regex then appears to permit shorter-than-5-digit regexes. Consider that `.....:.....` as the input could be treated as `^1234:2345$`.

See updated comment.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 27 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4OyY_d4LVvtE_Gw9uL:-O4QsqkV6Qji6MzD1aAL:b4i2g6v) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L27)):*

Whoops the last regex above should be `^^.*30:`
</blockquote></details>

What Todd said.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 41 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4P2Vrk3FwdSV9IuzyE:-O4QsV_93W4p6nzWWEMu:b-xuulq7) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L41)):*

you can pull this out in another method since it is the same code as below.
</blockquote></details>

eh. It's the same amount of code, but yeah if I end up doing more and more (probably tackle extended communities next) I'm sure we'll refactor.

For now, don't want to over-perfect this code.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4NtnggDFZLNdvlxR9e:-O4O3YHICFefPzgwrGit:b-ftd42o) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java#L50)):*

+1, especially if we end up with other types of violations too.
</blockquote></details>

This is not meant to be a super generic reusable function. It's specific to Junos semantics and kinda this use case. Even worse, if we have many types of violations the caller wouldn't know how to treat them differently -- so we'd have to change the type again.

That said, it made testing easier to return the list, so I did that. But I don't expect it to become reusable.
___
*[`projects/batfish/src/test/java/org/batfish/representation/juniper/RegexCommunityMemberTest.java` line 21 at r1](https://reviewable.io/reviews/batfish/batfish/9167#-O4PPWrTDKi5C5pXKK7W:-O4QrkLNBxTyE4NNwnFa:byemwx8) ([raw file](https://github.com/batfish/batfish/blob/6232c6f476aceee66228b08f85b3afa859eefd68/projects/batfish/src/test/java/org/batfish/representation/juniper/RegexCommunityMemberTest.java#L21)):*

Is something like
`community lst members 123:4` a case we will warn on? This will not be parsed as a regular expression community by Batfish, right?

(It's fine for this code to mark this as risky)
</blockquote></details>

This function will warn on that, but you're right that it should never be called with that argument.

nickgian (commented):
:lgtm:

> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167#-:-O4R1tG-B8jv2W4BjsIR:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @anothermattbrown and @millstein)

dhalperi (commented):
> Dismissed @millstein and @nickgian from 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9167)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #9170, opened 2024-08-16
SymbolicRegex: Dot should not match the start-of-string or end-of-string characters

millstein:
The automaton library that we use treats ^ and $ as ordinary characters, but they should not be allowed to match against the . regex, or else for instance we'd allow "^5:5$" as a valid instance of the community regex "..:..".


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9170)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @millstein)

PR #9242, opened 2024-10-09
Rename junos StaticRoute class to StaticRouteV4

SLarkworthy:
Just a quick rename CR, next CR will add a StaticRoute base class and StaticRouteV6 class and renaming in the same CR seemed noisy.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9242)* status: all files reviewed, 1 unresolved discussion (waiting on @SLarkworthy)

*[a discussion](https://reviewable.io/reviews/batfish/batfish/9242#-O8idYWg0sHBEyDDjOTj:-O8idYWhCy32mZ7p7ogn:b-h3puih) (no related file):*
Fine with it, tho historically we don't actually do this and instead we just have Foo and Foo6. Up to you.

___

PR #9243, opened 2024-10-09
fix_java_format.sh: redo to work better with pre-commit

dhalperi:
1. Pass filenames in and use it.
2. Run serially, as google-java-format already parallelizes internally. This
means that pre-commit will run all files in batches, with size controlled by
pre-commit. It would actually be ideal if it was only one batch, but the hit
is not so bad.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,11 @@ repos:
     hooks:
     -   id: java-format
         name: java-format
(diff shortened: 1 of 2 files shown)


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9243)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SLarkworthy)

PR #9339, opened 2025-02-24
TransferBDD: Support for static routes

millstein:
Adds support to the symbolic route-policy analysis (TransferBDD) for analyzing paths through a route policy that require an input route that is a static route (as opposed to a BGP route).

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
@@ -455,7 +455,7 @@ public final class BDDRoute implements IDeepCopy<BDDRoute> {
    * obtaining models would be incorrect, because it would not ensure that the well-formedness
    * constraints hold.
    *
-   * @param onlyBGPRoutes whether to constrain the input routes to only BGP routes
+   * @param onlyBGPRoutes whether to constrain the input routes to be BGP routes
    * @return the constraints
    */
   public BDD wellFormednessConstraints(boolean onlyBGPRoutes) {
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
@@ -443,10 +443,9 @@ public final class BDDRoute implements IDeepCopy<BDDRoute> {
   }
 
   /**
-   * Not all assignments to the BDD variables that make up a BDDRoute represent valid BGP routes.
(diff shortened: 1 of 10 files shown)


millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9339)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @nickgian)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 358 at r2](https://reviewable.io/reviews/batfish/batfish/9339#-OM950p_5UdCZSLSpqfv:-OM9TWV6Er3udJ9QFm4T:bb31nzs) ([raw file](https://github.com/batfish/batfish/blob/6e068439054bf70c1405bca715b5fe23674762f0/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L358)):*

IIRC you used to validate the output routes between SRP/TRP, do we still do that?
</blockquote></details>

Yes, every SRP result is validated with TRP.  And the tests in `TransferBDDTest` validate that the symbolic results agree with TRP (and so will fail if this code does not agree with TRP).
___
*[`projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java` line 3591 at r2](https://reviewable.io/reviews/batfish/batfish/9339#-OK3XBJ4CCCa0FPf5fOL:-OM9ToqE6gUu-2Y1OD_l:b-main36) ([raw file](https://github.com/batfish/batfish/blob/6e068439054bf70c1405bca715b5fe23674762f0/projects/minesweeper/src/test/java/org/batfish/minesweeper/question/searchroutepolicies/SearchRoutePoliciesAnswererTest.java#L3591)):*

I am a bit confused about this part (and the corresponding part in SearchRoutePolicies source code). Why is that the desired behavior?
TransferBDD seems to support an input static route that outputs a BGP route?
</blockquote></details>

For now I didn't update SRP or CRP because that would also require updates to their API.  For example, now an input route in a result can be sometimes a BGP route and sometimes a static route, so we'll also need a user-facing representation of a static route, etc.   So SRP and CRP still only consider route-map paths that can be taken by BGP input routes.  Once the support for static routes is more mature it would be worth considering to update those questions.

nickgian (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9339#-:-OM9V7xb6MgR9g8fb2gc:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @anothermattbrown)

PR #9357, opened 2025-03-17
Junos: Fix local-address grammar to require IP address

SirasornT:
The parser allows statement `local-address` to not have defined IP address. In the extractor, warn of fatal error if the statement exists without a value. 

`@JsonIgnore` was added due to the error `Caused by: java.lang.UnsupportedOperationException: Should never call set() on setterless property ('fatalRedFlagWarnings')` because Jackson marks fatalRedFlagWarnings as a property.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
@@ -193,7 +193,6 @@ public class Warnings implements Serializable {
   }
 
   /** Get all red flag warnings that are fatal error */
-  @JsonIgnore
   public SortedSet<Warning> getFatalRedFlagWarnings() {
     SortedSet<Warning> fatalWarnings = new TreeSet<>();
     for (Warning warning : _redFlagWarnings) {
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/common/Warnings.java
@@ -193,6 +193,7 @@ public class Warnings implements Serializable {
   }
 
   /** Get all red flag warnings that are fatal error */
+  @JsonIgnore
   public SortedSet<Warning> getFatalRedFlagWarnings() {
     SortedSet<Warning> fatalWarnings = new TreeSet<>();
     for (Warning warning : _redFlagWarnings) {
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9357)* status: 0 of 4 files reviewed, 2 unresolved discussions

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4408 at r1](https://reviewable.io/reviews/batfish/batfish/9357#-OLa7xuFCd70YwAT2NAt:-OLa7xuFCd70YwAT2NAu:btn8222) ([raw file](https://github.com/batfish/batfish/blob/341ce275c4ea2a78be8cc6852c276754aaced1ce/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4408)):*
> ```Java
>     } else if (ctx.ipv6_address() != null) {
>       // TODO: Extract ipv6 local address
>     } else {
> ```

I think this `else` should be unreachable. If it isn't, there's a bug in the grammar: https://www.juniper.net/documentation//us/en/software/junos/cli-reference/topics/ref/statement/local-address-edit-protocols-bgp.html

address is not optional in the grammar.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 4409 at r1](https://reviewable.io/reviews/batfish/batfish/9357#-OLa6EGD42YeVsdQ0pSK:-OLa6EGD42YeVsdQ0pSL:b-muupoo) ([raw file](https://github.com/batfish/batfish/blob/341ce275c4ea2a78be8cc6852c276754aaced1ce/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L4409)):*
> ```Java
>       // TODO: Extract ipv6 local address
>     } else {
>       _w.fatalRedFlag(
> ```

If you did want to keep this, I'd use an actual parse warning. That way it would have the source line and you wouldn't need to be redundant here.

SirasornT:
Revision 2: Detect the syntax error in the parsing stage by modifying the grammar file. If there is a statement `local-address`, an IP address has to be present

https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/local-address-edit-protocols-bgp.html

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9357)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

PR #9358, opened 2025-03-18
Junos: Detect bgp peer-as commit errors

SirasornT:
There are 4 cases of commit errors discovered regarding peer-as. For 1-3, see [Juniper doc](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/peer-as-edit-protocols-bgp.html#id-13310739__d671587e144)

1. EBGP peer AS must be configured
2. EBGP peer AS must not be the same as local AS
3. IBGP peer AS must not be different from different AS
4. Undefined (`ig.getType() == null`) peer AS must be configured. Juniper by default set group type to external if none is set. See [doc](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/type-edit-protocols-bgp.html).

The doc says: "Default: If you do not specify the BGP group type or assign a peer-as, then Junos OS assigns peer group type external by default." This allows peer-as of undefined type group can be both the same and different from local AS.

Commit 2:
- Initialize boolean for whether the peer is IBGP using the definition from [Juniper doc](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/peer-as-edit-protocols-bgp.html#id-13310739__d671587e144). This boolean is then used for the peer type `if` statements below for consistency


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9358)* status: 0 of 6 files reviewed, 3 unresolved discussions

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 552 at r1](https://reviewable.io/reviews/batfish/batfish/9358#-OLer2_s97lfyKq6IUZi:-OLer2_s97lfyKq6IUZj:b3n6e89) ([raw file](https://github.com/batfish/batfish/blob/5b5dc4c24d6a9d61003f694dab86b5c34ded2471/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L552)):*
> ```Java
>       Long remoteAs = ig.getType() == BgpGroupType.INTERNAL ? ig.getLocalAs() : ig.getPeerAs();
> 
>       // Undefined type defaults to external
> ```

based on your experiments, I think we need to do something slightly different.

First, we need to determine the type based on getType() and/or ASN values. This is also where we would warn/reject if both peer-as and peer type are unset.

Then the `ifs` below this can be based on the actual determined type.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 603 at r1](https://reviewable.io/reviews/batfish/batfish/9358#-OLesdWhEknTsfHv6DRC:-OLesdWhEknTsfHv6DRD:b-lqedec) ([raw file](https://github.com/batfish/batfish/blob/5b5dc4c24d6a9d61003f694dab86b5c34ded2471/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L603)):*
> ```Java
>       neighbor.setConfederation(routingInstance.getConfederation());
> 
>       boolean ibgp = Objects.equals(remoteAs, ig.getLocalAs());
> ```

Can we use the same logic about peer type for IBGP? 
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 803 at r1](https://reviewable.io/reviews/batfish/batfish/9358#-OLet43J8g2TQ8PvxrDI:-OLet43J8g2TQ8PvxrDJ:b-mti4iu) ([raw file](https://github.com/batfish/batfish/blob/5b5dc4c24d6a9d61003f694dab86b5c34ded2471/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L803)):*
> ```Java
> 
>       /* Inherit multipath */
>       if (ig.getType() == BgpGroupType.INTERNAL || ig.getType() == null) {
> ```

and we need to use the same logic down here.

I think it might be like 3 PRs.

dhalperi (commented):
> Reviewed 3 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9358)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 603 at r1](https://reviewable.io/reviews/batfish/batfish/9358#-OLesdWhEknTsfHv6DRC:-OLfdx1kCnTmtocZCRKR:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/5b5dc4c24d6a9d61003f694dab86b5c34ded2471/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L603)):*

Can we use the same logic about peer type for IBGP?
</blockquote></details>

Done.

PR #9359, opened 2025-03-18
Make Jackson ignore fatal red flag warnings

SirasornT:
`@JsonIgnore` was added due to the error `Caused by: java.lang.UnsupportedOperationException: Should never call set() on setterless property ('fatalRedFlagWarnings')` because Jackson marks fatalRedFlagWarnings as a property.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9359)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

PR #9374, opened 2025-03-31
Juniper: remove default vrf for PBR

dhalperi:
The 'default vrf' should be the routing-instance that we're already in. Otherwise, we
switch back to the master RI if the PBR policy is just 'accept'.


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9374)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9376, opened 2025-04-01
Junos: implement and test policy-statement from route-type

dhalperi:


progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9376)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/PsFromRouteType.java` line 30 at r1](https://reviewable.io/reviews/batfish/batfish/9376#-OMjX-6z6nrJ8YC-FJfA:-OMjX-6z6nrJ8YC-FJfB:b-oijn92) ([raw file](https://github.com/batfish/batfish/blob/c9eb04e375b84704b79c29170d58bc91fb6177d2/projects/batfish/src/main/java/org/batfish/representation/juniper/PsFromRouteType.java#L30)):*
I don't think so if I read https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-match-conditions-in-routing-policy-terms.html correctly 

for some reason they chose `match external` for OSPF/IS-IS but added `route-type` for BGP

> (OSPF and IS-IS only) Match IGP external routes. For IS-IS routes, the `external` condition also matches routes that are exported from one IS-IS level to another. The `type` keyword is optional and is applicable only to OSPF external routes. When you do not specify `type`, the `external` condition matches all IGP external (OSPF and IS-IS) routes. When you specify `type`, the `external` condition matches only OSPF external routes with the specified OSPF metric type. The metric type can either be 1 or 2.
> To match BGP external routes, use the `route-type` match condition.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9376)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9379, opened 2025-04-02
Junos: Don't crash on invalid wildcard in apply-path

SirasornT:
Juniper allows statements like 

```
set policy-options prefix-list FOO apply-path "protocols bgp group BAR neighbor <*>;" 
```
and even things like

```
set policy-options prefix-list FOO apply-path "protocols bgp group <*>a neighbor <*>b" 
```

These cases would not expand but will also not cause commit error so Batfish should not crash on them.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9379)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

PR #9380, opened 2025-04-02
Junos: Support group wildcard with comma

SirasornT:
Batfish is throwing exceptions when creating groups with `,` inside `<>` for lines such as `set groups g4 interfaces "<em[0,1]>" unit <*> family inet filter input FILTER`. This is valid syntax.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/parboiled/GroupWildcard.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/parboiled/GroupWildcard.java
@@ -68,7 +68,8 @@ public class GroupWildcard extends BaseParser<String> {
                 Ch('-'),
                 Ch('_'),
                 Ch(':'),
-                Ch('/'))),
+                Ch('/'),
+                Ch(','))),
         push(match()));
   }
 
--- a/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
@@ -5047,6 +5047,8 @@ public final class FlatJuniperGrammarTest {
     /* prefix-list p5 with trailing semicolon in apply-path should exist but be empty */
     assertThat(c, hasRouteFilterLists(hasKey("p5")));
     assertThat(c.getRouteFilterLists().get("p5").getLines(), empty());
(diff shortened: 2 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9380)* status: all files reviewed, 1 unresolved discussion (waiting on @SirasornT)

___
*[`projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-wildcards` line 26 at r1](https://reviewable.io/reviews/batfish/batfish/9380#-OMrloOH8PYoG4o2-cxf:-OMrloOH8PYoG4o2-cxg:bxt7296) ([raw file](https://github.com/batfish/batfish/blob/e6b3f133d09a7fc0af730650e445726079b42329/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-wildcards#L26)):*
> ```
> set firewall family inet filter FILTER term PERMIT-ALL from protocol tcp
> set firewall family inet filter FILTER term PERMIT-ALL then accept
> set groups g4 interfaces "<em[0,1]>" unit <*> family inet filter input FILTER
> ```

Odd grammar here. I wonder if a lot more is allowed in these character classes. Note that this would match `em,` but that is not a thing :)

SirasornT (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9380)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

___
*[`projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-wildcards` line 26 at r1](https://reviewable.io/reviews/batfish/batfish/9380#-OMrloOH8PYoG4o2-cxf:-OMrnIK22vPr_Wp6fxET:b-pwwtul) ([raw file](https://github.com/batfish/batfish/blob/e6b3f133d09a7fc0af730650e445726079b42329/projects/batfish/src/test/resources/org/batfish/grammar/juniper/testconfigs/juniper-wildcards#L26)):*

Odd grammar here. I wonder if a lot more is allowed in these character classes. Note that this would match `em,` but that is not a thing :)
</blockquote></details>

em0 and em1 seems to be valid interfaces https://supportportal.juniper.net/s/article/MX-Internal-interfaces-em0-and-em1-on-MX240-MX480-MX960-explained?language=en_US

PR #9395, opened 2025-04-15
InterfaceType: remove VPN in favor of TUNNEL

dhalperi:
We never used VPN intentionally and we can't think of a reason to keep it. In the meantime, VPN processing
has bitrotted compared to TUNNEL processing.


progwriter (commented):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9395)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9403, opened 2025-04-24
Add support for bandwidth extended community

SirasornT:
Add support for bandwidth, bandwidth-transitive, and bandwidth-non-transitive.

Source 
https://www.juniper.net/documentation/us/en/software/junos/ai-ml-evo/topics/topic-map/bgp-link-bandwidth.html for types of bandwidth community and default transitive behavior

https://community.juniper.net/blogs/moshiko-nayman/2024/05/13/bgp-link-bandwidth-with-junos for confirming that bandwidth value is in bytes

https://datatracker.ietf.org/doc/html/draft-ietf-idr-link-bandwidth-11#name-link-bandwidth-extended-com for parsing guideline regarding type, subtype, and value.

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
@@ -6,7 +6,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.primitives.Floats;
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
 import java.math.BigInteger;
@@ -95,7 +94,6 @@ public final class ExtendedCommunity extends Community {
     String subType = parts[0];
     String globalAdministrator = parts[1];
     String localAdministrator = parts[2];
-    boolean isBandwidth = false;
     // Try to figure out type/subtype first
     Byte typeByte = null;
     byte subTypeByte;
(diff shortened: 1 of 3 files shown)


SirasornT (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9403)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 181 at r1](https://reviewable.io/reviews/batfish/batfish/9403#-OOduHdp0GD89LcUsh-r:-OOduHdp0GD89LcUsh-s:b8xqqcx) ([raw file](https://github.com/batfish/batfish/blob/c70864584a11b64089b9fac2cefb031dcbd3e506/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L181)):*
> ```Java
>       }
>     }
>     if (isBandwidth) {
> ```

Did not want to put this here, but the alternative case is to bring all the validation from above into bandwidth()

progwriter (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9403)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SirasornT)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java` line 181 at r1](https://reviewable.io/reviews/batfish/batfish/9403#-OOduHdp0GD89LcUsh-r:-OOe1t6q97M8gks_5tUK:b-4kc16w) ([raw file](https://github.com/batfish/batfish/blob/c70864584a11b64089b9fac2cefb031dcbd3e506/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java#L181)):*
<details><summary><i>Previously, SirasornT (Sirasornt) wrote…</i></summary><blockquote>

Did not want to put this here, but the alternative case is to bring all the validation from above into bandwidth()
</blockquote></details>

so I'm gonna claim you don't need this.... and you can get away with just storing the long value.

if this was a specified as a float in string form `laLong = Long.parseUnsignedLong(localAdministrator);` would have failed on line 131, right? (if such a string value is possible, then likely a missing test :P )

I will also claim we don't have to worry about super-accurate bit representation, unless of course, you know there is a way to write a match/set policy expression in routing policy where we need to handle float values accurately.

progwriter (commented):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9403)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SirasornT)

PR #9407, opened 2025-04-28
TransferBDD: MatchPeerAddress

millstein:
Adds support for the MatchPeerAddress statement to the symbolic routing analysis.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/ConfigAtomicPredicates.java
@@ -15,6 +15,7 @@ import java.util.TreeSet;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.Ip;
 import org.batfish.datamodel.bgp.TunnelEncapsulationAttribute;
 import org.batfish.datamodel.bgp.community.StandardCommunity;
 import org.batfish.datamodel.routing_policy.RoutingPolicy;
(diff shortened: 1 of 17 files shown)


progwriter (commented):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9407)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 461 at r1](https://reviewable.io/reviews/batfish/batfish/9407#-OP1YVY19JtbTTZin2fp:-OP1YVY19JtbTTZin2fq:b-9rckzl) ([raw file](https://github.com/batfish/batfish/blob/027e53c2bd9325079a0c103c0727af07a5a5e59b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L461)):*
> ```Java
>       if (configAPs.getPeerAddresses().isEmpty()) {
>         // we are not tracking any peer addresses, so just pick a dummy value to use
>         remoteIp = Ip.parse("2.2.2.2");
> ```

Naive q: is it safe to pick a dummy value because this is equivalent to the config containing to BGP peers at all?

millstein (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9407)* status: all files reviewed, 1 unresolved discussion (waiting on @anothermattbrown and @progwriter)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java` line 461 at r1](https://reviewable.io/reviews/batfish/batfish/9407#-OP1YVY19JtbTTZin2fp:-OP1esYB0sWgor-VI8qU:bxecahe) ([raw file](https://github.com/batfish/batfish/blob/027e53c2bd9325079a0c103c0727af07a5a5e59b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java#L461)):*

Naive q: is it safe to pick a dummy value because this is equivalent to the config containing to BGP peers at all?
</blockquote></details>

If the list of peer addresses is empty it just means that the route maps never match on any peer addresses.  Hence it's safe to pick any value since the route maps' behavior does not depend on it.  But there can still be BGP peers in the config.

(Actually we can also get rid of this special case and just solve for an IP address like we do in the non-empty case.)

PR #9438, opened 2025-05-21
Validation for symbolic route analysis

millstein:
A question to validate the results of symbolic route-map analysis (TransferBDD) against concrete route simulation (TestRoutePolicies).  For each feasible path through a route map, we solve for a random input route that causes that path to be executed and compare the results from concrete simulation against what the symbolic analysis expects.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9438)* status: all files reviewed, 2 unresolved discussions (waiting on @millstein)

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/question/transferbddvalidation/TransferBDDValidationAnswerer.java` line 286 at r1](https://reviewable.io/reviews/batfish/batfish/9438#-OQz0gXu4lcuLGGuC2uE:-OQz0gXu4lcuLGGuC2uF:bcx76hp) ([raw file](https://github.com/batfish/batfish/blob/b21b875a9df53cabc76c571441eead4048fa1ac0/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/transferbddvalidation/TransferBDDValidationAnswerer.java#L286)):*
> ```Java
>     List<ColumnMetadata> columnMetadata =
>         ImmutableList.of(
>             new ColumnMetadata(COL_NODE, NODE, "The node that has the policy", true, false),
> ```

Strong suggestion that you include the seed in the output. Was invaluable when I did this for traceroute/symbolic tracerouter.

PR #9460, opened 2025-06-19
Basic tracking of "to" conditions in Junos policy-statement terms

SLarkworthy:
- Add todo's for further extraction
- Otherwise just add a flag to track if there are any "to"'s, to enable proper logging of risky terminal actions.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -583,6 +583,8 @@ import org.batfish.grammar.flatjuniper.FlatJuniperParser.Popstnh_ipv6Context;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Popstnh_peer_addressContext;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Popstnh_rejectContext;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Popstnh_selfContext;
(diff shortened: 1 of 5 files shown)


SLarkworthy (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9460)* status: all files reviewed, 1 unresolved discussion (waiting on @dhalperi)

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 4442 at r1](https://reviewable.io/reviews/batfish/batfish/9460#-OT8c8Uq0QhowVmnC6Y7:-OT8jvs5CEg5p9qyAD3h:backbgq) ([raw file](https://github.com/batfish/batfish/blob/1f9bd7dd9435a6be8495ef8f41c4b58e15040eb1/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L4442)):*

Being mildly more explicit -- I wonder if we should kill the separate `hasFrom` and `hasTo` in order to just say `hasMatch`. I worry with 2 different functions we'll spend our time whacking moles.
</blockquote></details>

thinking - it's used in two places:
one to check if there's a default term:  boolean hasDefaultTerm =  
    ps.getDefaultTerm().hasAtLeastOneFrom() || !ps.getDefaultTerm().getThens().isEmpty(); 
and one to add new `if` statements based on the froms:
if (term.hasAtLeastOneFrom()) {  
  If ifStatement = new If();
...

My hesitation here is that I'm unsure how/when we'll apply `to` terms if we ever implement them and if we'll even want to use the same strategy. It doesn't seem like we'd have many more moles to whack from leaving it separate like this.

PR #9462, opened 2025-06-20
Don't warn for an unreachable terminal term if there is exactly one simple terminal term

SLarkworthy:
Make a special case for terminal term warnings - don't worry if there is exactly one terminal term that is unreachable, as long as it's "simple" - ie, has only one "then". 

Default `reject` or `next policy` is common at the end of a policy statement - but if that `reject` is coupled with something mutating, it is more likely to be an accidentally unreachable term.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9462)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SLarkworthy)

PR #9493, opened 2025-07-30
Add label-triggered lab validation for PRs

dhalperi:
## Summary

Adds a workflow that triggers lab validation when the `lab-validation` label is applied to PRs.

## Usage

Add the `lab-validation` label to a PR. Lab validation will run and rerun automatically on new commits until the label is removed.

## Implementation

- **Trigger**: PR label events and commit pushes
- **Behavior**: Runs when `lab-validation` label is added or when commits are pushed to PRs with the label
- **Testing**: Builds Batfish JAR from PR branch and runs all available lab scenarios
- **Results**: Posts status comments to PR with workflow links

## Workflow

1. Apply `lab-validation` label to PR
2. Workflow builds Batfish JAR from PR branch
3. Runs lab validation using reusable workflow from lab-validation repository
4. Posts results as PR comment
5. Reruns steps 2-4 on new commits while label is present
6. Remove label to stop automatic execution

## Testing

After merge:
1. Add `lab-validation` label to this PR
2. Push a commit to verify rerun behavior
3. Remove label to verify execution stops

### DIFF
--- a/.github/workflows/lab-validation-pr.yml
+++ b/.github/workflows/lab-validation-pr.yml
@@ -1,9 +1,9 @@
 name: Lab Validation for PR
 
 on:
-  # Trigger when label is added to PR
+  # Trigger when label is added to PR or new commits are pushed (if label exists)
   pull_request:
-    types: [labeled]
+    types: [labeled, synchronize]
   # Trigger when someone comments on PR
   issue_comment:
     types: [created]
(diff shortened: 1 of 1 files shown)


.github/workflows/lab-validation-pr.yml
```diff
@@ -0,0 +1,172 @@
...
+        if [ "${{ github.event.action }}" = "labeled" ]; then
+          # Check if 'lab-validation' label was added
+          if [ "${{ github.event.label.name }}" = "lab-validation" ]; then
+            SHOULD_RUN="true"
+            PR_NUMBER="${{ github.event.number }}"
+            echo "🏷️ Label 'lab-validation' added to PR #$PR_NUMBER"
+          fi
+        elif [ "${{ github.event.action }}" = "synchronize" ]; then
+          # Check if PR has 'lab-validation' label when new commits are pushed
+          LABELS=$(gh api repos/batfish/batfish/pulls/${{ github.event.number }}/labels --jq '.[].name')
```
dhalperi, line 33: Is there really no other way to get labels from inside github actions?!

.github/workflows/lab-validation-pr.yml
```diff
@@ -0,0 +1,172 @@
...
+        echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
+
+  # Build Batfish JAR from the PR branch
+  build-batfish:
+    runs-on: ubuntu-latest
+    needs: check-trigger
+    if: needs.check-trigger.outputs.should_run == 'true'
+    outputs:
+      pr_ref: ${{ steps.get_pr.outputs.pr_ref }}
+      pr_title: ${{ steps.get_pr.outputs.pr_title }}
```
dhalperi, line 51: Is none of this in the input event? We have to actually pull to from the API?

.github/workflows/lab-validation-pr.yml
```diff
@@ -0,0 +1,172 @@
...
+  check-trigger:
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+      pr_number: ${{ steps.check.outputs.pr_number }}
+    steps:
+    - name: Check trigger conditions
+      id: check
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
dhalperi: You're right to question this. The GITHUB_TOKEN doesn't actually need to be explicitly declared in the env section since it's automatically available to all actions. However, declaring it explicitly is considered a best practice for clarity and doesn't introduce additional security risk - the token has the same minimal permissions either way (read access to the repository). I can remove the explicit declaration if you prefer.

.github/workflows/lab-validation-pr.yml
dhalperi: Unfortunately, GitHub Actions doesn't provide PR labels directly in the event data for 'synchronize' events. The event only contains commit information, not the current labels. The GitHub API call is the standard approach - you'll see this pattern in many popular actions like 'actions/labeler' and similar workflows that need to check current PR state.

.github/workflows/lab-validation-pr.yml
dhalperi: The PR title and branch ref aren't available in 'labeled' events (which only contain the label that was added). We need the API call to get this metadata for consistent messaging across both trigger types. This is also standard practice in GitHub workflows that need PR details beyond what's in the triggering event.

.github/workflows/lab-validation-pr.yml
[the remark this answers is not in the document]
dhalperi: Remove it.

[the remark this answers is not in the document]
dhalperi: Done! Removed all explicit GITHUB_TOKEN environment variable declarations. The token is automatically available to GitHub Actions steps without needing explicit declaration.

PR #9499, opened 2025-07-31
Fix GitHub API endpoint for PR labels

dhalperi:
The lab validation workflow was failing with a 404 error when trying to check PR labels on synchronize events.

**Issue:** The workflow was using  which doesn't exist in GitHub's API.

**Fix:** Changed to  which is the correct endpoint, as PRs are treated as issues for label operations in GitHub's REST API.

**Testing:** Verified the API call works locally:
```bash
gh api repos/batfish/batfish/issues/9494/labels --jq '.[].name'
# Returns: lab-validation  
```

This should resolve the 404 errors when the workflow runs on PR synchronize events.

### DIFF
--- a/.github/workflows/lab-validation-pr.yml
+++ b/.github/workflows/lab-validation-pr.yml
@@ -28,7 +28,7 @@ jobs:
           fi
         elif [ "${{ github.event.action }}" = "synchronize" ]; then
           # Check if PR has 'lab-validation' label when new commits are pushed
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9499)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9551, opened 2025-10-17
docs: add BDD memory management best practices

dhalperi:
Add comprehensive documentation on BDD memory management, covering
reference counting rules, ownership patterns, and performance impact.
Includes real-world examples showing order-of-magnitude improvements
from proper memory management.

---
Prompt:
```
I want to add good documentation on avoiding garbage in the BDD libraries in the right place in the docs/ subfolder. Here's my high level explanation:

---

The JavaBDD library doesn't use Java's garbage collection, instead it requires users to perform reference counting at the level of the BDDFactory that is used for all BDD operations. When references leak, BDDs cannot be removed from memory, occuping heap space and cache space and slowing down operations. Because BDDs are recursive, the number of BDDs leaked can be effectively unbounded -- and we've seen cases where adding a single missing free took a computation from 100 GiB heap overflow and running for hours, to running in milliseconds with less than 100MiB of RAM.

Here are the patterns for BDD memory management:

- base functions, like bdd.or(other), bdd.and(other) bdd.xor(other), bdd.not(), bdd.fullSatOne(), etc., create NEW references and destroy none. Both java BDD objects need to be freed.
- There are alternatives like bdd.orEq(other) which create no new references, instead changing the value of \`bdd\` to the result.
- There's also bdd.orWith(other), which is equivalent to bdd.orEq(other) followed by other.free().
- Sometimes, you may want to copy a BDD in order to create an intermediate value that you can free later (e.g., as the accumulator in a \`for\` loop). If so, use \`bdd.id()\`.

Using these correctly is extremely important. Since BDD-heavy code often accepts and creates BDDs willy-nilly, it's important to have clear documented ownership of BDD objects. It's also important for core computation loops to ensure they don't create any garbage in the process of producing their output. Sometimes, the code is a little bit more complicated as a result, but it pays off in runtime and memory benefits.

An few PRs that reduced BDD garbage and made a huge difference: https://github.com/batfish/batfish/pull/7852 and https://github.com/batfish/batfish/pull/7820. Here's one where we made the BDD factory itself not produce garbage internally: https://github.com/batfish/batfish/pull/7847.

---

Side note: We're a little bit more relaxed about this for short-lived BDDFactories, but that's more from laziness rather than a desire to do it. We should imrpove this code over time, though it's not urgent.
Side note: I think we should make sure that orEq and other functions like it gracefully handle the case when the input BDD is the same object. Should this be a crash? E.g., a.notEq(a) -- this will change the RHS as well as the left!
Side note: I was lazy about \`code font\` above, but in the docs we should be rigorous about it.

---

Can you turn this into a useful document (token concise and useful for both humans and agents) and add it to a good place in the docs/ folder? Feel free to ask me any questions about the above, read the PRs, etc.
```

### DIFF
--- a/docs/development/bdd_best_practices.md
+++ b/docs/development/bdd_best_practices.md
@@ -1,0 +1,97 @@
+# BDD Best Practices
+
+This guide covers essential practices for working with Binary Decision Diagrams (BDDs) in Batfish. BDDs are a core component of Batfish's symbolic analysis engine, and proper usage is critical for performance and correctness.
+
+## Memory Management
+
+### Overview
+
+The JavaBDD library does not use Java's garbage collection. Instead, it requires explicit reference counting at the `BDDFactory` level for all BDD operations. Proper memory management is critical for BDD-heavy code—reference leaks prevent BDDs from being freed, consuming heap and cache space, and severely degrading performance.
+
+Because BDDs are recursive structures, leaked references can create effectively unbounded memory growth. Real-world cases exist where adding a single missing `free()` call reduced memory usage from 100+ GiB (with heap overflow) and hours of runtime to under 100 MiB and milliseconds of execution time.
+
+### Reference Counting Rules
+
+#### Base Operations (Create New Reference)
+
+Base BDD operations create **new** references and destroy **none**. Both the input and output BDD objects must be freed separately.
+
+Examples: `bdd.or(other)`, `bdd.and(other)`, `bdd.not()`, `bdd.fullSatOne()`
+
+#### In-Place Operations (No New Reference)
+
+In-place operations modify the target BDD and create **no** new references. The suffix `Eq` indicates "equals" semantics.
+
+Examples: `bdd.orEq(other)`, `bdd.andEq(other)`, `bdd.notEq()`
+
+**CRITICAL**: Do not use with the same object as both operands (e.g., `a.andEq(a)`). Currently produces incorrect results; ideally the code should reject this.
+
+#### Combined Operations (Consume One Reference)
+
+Combined operations create no new references and automatically free one input. The suffix `With` indicates "consume with" semantics.
+
+Example: `bdd.andWith(other)` is equivalent to `bdd.andEq(other); other.free();`
+
+Note: `bdd.andWith(bdd)` is safe (unlike `bdd.andEq(bdd); bdd.free()` which frees `bdd` twice).
+
+#### Copying References
+
+Use `bdd.id()` to create a copy of a BDD reference. This is useful when you need an intermediate value that can be freed independently (e.g., as an accumulator in a loop).
+
+```java
+BDD accumulator = factory.zero().id();  // Create a copy we can mutate
+for (...) {
+  BDD term = ...;
+  accumulator.orWith(term);  // Updates accumulator, frees term
+}
+// Use accumulator, then free it when done
+accumulator.free();
+```
+
+### Ownership and Documentation
+
+Default assumption: callees never free BDD parameters or return values unless documented. For complex call patterns, document ownership explicitly—particularly whether the caller must free the return value.
+
+Example:
+```java
+/**
+ * @return new BDD - caller must free
+ */
+```
+
+Core computation loops must not create BDD garbage—the result is returned, intermediate values are freed.
+
+### Real-World Examples
+
+These pull requests demonstrate the impact of proper BDD memory management:
+
+1. [PR #7852](https://github.com/batfish/batfish/pull/7852) - Fixed missing `free()` calls in `Composite` transition
+2. [PR #7820](https://github.com/batfish/batfish/pull/7820) - Fixed BDD garbage in reachability fixpoint computation
+3. [PR #7847](https://github.com/batfish/batfish/pull/7847) - Fixed BDD factory to avoid internal garbage generation
+
+### Guidelines
+
+Long-lived `BDDFactory` instances (shared across an entire analysis) **must** eliminate all BDD garbage:
+
+1. Use `With` operations when consuming temporary values
+2. Use `Eq` operations when updating accumulators in-place
+3. Explicitly `free()` all intermediate BDDs before returning
+4. Use `id()` when you need independent copies for mutation
+5. Never leak BDD references from loops or recursive calls
+
+Short-lived `BDDFactory` instances can be more relaxed, but we should improve this code over time.
+
+### Common Pitfalls
+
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9551)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @anothermattbrown, @millstein, and @nickgian)

___
*[`docs/development/bdd_best_practices.md` line 19 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn5Oy-AN-b4vmbQoGu:-Obn5Oy-AN-b4vmbQoGv:btwicn7) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L19)):*
> ```Markdown
> Base BDD operations create **new** references and destroy **none**. Both the input and output BDD objects must be freed separately.
> 
> Examples:
> ```

Are all these examples needed, or are some of them redundant?
___
*[`docs/development/bdd_best_practices.md` line 31 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn5TSEEZdK3U397x6p:-Obn5TSEEZdK3U397x6q:btwicn7) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L31)):*
> ```Markdown
> 
> Examples:
> - `bdd.orEq(other)` - changes `bdd` to result of `bdd | other`, returns void
> ```

Are all these examples needed, or are some of them redundant?
___
*[`docs/development/bdd_best_practices.md` line 36 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn5UR25qWx5mrTRqkT:-Obn5UR25qWx5mrTRqkU:bol2iel) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L36)):*
> ```Markdown
> - `bdd.notEq()` - changes `bdd` to result of `!bdd`, returns void
> 
> **CRITICAL**: These operations **modify the BDD in-place**. Do not use with the same object as both operands (e.g., `a.notEq(a)`), as this will produce incorrect results.
> ```

What actually happens today? I think that ideally, the code should reject this rather than silently producing bad data, I think.
___
*[`docs/development/bdd_best_practices.md` line 43 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn5b7dC8hPZ5xMH-9Z:-Obn5b7dC8hPZ5xMH-9_:b-1cira8) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L43)):*
> ```Markdown
> 
> Examples:
> - `bdd.orWith(other)` - equivalent to `bdd.orEq(other); other.free();`
> ```

Are all these examples necessary?

Also, there's a nuance that `bdd.andWith(bdd)` is actually okay (unlike bdd.andEq(bdd); bdd.free() which should both crash, since the final state of `bdd` is freed – once.)
___
*[`docs/development/bdd_best_practices.md` line 67 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn61VE1x5_-LTxSlvE:-Obn61VE1x5_-LTxSlvF:b-k85g6i) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L67)):*
> ```Markdown
> BDD-heavy code frequently passes and creates BDD objects throughout the codebase. Clear ownership documentation is essential:
> 
> 1. **Document ownership**: Specify in Javadoc which methods take ownership of BDD parameters and which methods return BDDs that callers must free
> ```

This feels a bit heavy handed. I'd say that we should assume that callees never free objects they were called with or return, unless specifically documented.

Instead, I'd say that in complicated call patterns we should document who owns what. The return value is the important one to document -- whether the callee must free it or it may not.
___
*[`docs/development/bdd_best_practices.md` line 83 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn6Knb-3HxLklFPYk7:-Obn6Knb-3HxLklFPYk8:b-nxzto6) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L83)):*
> ```Markdown
> ### Performance Impact
> 
> Proper BDD memory management is not optional—it directly determines whether computations are feasible:
> ```

Can this section be made less redundant with the intro or eliminataed entirely? It feels wordy and token heavy.
___
*[`docs/development/bdd_best_practices.md` line 102 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn6PZaEkwdY9hcPVbI:-Obn6PZaEkwdY9hcPVbJ:b-beatvz) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L102)):*
> ```Markdown
> ### Guidelines
> 
> #### For Long-Lived BDDFactories
> ```

Do we need this section or is it redundant with the above?
___
*[`docs/development/bdd_best_practices.md` line 112 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn6Rz52pC31JIRGwnc:-Obn6Rz6AbF2KJHFR3Vb:b-vdkbu0) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L112)):*
> ```Markdown
> 5. Never leak BDD references from loops or recursive calls
> 
> #### For Short-Lived BDDFactories
> ```

This is an exception worth documenting, but can we be more concise?
___
*[`docs/development/bdd_best_practices.md` line 135 at r1](https://reviewable.io/reviews/batfish/batfish/9551#-Obn6_F26UbuAG3ATFBF:-Obn6_F26UbuAG3ATFBG:b93l2n) ([raw file](https://github.com/batfish/batfish/blob/0431f2416a0a127fcb931e985e0704881ef05b2f/docs/development/bdd_best_practices.md#L135)):*
> ```Markdown
> 2. Look for loops—ensure no garbage is created per iteration
> 3. Check recursive calls—ensure all paths free intermediate values
> 4. Test with large inputs—memory leaks become obvious at scale
> ```

drop line 4 and 5. We shouldn't need to try it out, code should be obviously correct. (We don't need to write that tho).

PR #9554, opened 2025-10-17
Add verifyRoutingPolicyReferences to finalizeConfiguration

dhalperi:
When assertions are enabled, crashes if a CallExpr or CallStatement
refers to a nonexistent routing policy. This would have caught the
issue fixed by commit 1bc0b0dd77 where PsFromPolicyStatement was
creating a CallExpr for a nonexistent policy.

---
Prompt:
```
I'd like to add a verifyRoutingPolicyReferences(c) function in finalizeConfiguration. It is styled after verifyCommunityStructures, but for RoutingPolicies. When assertions are enabled, it crashes if a CallExpr or CallStatement is present that refers to a non-existing routing policy. Crucially, it would have caught the Juniper PsFromPolicyStatement issue fixed by the most recent commit.
```

---

**Stack**:
- #9563
- #9554 ⬅

### DIFF
--- a/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifier.java
+++ b/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifier.java
@@ -1,0 +1,489 @@
+package org.batfish.datamodel.routing_policy;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.batfish.common.VendorConversionException;
+import org.batfish.datamodel.Configuration;
+import org.batfish.datamodel.routing_policy.expr.BooleanExprVisitor;
+import org.batfish.datamodel.routing_policy.expr.BooleanExprs.StaticBooleanExpr;
+import org.batfish.datamodel.routing_policy.expr.CallExpr;
+import org.batfish.datamodel.routing_policy.expr.Conjunction;
+import org.batfish.datamodel.routing_policy.expr.ConjunctionChain;
+import org.batfish.datamodel.routing_policy.expr.Disjunction;
+import org.batfish.datamodel.routing_policy.expr.FirstMatchChain;
+import org.batfish.datamodel.routing_policy.expr.HasRoute;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9554)* status: all files reviewed, 1 unresolved discussion (waiting on @abanghar)

___
*[`projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifier.java` line 275 at r2](https://reviewable.io/reviews/batfish/batfish/9554#-Oc7PXwnCSw1GTyuM-Sb:-Oc7_4u3E8AR_J84e6as:bpg049f) ([raw file](https://github.com/batfish/batfish/blob/0a9e8dd3d240269d7d9aa872ebd3a294806b1081/projects/batfish-common-protocol/src/main/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifier.java#L275)):*
<details><summary><i>Previously, abanghar wrote…</i></summary><blockquote>

Is it better to do this or call `forEach` without converting to stream `withEnvironmentExpr.getPostTrueStatements().forEach(s -> s.accept(STATEMENT_VERIFIER, arg));`?
</blockquote></details>

I asked the same question -- the answer is that this is copied from somewhere else and so it's better to copy than make a design change.
___
*[`projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifierTest.java` line 29 at r2](https://reviewable.io/reviews/batfish/batfish/9554#-Oc7RoTF6AtjYlK0c3CO:-Oc7ZvNg1qmcmip0zxNf:b4ozowa) ([raw file](https://github.com/batfish/batfish/blob/0a9e8dd3d240269d7d9aa872ebd3a294806b1081/projects/batfish-common-protocol/src/test/java/org/batfish/datamodel/routing_policy/RoutingPolicyReferencesVerifierTest.java#L29)):*
<details><summary><i>Previously, abanghar wrote…</i></summary><blockquote>

Nit: could just be declared once for all tests (either as static variable or with `@Before` method)
</blockquote></details>

Usually not true in Batfish, as each test may change it. For example, the test on the next line changes it by adding `existingPolicy`.

PR #9557, opened 2025-10-20
BDD: add iteWith method to reduce garbage

dhalperi:
Add BDD.iteWith() method that consumes both then and else parameters,
following the pattern of existing With methods like andWith, orWith, etc.

This enables more efficient BDD manipulation patterns by avoiding
intermediate BDD allocations. The implementation properly handles edge
cases where parameters may be the same object to avoid double-free.

Tests verify both correctness and proper memory management using
numOutstandingBDDs tracking.

---
Prompt:
```
Can you make a BDD#iteWith function like orAllAndFree? That might be good as a pre-step to this change, making it much more clear. Or iteWith, actually
```

---

**Stack**:
- #9560
- #9559
- #9558
- #9557 ⬅

### DIFF
--- a/projects/bdd/src/main/java/net/sf/javabdd/BDD.java
+++ b/projects/bdd/src/main/java/net/sf/javabdd/BDD.java
@@ -508,6 +508,18 @@ public abstract class BDD implements Serializable {
   public abstract BDD ite(BDD thenBDD, BDD elseBDD);
 
   /**
+   * Makes this BDD be the if-then-else of three BDDs. The {@code thenBDD} and {@code elseBDD}
+   * parameters are consumed, and can no longer be used.
+   *
+   * <p>Compare to bdd_ite and bdd_delref.
+   *
+   * @param thenBDD the 'then' BDD
+   * @param elseBDD the 'else' BDD
+   * @return the result of the if-then-else operator on the three BDDs
+   */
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9557)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @nickgian)

___
*[`projects/bdd/src/main/java/net/sf/javabdd/JFactory.java` line 265 at r1](https://reviewable.io/reviews/batfish/batfish/9557#-Oc23cfU5FKI46HE-6o8:-Oc26TQh9iUqy0RNg8Al:b-bwbvsr) ([raw file](https://github.com/batfish/batfish/blob/519ce43ea50c0f7240f5cd897a0ee798cdaada7f/projects/bdd/src/main/java/net/sf/javabdd/JFactory.java#L265)):*

For my own understanding, is the condition `thenBDD != elseBDD` necessary?

If `thenBDD == elseBDD` then wouldn't it be the case that `thenBDD == elseBDD == this` and we would skip both if statements anyway?
</blockquote></details>

I think this is necessary specifically when `this != then` but `then == else`. Like `a.ite(b,b)`. We'll free `b` 2 lines up, because `a != b`, but we can't free `b` again so we need this else.

PR #9571, opened 2025-10-25
client: remove JLine dependency and interactive mode

dhalperi:
Remove JLine terminal library and interactive mode support from the
client to reduce deployment size. Interactive mode has been deprecated
for years in favor of pybatfish.

Changes:
- Delete CommandCompleter.java (JLine-based tab completion)
- Remove JLine imports and interactive mode implementation from Client.java
- Replace interactive mode with clear error directing users to pybatfish
- Remove unused fields: ENV_HOME, HISTORY_FILE, _exit, _reader
- Update client tests to use batch mode instead of interactive mode
- Remove org.jline:jline:3.30.6 from library_deps.bzl
- Regenerate maven_install.json

Results:
- Deploy jar reduced from 69 MB to 68 MB (~1.3 MB savings)
- All tests pass (client, allinone, ref_tests)
- Interactive mode enum and exit() method retained with error messages
for API compatibility

---
Prompt:
```
So allinone is our utility for everything -- it's used for ref_Tests (see
skylark/ref_tests.bzl or so) and for normal service mode (we use
'-runclient false'). Right now, allinone depends on all the modules,
including client, but in reality the client is only needed for ref_tests.
I'd like to preserve ref_tests but otherwise simplify the code and
streamline the dependencies we deploy with the server (currently,
allinone_main_deploy.jar). I want to know a few ideas for how we can do
this and also how much this would save off of the allinone_main_deploy.jar
size and which dependencies it would remove.
```

Follow-up discussion determined that removing JLine (only used for
interactive mode, which is deprecated) would provide nearly all the
savings while being much simpler than splitting client/server binaries.


dhalperi (commented):
> @dhalperi reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9571)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9572, opened 2025-10-25
client: remove unused interactive mode functionality

dhalperi:
After removing interactive mode in batfish/batfish#9563, there remained several unused components:
- RunMode enum with only batch mode remaining
- EXIT and QUIT commands
- RunMode switches in Client constructor and run()
- RunMode parameter passing in Allinone

Remove these now-unnecessary abstractions and simplify the client to assume batch mode.

---
Prompt:
```
In the recent commit we deleted the interactive mode in client, but we mostly made things crash intead of removing them. Can we now go to delete now-unused functionality? That means commands in client that don't make sense, client runmode if there's only one value, simplifying client and runmode handling in Allinone, etc.
```

### DIFF
--- a/projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
+++ b/projects/allinone/src/main/java/org/batfish/allinone/AllInOne.java
@@ -47,12 +47,10 @@ public class AllInOne {
 
     String argString =
         String.format(
-            "%s -%s %s -%s %s",
+            "%s -%s %s",
             _settings.getClientArgs(),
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> @dhalperi reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9572)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dhalperi)

PR #9749, opened 2026-01-31
feat: add grammar validation tool and pre-commit hook

natl-set:
## Summary

Add automated grammar validation tool to ensure ANTLR grammar files comply with conventions documented in `docs/parsing/README.md`.

## Features

The grammar validation tool (`tools/validate_grammar_rules.py`) checks:

- **Parser rule naming conventions**
  - Rules must be lowercase
  - Top-level rules should use `s_` prefix
  - `_null` suffix rules must be leaf rules (don't call other rules)
  - Non-leaf rules should NOT have `_null` suffix

- **Lexer rule naming conventions**
  - Lexer rules must be ALL_CAPS with underscores
  - Fragments should use `F_CamelCase` prefix

- **NEWLINE placement**
  - NEWLINE should be at leaf level, not parent level
  - Helps with error recovery

- **LL(1) hints**
  - Warns if multiple alternatives start with same token
  - Helps identify potential LL(1) violations

- **Import structure**
  - Warns about potential circular imports

## Two modes of operation

1. **Default mode** (for pre-commit): Only shows errors
2. **Verbose mode** (`-v` or `--verbose`): Shows warnings too

## Usage

```bash
# Validate specific grammar file
python3 tools/validate_grammar_rules.py path/to/file.g4

# Validate all grammar files (verbose mode, shows warnings)
python3 tools/validate_grammar_rules.py --verbose

# Pre-commit automatically runs on .g4 files
pre-commit run validate-grammar-rules --files <file.g4>
```

## Integration

Added to `.pre-commit-config.yaml` to automatically validate all `.g4` files before commit.

## Documentation

See `tools/VALIDATE_GRAMMAR_RULES.md` for detailed usage guide and examples.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,13 @@ repos:
         language: system
         require_serial: true
         pass_filenames: false # this is a global check, only run it once
+      - id: validate-grammar-rules
+        name: validate-grammar-rules
+        entry: ./tools/validate_grammar_rules.py --changed-only
+        language: python
+        files: '\.g4$'
+        require_serial: true
+        pass_filenames: true
   - repo: https://github.com/psf/black
     rev: 25.1.0
     hooks:
--- a/docs/parsing/README.md
+++ b/docs/parsing/README.md
@@ -996,6 +996,34 @@ For the example, you can:
 
 This section is still in progress. Check back later!
 
+## Grammar Validation Tool
+
+Batfish includes an automated grammar validation tool (`tools/validate_grammar_rules.py`) that checks ANTLR grammar files against the conventions documented in this file.
+
+### What it validates
+
+- **Parser rule naming**: Rules must be lowercase
+- **`_null` suffix usage**: Leaf rules should have `_null` suffix, non-leaf rules should not
+- **NEWLINE placement**: NEWLINE should be at leaf level, not parent level
+- **Lexer rule naming**: Default mode tokens should be ALL_CAPS with underscores
+- **Mode token naming**: Mode tokens should follow `M_ModeName_TOKEN_NAME` pattern
+- **Fragment naming**: Fragments should use `F_CamelCase` naming
+- **LL(1) hints**: Warns about potential LL(1) violations
+
+### Usage
+
+```bash
+# Validate all grammar files (verbose mode shows warnings)
+python3 tools/validate_grammar_rules.py --verbose
+
+# Validate specific file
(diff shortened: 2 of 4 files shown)


dhalperi (commented):
> @dhalperi made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9749)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on natl-set).

___
*[`.pre-commit-config.yaml` line 25 at r1](https://reviewable.io/reviews/batfish/batfish/9749#-OleAfCKCvDG-wBgqDkQ:-OleAfCKCvDG-wBgqDkR:bqorx63) ([raw file](https://github.com/batfish/batfish/blob/72ab00111c45a9d30b2213889888f385877699f2/.pre-commit-config.yaml#L25)):*
> ```YAML
>         files: '\.g4$'
>         require_serial: true
>         pass_filenames: true
> ```

I really like the idea, but I have some concerns.

The main one is that I am concerned about the brittleness of this setup. Passing `--changed-only` globally implies that it doesn't really work on everything it "looks like" it should. Instead, it should work like every other check and b able to be run both differentially and globally.

Here's an idea: make 1 (optional) grammar validation tool per OS, where the tool "works correctly"  on the entire grammar for that OS,. using the `files` argument to scope it. That would let us deploy this strategy immediately for new grammars or and incrementally as we fixup old ones.

It might be good to start that for your new grammars, such as Huawei.

(And of course nothing stops us from putting all these as different vendor arguments to the same program.)
___
*[`docs/parsing/README.md` line 1007 at r1](https://reviewable.io/reviews/batfish/batfish/9749#-OleB__YCRG1G_3j8g1a:-OleB__YCRG1G_3j8g1b:b438std) ([raw file](https://github.com/batfish/batfish/blob/72ab00111c45a9d30b2213889888f385877699f2/docs/parsing/README.md#L1007)):*
> ```Markdown
> - **Parser rule naming**: Rules must be lowercase
> - **`_null` suffix usage**: Leaf rules should have `_null` suffix, non-leaf rules should not
> - **NEWLINE placement**: NEWLINE should be at leaf level, not parent level
> ```

This is not vendor-independent: for Junos, newline is at the parent level. I think in general a lot of this is going to be specific to one vendor.

natl-set:
I have two levels of checks.... errors and warnings. I went through existing grammars and found that there are 11 errors and around 60 warnings. If you want, I can raise a different PR to fix the errors first.

PR #9777, opened 2026-02-16
PaloAlto: support address object references in BGP peer-address and static-route nexthop

### LINKED ISSUE #8934: PaloAlto - parsing error when using address object as routing peer or nexthop
When using an address object as BGP peer-address (line 2 of the sample configuration) or static-route nexthop (line 4), batfish throws a parsing error.
This syntax is valid on a PaloAlto box running PAN-OS 10.2 version.

**Runnable example**
```py
TXT = """
set deviceconfig system hostname FW1
set network virtual-router VR1 protocol bgp peer-group PG1 peer PEER peer-address ip PEER-OBJECT
set network virtual-router VR1 protocol bgp peer-group PG1 peer PEER peer-address ip 10.10.10.10
set network virtual-router VR1 routing-table ip static-route ROUTE-NAME nexthop ip-address PEER-OBJECT
set address PEER-OBJECT ip-netmask 10.10.10.10
"""

bf = Session()
bf.set_network("github-bug-report")
bf.init_snapshot_from_text(TXT)
print(bf.q.parseWarning().answer())
```

When replacing the address object (PEER-OBJECT in the sample configuration) by its IP address the line is parsed without error (see line 3).
I also tried to raise the address object definition (line 5) before the routing configuration without success.

natl-set:
Fixes #8934

## Summary

PaloAlto allows using address objects (named references) as BGP peer-address or static-route nexthop, not just literal IP addresses. For example:

```
set address PEER-OBJ ip-netmask 5.4.3.2/32
set network virtual-router BGP protocol bgp peer-group PG peer PEER peer-address ip PEER-OBJ
```

This PR adds support for resolving address object references in these contexts.

## Changes

- **Grammar**: Updated `bgppgp_peer_address` and `vrrtn_ip` rules to use `interface_address_or_reference`
- **Model**: Changed `BgpPeer._peerAddress` and `StaticRoute._nextHopIp` from `Ip` to `InterfaceAddress` type
- **Builder**: Updated extraction to use `toInterfaceAddress` and track references
- **Conversion**: Added `interfaceAddressToIp` helper to resolve address object references to IPs
- **Structure tracking**: Added `BGP_PEER_ADDRESS` and `STATIC_ROUTE_NEXTHOP_IP` usage types
- **Tests**: Added test config and `testAddressObjectNexthop` test method

## Test Plan

- [x] New test `testAddressObjectNexthop` verifies address object resolution works
- [x] Existing BGP and static route tests continue to pass
- [x] All PaloAlto grammar tests pass


dhalperi (commented):
> @dhalperi reviewed 8 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9777)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on natl-set).

___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 1989 at r1](https://reviewable.io/reviews/batfish/batfish/9777#-Olcvw_D0LsUzaGtGrSu:-Olcvw_D0LsUzaGtGrSv:b-qv8mob) ([raw file](https://github.com/batfish/batfish/blob/7959625b97766ac3dfad3724898379ba1538ba76/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1989)):*
> ```Java
>     }
>     // If not resolved as object, try to parse as IP directly
>     if (address.getType() == InterfaceAddress.Type.IP_ADDRESS) {
> ```

Why is this reachable? Looks like it will be handled in the function above.

```
        return switch (address.getType()) {
          case IP_ADDRESS ->
              ConcreteInterfaceAddress.create(Ip.parse(addressText), Prefix.MAX_PREFIX_LENGTH);
          case IP_PREFIX -> ConcreteInterfaceAddress.parse(addressText);
          case REFERENCE ->
              // Assume warning is surfaced in undefined references or in conversion to concrete
              // addr
              null;
```

dhalperi (commented):
> @dhalperi reviewed 2 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9777)* status: all files reviewed, 2 unresolved discussions (waiting on natl-set).

___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 1985 at r1](https://reviewable.io/reviews/batfish/batfish/9777#-Olcw96YE2zv0Uqfesjq:-Olcw96YE2zv0Uqfesjr:bg4zri8) ([raw file](https://github.com/batfish/batfish/blob/7959625b97766ac3dfad3724898379ba1538ba76/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1985)):*
> ```Java
>     ConcreteInterfaceAddress concreteAddress =
>         interfaceAddressToConcreteInterfaceAddress(address, vsys, _w);
>     if (concreteAddress != null) {
> ```

I think we probably need to check the netmask here... should be /32.

natl-set (commented):
## Addressed Review Comments

I've addressed both review comments in the latest commit (fb38ad63da):

### 1. Redundant IP_ADDRESS check (line 1989)
You're right - the `if (address.getType() == InterfaceAddress.Type.IP_ADDRESS)` check was redundant since `interfaceAddressToConcreteInterfaceAddress` already handles all three types (IP_ADDRESS, IP_PREFIX, REFERENCE). I've removed that code and simplified `interfaceAddressToIp` to just call `interfaceAddressToConcreteInterfaceAddress` and handle the result.

### 2. /32 netmask check (line 1985)
Added the netmask validation in `interfaceAddressToIp` to ensure the resolved address is a host address (/32):
```java
if (concreteAddress.getPrefix().getPrefixLength() != Prefix.MAX_PREFIX_LENGTH) {
    _w.redFlagf("Address %s has non-/32 mask %s, expected a host address", ...);
    return null;
}
```

### Additional fixes
- Added try-catch around `addrObject.toConcreteInterfaceAddress(_w)` to handle `AssertionError` for invalid address object formats
- Added skip logic for static routes with failed nexthop resolution when no fallback exists
- Added test `testAddressObjectNexthopWarnings` to verify warning scenarios (undefined references, non-/32 masks)

natl-set (commented):
## Addressed Review Comments

I've addressed both review comments in the latest commit (83bae8485c):

### 1. Redundant IP_ADDRESS check (line 1989)
You're right - the `if (address.getType() == InterfaceAddress.Type.IP_ADDRESS)` check was redundant since `interfaceAddressToConcreteInterfaceAddress` already handles all three types (IP_ADDRESS, IP_PREFIX, REFERENCE). I've removed that code and simplified `interfaceAddressToIp` to just call `interfaceAddressToConcreteInterfaceAddress` and handle the result.

### 2. /32 netmask check (line 1985)
Added the netmask validation in `interfaceAddressToIp` to ensure the resolved address is a host address (/32):
```java
if (concreteAddress.getPrefix().getPrefixLength() != Prefix.MAX_PREFIX_LENGTH) {
    _w.redFlagf("Address %s has non-/32 mask %s, expected a host address", ...);
    return null;
}
```

### Additional fixes
- Added try-catch around `addrObject.toConcreteInterfaceAddress(_w)` to handle `AssertionError` for invalid address object formats
- Added skip logic for static routes with failed nexthop resolution when no fallback exists
- Added test `testAddressObjectNexthopWarnings` to verify warning scenarios (undefined references, non-/32 masks)

dhalperi (commented):
> @dhalperi reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9777)* status: 8 of 11 files reviewed, 4 unresolved discussions (waiting on natl-set).

___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 1945 at r3](https://reviewable.io/reviews/batfish/batfish/9777#-OldtmDrBkqopAsNINhS:-OldtmDs-o2UIZ_3rEFF:bjz5ih) ([raw file](https://github.com/batfish/batfish/blob/83bae8485c011549c4d52641af121461621d5d03/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L1945)):*
> ```Java
>       try {
>         concreteIfaceAddr = addrObject.toConcreteInterfaceAddress(_w);
>       } catch (AssertionError e) {
> ```

What's going on here? This seems unwise.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 2916 at r3](https://reviewable.io/reviews/batfish/batfish/9777#-Oldu4oB1kP0SsS6M-Pd:-Oldu4oB1kP0SsS6M-Pe:buurdx3) ([raw file](https://github.com/batfish/batfish/blob/83bae8485c011549c4d52641af121461621d5d03/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L2916)):*
> ```Java
>       Ip nextHopIp =
>           interfaceAddressToIp(sr.getNextHopIp(), _virtualSystems.get(DEFAULT_VSYS_NAME));
>       // Skip if nexthop IP resolution failed and no interface/nextVr specified
> ```

You're duplicating the inline logic in `.setNextHop` below. Find a way to do it once -- plausibly, a helper method that computes the next hop given a staticRoute and a System like `static Optional<NextHop> staticRouteNextHop(sr, vs)` that returns Optional.empty for this case, and then you continue.

natl-set:
Oh you're right. I forgot about that. There is definitely a better way to handle that.

natl-set:
Should be all set? At first, I was trying to minimize the change. Now I refactored it more extensively. Btw, I wonder about the Vsys stuff. Do we need to resolve it from the default Vsys? Is there a case where other vsys should be used?

dhalperi (commented):
Yeah I'm also worried about the vsys. Have you had a chance to look?

> @dhalperi partially reviewed 1 file, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9777#-:-OmSoOPx-oMLiaxFJIZN:bm4hbx1)* status: 9 of 12 files reviewed, 2 unresolved discussions (waiting on natl-set).

natl-set:
I don't know how to test that properly but I tried to take a look at how vsys is resolved. I hope this is the right way to do it.

gh-ajinks:
We've been running into both #8934 and a couple of related gaps in production PAN-OS configs. Sharing our findings in case they're worth folding in here.

**1. Edge case: address object named after a bare IPv4 literal**

We have configs where an address object name is identical to a bare IP:

```
set address 10.8.0.2 ip-netmask 10.8.0.2/29
set ... peer-address ip 10.8.0.2
```

With the current `interfaceAddressToIp()` logic, the lookup hits the address object map first, finds the /29 object, fails `isValidHostAddress()`, and returns null — peer loses its address and the session is dropped.

A guard that tries `Ip.tryParse(addressText)` first (use the literal if it parses cleanly, fall back to object lookup only if it doesn't) would handle this without changing the happy path.

**2. Address object references in `local-address ip` (same gap as peer-address)**

The grammar already supports the `interface` sub-field of `local-address` via `bgppgp_la_interface`, but `bgppgp_la_ip` still uses `interface_address` (requires a literal `ip/prefix`). If an address object is referenced in `local-address ip`, the parser fails the same way peer-address did before this PR.

The fix mirrors what you've done here — change `bgppgp_la_ip` from:
```antlr
bgppgp_la_ip : IP addr = interface_address ;
```
to:
```antlr
bgppgp_la_ip : IP addr = interface_address_or_reference ;
```
and apply the same `interfaceAddressToIp()` resolution in `exitBgppgp_la_ip`. Happy to open a follow-up issue if you'd prefer to keep this PR scoped, but since the builder plumbing is already being added here it seemed worth raising.

gh-ajinks:
I've opened #9966, which rebases this onto current master and carries it forward: it adds `local-address ip` object support (the remaining BGP address position) and resolves interface-subnet-masked objects like `10.0.0.1/29` to their host IP. Your commits are credited via `Co-authored-by`. Opened a fresh PR rather than pushing here since this branch has been quiet a while and had drifted from master — happy to go whichever way you and the maintainers prefer.

dhalperi:
Tanks, merged in #9966

PR #9823, opened 2026-03-04
Remove Ip.AUTO sentinel value and use null for unset IPs

dhalperi:
Replace artificial sentinel values with proper null semantics
throughout the codebase. The sentinel value Ip.AUTO (created as
Ip.create(-1L)) was used to represent "no IP address" while
satisfying @Nonnull contracts. This led to confusing code with
double-null checks and artificial sentinels propagating through
APIs.

Key changes:

1. API signatures made nullable:
- AbstractRoute.getNextHopIp(): @Nonnull Ip -> @Nullable Ip
- AbstractRouteBuilder.getNextHopIp(): @Nonnull Ip -> @Nullable Ip
- BgpAdvertisement.getOriginatorIp(): @Nonnull Ip -> @Nullable Ip
- IpExpr.evaluate(): Ip -> @Nullable Ip
- PrefixExpr.evaluate(): Prefix -> @Nullable Prefix

2. NextHop made Comparable<NextHop>:
- Implemented strict total ordering: Discard < Ip < Interface <
Vrf < Vtep
- Type-specific comparison within each NextHop type
- Route comparators now use getNextHop() instead of nullable
getNextHopIp()

3. Builder behavior:
- AbstractRouteBuilder.setNextHopIp(null) clears _nextHop
(sets to null)
- Callers that want discard use the explicit
setNextHop(NextHopDiscard.instance()) API
- EigrpExternalRoute.toBuilder() uses setNextHop(getNextHop())
to preserve non-IP next hops correctly

4. Ip validation strengthened:
- Ip.create() now validates both bounds: ipAsLong >= 0L &&
ipAsLong <= 0xFFFFFFFFL
- Prevents creation of invalid IPs like create(-1L)
- Deleted Ip.valid() method (always returned true after
validation fix)

5. Deleted sentinel constants:
- Ip.AUTO
- Route.UNSET_ROUTE_NEXT_HOP_IP
- BgpAdvertisement.UNSET_ORIGINATOR_IP

6. Updated 103+ usage sites:
- Comparisons: .equals(UNSET_ROUTE_NEXT_HOP_IP) -> == null
- Builders: .setNextHopIp(UNSET_ROUTE_NEXT_HOP_IP) ->
.setNextHopIp(null)
- Tests: equalTo(UNSET_ROUTE_NEXT_HOP_IP) -> nullValue()
- Removed firstNonNull(x, Ip.AUTO) patterns
- Removed Ip.AUTO guards from NextHop constructors
- Removed TODO assertions in FibImpl, FibForward,
PrefixTrieMultiMap, MockFib

JSON output now shows proper null instead of "AUTO/NONE(-1l)" for
routes without next hop IPs (e.g., discard routes, VRF hops).

----

Prompt:
```
I want to get rid of Ip.AUTO. It's a placeholder for null that we
used for dumb reasons 8 years ago. This will be a very large change
as there are countless places we use Ip.AUTO or
Route.UNSET_NEXT_HOP_IP = Ip.AUTO when in reality just null would be
fine. Analyze the entire codebase and make an accounting of the
different classes of problems we see. You'll see many repeats
because much of the code is copied and pasted around or using
similar patterns.

Pay attention to @Nullable, @Nonnull, @ParametersAreNonNullByDefault
annotations as these indicate places we need to track down callers /
usages of properties or functions that might now be null or be
accepting null arguments, etc.
```

Followup: fixed setNextHopIp(null) to clear _nextHop instead of
creating NextHopDiscard, and updated all callers that intended
discard to use the explicit setNextHop(NextHopDiscard.instance())
API.


progwriter (commented):
> @progwriter reviewed 93 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9823)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9824, opened 2026-03-08
BGP: reset local preference and weight in post-export

### LINKED ISSUE #9262: EBGP local preference anomaly
**Describe the bug and expected behavior**
I have two Cisco IOS router configurations. Router R1 is in AS 40 and router R2 is in AS 30. R1 sends a route 100.0.0.0/8 to R2 after setting the local preference to 50. According to EBGP rules, when the route is installed in R2's BGP RIB, it should reset the local preference to 100. But, when I check the BGP RIBS using the Batfish question **bf.q.bgpRib**, I see that the route installed in R2 still has a local preference of 50.

**Runnable example**

Here are the two configurations:
R1:
```
!
version 15.2
service timestamps debug datetime msec
service timestamps log datetime msec
!
hostname R1
!
boot-start-marker
boot-end-marker
no aaa new-model
no ip icmp rate-limit unreachable
ip cef
no ip domain lookup
ip domain name lab.local
no ipv6 cef
multilink bundle-name authenticated
ip tcp synwait-time 5
!
interface eth0/1
 ip address 3.0.0.3 255.0.0.0
interface eth0/2
 ip address 100.0.0.1 255.0.0.0

route-map RM1 permit 10
 set local-preference 50

router bgp 400
 neighbor 3.0.0.2 remote-as 512
 neighbor 100.0.0.2 remote-as 400
 network 100.0.0.0
 neighbor 3.0.0.2 route-map RM1 out
```

R2:
```
!
version 15.2
service timestamps debug datetime msec
service timestamps log datetime msec
!
hostname R2
!
boot-start-marker
boot-end-marker
no aaa new-model
no ip icmp rate-limit unreachable
ip cef
no ip domain lookup
ip domain name lab.local
no ipv6 cef
multilink bundle-name authenticated
ip tcp synwait-time 5
!
interface eth0/1
 ip address 3.0.0.2 255.0.0.0

router bgp 512
 neighbor 3.0.0.3 remote-as 400
```

Save these two configs under the directory **snapshot/configs**. Next, run the following the Python code. I saved it as test.py

```py
from pybatfish.datamodel import *
from pybatfish.datamodel.answer import *
from pybatfish.datamodel.flow import *
import pandas as pd

bf = Session()
NETWORK_NAME = 'example'
SNAPSHOT_NAME = 'example_snapshot'
SNAPSHOT_PATH = 'snapshot'

bf.set_network(NETWORK_NAME)
bf.init_snapshot(SNAPSHOT_PATH, name=SNAPSHOT_NAME, overwrite=True)

df1 = bf.q.bgpRib(nodes='R1').answer().frame()
df2 = bf.q.bgpRib(nodes='R2').answer().frame()

newdf = (pd.concat([df1, df2])).reset_index(drop=True)
newdf = newdf[newdf['Network'] == '100.0.0.0/8'][["Node", "AS_Path", "Local_Pref"]].reset_index(drop=True)
 
print(new_df)
```

I got the following output:
```bash
$ python3 test.py
  Node  AS_Path Local_Pref
0   r1                 100
1   r2      400         50
```

dhalperi:
Local preference and weight are non-transitive BGP attributes that
should not be propagated to peers. Previously, both were cleared in
transformBgpRoutePreExport, before the export policy ran. If the
export policy set either attribute (e.g., via `set local-preference`
or `set weight`), the value would persist into the exported route.

For local preference, this meant an EBGP peer would install a route
with the sender's local-pref instead of the default (100).

Move both resets to transformBgpRoutePostExport so they run after
the export policy. Local-pref is preserved for IBGP and
within-confederation sessions via an `advertiseUnchangedLocalPref`
parameter. Weight is always cleared since it is never carried in
BGP UPDATE messages.

Fixes https://github.com/batfish/batfish/issues/9262

----

Prompt:
```
Fix https://github.com/batfish/batfish/issues/9262
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
@@ -85,8 +85,6 @@ public final class BgpProtocolHelper {
     builder.setProtocol(outgoingProtocol);
     builder.setSrcProtocol(routeProtocol);
 
-    // Clear a bunch of non-transitive attributes
-    builder.setWeight(0);
     if (!(route instanceof EvpnRoute<?, ?>)) {
       // These attributes are constants for EVPN routes and cannot be set
       builder.setNonRouting(false);
@@ -187,12 +185,6 @@ public final class BgpProtocolHelper {
     if (!localSessionProperties.advertiseUnchangedMed() && !routeOriginatedLocally) {
       builder.setMetric(0);
     }
-
-    // Local preference: only transitive for iBGP or within a confederation
-    builder.setLocalPreference(
-        localSessionProperties.advertiseUnchangedLocalPref()
-            ? route.getLocalPreference()
-            : DEFAULT_LOCAL_PREFERENCE);
 
     return builder;
   }
@@ -419,6 +411,7 @@ public final class BgpProtocolHelper {
     transformBgpRoutePostExport(
         routeBuilder,
         ourSessionProperties.isEbgp(),
+        ourSessionProperties.advertiseUnchangedLocalPref(),
         af.getAddressFamilyCapabilities().getSendCommunity(),
         af.getAddressFamilyCapabilities().getSendExtendedCommunity(),
         ourSessionProperties.getConfedSessionType(),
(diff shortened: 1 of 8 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9824)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on progwriter).

PR #9836, opened 2026-03-27
Ip: store as int instead of long for memory savings

dhalperi:
Change Ip's internal representation from long to int, treating it
as an unsigned 32-bit value. This shrinks each Ip instance from 24
to 16 bytes (33% reduction) by fitting the int into the alignment
padding gap after the compressed class pointer.

Public API is preserved: create(long) and asLong() still work via
Integer.toUnsignedLong/compareUnsigned. Added create(int) overload
and getBitAtPosition(int) member method. Simplified several methods
that no longer need long masking (inverted, isValidNetmask1sLeading,
numSubnetBits, hashCode). Migrated PrefixTrieMultiMap to use new
member getBitAtPosition. Simplified Prefix.SerializedForm to use
Ip.create(int) directly.

----

Prompt:
```
Suppose I changed Ip to store an int instead of a long, but then
treated it like an unsigned int in all places we have to
(comparator, etc). I think guava has helpers for this if not other
places. Would that save memory (use //tools/jol)? How would it
impact performance?
```

---
**Stack**:
- #9840
- #9839
- #9838
- #9837
- #9836 ⬅
---

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/Ip.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/Ip.java
@@ -19,7 +19,7 @@ public class Ip implements Comparable<Ip>, Serializable {
 
   // Soft values: let it be garbage collected in times of pressure.
   // Maximum size 2^20: Just some upper bound on cache size, well less than GiB.
-  //   (8 bytes seems smallest possible entry (long), would be 8 MiB total).
+  //   (4 bytes seems smallest possible entry (int), would be 4 MiB total).
   private static final LoadingCache<Ip, Ip> CACHE =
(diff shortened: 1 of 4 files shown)


SLarkworthy (commented):
:lgtm:

> @SLarkworthy reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9836#-:-OozyHFz1BfTeeWEItJL:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9837, opened 2026-03-27
AbstractRoute: store admin and tag as int for memory savings

dhalperi:
Change _admin and _tag from long to int in AbstractRoute, treating
them as unsigned 32-bit values. This shrinks every route instance by
8 bytes (e.g., Bgpv4Route 80 -> 72 bytes, measured with JOL). A new
_hasTag boolean distinguishes UNSET_ROUTE_TAG (-1L) from MAX_TAG
(0xFFFFFFFFL), which both map to int -1. The boolean fits in
alignment padding so costs no additional memory.

Both fields are now private. Subclasses use getAdministrativeCost()
and getTag() for all access, eliminating the risk of accidental
sign-extension when passing the raw int to a long parameter.

Also: remove redundant EigrpRoute._admin that shadowed the base
class field, fix ConnectedRoute/KernelRoute JsonCreators that used
int instead of long for admin, add tag range validation in the
constructor, and fix a test that used a 37-bit tag value.

----

Prompt:
```
The most recent commit used JOL analysis to shrink Ip in memory
representation by adding a bit of complexity internal to that
class (ints as UnsignedInts, etc.).

Look at Prefix and all variants of AbstractRoute to see if there
are similar opportunities
```

---
**Stack**:
- #9840
- #9839
- #9838
- #9837 ⬅
---

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/AbstractRoute.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/AbstractRoute.java
@@ -38,10 +38,13 @@ public abstract class AbstractRoute implements AbstractRouteDecorator, Serializa
   static final String PROP_TAG = "tag";
 
   protected final @Nonnull Prefix _network;
-  protected final long _admin;
+  // Stored as unsigned 32-bit int for memory savings. Use getAdministrativeCost() to read.
+  private final int _admin;
   private final boolean _nonRouting;
   private final boolean _nonForwarding;
-  protected final long _tag;
+  private final boolean _hasTag;
+  // Stored as unsigned 32-bit int for memory savings. Use getTag() to read.
+  private final int _tag;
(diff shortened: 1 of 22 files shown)


ritiagrawal (commented):
> @ritiagrawal reviewed 22 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9837)* status: 0 of 22 files reviewed, all discussions resolved (waiting on ritiagrawal).

dhalperi (commented):
> @dhalperi reviewed 22 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9837)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9840, opened 2026-03-27
BgpRoute: store pathId as int with 0 sentinel for null

dhalperi:
Replace the boxed @Nullable Integer _pathId with a primitive int,
using 0 as the sentinel for null (path IDs are generated starting
at 1, so 0 is never a valid value). This eliminates both the 8-byte
object reference and the ~16-byte Integer object per route.

JOL: Bgpv4Route shrinks from 72 to 64 bytes. The int slots into
alignment padding that was previously wasted, achieving zero padding
loss (1 byte internal only).

The public API (getPathId() returning @Nullable Integer) is
preserved; the getter boxes on return only when non-null.

----

Prompt:
```
Convert BgpRoute's boxed @Nullable Integer _pathId to a primitive
int with 0 as sentinel for null, eliminating the object reference
and Integer allocation per route.
```

---
**Stack**:
- #9840 ⬅
---

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/BgpRoute.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/BgpRoute.java
@@ -496,7 +496,8 @@ public abstract class BgpRoute<B extends Builder<B, R>, R extends BgpRoute<B, R>
   static final String PROP_WEIGHT = "weight";
 
   protected final @Nonnull BgpRouteAttributes _attributes;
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9840)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9857, opened 2026-04-02
Add missing readResolve() to Serializable singletons and cached classes

dhalperi:
Many Serializable classes with INSTANCE singletons or LoadingCache
interning lacked readResolve(), causing deserialization to bypass
the cache or create duplicate singleton instances. Add readResolve()
to all 33 such classes across the datamodel package, following the
existing pattern used by Ip, Prefix, EmptyIpSpace, etc.

For cache-interned classes (Ip6, Prefix6), readResolve re-interns
through the cache. For pure singletons, it returns INSTANCE. For
acl FalseExpr/TrueExpr which have a parameterized variant, it
conditionally returns INSTANCE only when traceElement is null.

----

Prompt:
```
I would expect that every serializable singleton or cached object
has special handling for serialization to ensure that it comes back
out as cached/singleton. Look for classes that have properties like
INSTANCE or Guava Cache or Caffeine Cache. Audit this package for
all such classes and tell me the state of all of them in a table.
[Then: Fix all 36 issues reported]
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/FibNullRoute.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/FibNullRoute.java
@@ -1,5 +1,7 @@
 package org.batfish.datamodel;
 
+import java.io.ObjectStreamException;
+import java.io.Serial;
 import org.batfish.datamodel.visitors.FibActionVisitor;
 
 /** A {@link FibAction} that discards a packet. */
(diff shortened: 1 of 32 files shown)


dhalperi (commented):
> @dhalperi reviewed 32 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9857)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9858, opened 2026-04-03
Driver: remove broken thread timeout mechanism from runBatfish

dhalperi:
runBatfish() spawned an inner thread solely to support a timeout via
Thread#join(maxRuntimeMs) + Thread#stop(). This was broken:
- maxRuntimeMs defaulted to 0, meaning join() waits forever
- Thread#stop() was deprecated in Java 1.2 and removed in Java 18+
- The code's own comment acknowledged this ("AF: This doesn't do what
you think it does")

The inner thread was unnecessary because runBatfishThroughService()
already runs runBatfish() on its own thread. Remove the thread wrapper
and run the batfish work inline. Deprecate the maxruntime setting.

----

Prompt:
```
Can you analyze the threading logic in Driver and rwwrite whatever
crazy code is there? It seems over engineered, using maybe outdated
work, and using Thread#stop which we know doesn't work. I don't even
know/remember what this code is for.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/config/Settings.java
+++ b/projects/batfish/src/main/java/org/batfish/config/Settings.java
@@ -52,8 +52,6 @@ public final class Settings extends BaseSettings implements GrammarSettings {
 
   private static final String ARG_MAX_PARSE_TREE_PRINT_LENGTH = "maxparsetreeprintlength";
 
-  private static final String ARG_MAX_RUNTIME_MS = "maxruntime";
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9858)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9865, opened 2026-04-09
Inline *MatchersImpl classes into their *Matchers counterparts

dhalperi:
Eliminate the FooMatchers/FooMatchersImpl split pattern across the
test codebase. Each MatchersImpl's inner classes are moved into the
corresponding Matchers file as private static final classes, and the
MatchersImpl files are deleted.

Also:
- Fix factory methods that returned concrete inner class types (e.g.,
HasBandwidth) to return Matcher<T> instead, since the inner classes
are now private.
- Add missing factory methods to ConfigurationMatchers,
HeaderSpaceMatchers, InterfaceMatchers, OspfProcessMatchers, and
VrfMatchers for methods previously only accessible via
DataModelMatchers.
- Update DataModelMatchers to delegate to typed Matchers rather than
constructing inner classes directly.
- Update BUILD.bazel glob in common/bdd to remove MatchersImpl
pattern.

----

Prompt:
```
Look through the matchers in this project. Sometimes we have
FooMatcher and FooMatchersImpl or something like that. is there any
good reason or can we just inline them? I think matchers compose
naturally via the public API anyway, there's no better way to test
them, etc.
```
(followed by "do it")

---

**Stack**:
- #9866
- #9865 ⬅


dhalperi (commented):
> @dhalperi reviewed 189 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9865)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9868, opened 2026-04-10
Juniper: fix forwarding-table export default action to accept

dhalperi:
On Junos, the default forwarding-table export action is accept: routes
that fall through the policy without a terminal action (accept/reject)
are installed in the FIB. Batfish was incorrectly defaulting to reject,
which meant a policy with only `then load-balance per-packet` (no
explicit accept) would reject all routes from the FIB.

Fix by wrapping the user's policy in a generated policy that sets
SetDefaultActionAccept before calling the user's policy, matching the
pattern used for BGP import policies.

----

Prompt:
```
We recently added support for junos forwarding-table export. I learned
that a policy could only have one term with "then load-balance
per-packet" and no accompanying "then accept". It seems that this is
treated as an accept even without an accompanying explicit accept, but
I think if there were an accompanying then reject it would get
rejected.
```

Further investigation of the Junos routing policy manual confirmed that
`load-balance per-packet` is an attribute manipulation action (not a
flow control action), so it does not terminate policy evaluation. The
default forwarding-table export action is accept, meaning routes that
fall through without a terminal action are accepted into the FIB.

### DIFF
--- a/docs/conversion/juniper_routing_policy.md
+++ b/docs/conversion/juniper_routing_policy.md
@@ -1,0 +1,21 @@
+# Juniper Routing Policy Conversion
+
+## Default Fall-Through Actions
+
+When wiring a Juniper `policy-statement` to a VI context, always
+consider the **default fall-through action** for that context.
+
+In the VI model, `RoutingPolicy.call()` returns
+`Environment.getDefaultAction()` when a policy falls through without a
+terminal action. The `Environment` defaults to `false` (reject). This is
+correct for contexts where Junos defaults to reject (e.g., OSPF export),
+but wrong for contexts where Junos defaults to accept.
+
+For accept-default contexts, wrap the user's policy in a generated
+policy that sets `SetDefaultActionAccept`. See
+`generatedBgpPeerImportPolicyName` (BGP import) and
+`generatedFibExportPolicyName` (forwarding-table export) for examples.
+
+Junos default actions per context are documented in the Junos routing
+policy manual, Table 4 ("Default Import and Export Policies for
+Protocols").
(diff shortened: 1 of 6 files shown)


dhalperi:
*[`docs/conversion/juniper_routing_policy.md` line 4 at r1](https://reviewable.io/reviews/batfish/batfish/9868#-OptgBZlAKPoaDJQ_auu:-OptgBZlAKPoaDJQ_auv:b6byvub) ([raw file](https://github.com/batfish/batfish/blob/0034801fa7f49089e3149db11fe0eb44deed0ead/docs/conversion/juniper_routing_policy.md#L4)):*
> ```Markdown
> 
> ## Default Fall-Through Actions
> ```

I don't like this.

dhalperi (commented):
> @dhalperi resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9868)* status: 0 of 7 files reviewed, all discussions resolved.

SLarkworthy (commented):
:lgtm:

> @SLarkworthy reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9868#-:-OpthXXNEy5jJQKJhIMG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9868)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9881, opened 2026-04-15
RouteFilterList: make immutable; fix O(N^2) addLine; delete dead code

dhalperi:
RouteFilterList._lines was in a broken half-mutable state: addLine()
rebuilt the entire ImmutableList on every call (O(N) per call, O(N^2)
in loops), caches were never invalidated on mutation, and setLines()
accepted any list with no immutability guarantee.

- Make _lines final ImmutableList, set once at construction
- Remove addLine() and setLines() mutation methods
- Remove getMatchingIps() which had zero callers
- Change Route4FilterLine.applyTo(RouteFilterList) to
toRouteFilterLines() returning a List
- Update all 21 call sites across Juniper, Cisco, CiscoXR, ASA,
Arista, FRR, Cumulus, VyOS, and Common to build lines first then
pass to constructor

----

Prompt:
```
Make a comprehensive look at RouteFilterList. I have many questions:

1. Why does it take IpWildcard instead of Prefix? The semantics it
has seem to apply specifically to prefixes.
2. Can we delete unused code (getMatchingIps)?

Most importantly:

3. What are the semantics in terms of immutability or not, especially
of the _lines. I have some evidence that VI conversion is causing
O(N^2) work in the addLine function which is dumb. Should we rewrite
it as an immutable class with a builder, or just make folks pass in
the lines, or let it be mutable during construction but change it to
immutable in finalizeConfiguration?
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cisco/CiscoConfiguration.java
@@ -1960,13 +1960,12 @@ public final class CiscoConfiguration extends VendorConfiguration {
       Map<Prefix, OspfAreaSummary> summaries = e1.getValue();
       OspfArea.Builder area = areas.get(areaLong);
       String summaryFilterName = "~OSPF_SUMMARY_FILTER:" + vrfName + ":" + areaLong + "~";
-      RouteFilterList summaryFilter = new RouteFilterList(summaryFilterName);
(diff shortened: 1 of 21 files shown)


dhalperi (commented):
> @dhalperi reviewed 21 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9881)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9885, opened 2026-04-15
Configuration: return views from getAllInterfaces(vrf) and getActiveInterfaces(vrf)

dhalperi:
Both methods built a new ImmutableMap on every call by streaming
and filtering all interfaces. Replace with Maps.filterValues
views, which avoid the copy and support O(1) get() lookups
against the underlying map. No callers iterate the result more
than once, so re-evaluating the predicate is not a concern.

These methods are called per-BGP-neighbor during vendor
conversion in Cisco IOS, IOS-XR, ASA, Arista, NX-OS, FRR, and
Cumulus. The per-call map copy was the dominant cost; the view
eliminates it for all vendors at once.

----

Prompt:
```
Read HEAD. This seems likely to be a problem in many vendors,
and if it's expensive then we should consider fixing it
everywhere. Is this the right fix or should we actually change
the function to take a list of interfaces, or something else?
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/Configuration.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/Configuration.java
@@ -12,6 +12,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.ImmutableSortedSet;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9885)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9887, opened 2026-04-17
Juniper: detect fatal xSTP error for interfaces without ethernet-switching

dhalperi:
On Juniper devices, interfaces referenced in `protocols vstp`,
`protocols rstp`, or `protocols stp` must have
`family ethernet-switching` configured. Without it, the device
rejects the commit with a fatal error:
"XSTP : Interface <name> is not enabled for Ethernet Switching"

Batfish previously silently discarded all xSTP configuration via
the `p_null` grammar rule. This change extracts interface names
from VSTP, RSTP, and STP protocol blocks, starts a minimal grammar, and
validates during conversion that each referenced interface has
`family ethernet-switching`.

----

Prompt:
```
XSTP : Interface xe-0/2/1 is not enabled for Ethernet Switching

This is a fatal error not caught by Batfish today on Juniper
devices. Here's a sample config.

Investigate (using junos manuals in ../batfish/working if needed)
and fix
```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -674,6 +674,8 @@ ECHO_REPLY: 'echo-reply';
 
 ECHO_REQUEST: 'echo-request';
 
+EDGE: 'edge';
+
 EGP: 'egp';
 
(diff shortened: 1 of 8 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9887)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9888, opened 2026-04-17
Flatten, Annotate: support files as well as directories

dhalperi:
Bring both tools in line with Preprocess: if the input is a regular
file, flatten/annotate it directly and write the result to the given
output path; otherwise, treat the input as a snapshot directory as
before. Usage strings updated to <input_path> <output_path>. Adds
file-based tests mirroring PreprocessTest.

----

Prompt:
```
Only one of //tools:{preprocess,flatten,annotate} gracefully takes in
a file or a directory. Fix the other two? Also make sure they
properly use relative links from pwd, etc.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/Flatten.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Flatten.java
@@ -13,6 +13,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9888)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9891, opened 2026-04-20
AWS IPSec: fix VPN connection parsing when TunnelOptions is incomplete

### LINKED ISSUE #74: onug_cost_optimized_s3_path: broken by batfish/batfish#9381
In batfish/batfish#9381 we overhauled some AWS work, now Batfish isn't establishing the tunnels. This could be a problem with the lab if the lab was setup incomplete, but more likely some kind of regression in that PR. Will need to dive deeper.

dhalperi:
Two bugs in VPN connection handling introduced by PR #9381:

1. TunnelInsideCidr was required to be non-null in TunnelOptions
deserialization, but real AWS snapshots from older API versions only
have OutsideIpAddress and PreSharedKey. The inside addresses are
already available in the CustomerGatewayConfiguration XML, so this
field is supplementary. Remove the assertion.

2. When TunnelOptions is present but Phase1/Phase2 crypto algorithms
are null, use permissive defaults instead of asserting non-null.

Without these fixes, VPN connections fail to parse entirely, causing
missing IPSec tunnels, missing BGP sessions, and missing routes.

Fixes: batfish/lab-validation#74

----

Prompt:
```
work on https://github.com/batfish/lab-validation/issues/74
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/aws/IpsecTunnel.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/aws/IpsecTunnel.java
@@ -140,19 +140,24 @@ final class IpsecTunnel implements Serializable {
             Utils.textOfFirstXmlElementWithTag(ipsecElement, AwsVpcEntity.XML_KEY_LIFETIME)));
     builder.setIpsecMode(
         Utils.textOfFirstXmlElementWithTag(ipsecElement, AwsVpcEntity.XML_KEY_MODE));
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9891)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9896, opened 2026-04-21
HeaderConstraintsToFlows: generalize TracerouteAnswererHelper to question-independent utility

dhalperi:
Rename and move from the traceroute subpackage to
org.batfish.question, since it is a general utility for converting
PacketHeaderConstraints into concrete Flow objects for any question
that uses TracerouteEngine.

Add a FlowPreference parameter to the constructor (previously
hardcoded to TRACEROUTE) so non-traceroute questions can reuse it.
Make getFlows() public.

----

Prompt:
```
I want to make TracerouteAnswererHelper::getFlows available for
others to use, but it's only visible for testing. Does it make
sense to add as part of the public API? is there a better
abstraction for the role that TracerouteAnswererHelper is filling
-- not a helper for the specific TracerouteAnswerer question, but
in general a tool to help in questions that use TracerouteEngine
to map packet forwarding behavior?
```

### DIFF
--- a/docs/question_development/traceroute.md
+++ b/docs/question_development/traceroute.md
@@ -30,7 +30,7 @@ The 2023 SIGCOMM paper discusses how forwarding analysis ("Lesson 4: Faithfully 
 User inputs (startLocation, headers)
         |
         v
-[TracerouteAnswererHelper] -- resolves specifiers, builds Flow objects
+[HeaderConstraintsToFlows] -- resolves specifiers, builds Flow objects
         |
         v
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> @dhalperi partially reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9896)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9843, opened 2026-03-27
Junos: EVPN Type 5

jtwb:
Add support for EVPN Type 5 routes in Junos and import, export policies (route leaking). A minimal configuration (excluding underlay / overlay) might look like:

```
policy-options {
    policy-statement policy1 {
        then accept;
    }
    policy-statement policy2 {
        then accept;
    }
}
routing-options {
    router-id 1.1.1.1;
    autonomous-system 65000;
}
routing-instances {
    test_vrf {
        instance-type vrf;
        protocols {
            evpn {
                ip-prefix-routes {
                    import policy2;
                    vni 1000;
                    advertise direct-nexthop;
                    encapsulation vxlan;
                }
            }
        }
        route-distinguisher 1:1;
        vrf-target target:1:1;
        vrf-import policy1;
    }
}
```

### New Statements Supported

- routing-instances *routing-instance* evpn ip-prefix-routes export *policy-name*
- routing-instances *routing-instance* evpn ip-prefix-routes import *policy-name*
- routing-instances *routing-instance* evpn ip-prefix-routes vni *number*
- routing-instances *routing-instance* route-distinguisher *identifier*
- routing-instances *routing-instance* vrf-target *community*
- routing-instances *routing-instance* vrf-target import *community*
- routing-instances *routing-instance* vrf-target export *community*
- routing-instances *routing-instance* vrf-import *policy-name*
- switch-options vtep-source-interface *interface-name*
- (various BGP hierarchies) family evpn signaling

### Notable Gaps (TODO)

- routing-instances *routing-instance* vrf-export
- routing-options route-distinguisher-id

### Testing

The testrig `bgp-evpn-type5-route-test` which tests CE -> PE -> RR -> PE EVPN Type 5 route distribution was loaded into containerlab. RIBs were compared in detail against Batfish. Connectivity (traceroute) from clients was tested and compared to Batfish.

### How Statements Interact

**Import**

If `ip-prefix-routes import` and `vrf-import` are supplied, they are called conjunctively against candidate routes. If one or both are supplied, `vrf-target import` and `vrf-target` are ignored for import purposes.

Otherwise we fall back to simple community-based route acceptance using `vrf-target import`, falling back to `vrf-target` if `vrf-target import` is not supplied.

Importantly, routes with the same route distinguisher value as the IP-VRF are not eligible for import. This prevents self-import loops. Junos devices have some mechanism to prevent EVPN Type 5 self-import loops but it's unclear if this is the actual scheme used on-device. (See import loops below)

[...]

PR #9898, opened 2026-04-22
Junos: apply autostate to VNI-backed VLANs

dhalperi:
Previously, VLANs with a VXLAN VNI were excluded from autostate
(normalVlanRange), keeping their IRB interfaces active even when no
switchport was assigned to the VLAN. This was added in #9843 under the
assumption that VXLAN overlay traffic would provide connectivity without
local physical members.

In practice, if the underlay links are down (e.g., in a containerlab
environment), the IRBs should be deactivated — the real device marks
them as hardware-down and does not install connected routes. Removing
the exclusion lets autostate deactivate these IRBs, matching on-device
behavior when runtime data reflects the actual link state.

This fixes a regression from batfish/batfish#9843.

Prompt:
```
Let's make the batfish change first. Ideally, some test will fail
from this change and we can confirm that test was testing the wrong
behavior.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
@@ -2168,12 +2168,6 @@ public final class JuniperConfiguration extends VendorConfiguration {
       if (vxlan.getVniId() == null) {
         continue;
       }
-      // Exclude VNI VLANs from normalVlanRange so their IRB interfaces are not deactivated by
-      // autostate (these VLANs carry VXLAN overlay traffic and may have no physical members).
-      if (vxlan.getVlanId() != null) {
-        _c.setNormalVlanRange(
-            _c.getNormalVlanRange().difference(IntegerSpace.of(vxlan.getVlanId())));
-      }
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9898)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

jtwb:
I think this Juniper KB article supports applying autostate to VNI-backed VLANs as well:

https://supportportal.juniper.net/s/article/MX-Troubleshooting-Checklist-IRB-Interface#:~:%20UP%20state.-,Tips:,domains%20and%20their%20operational%20status.&text=In%20the%20output%20above%2C%20the,down%20in%20the%20bridge%2Ddomain.


> Tips:
> 
> - The IRB logical interface is operationally up if at least one Layer-2 logical interface is up in the bridge-domain.
> - The IRB logical interface is operationally down if all L2 logical interfaces are operationally down in the bridge-domian.
> 
> If the Admin status of the IRB logical interface is down, then verify the configuration and correct the problem.  As per the above tips, the IRB logical interface Link status is down if all the l2-interfaces are down in the bridge-domain.

PR #9900, opened 2026-04-23
VXLAN-aware autostate: activate IRBs when L2 VNI tunnels come up

dhalperi:
Previously, VLAN interfaces (IRBs) with no physical switchport
members were permanently deactivated by autostate during
pre-dataplane post-processing. VXLAN tunnels are established later
during topology iteration, but there was no feedback path to
reactivate IRBs once remote VTEPs became reachable. NX-OS worked
around this by excluding VNI-backed VLANs from autostate entirely,
which was wrong when the underlay was down (the same issue fixed
for Juniper in #9898).

This change makes autostate VXLAN-aware during topology iteration.
When L2 VNI edges change, affected VLAN interfaces are re-evaluated:
- Gained edge + AUTOSTATE_FAILURE IRB → reactivate
- Lost edge + zero physical members → deactivate

Connected and local routes are recomputed via delta when interface
status changes, so newly-active IRBs contribute routes to the RIBs.

Also removes the NX-OS normalVlanRange exclusion workaround, which
is no longer needed.

Prompt:
```
Read https://github.com/batfish/lab-validation/issues/125 and
d255adc5c2f337783496161aa92834aa81dd38db
(batfish/batfish#9898). I think that code, also used in other
vendors, is a misguided attempt to work around the fact that we
don't dynamically bring IRBs up and down during
TopologyIteration based on the presence of VTEPs. Do your
analysis and see if you agree? If so, how should we fix?
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
@@ -37,6 +37,7 @@ import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.batfish.common.BdpOscillationException;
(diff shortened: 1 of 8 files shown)


dhalperi:
*[`projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java` line 545 at r1](https://reviewable.io/reviews/batfish/batfish/9900#-Oqvv3su5UF-tZg9ZJ8n:-Oqvv3svBDbw3LPGihsR:b-bwy6tw) ([raw file](https://github.com/batfish/batfish/blob/bc2584186701bb9faeae7d2722661b275c382ff5/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java#L545)):*
> ```Java
>       if (autostateChanged) {
>         // Recompute connected routes for affected VRFs so newly-active IRBs contribute routes
>         vrs.parallelStream()
> ```

It looks like we do the updates for ALL VRs, not just the affected ones. Can we improve this?

dhalperi:
*[`projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java` line 545 at r1](https://reviewable.io/reviews/batfish/batfish/9900#-Oqvv3su5UF-tZg9ZJ8n:-OqwEBE81R1fuFRC5nVQ:b1h19za) ([raw file](https://github.com/batfish/batfish/blob/bc2584186701bb9faeae7d2722661b275c382ff5/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java#L545)):*

It looks like we do the updates for ALL VRs, not just the affected ones. Can we improve this?
</blockquote></details>

Done in r2 - limiting to affected hosts as a middle step between all vrfs and "only vrfs on select hosts".

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9900)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9904, opened 2026-04-24
Group async-profiler packages in bazel-steward config

dhalperi:
----

Prompt:
```
Fix the bazel-steward config to consider all the async-profiler
packages as part of the same group. We do it for other things, like
jackson and jersey
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9904)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9905, opened 2026-04-24
BOMInputStream: migrate from deprecated constructor to builder API

dhalperi:
The deprecated BOMInputStream varargs constructor now throws IOException
in commons-io 2.21.0, causing compilation failures. Use the builder API
(BOMInputStream.builder()) which is the recommended replacement.

----

Prompt:
```
https://github.com/batfish/batfish/actions/runs/24835593475/job/72731602753?pr=9899
fails to compile against new commons-io - because we are using a
deprecated method [...] and in the new version they added an exception
to it.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/StreamDecoder.java
+++ b/projects/batfish/src/main/java/org/batfish/main/StreamDecoder.java
@@ -31,14 +31,17 @@ final class StreamDecoder {
     }
   }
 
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9905)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9912, opened 2026-04-27
Arista: consolidate concrete BGP neighbors into a shared base class

dhalperi:
Introduce AristaBgpConcreteNeighbor as the common parent of V4,
V4Dynamic, Interface, and V6 neighbors (but not peer-group neighbors).
It holds _peerGroup, provides the shared inherit() and isActive()
implementations, and declares abstract getPeerString() /
getAfSettings() / getOrCreateAfSettings() hooks that each subclass
fills in with its identifier (Ip, Prefix, interface name).

This removes three copies of inherit(), three copies of isActive(),
and three getTextDesc overloads. In AristaConversions.toBgpNeighbor
it also collapses the three-way instanceof ladders that picked naf4,
nEvpn, and peerStrRepr into single polymorphic calls.

Also:
- Replace Stream.peek(...) side effects with explicit forEach before
the stream pipeline, matching Javadoc guidance on peek.
- Lift the 1000-entry range cap out of expandBgpIfaceRange into a
named constant with a comment; clarify the skip warning message.
- Add deliberate-warning assertions for the existing Xy1,
Et200-2000, and Et100 test cases, plus coverage for
`no neighbor interface` with a range.

V6 neighbors inherit the base class with stub AF hooks; no converter
iterates them today.

----

Prompt:
```
Read HEAD, an external submission. What review would you have given
in terms of idiomatic Batfish contibutions and clean maintainable
code?
```

Followups in conversation: fix review items 2 (warning assertions),
3 (range-removal test coverage), 4 (dedup inherit + unify peer-string
accessor), 5 (peek to forEach), 7 (magic number), 8 (move isActive
to the class); also make V6 extend the new base class with stubs.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
@@ -2890,6 +2890,13 @@ public class AristaControlPlaneExtractor extends AristaParserBaseListener
   }
 
   /**
+   * Upper bound on the size of a single {@code neighbor interface} range entry. A typical Arista
+   * chassis has far fewer physical interfaces than this; any range that exceeds it is almost
+   * certainly a typo or misuse of the syntax.
+   */
+  private static final int BGP_IFACE_RANGE_MAX = 1000;
+
+  /**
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9912)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9913, opened 2026-04-27
Arista: do not set weight 32768 on locally-originated BGP paths

dhalperi:
Unlike IOS/IOS-XE, Arista EOS does NOT apply the industry-common
weight 32768 to locally-originated BGP paths; the effective weight
is 0. Verified empirically on EOS 4.36 in both multi-agent and ribd
modes: with a received path carrying `neighbor ... weight 10000`
and a matching local origin, the received path wins best-path
selection and the device emits "Not best: Path weight" on the local
path. A route-map on a `network` statement or aggregate
attribute-map can still set weight as usual.

Remove the `SetWeight(32768)` statement from the Arista BGP
redistribution policy. Aggregate routes are still generated with
weight 32768 via the shared BgpProtocolHelper.toBgpv4Route code
path; fixing that requires a vendor-specific override and is left
for a follow-up. Test expectations are updated: non-aggregate local
paths no longer assert weight 32768; aggregate paths retain the
32768 expectation with a TODO pointer.

Reproducer lab: batfish/lab-validation snapshot
eos_ceos_bgp_weight_experiment. See
https://github.com/batfish/lab-validation/issues/152 for the
broader Arista local-path attribute divergence (this change fixes
the weight half; local-preference needs further investigation).

----

Prompt:
```
This issue https://github.com/batfish/lab-validation/issues/6
seems to be responsible for a lot of lab issues. Can you dive
into it? Note that there's an EOS manual in ./batfish/working/
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
@@ -168,7 +168,6 @@ import org.batfish.datamodel.routing_policy.expr.Conjunction;
 import org.batfish.datamodel.routing_policy.expr.DestinationNetwork;
 import org.batfish.datamodel.routing_policy.expr.Disjunction;
 import org.batfish.datamodel.routing_policy.expr.ExplicitPrefixSet;
-import org.batfish.datamodel.routing_policy.expr.LiteralInt;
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9913)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9915, opened 2026-04-27
Arista: do not set local-preference 0 on locally-originated BGP paths

dhalperi:
The previous commit (#9913) fixed the weight of locally-originated
Arista BGP paths from 32768 to 0. But the redistribution policy also
set local-preference to 0, overriding the standard BGP default of
100. Empirical testing on EOS 4.36 showed that locally-originated
paths use the default local-preference of 100: a received path with
`set local-preference 50` loses, and one with `set local-preference
150` wins. See lab-validation#152 for the reproducer lab.

Removing the explicit SetLocalPreference(0) also required updating
the `export-localpref` guard in AristaConversions, which previously
checked for LP==0 as a proxy for "unset default" — now it checks
for LP==DEFAULT_LOCAL_PREFERENCE (100) since that is the new
effective default for locally-originated routes.

Without this fix, the weight-only change in #9913 caused a
regression in eos_private_asn: on DuplicatePrivate (which has
`allowas-in 3`), the locally-originated route for
`192.168.123.8/32` had weight=0 and LP=0. The eBGP peer
(R-PublicASN100) sent the route back, and the received copy arrived
with LP=100 (default eBGP import). Since weight tied (both 0) and
LP 100 > 0, the received route replaced the local route as
best-path. But the received route failed the export filter
(not ReceivedFromSelf), causing a withdrawal that removed the
route from the entire network. With LP=100 on the locally-
originated route, LP ties and the "prefer locally originated" step
keeps the local route as best-path, preventing the oscillation.

----

Prompt:
```
So we did a deep analysis and concluded that the issue was LP --
not being set on the local route, but being set on the re-received
route. Why did we have to undo the LP 100 stuff?
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
@@ -174,7 +174,6 @@ import org.batfish.datamodel.routing_policy.expr.MatchPrefixSet;
 import org.batfish.datamodel.routing_policy.expr.MatchProtocol;
 import org.batfish.datamodel.routing_policy.expr.Not;
 import org.batfish.datamodel.routing_policy.statement.If;
-import org.batfish.datamodel.routing_policy.statement.SetLocalPreference;
 import org.batfish.datamodel.routing_policy.statement.SetMetric;
 import org.batfish.datamodel.routing_policy.statement.SetOrigin;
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9915)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9916, opened 2026-04-28
Junos: model advertise-peer-as as EXCEPT_RECEIVED_FROM, not ALWAYS

dhalperi:
Junos `advertise-peer-as` disables AS-path-based loop prevention for
outbound eBGP advertisements but still blocks re-advertisement back
to the specific peer the route was received from. Previously, Batfish
mapped this to AllowRemoteAsOutMode.ALWAYS which has no per-peer
check, causing routes to be incorrectly sent back to originating
peers.

Add a new AllowRemoteAsOutMode.EXCEPT_RECEIVED_FROM that skips
AS-path filtering but checks the route's ReceivedFrom against the
export session's remote peer IP. Map Junos advertise-peer-as to
this new mode.

This is for batfish/lab-validation#44.

----

Prompt:
```
Read up on https://github.com/batfish/lab-validation/issues/44 and
dig into both Batfish and the affected labs. Also note that there
are junos manuals in ./batfish/working/
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
@@ -106,11 +106,16 @@ public final class BgpProtocolHelper {
     }
 
     // For eBGP, do not export if AS path contains the peer's AS in a disallowed position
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9916)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9917, opened 2026-04-28
LargeCommunity: add @JsonValue for correct JSON serialization

dhalperi:
LargeCommunity was missing the @JsonValue annotation on toString(),
unlike StandardCommunity and ExtendedCommunity. Without it, Jackson
serialized LargeCommunity as a JSON object with field names instead
of its string representation, causing deserialization failures (e.g.,
viModel question returning HTTP 400).

Also fix testJsonSerialization which was testing Java serialization
instead of JSON round-trip, and add a test for round-tripping through
the Community base class.

----

Prompt:
```
go in ../batfish and fix the large community issue
[Batfish viModel question fails to deserialize LargeCommunity —
batfish/lab-validation#160]
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9917)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9918, opened 2026-04-29
Arista: support `ip address unnumbered` with new UnnumberedAddress type

dhalperi:
Add UnnumberedAddress, a new InterfaceAddress subclass for interfaces
that borrow their IP from another interface (e.g., `ip address
unnumbered Loopback0`). Unlike copying the source's
ConcreteInterfaceAddress, UnnumberedAddress naturally suppresses
connected route generation and IP ownership claims, and provides an
explicit signal for topology computation.

Parse and extract the `ip address unnumbered <interface>` syntax on
Arista EOS interfaces. During conversion, resolve the source
interface's IP and set UnnumberedAddress on the VI interface. Set
OspfAddresses on the OSPF interface settings so that OSPF neighbor
init can find the borrowed IP for adjacency formation.

Extend TopologyUtil.computeRawLayer3Topology() to create L3 edges for
interfaces with UnnumberedAddress (or LinkLocalAddress) via physical
adjacency, since neither type participates in subnet-based L3
topology synthesis.

Fixes https://github.com/batfish/batfish/issues/7227

----

Prompt:
```
Investigate https://github.com/batfish/batfish/issues/7227. Create a
ceos lab that exhibits the behavior, test it, triage it against
current batfish (origin/master) and commit it with a complete sickbay
including new bugs if appropriate, referencing the original. Then
start closing the gaps in Batfish. Work autonomously, until complete.
```

PR #9919, opened 2026-04-29
TopologyUtil: create L3 edges for /32-only point-to-point interfaces

dhalperi:
When physical (L1) topology is available, create L3 edges for active
non-loopback interfaces that have only /32 concrete addresses (e.g.,
unnumbered interfaces borrowing a loopback address). These interfaces
are excluded from subnet-based L3 synthesis because /32 prefixes are
never bucketed, but they should still get edges via physical
adjacency for OSPF point-to-point adjacencies to form.

----

Prompt:
```
Investigate https://github.com/batfish/batfish/issues/7227. Create a
ceos lab that exhibits the behavior, test it, triage it against
current batfish (origin/master) and commit it with a complete sickbay
including new bugs if appropriate, referencing the original. Then
start closing the gaps in Batfish. Work autonomously, until complete.
```

---

**Stack**:
- #9919 ⬅
- #9918

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/topology/TopologyUtil.java
+++ b/projects/common/src/main/java/org/batfish/common/topology/TopologyUtil.java
@@ -502,8 +502,34 @@ public final class TopologyUtil {
                   return Stream.of(new Edge(nip, paired), new Edge(paired, nip));
                 });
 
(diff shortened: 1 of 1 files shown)


dhalperi:
Decided not to go with this approach, instead to add an UnnumberedAddress VI type.

PR #9920, opened 2026-04-29
checkstyle: catch @Nullable/@Nonnull before transient, abstract, synchronized, etc.

dhalperi:
The nullity annotation placement rules only checked modifiers
private/public/protected/static/final. Add transient, volatile, abstract,
default, synchronized, and native. Fix all 31 existing violations.

Prompt:
```
We have a checkstyle or pmd rule that keeps @Nullable/@Nonnull next to
the type they annotate. I saw one like `private @Nullable transient Set`
-- can we fix the rule so it catches these, and then fix the violations
```


dhalperi (commented):
> @dhalperi reviewed 19 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9920)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9921, opened 2026-04-29
Arista: accept IP address config regardless of position relative to `no switchport`

dhalperi:
Batfish was rejecting `ip address` on Arista Ethernet interfaces when it
appeared before `no switchport` in the config. Physical interfaces default
to switchport mode, and the extraction code was checking switchport state
at parse time — so IP addresses were silently dropped if they preceded
`no switchport`.

Real EOS devices accept L3 config in either order. Confirmed with a cEOS
4.36.0.1F lab: configuring `ip address` before `no switchport` results in
a working L3 interface. The device normalizes `show running-config` to
display `no switchport` first regardless of input order.

The fix removes the premature switchport guards from IP address extraction.
The conversion layer already correctly ignores addresses for interfaces
that remain in switchport mode.

Prompt:
```
I can't find the report, but a user reported that we are incorrectly
requiring switchport to be disabled before accepting L3 config for Arista
interfaces, but in reality devices may well have L3 config before
`no switchport` in show data. Can you confirm with a new cEOS lab and if
so, fix Batfish behavior.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
@@ -5150,10 +5150,6 @@ public class AristaControlPlaneExtractor extends AristaParserBaseListener
     ConcreteInterfaceAddress addr = toAddress(ctx.addr);
     _currentInterfaces.forEach(
         i -> {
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9921)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9923, opened 2026-04-29
Arista: fix OSPF default interface cost when auto-cost is not configured

### LINKED ISSUE #9922: Arista OSPF: default interface cost is 10 when auto-cost reference-bandwidth is not configured
## Problem

When `auto-cost reference-bandwidth` is not explicitly configured, EOS uses a flat default OSPF interface cost of 10 for all Ethernet interfaces, rather than computing cost from a reference bandwidth. Batfish always computes cost from reference bandwidth, producing incorrect OSPF metrics.

## Technical Details

Per the EOS 4.36 User Manual, Chapter 27, "auto-cost reference-bandwidth (OSPFv2)":

> The `no auto-cost reference-bandwidth` and `default auto-cost reference-bandwidth` command removes the auto-cost reference-bandwidth command from running-config. **When this parameter is not set, the default cost for Ethernet interfaces is the default ip ospf cost value of 10.**

The `auto-cost reference-bandwidth` default rate is 100 (meaning 100 Mbps), but this only applies when the command is explicitly present in the config. When absent, the cost formula is not used at all — interfaces get a flat cost of 10.

Batfish currently always computes cost from a reference bandwidth (`OspfProcess.computeInterfaceCost()` at `projects/common/.../ospf/OspfProcess.java:366-378`). The Arista default reference bandwidth is set to `10E6` (10 Mbps) in `projects/batfish/.../vendor/arista/representation/OspfProcess.java:32`, which produces `max(10E6 / bandwidth, 1) = 1` for typical interface speeds — not 10.

## Observed In

`eos_ospf_unnumbered` lab in batfish/lab-validation. On a 3-node OSPF topology with 1 Gbps cEOS interfaces and no `auto-cost reference-bandwidth` configured:
- Device metric for a 3-hop route: 30 (cost 10 per hop)
- Batfish metric for the same route: 3 (cost 1 per hop)

## Code References

- `projects/batfish/src/main/java/org/batfish/vendor/arista/representation/OspfProcess.java:32` — `DEFAULT_REFERENCE_BANDWIDTH_10_MBPS = 10E6D`
- `projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java:1798-1803` — sets cost to null when not explicitly configured
- `projects/common/src/main/java/org/batfish/datamodel/ospf/OspfProcess.java:366-378` — `computeInterfaceCost()` always falls back to `referenceBandwidth / bandwidth`

## Suggested Fix

Track whether `auto-cost reference-bandwidth` was explicitly configured. When it was not, set a default interface cost of 10 instead of computing from reference bandwidth. When it was explicitly configured (including `auto-cost reference-bandwidth 100`), compute cost as `max(rate * 1Mbps / bandwidth, 1)`.

dhalperi:
Per the EOS 4.36 manual, when `auto-cost reference-bandwidth` is not
configured, all interfaces get a flat default OSPF cost of 10 (the
"default ip ospf cost value"). The reference-bandwidth formula
(`rate * 1 Mbps / bandwidth`) is only used when `auto-cost
reference-bandwidth` is explicitly present in the config.

Batfish previously always computed cost from a reference bandwidth,
using an incorrect default of 10 Mbps. This produced cost 1 for
typical interfaces instead of cost 10.

The fix:
- Make the VS OspfProcess reference bandwidth nullable to distinguish
"auto-cost configured" from "not configured"
- When not configured, set all interface costs to 10 during conversion
- Fix the default reference bandwidth from 10 Mbps to 100 Mbps (the
documented default when the command IS explicitly configured)

Fixes batfish/batfish#9922

Prompt:
```
Let's try to fix the batfish issue first.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/AristaConfiguration.java
@@ -1584,7 +1584,8 @@ public final class AristaConfiguration extends VendorConfiguration {
     org.batfish.datamodel.ospf.OspfProcess newProcess =
         org.batfish.datamodel.ospf.OspfProcess.builder()
             .setProcessId(proc.getName())
-            .setReferenceBandwidth(proc.getReferenceBandwidth())
+            .setReferenceBandwidth(
+                firstNonNull(proc.getReferenceBandwidth(), OspfProcess.DEFAULT_REFERENCE_BANDWIDTH))
             .setAdminCosts(
                 org.batfish.datamodel.ospf.OspfProcess.computeDefaultAdminCosts(
                     c.getConfigurationFormat()))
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9923)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9924, opened 2026-05-01
PaloAlto: fix EBGP session type — use SINGLEHOP when multihop not configured

gh-ajinks:
## Summary

`PaloAltoConfiguration.java` unconditionally calls `.setEbgpMultihop(true)` for every PAN-OS EBGP peer. This causes a session type mismatch when a PAN-OS firewall peers directly (single hop) with a device that correctly classifies the session as `EBGP_SINGLEHOP` (e.g. Arista EOS without `ebgp-multihop`), resulting in `NOT_ESTABLISHED` sessions in `bgpSessionStatus` even when `bgpSessionCompatibility` returns `UNIQUE_MATCH`.

The original comment cites a [PAN-OS KB article](https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000ClKkCAK) about a TTL+2 computation to justify always setting multihop. However, that adjustment only applies when multihop is **explicitly enabled** with a value > 0. `multihop 0` means disabled in PAN-OS; the absence of multihop config means the peer is a directly adjacent single-hop neighbor.

Additionally, `peer.getMultihop()` exists on the `BgpPeer` data model but was never consulted — the parsed value was silently discarded.

The fix uses the parsed value, consistent with how `CiscoConfiguration.java` handles the equivalent:

```java
// Before
.setEbgpMultihop(true)

// After
.setEbgpMultihop(peer.getMultihop() != null && peer.getMultihop() > 0)
```

## Repro

```python

PAN_CFG = """
set deviceconfig system hostname pan-fw
set network interface ethernet ethernet1/1 layer3 units ethernet1/1.10 ip 192.0.2.1/30
set network virtual-router default interface ethernet1/1.10
set network virtual-router default protocol bgp enable yes
set network virtual-router default protocol bgp router-id 192.0.2.1
set network virtual-router default protocol bgp local-as 65001
set network virtual-router default protocol bgp peer-group ebgp type ebgp
set network virtual-router default protocol bgp peer-group ebgp peer peer1 peer-address ip 192.0.2.2
set network virtual-router default protocol bgp peer-group ebgp peer peer1 peer-as 65002
set network virtual-router default protocol bgp peer-group ebgp peer peer1 local-address ip 192.0.2.1
set network virtual-router default protocol bgp peer-group ebgp peer peer1 enable yes
"""

bf = Session()
bf.set_network("pan-multihop-bug")
bf.init_snapshot_from_text(PAN_CFG, platform="paloalto")
df = bf.q.bgpSessionCompatibility().answer().frame()
print(df[["Node", "Session_Type"]])
# Expected: EBGP_SINGLEHOP (no multihop configured)
# Actual:   EBGP_MULTIHOP  (hardcoded true)
```

## Changes

- `PaloAltoConfiguration.java`: set `ebgpMultihop` based on parsed `peer.getMultihop()` value
- `PaloAltoGrammarTest.java`: update `testBgpMultihopConversion` (`multihop 0` = disabled = singlehop); add `testBgpMultihopEnabledConversion` for `multihop 5` case; add singlehop assertion to `testBgpConversion`
- Add test fixture `bgp-multihop-enabled` with `multihop 5`


gh-ajinks:
Originated from  https://batfish-org.slack.com/archives/C8XKQNHAQ/p1777589938862729

dhalperi (commented):
> @dhalperi reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9924)* status: all files reviewed, 3 unresolved discussions (waiting on gh-ajinks).

___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 2614 at r1](https://reviewable.io/reviews/batfish/batfish/9924#-OrswiMp9XniI5mo-cO2:-OrswiMp9XniI5mo-cO3:b-n52hbo) ([raw file](https://github.com/batfish/batfish/blob/efaf0d74f9ef4ba5d5d925f83a733fe4c3e5740c/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L2614)):*
> ```Java
>             // Use multihop only when explicitly configured with a TTL > 0. PAN-OS "multihop 0"
>             // means disabled; absence of multihop config means single-hop (directly adjacent peer).
>             .setEbgpMultihop(peer.getMultihop() != null && peer.getMultihop() > 0)
> ```

This should be >1, right? TTL of 1 would still be singlehop.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java` line 4275 at r1](https://reviewable.io/reviews/batfish/batfish/9924#-Orswvbh3gL5EHqmrkhz:-Orswvbh3gL5EHqmrki-:b-gw4fnk) ([raw file](https://github.com/batfish/batfish/blob/efaf0d74f9ef4ba5d5d925f83a733fe4c3e5740c/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java#L4275)):*
> ```Java
>   @Test
>   public void testBgpMultihopEnabledConversion() {
>     // multihop > 0 means enabled in PAN-OS; peer should be classified as multi-hop
> ```

same note about >0 vs > 1.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java` line 4277 at r1](https://reviewable.io/reviews/batfish/batfish/9924#-Orsx-mJ1PD3bu2DVORH:-Orsx-mJ1PD3bu2DVORI:b-bxic1u) ([raw file](https://github.com/batfish/batfish/blob/efaf0d74f9ef4ba5d5d925f83a733fe4c3e5740c/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java#L4277)):*
> ```Java
>     // multihop > 0 means enabled in PAN-OS; peer should be classified as multi-hop
>     String hostname = "bgp-multihop-enabled";
>     Configuration c = parseConfig(hostname);
> ```

Can we put multiple peers in the same file, with different settings? That should simplify our code.

gh-ajinks:
Thanks for the thorough review. All three points addressed:

**`> 0` → `> 1`**: Confirmed against PAN-OS docs — `multihop 1` sets TTL to 1, which is still directly adjacent (single-hop). TTL ≥ 2 is the minimum for true multi-hop ([PAN-OS 10.2 BGP config reference](https://docs.paloaltonetworks.com/pan-os/10-2/pan-os-networking-admin/bgp/configure-bgp), [KB article on multihop TTL defaults](https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000PPm8CAG)). Fixed the condition and comment in `PaloAltoConfiguration.java`, and updated the test comment accordingly.

**Consolidated test configs**: Merged `bgp-multihop` and `bgp-multihop-enabled` into a single config with three peers (`peer-disabled` multihop 0, `peer-ttl1` multihop 1, `peer-enabled` multihop 5). Also collapsed the two conversion test methods into one — `testBgpMultihopConversion` now asserts all three cases in a single parse. The `multihop 1` peer adds coverage for the edge case your comment identified.

dhalperi (commented):
> @dhalperi reviewed 4 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9924)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #9928, opened 2026-05-04
Junos: fatal red flag error for unnormalized static routes

anothermattbrown:
Junos commit checks require that static route prefixes be normalized (i.e. not host bits set). Add a fatal red flag warning for that. 

note: I suspect this is enforced other places too. For now, I'm limiting to the static route case that I know about

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -3273,7 +3273,7 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
     _currentAreaRangeRestrict = false;
 
     if (ctx.prefix != null) {
-      _currentAreaRangePrefix = toPrefix(ctx.prefix);
+      _currentAreaRangePrefix = toNormalizedPrefix(ctx.prefix, "OSPF area-range");
     } else {
       todo(ctx);
     }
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> @dhalperi reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9928)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on anothermattbrown).

___
*[`projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java` line 3804 at r1](https://reviewable.io/reviews/batfish/batfish/9928#-OrstlHf-H8iBCAaeE96:-OrstlHf-H8iBCAaeE97:b8d689e) ([raw file](https://github.com/batfish/batfish/blob/ea5875656c7874226ad8cb2a15bb72a8bfd00d40/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java#L3804)):*
> ```Java
>   public void enterRos_route4(Ros_route4Context ctx) {
>     Prefix prefix = toPrefix(ctx.prefix);
>     if (ctx.prefix.IP_PREFIX() != null) {
> ```

1. Add a dedicated grammar rule for this, not `ip_prefix`, so that we can put it all in a common helper.
2. Use `Prefix#strict` (or add a `Prefix#tryStrict` or similar, if you prefer not to try/catch)

dhalperi (commented):
> @dhalperi reviewed 5 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9928)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on anothermattbrown).

PR #9932, opened 2026-05-05
Arista: fix BGP prefix-list AF tracking in undefined-reference detection

MicahFulton:
## Summary

`eos_rbafnc_prefix_list` (the per-neighbor prefix-list binding inside `address-family` blocks) always references the binding as `PREFIX_LIST` (IPv4) regardless of which address-family block it appears in. A config that defines an `ipv6 prefix-list` and binds it under `address-family ipv6` produces symmetric false positives:

```
undefinedReferences: ipv4 prefix-list X referenced at bgp inbound prefix-list but not defined
unusedStructures:    ipv6 prefix-list X defined but unused
```

Reference `PREFIX6_LIST` when `_currentAristaBgpVrfAf` is an `AristaBgpVrfIpv6UnicastAddressFamily`.

## Scope

Only the AF-block binding case changes. Top-level neighbor bindings (`neighbor X prefix-list Y in` outside any AF block, where `_currentAristaBgpVrfAf` is null and the prefix-list could be v4 or v6) and non-unicast AFs are unchanged — those cases predate this fix and have their own ambiguities.

## Reproduction

```
ipv6 prefix-list PL-V6
   seq 10 permit ::/0
router bgp 65000
   neighbor PG peer group
   neighbor PG remote-as 65001
   neighbor 192.0.2.1 peer group PG
   address-family ipv6
      neighbor PG activate
      neighbor PG prefix-list PL-V6 out
```

`undefinedReferences` reports `ipv4 prefix-list "PL-V6" referenced at bgp outbound prefix-list but not defined`. EOS resolves `prefix-list` references to `ipv6 prefix-list` definitions when bound under `address-family ipv6` (verified against cEOS 4.35.1F via session-diff).

## Test

`testBgpNeighborPrefixListIpv6AddressFamily` parses a config with both `ip` and `ipv6` prefix-lists bound under their respective AF blocks and asserts `hasNoUndefinedReferences()`. The test fails without the extractor change.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
@@ -999,6 +999,7 @@ import org.batfish.vendor.arista.representation.eos.AristaBgpVlanBase;
 import org.batfish.vendor.arista.representation.eos.AristaBgpVrf;
 import org.batfish.vendor.arista.representation.eos.AristaBgpVrfAddressFamily;
 import org.batfish.vendor.arista.representation.eos.AristaBgpVrfIpv4UnicastAddressFamily;
+import org.batfish.vendor.arista.representation.eos.AristaBgpVrfIpv6UnicastAddressFamily;
 import org.batfish.vendor.arista.representation.eos.AristaEosVxlan;
 import org.batfish.vendor.arista.representation.eos.AristaRedistributeType;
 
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> @dhalperi reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9932)* status: all files reviewed, 2 unresolved discussions (waiting on MicahFulton).

___
*[`projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java` line 2544 at r1](https://reviewable.io/reviews/batfish/batfish/9932#-OrthMTs-Kg4FOC7hBlu:-OrthMTs-Kg4FOC7hBlv:blk07pm) ([raw file](https://github.com/batfish/batfish/blob/d1a0836c101351efc72ecbe995051940487c0a89/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java#L2544)):*
> ```Java
>   public void exitEos_rbafnc_prefix_list(Eos_rbafnc_prefix_listContext ctx) {
>     String name = ctx.name.getText();
>     // The grammar rule eos_rbafnc_prefix_list is shared by both
> ```

delete. Accurate, but this is true for many features in the grammar and is a standard pattern.
___
*[`projects/batfish/src/test/java/org/batfish/vendor/arista/grammar/AristaGrammarTest.java` line 640 at r1](https://reviewable.io/reviews/batfish/batfish/9932#-OrthCWJDgZnMr0lT1VX:-OrthCWJDgZnMr0lT1VY:b8anp7n) ([raw file](https://github.com/batfish/batfish/blob/d1a0836c101351efc72ecbe995051940487c0a89/projects/batfish/src/test/java/org/batfish/vendor/arista/grammar/AristaGrammarTest.java#L640)):*
> ```Java
>     // binding as PREFIX_LIST regardless of the enclosing address-family
>     // block, so PL-V6 (an ipv6 prefix-list bound under address-family
>     // ipv6) showed up as an ipv4 undefined reference.
> ```

Please remove comments that don't make sense when checked in, like this.



_Code quote:_
```Java
    // Before this commit, eos_rbafnc_prefix_list always referenced the
    // binding as PREFIX_LIST regardless of the enclosing address-family
    // block, so PL-V6 (an ipv6 prefix-list bound under address-family
    // ipv6) showed up as an ipv4 undefined reference.
```

dhalperi (commented):
:lgtm:

> @dhalperi reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9932#-:-Oru1crNDKLK9spkcFVe:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on MicahFulton).

PR #9944, opened 2026-05-15
routes: drop legacy Next_Hop_IP and Next_Hop_Interface columns

dhalperi:
The unified Next_Hop column (Schema.NEXT_HOP) has carried the structured
next hop for every routes/bgpRib/evpnRib row -- both the non-diff and
diff variants -- so the legacy decomposed columns are pure duplication.
Remove them along with the now-orphaned plumbing:

- INTERNAL_USE_IPS, which only existed to mask the BGP-unnumbered IP
out of the Next_Hop_IP column.
- RouteRowAttribute._nextHopInterface; the secondary key already pins
NextHop within a sub-group, so the field was redundant for both
equality and comparison.

----

Prompt:
```
I want to remove the columns like Next_Hop_Ip from all questions. No
Next_Hop_Interface, etc. They should all just hae a Next_Hop column.
Fan out agents to find all questions that have such a column,
especially any that don't have a Next_Hop column and need to be
upgraded
```

Investigation found the legacy columns lived only in RoutesAnswerer
(routes, bgpRib, evpnRib), and every emitting site already produced
Next_Hop alongside them, so no question needed an upgrade. Follow-up
in-conversation: confirmed NextHop is comparable (NextHopComparator)
and the secondary keys hash on it, so dropping _nextHopInterface from
RouteRowAttribute does not lose any ordering or grouping.

---
**Stack**:
- #9944 ⬅
---

### DIFF
--- a/projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java
+++ b/projects/question/src/main/java/org/batfish/question/routes/RouteRowAttribute.java
@@ -31,7 +31,6 @@ import org.batfish.datamodel.questions.BgpRouteStatus;
  */
 @ParametersAreNullableByDefault
 public class RouteRowAttribute implements Comparable<RouteRowAttribute> {
-  private final @Nullable String _nextHopInterface;
   private final @Nullable AsPath _asPath;
   private final @Nullable Long _adminDistance;
   private final @Nonnull Set<Long> _clusterList;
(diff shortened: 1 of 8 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9944)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9948, opened 2026-05-18
AWS: model Direct Connect Gateway end-to-end

### LINKED ISSUE #9938: AWS Direct Connect not modeled — DX Gateway TGW attachments silently dropped, VIFs/BGP unsupported
## Summary

Batfish's AWS parser has no support for AWS Direct Connect resources. DX Gateway attachments to Transit Gateways are silently dropped, and there are no resource classes for Virtual Interfaces, DX Connections, or DX Gateways. Hybrid networks using DX for on-prem ↔ AWS connectivity cannot be correctly modeled.

## Evidence in source

`TransitGatewayAttachment.ResourceType` (`TransitGatewayAttachment.java:25`) defines `DIRECT_CONNECT_GATEWAY` as a known enum value, so the JSON parser won't crash when it reads a DX attachment from `TransitGatewayAttachments.json`. However all four `switch` statements in `TransitGateway.java` that dispatch on `ResourceType` only handle `VPC`, `VPN`, and `PEERING` — `DIRECT_CONNECT_GATEWAY` falls to `default` and is not modeled:

```java
// TransitGateway.java line 331
switch (attachment.getResourceType()) {
  case PEERING: ...
  case VPC:     ...
  case VPN:     ...
  default:      // DIRECT_CONNECT_GATEWAY lands here — no routing or interface created
}
```

There are no classes for `DirectConnectGateway`, `VirtualInterface`, `DirectConnectConnection`, or any DX-specific resource in `org.batfish.representation.aws`.

## Missing resource types

| AWS API resource | Batfish class | Status |
|---|---|---|
| `dx:DescribeConnections` | none | ❌ not parsed |
| `dx:DescribeVirtualInterfaces` (Transit VIFs) | none | ❌ not parsed |
| `dx:DescribeDirectConnectGateways` | none | ❌ not parsed |
| `dx:DescribeDirectConnectGatewayAttachments` | none | ❌ not parsed |
| TGW attachment type `direct-connect-gateway` | enum value only | ❌ not modeled |

## Impact

**1. Sessions:** BGP sessions from on-prem routers into AWS via DX Transit VIFs cannot be modeled natively. Users must hand-craft synthetic router stubs representing the Amazon side of each VIF, manually encoding per-VIF ASNs, BGP peer IPs, and VLANs.

**2. Route preference:** AWS TGW enforces a documented route preference order: static > propagated from Direct Connect > propagated from VPN. When both a DX attachment and a VPN attachment propagate the same prefix, AWS always prefers DX. Batfish has no model of this preference — it treats routes from a hand-crafted DX stub identically to VPN-originated routes, producing incorrect path answers for any network that uses both DX and VPN to the same on-prem destinations.

**3. Path tracing:** End-to-end paths from on-prem through DXGW → TGW → VPC cannot be traced at all without the DX attachments and VIF model.

## What correct support would look like

1. Parse `DirectConnectGateways`, `VirtualInterfaces`, and `DirectConnectGatewayAttachments` from AWS JSON exports
2. In `TransitGateway.java`, handle `DIRECT_CONNECT_GATEWAY` attachment type — create a virtual interface toward the DXGW, wire BGP sessions over the VIF subnet, propagate TGW routes back toward the DX peer
3. Model the BGP session between the on-prem router and the VIF IP with the correct per-VIF Amazon ASN
4. Implement TGW route preference: DX-propagated routes should be preferred over VPN-propagated routes for the same prefix

dhalperi:
Models AWS Direct Connect Gateways attached to Transit Gateways. Previously, `TransitGatewayAttachment` entries with `ResourceType: direct-connect-gateway` were silently dropped. This PR covers Transit VIFs end-to-end (DXGW → TGW → VPC). Private VIFs landing on a DXGW are partially wired (customer-side BGP only, no path through the VGW); Private VIFs landing directly on a VGW and Public VIFs are not modeled.

The DXGW node runs eBGP toward the customer over the VIF subnet and eBGP unnumbered toward each TGW peer over link-local. Routes flow via BGP, not pre-installed statics — reachability tracks session state. Each association's `AllowedPrefixesToDirectConnectGateway` is originated as static null-routes (advertised to on-prem, matching AWS) and used as the import filter for routes redistributed to each TGW.

AWS's `static > DX > VPN` preference is encoded as BGP local-preference on the TGW's DX import policy. The 7224:7300/7200/7100 communities map to 300/200/150; all exceed VPN's default 100, so DX wins unconditionally while customers retain control over active/active vs active/passive among DX paths. Each TGW route table is a separate VRF; the DXGW↔TGW BGP session runs only in the VRF the DX attachment is associated with, so VPCs propagated to a different route table are not visible on the DXGW.

DXGWs are global; the same JSON entry appears in every region while associations and VIFs are regional. `DirectConnectGatewayConverter` runs after the per-region pass, dedups DXGWs by id, aggregates associations and VIFs across regions, and produces one Configuration node per unique DXGW. `TransitGateway.connectDirectConnect` and the AmazonSideAsn lookups also query across regions, since the matching association may live in a region different from the TGW's "winning" region after `TransitGatewayConverter` deduplication.

DXGW node names are prefixed with `dxgw-` because AWS returns IDs as bare UUIDs and the pybatfish node specifier grammar disallows leading digits in unquoted names. The `DirectConnectGatewayName` is set as the Configuration human name when no `Name` tag is present. VIF runtime status fields (`BgpStatus`, `BgpPeerState`) are intentionally ignored — Batfish models configuration, and reading operational status would block what-if analysis.

`AwsConfigurationDirectConnectTest` covers an end-to-end single-region snapshot (on-prem, DXGW, TGW, two VPCs, a competing VPN), verifying path traces, multi-RT isolation, DX local-pref tagging, and the community-to-local-pref mapping. `AwsConfigurationDirectConnectMultiRegionTest` adds a two-region snapshot with the DXGW duplicated and associations only in us-east-1 listing both `(DXGW↔tgw-east)` and `(DXGW↔tgw-west)`, asserting the DXGW has interfaces toward both TGWs. `DirectConnectGatewayConverterTest` adds unit coverage for the global converter.

`docs/aws_network_modeling.md` adds section 6.9.

Fixes #9938.


gh-ajinks (commented):
Hi @dhalperi — thanks for this PR. I tested it against our full snapshot (7 DXGWs / 17 available VIFs / 3 TGWs / ~30 on-prem configs) and the AWS-internal modeling looks great: DXGWs become nodes, both us-west-1 and us-west-2 TGWs see the DX attachments, and the route propagation through DXGW shows up alongside VPC/VPN attachments. This closes the structural gap called out in #9938.

There's a piece I think is needed before the new behaviors are actually exercisable, though: per-VIF customer-side BGP endpoints. Concretely:

- All 17 `available` VIFs in our snapshot are read and stored in `_directConnectVirtualInterfaces`, but no per-VIF interfaces appear on the DXGW node — only `[bgp-loopback]` plus one `[tgw-<id>-tgw-rtb-<id>]` per associated TGW route table.
- The customer side is already a real device in our snapshot. The Arista switches that terminate our DX cross-connects have the per-VIF dot1Q sub-interfaces (one per VLAN ID returned by the VIF API), BGP peer config pointing at the `AmazonAddress` /30 with the expected `local-as` / `remote-as`, and route-maps for community-based path selection. Batfish parses all of this from EOS today. The only reason these BGP sessions come back `NOT_COMPATIBLE` with `Remote_Node = None` is that no Batfish node has the corresponding `AmazonAddress` IP, so the on-prem peer config has nothing to resolve against.
- The two new behaviors this PR ships — DX-over-VPN route preference, and AWS TE community modeling (7224:7100/7200/7300) — are technically active but unobservable in our snapshot. TGW route tables show 0 DX-attached prefixes across all 3 TGWs (the DXGW attachment has nothing to propagate without an upstream BGP peer), and the DXGW BGP RIB has 41 entries with 0 carrying any 7224 community. So #9938 is closed structurally, but a regression test that exercises the preference logic against a real customer snapshot needs the VIF customer endpoint to be wired up.

The natural fit looks a lot like the existing `VpnConnections` pattern: each VPN connection produces a customer-gateway-facing interface on the AWS side (CGW interface IP, tunnel inside-IP, ASN) and operators add an L1 topology edge from their on-prem device to that interface. Mirroring that for DX would be roughly:

- For each `DirectConnectVirtualInterface` in `available` state, add an interface to the node it terminates on (DXGW for `transit`/`private` with a `DirectConnectGatewayId`, VGW for `private` with a `VirtualGatewayId`) using `AmazonAddress` as the interface IP and the VIF's `Vlan` as a dot1Q encapsulation hint.
- Create a BGP peer on that node bound to that interface, with the `AmazonSideAsn` from the parent DXGW and an expected-remote of the VIF's `Asn` and `CustomerAddress`.
- Name the interface after the VIF ID (e.g. `dxvif-fgrh4855`) so operators can reference it from their L1 topology JSON.

With those pieces in place, no synthetic customer-side modeling would be needed — the Arista (or any vendor's) parsed BGP config would resolve directly against the AWS-side endpoint via L1 edges, and BGP would establish naturally. Happy to either fold this into this PR or pick it up as a follow-up — I have a reproducible snapshot with mixed transit/private VIFs across four physical DX connections and seven DXGWs that would make useful test coverage.

(Also left three inline comments on the diff for a couple of smaller items I hit during testing.)

gh-ajinks:
Quick update after deeper testing — the gap I described above is real but the framing was off.

Re-reading your description: "The DXGW Configuration node runs eBGP toward the customer router over the VIF subnet" + "covers Transit VIFs end-to-end." That's exactly what we want, and `configureVifBgpSession` does add the VIF interface with `AmazonAddress` + BGP peer toward `CustomerAddress`. The asked-for design is shipped.

What I'm actually seeing in our snapshot: `configureVifBgpSession` is not firing for our Transit VIFs.

Concretely, we have 3 `available` Transit VIFs whose `DirectConnectGatewayId` matches the DXGW id `085f3c96-...`. After running this PR's image against our snapshot, the resulting DXGW node has only 3 interfaces (bgp-loopback plus one per associated TGW route table) — no `dxvif-*` interfaces, and the BGP peer count on the node is 2 (the TGW peers) instead of 5 (TGW peers + 3 VIF peers).

I think the cause is at `Region.java:861`:

```java
for (DirectConnectGateway dcGw : getDirectConnectGateways().values()) {
    Configuration cfgNode = dcGw.toConfigurationNode(this, awsConfiguration, warnings);
    awsConfiguration.addNode(cfgNode);
}
```

The DXGW is built per-Region using the calling Region's `_directConnectVirtualInterfaces`. DXGWs are global — the same DXGW JSON entry shows up in every region we collect from (us-west-1, us-west-2, eu-west-2 for us). VIFs are regional and only exist in us-west-1 in our case. So the build runs three times, and only the us-west-1 invocation has matching VIFs to wire. The us-west-2 / eu-west-2 invocations match zero VIFs and produce a DXGW without VIF interfaces. Whichever region processes last via `awsConfiguration.addNode(cfgNode)` overwrites the earlier one, and we lose the wired version.

Your unit test (`AwsConfigurationDirectConnectTest`) uses a single region with the DXGW + VIFs + TGW + VPCs all co-located, which wouldn't surface this. Easiest repro is probably to duplicate the DXGW JSON into a second region's `aws_configs/` directory in the test fixture and assert that the VIF BGP session still configures.

Fix likely needs to either (a) hoist DXGW configuration node creation out of the per-Region loop into `AwsConfiguration` and aggregate VIFs across regions before wiring, or (b) make the DXGW configuration node merge VIF interfaces across regions when `addNode` sees a duplicate id.

The other 6 DXGWs in our snapshot carry only Private VIFs, which the PR intentionally doesn't wire yet (per your description), so they're not relevant to the bug report above. As a separate ask, Private VIF wiring would matter for us — Private VIFs are the bulk of our DX surface (13 of 17 available across multiple per-tenant DXGWs), and one of them carries a workload where end-to-end modeling has significant interest at this time. Curious if the lift to pull that into this PR is possible.

---

**Update (verified):** stripped the duplicate `DirectConnectGateways.json` and `VirtualInterfaces.json` from non-VIF regions in the snapshot — keeping the global DXGW + VIF data only in the region where the physical DX terminates. The DXGW now has the expected per-VIF interfaces (named after the VIF IDs, each with the correct `AmazonAddress`) and the matching customer-facing eBGP sessions on the /29 subnets. So `configureVifBgpSession` fires correctly when the DXGW is built in a region whose `_directConnectVirtualInterfaces` is non-empty — confirming the per-region build at `Region.java:861` is the issue.

One side observation that makes me think the region-scope leakage is broader than just VIFs: the original snapshot wired 2 TGW interfaces on the DXGW; the workaround wires only 1, and which one depends on whether `DirectConnectGatewayAssociations.json` is present in non-VIF regions. The DXGW↔TGW interface creation appears to depend on the TGW being built in a region that also has the association data. This suggests fix (a) — hoist DXGW build out of the per-Region loop and aggregate across regions — is probably the more correct shape than (b).

gh-ajinks (commented):
`State` / `AssociationState` field mismatch — the parser reads a field that doesn't exist in real boto3 output, and the test fixture includes both forms which hides the bug from CI.

projects/batfish/src/main/java/org/batfish/representation/aws/Region.java
```diff
@@ -328,6 +344,36 @@ public static RegionBuilder builder(String name) {
...
+          String state = json.get("DirectConnectGatewayState").textValue();
+          if (state.equals(AwsVpcEntity.STATE_AVAILABLE)) {
+            DirectConnectGateway dcGateway =
+                BatfishObjectMapper.mapper().convertValue(json, DirectConnectGateway.class);
+            _directConnectGateways.put(dcGateway.getId(), dcGateway);
+          }
+        };
+      case DirectConnectGatewayAssociation.JSON_KEY_DIRECT_CONNECT_GATEWAY_ASSOCIATIONS:
+        return json -> {
+          String state = json.get(AwsVpcEntity.JSON_KEY_STATE).textValue();
```
gh-ajinks: NPE here when parsing real boto3 output. `DescribeDirectConnectGatewayAssociations` returns `associationState`, never `state` — your test fixture has both fields, which masks this. Suggest reading `AssociationState` instead (e.g. add `JSON_KEY_ASSOCIATION_STATE = "AssociationState"` on `DirectConnectGatewayAssociation`).

projects/batfish/src/test/resources/org/batfish/representation/aws/test-direct-connect-gateway/aws_configs/DirectConnectGatewayAssociations.json
```diff
@@ -0,0 +1,21 @@
...
+   "AssociationId": "dxgw-assoc-dx01",
+   "DirectConnectGatewayId": "dxgw-dx01",
+   "AssociatedGateway": {
+    "Id": "tgw-dx01",
+    "Type": "transitGateway",
+    "OwnerAccount": "123456789012",
+    "Region": "us-east-1"
+   },
+   "AssociationState": "associated",
+   "State": "associated",
```
gh-ajinks: This field doesn't exist in real `DescribeDirectConnectGatewayAssociations` responses (only `AssociationState` does). Including both here is what hides the NPE in `Region.java:358` — recommend dropping it so the fixture matches what `boto3` actually returns.

projects/batfish/src/main/java/org/batfish/representation/aws/DirectConnectGateway.java
```diff
@@ -0,0 +1,324 @@
...
+  public long getAmazonSideAsn() {
+    return _amazonSideAsn;
+  }
+
+  public @Nonnull Map<String, String> getTags() {
+    return _tags;
+  }
+
+  static String nodeName(String directConnectGatewayId) {
+    return directConnectGatewayId;
```
gh-ajinks: Using the raw DXGW ID as a node name produces UUID-form names like `085f3c96-59d9-4e6a-a37a-982e8f5653fc`, which break the pybatfish node specifier grammar — every reference needs `'"..."'` quoting or returns HTTP 400 "Error parsing ... as nodeSpecifier after index 0". Suggest prefixing like `tgw-...` already does (e.g. `dxgw-085f3c96` from the first 8 hex of the ID), or preferring the `Name` tag if present. Same applies to `DirectConnectGatewayAssociation` IDs which are also UUIDs.

gh-ajinks (commented):
DXGW node names use the raw UUID returned by the AWS API, which breaks the pybatfish node specifier grammar — every reference needs explicit quoting or returns HTTP 400.

ratulm (commented):
:lgtm:

> @ratulm reviewed 39 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9948#-:-Ot8ArH0APA02MVdhwH5:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on dhalperi).

projects/batfish/src/main/java/org/batfish/representation/aws/Region.java
dhalperi: Fixed in d1375bf655 — the parser now reads `JSON_KEY_ASSOCIATION_STATE` ("AssociationState") on `DirectConnectGatewayAssociation`. Verified against the [API docs](https://docs.aws.amazon.com/directconnect/latest/APIReference/API_DirectConnectGatewayAssociation.html): only `associationState` exists on this object — there is no `state` field. Thanks for catching this.

projects/batfish/src/test/resources/org/batfish/representation/aws/test-direct-connect-gateway/aws_configs/DirectConnectGatewayAssociations.json
dhalperi: Fixed in d1375bf655 — dropped the spurious `State` field from this fixture and from `DirectConnectGatewayAssociationTest.json`. Combined with the parser fix above, the test fixtures now exercise the same field name as real API output.

projects/batfish/src/main/java/org/batfish/representation/aws/DirectConnectGateway.java
dhalperi: Fixed in d95e7aa938. `DirectConnectGateway.nodeName()` now prefixes the ID with `dxgw-` so the node name starts with a letter. The actual specifier-grammar constraint is the leading digit (it disallows leading digits in unquoted names to avoid ambiguity with IP addresses) — embedded hyphens parse fine. Also set the human name from `DirectConnectGatewayName` (always present on the API response) when no `Name` tag is set, so users see a meaningful label rather than just the UUID. Updated the integration- and unit-test fixtures to use UUID-shaped IDs matching real AWS output, so the test exercises the leading-digit case. DXGW association IDs already carry a `dxgw-assoc-` prefix in real API output, so they parse fine without further treatment.

dhalperi:
Re: the multi-region scoping issue you flagged in the follow-up — fixed in 0b9a9182ab (`AWS: build Direct Connect Gateways once globally, not per-region`).

You diagnosed it exactly: the per-region `toConfigurationNodes()` was rebuilding the DXGW node from each regions local VIFs/associations only, and the last `addNode` won. Same root cause for the DXGW↔TGW interface gap: `TransitGateway.connectDirectConnect` was looking up the association in the TGWs region only, which after `TransitGatewayConverter` deduplication might not be the region that actually carries the association.

Fix mirrors `TransitGatewayConverter`: a new `DirectConnectGatewayConverter` runs after the per-region pass, walks all accounts/regions, dedups DXGWs by id, aggregates VIFs and associations across regions, and produces one Configuration node per unique DXGW. `connectDirectConnect` and the ASN lookups now query across all regions via two new `AwsConfiguration` helpers (`getDirectConnectGateway`, `getDirectConnectGatewayAssociations`).

Two new tests:
- `DirectConnectGatewayConverterTest` — DXGW in 3 regions, VIFs/associations in 1; asserts the resulting node has the VIF interface and allowed-prefix static.
- `AwsConfigurationDirectConnectMultiRegionTest` — 2-region snapshot, global DXGW duplicated, TGWs in each, associations file only in us-east-1 listing both `(DXGW↔tgw-east)` and `(DXGW↔tgw-west)`. Asserts the DXGW has interfaces toward both TGWs. Confirmed this test fails when `connectDirectConnect` is reverted to per-region scope.

The remaining piece you flagged — per-VIF customer-side BGP endpoints (so on-prem routers can resolve against the AWS-side `AmazonAddress`) — is more substantial and likely a follow-up PR. Happy to fold it in here if you prefer; otherwise Ill open a separate one once this lands.

dhalperi (commented):
> @dhalperi reviewed 52 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9948)* status: all files reviewed, 3 unresolved discussions (waiting on gh-ajinks).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9948)* status: all files reviewed, 3 unresolved discussions (waiting on gh-ajinks).

dhalperi (commented):
> @dhalperi resolved 3 discussions and dismissed @gh-ajinks from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9948)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9949, opened 2026-05-20
Junos: model commit-time and runtime semantics of policy-statement then actions

### LINKED ISSUE #1551: Juniper: policy terms should use slots instead of a list of PsFrom/PsThen
The following 3 lines can be typed at an SRX console:

```
set policy-options policy-statement POLICY-NAME term TERM then metric 10
set policy-options policy-statement POLICY-NAME term TERM then metric add 20
set policy-options policy-statement POLICY-NAME term TERM then metric add 20
```

The running config after these three steps will be:

1. only `metric 10`
2. only `metric add 20` . (not 10 + 20)
3. only `metric add 20` . (not 10 + 20 + 20 and not 20 + 20)

That is, these lines all replace each other.

Batfish will add the 3 lines as three consecutive statements, which is not a faithful implementation.

This bug is currently low priority as the current behavior _is correct_ on configs pulled from routers (either archived or post-validation).

dhalperi:
Refactors PsThens to faithfully model Junos commit-time collapse and
runtime ordering for actions within a single `then` block, as
characterized by the junos_then_action_conflicts lab in lab-validation.

Last-wins / cumulative families:

- Scalar/numeric families (origin, preference, tag, next-hop, external,
local-preference, metric, load-balance, as-path-prepend,
as-path-expand, source-class, etc.) use last-wins semantics: only the
last action is retained, exact duplicates collapse to one.
- Community actions (set/add/delete) and tunnel-attribute set are
cumulative: all distinct actions are retained, duplicates dedup.
Multiple community-set actions in one term are converted with union
semantics (matching Junos, where every set contributes cumulatively).
- getAllThens() returns actions in a canonical order matching Junos
commit-time normalization (as-path-prepend before as-path-expand,
side-effects before terminators).
- Adds equals/hashCode to PsThen subclasses that lacked it, enabling
proper dedup detection.

Flow-control terminators:

- Bare `accept`/`reject` form a "bare-terminator" family with last-wins
semantics: `accept; reject` collapses to `reject`-only, and
vice-versa. `default-action accept`/`reject` and `next policy` are
separate families and are retained alongside bare terminators.
- `then next term` is now modeled (was a no-op TODO since 2017,
closes #1551). It is its own family. At runtime, `next term`
suppresses any bare `accept`/`reject` in the same term: both lines
are retained at commit, but the route falls through to the next term
regardless of source order. This is enforced in
JuniperConfiguration.toStatements (which strips bare terminators
when next-term is present) and in getTerminalAction (which no longer
flags such terms as terminal for dead-code detection).

Risky parse warnings:

- Same-family overwrites emit "Overwriting existing then <family>".
- Same-family duplicates emit "Duplicate then <family>".
- Community add/delete on the same name emit "Conflicts with prior
then community add/delete <name>" when both are retained but their
combined runtime effect cancels out.
- Adding `then next term` after a prior bare `accept`/`reject` emits
"then next term suppresses prior then accept/reject ... route falls
through". Adding `then accept`/`reject` after a prior `then next
term` emits the converse "then accept/reject has no effect when then
next term is also present". These asymmetric messages point at the
removable line in each direction.

----

Prompt:
```
Read the new junos_then_action_conflicts lab in lab-validation
(../lab-validation). That has a lot to say about Juniper
policy-statements with duplicate/conflicting terms. Let's add unit
tests to Batfish that mirror the lab and make sure Batfish implements
the ordering correctly. A few notes:

1. We should emit RISKY warnings every time one of these overwrites,
including dedupes, happens.
1. We should emit RISKY warnings if both statements are preserved, but
don't both have an effect. E.g., two community set lines.
1. We should match Junos behavior, but it's preferred to eliminate
statements that don't have an effect (like lines that have been
overwritten) during conversion.
```

Follow-ups in the same commit responding to lab-validation README
updates: add bare-terminator last-wins family, model `then next term`
and its conflict with bare terminators (both retained at commit, route
falls through at runtime), add unit tests for `default-action`
overridden by bare terminator, and add asymmetric cross-family RISKY
warnings for `next term` + bare terminator interactions.

---
**Stack**:
- #9949 ⬅
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -1214,6 +1214,7 @@ import org.batfish.representation.juniper.PsThenNextHopPeerAddress;
 import org.batfish.representation.juniper.PsThenNextHopReject;
 import org.batfish.representation.juniper.PsThenNextHopSelf;
 import org.batfish.representation.juniper.PsThenNextPolicy;
+import org.batfish.representation.juniper.PsThenNextTerm;
 import org.batfish.representation.juniper.PsThenOrigin;
 import org.batfish.representation.juniper.PsThenPreference;
 import org.batfish.representation.juniper.PsThenReject;
@@ -2569,12 +2570,25 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
 
   private void addPsThen(PsThen then, ParserRuleContext ctx) {
     List<String> cleared = _currentPsThens.addPsThen(then);
-    if (!cleared.isEmpty()) {
-      warnRisky(
-          ctx,
-          String.format(
-              "Overwriting existing %s %s",
-              cleared.size() > 1 ? "thens" : "then", String.join(", ", cleared)));
+    for (String entry : cleared) {
+      if (entry.contains("(suppressed-by-next-term-or-policy)")) {
+        warnRisky(
+            ctx,
+            "then next term/next policy suppresses prior then accept/reject in the same term:"
+                + " accept/reject does not fire");
+      } else if (entry.contains("(dead-after-next-term-or-policy)")) {
+        warnRisky(
+            ctx,
+            "then accept/reject has no effect when then next term/next policy is also present in"
+                + " the same term: accept/reject does not fire");
+      } else if (entry.contains("(dedup)")) {
+        warnRisky(ctx, String.format("Duplicate then %s", entry.replace(" (dedup)", "")));
+      } else if (entry.contains("(conflict)")) {
+        warnRisky(
+            ctx, String.format("Conflicts with prior then %s", entry.replace(" (conflict)", "")));
+      } else {
+        warnRisky(ctx, String.format("Overwriting existing then %s", entry));
+      }
     }
   }
 
@@ -7012,10 +7026,7 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
 
   @Override
   public void exitPopst_next_term(Popst_next_termContext ctx) {
-    // The next-term action itself is a no-op in Batfish, so we do not model this behavior.
-    //
-    // TODO(https://github.com/batfish/batfish/issues/1551): need to implement next_term replacing
-    // any existing flow control action.
+    addPsThen(PsThenNextTerm.INSTANCE, ctx);
   }
 
   @Override
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
@@ -3329,11 +3329,22 @@ public final class JuniperConfiguration extends VendorConfiguration {
         || then instanceof PsThenNextPolicy;
   }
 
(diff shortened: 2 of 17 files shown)


dhalperi (commented):
> @dhalperi reviewed 17 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9949)* status: all files reviewed, 3 unresolved discussions.

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/PsThens.java` line 188 at r1](https://reviewable.io/reviews/batfish/batfish/9949#-Ot68-Yf2asi76jy3Znk:-Ot68-Yf2asi76jy3Znl:b-li0kih) ([raw file](https://github.com/batfish/batfish/blob/0a0d432261c3e2c17154b63034fafc0ec50c17a4/projects/batfish/src/main/java/org/batfish/representation/juniper/PsThens.java#L188)):*
> ```Java
>   }
> 
>   private static boolean isCumulativeFamily(String family) {
> ```

I think this is bogus, there's only one tunnel attribute on a BGP message so this is like duplicate community set. They both stay in config, but the last one wins.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java` line 4967 at r1](https://reviewable.io/reviews/batfish/batfish/9949#-Ot67FC_3Iv5eqFGY8Ad:-Ot67FC_3Iv5eqFGY8Ae:br7kel3) ([raw file](https://github.com/batfish/batfish/blob/0a0d432261c3e2c17154b63034fafc0ec50c17a4/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java#L4967)):*
> ```Java
>     // `next term` overrides any bare accept/reject in the same term: at runtime the route falls
>     // through, so the term is not actually terminal. next-policy still terminates this policy.
>     if (allThens.stream().anyMatch(t -> t instanceof PsThenNextTerm)) {
> ```

I don't get this. Why are we picking next policy  only if next-term is present. We don't even know which one wins do we?

I kinda think we need to model these terminal actions as separate fields on PsThen rather than lumping them in with the other thens.
___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/PsThens.java` line 249 at r1](https://reviewable.io/reviews/batfish/batfish/9949#-Ot68Gkn4gZdt7deqXZW:-Ot68Gkn4gZdt7deqXZX:b-lg0ka7) ([raw file](https://github.com/batfish/batfish/blob/0a0d432261c3e2c17154b63034fafc0ec50c17a4/projects/batfish/src/main/java/org/batfish/representation/juniper/PsThens.java#L249)):*
> ```Java
>     // Exact duplicate of any prior community action: dedup, drop the new one
>     if (_communityActions.contains(then)) {
>       return ImmutableList.of(communityLabel(then) + " (dedup)");
> ```

don't buy it. add 5, delete 5, add 5 will not be deduped, it will result in add.

dhalperi (commented):
> @dhalperi reviewed 4 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9949)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9949)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9950, opened 2026-05-21
Junos: warn on dead-config in policy-statement then actions

dhalperi:
Adds two RISKY warnings for `policy-statement` `then` blocks where
multiple actions in one term combine to leave config that does not fire
at runtime:

- **mutating action + bare `reject`**: when a term contains a bare
`reject` that actually fires (i.e. not suppressed by `next term`/
`next policy`), every mutating action in the same term (metric,
community add/delete/set, local-preference, etc.) is dead — the route
is discarded so changes don't propagate. One warning per dead line:
"then X has no effect on the propagated route when then reject is
also present in the same term".

- **`default-action` + bare terminator**: when a term contains both
`default-action accept`/`reject` and a bare `accept`/`reject` (in
either order), the bare terminator wins at runtime and the
`default-action` is dead. Warning: "then default-action is overridden
by bare then accept/reject in the same term: default-action does not
fire".

While here, treat `default-action accept`/`reject` as their own
last-wins family so that `default-action accept; default-action reject`
emits an "Overwriting existing then default-action" warning and
duplicates dedup. This catches a real conflict in the existing
`juniper_policy_statement` parsing fixture (lines 26-27).

`PsThenDefaultAction{Accept,Reject}` gain `equals`/`hashCode` so the
dedup detection works.

----

Prompt:
```
In the recent commit, we really improved Junos modeling of PsThen. I
had a few questions about risky warnings.

- If the terminal action of a policy is "then reject", but there are
mutating actions (not logging or sampling, but then metric for
example) this should be a warning -- mutating in the same term as
rejecting.
- If the term is definitive (accept/reject) then
default-action(-accept/-reject) is a no-op and should be a warning.
- Look around - are there other candidates?

Research and implement if missing
```

---
**Stack**:
- #9950 ⬅
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -2581,6 +2581,18 @@ public class ConfigurationBuilder extends FlatJuniperParserBaseListener
             ctx,
             "then accept/reject has no effect when then next term/next policy is also present in"
                 + " the same term: accept/reject does not fire");
+      } else if (entry.contains("(dead-with-bare-terminator)")) {
+        warnRisky(
+            ctx,
+            "then default-action is overridden by bare then accept/reject in the same term:"
+                + " default-action does not fire");
+      } else if (entry.contains("(dead-with-bare-reject)")) {
+        warnRisky(
+            ctx,
+            String.format(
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9950)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9951, opened 2026-05-21
Junos: warn when then community add/set overlaps then community delete

dhalperi:
Detects same-term `then community add X; then community delete Y` (and
`set X; delete Y`) sequences where X and Y can match overlapping
communities, so the just-added community is silently removed by the
later delete. The bug is easy to write by accident when `delete Y` is a
broad regex (e.g. `.*:.*` or `^65000:`) intended to scrub inbound
communities while a later term re-adds policy-specific markers.

Junos asymmetry between set/add and delete is honored:

- `then community set/add` only contributes literal community members;
regex members are skipped at runtime, so they are not flagged as
adders here.
- `then community delete` matches both literal members (by equality)
and regex members (substring `Pattern.find()` on the colon-separated
community string, matching the runtime evaluator).

Regex-vs-regex (e.g. `add ALL_65000; delete ALL`) is intentionally not
flagged: Junos treats regex `add`/`set` as a no-op, so there is no
literal in the route to flag. Extended/large communities are not
covered for the same reason as the existing risky-regex check.

Wired in via `JuniperConfiguration.toRoutingPolicy` next to
`detectMisplacedTerminalActions`, so existing PsThens parse-time
warnings (overwrite/dedup/conflict) are unchanged.

----

Prompt:
```
Can we extend our regex-based validation to detect bugs like this?

In Junos there are `community set`, `community delete`, and `community
add` actions that you can do in a policy term. There are times when
performing some of these is effectively a no-op, or can be used in an
incorrect order leading to unwanted results.

For example, when adding (or setting) a community then deleting via a
regex that matches it, we end up removing the community even though we
clearly wanted to keep it on the route. Two examples:

term A {
then {
community set COMM1;
community delete ALL-COMMUNITIES;
}
}

term A {
then {
community add COMM1;
community delete ALL-COMMUNITIES;
}
}

Where `ALL-COMMUNITIES` is a regex `.*:.*`, we have effectively deleted
COMM1 even though we clearly wanted to set/add it. The same applies if
the deleting community is a more specific regex but happens to also
match the value of COMM1.

Implement this as a risky warning, including overlap with set.
```

---
**Stack**:
- #9951 ⬅
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/CommunityActionOverlap.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/CommunityActionOverlap.java
@@ -1,0 +1,75 @@
+package org.batfish.representation.juniper;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
+import org.batfish.datamodel.bgp.community.Community;
+
+/**
+ * Detects overlap between {@link NamedCommunity} structures used in same-term {@code then community
+ * add/set} and {@code then community delete} actions in a Junos {@code policy-statement}.
+ *
+ * <p>Junos applies {@code then} actions in declared order at runtime. When a {@code community
+ * add}/{@code community set} contributes a community that a later {@code community delete} also
+ * matches, the result is that the just-added community is silently removed. This is rarely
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9951)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

Issue #9933, opened 2026-05-05
PAN-OS: vlan-type (SVI) interfaces on layer2 ae trunks not modeled


gh-ajinks:
## Summary

Batfish does not model PAN-OS `vlan-type` interfaces — the SVI architecture
where `ae` interfaces operate in `layer2` mode and each VLAN's IP address is
carried on a separate `vlan.X` unit. This is a common design on PA firewalls
that aggregate multiple VLANs through a single port-channel.

## Configuration pattern (anonymized)

```
ethernet-interface {
  ae2 {
    layer2 { ... }          # ae in layer2 mode
  }
}
vlan {
  vlan.10 {                 # vlan-type SVI — carries the IP
    ip { 192.0.2.1/29; }
  }
  vlan.20 {
    ip { 198.51.100.1/24; }
  }
}
zones {
  production {
    network { layer3 vlan.10; }   # zone references the vlan-type interface
  }
}
virtual-router {
  default {
    interface [ ... vlan.10 vlan.20 ... ];  # VR also references vlan-type
  }
}
```

## Impact

1. **Zero routed interfaces modeled** — `vlan-type` interfaces are silently
   ignored, so the device appears to have no routed interfaces for those VLANs.
2. **Zone membership broken** — zones reference `vlan.X` by name; since the
   interface is not modeled, Batfish marks any ae sub-interfaces on the same
   device as zoneless and denies all egress traffic (including BGP).
3. **BGP sessions NOT_ESTABLISHED** — BGP peers whose local address is on a
   `vlan-type` interface are blocked at the egress filter:
   `"Cannot exit interface ... because it is not in a zone"`.

## Expected behavior

`vlan-type` interfaces should be modeled as routed L3 interfaces with their
configured IP addresses, VRF memberships, and zone assignments respected —
equivalent to how `ae1.X` tagged layer3 sub-interfaces are already handled.

## Relationship to #3952

PR #3952 (2019) added grammar rules for `sni_vlan` / `sniv_unit` so configs
containing `vlan` interface blocks no longer produce parse warnings. However,
the vlan-type interface is still silently ignored in the VI model — the test
case in that PR only exercises a `comment` inside the vlan block. No IP
address extraction, VRF membership, zone assignment, or routing was
implemented. This issue tracks the missing modeling step.

## Workaround

Pre-process configs before import: extract each `vlan.X` IP and inject it into
the corresponding `ae.X` tagged sub-interface, convert the ae from `layer2` to
[...]

PR #9952, opened 2026-05-21
PAN-OS: model vlan-type interface units with IP, MTU, and SVI VLAN ID

gh-ajinks:
PR #3952 (2019) added grammar for `vlan.X` units under `network interface vlan`, but the rule only accepted `if_common` (comments) and the visitor emitted a TODO without building anything. As a result, configs of the form

```
vlan {
  units {
    vlan.10 { ip { 192.0.2.1/29; } }
  }
}
```

parsed without warnings but produced no L3 interface — BGP local-address lookups against `vlan.X` failed to resolve, zones referencing `layer3 vlan.X` silently dropped the membership, and any traffic targeted at the SVI's IP was dropped at zone egress with `Cannot exit interface ... because it is not in a zone`.

This extends `sniv_unit` to accept `sniv_ip` and `sniv_mtu` (mirroring `snit_unit` for tunnel.X), wires the visitor to call `addAddress` / `setMtu` on the already-created `Interface(VLAN_UNIT)`, adds a `VLAN_INTERFACE_ADDRESS` structure usage for reference tracking, and in `PaloAltoConfiguration.toInterface` parses the numeric suffix of the unit name as the SVI's VLAN ID and surfaces it via `Interface.setVlan`. The existing VI→Configuration translation and the `Type.VLAN_UNIT → InterfaceType.VLAN` mapping already do the rest: the SVI shows up in `Configuration.getAllInterfaces`, gets its VRF binding from the virtual-router's interface list, and gets its zone binding from `zones.<name>.network.layer3`.

Out of scope: the broadcast-domain → SVI bridge through PAN-OS layer2 ae trunks. Today PA models each `ae2.X` sub-interface as its own `Type.LAYER2` access-mode interface (`accessVlan=X`) rather than treating the parent ae as a trunk port with `allowedVlans` aggregated from its sub-interfaces. With this PR the SVI carries the right VLAN ID, but Batfish's broadcast-domain logic typically bridges via TRUNK ports, so the data-plane hand-off from a layer2 ae to `vlan.X` will need a follow-up reshape on the layer2 side. Tracked separately in #9953.

Addresses #9933 (control-plane portion: BGP local-address resolution, zone membership, VRF binding, structure-reference tracking).

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4
@@ -271,12 +271,12 @@ snit_units
     UNITS snit_unit?
 ;
 
-sniv_ip
+snivu_ip
 :
     IP address = interface_address_or_reference
 ;
 
-sniv_mtu
+snivu_mtu
 :
     MTU mtu = uint32
 ;
@@ -286,8 +286,8 @@ sniv_unit
     name = variable
     (
         if_common
-        | sniv_ip
-        | sniv_mtu
+        | snivu_ip
+        | snivu_mtu
     )?
 ;
 
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4
@@ -271,11 +271,23 @@ snit_units
     UNITS snit_unit?
 ;
 
+sniv_ip
+:
+    IP address = interface_address_or_reference
+;
+
+sniv_mtu
+:
+    MTU mtu = uint32
+;
+
 sniv_unit
 :
     name = variable
     (
         if_common
+        | sniv_ip
+        | sniv_mtu
     )?
 ;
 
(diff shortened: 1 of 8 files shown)


dhalperi (commented):
> @dhalperi reviewed 6 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9952)* status: all files reviewed, 1 unresolved discussion (waiting on gh-ajinks).

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4` line 289 at r1](https://reviewable.io/reviews/batfish/batfish/9952#-OtCuiwG6lrCcY8z-CVR:-OtCuiwG6lrCcY8z-CVS:begyazk) ([raw file](https://github.com/batfish/batfish/blob/85fd07d3a53d4cbd25ffe7407ce591c0c3a86e84/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_interface.g4#L289)):*
> ```ANTLR
>     (
>         if_common
>         | sniv_ip
> ```

nit: this should be `snivu_ip` as it's under `sniv_unit`. I don't believe this will be valid at the `sniv_` level, right? You would have to be on a numbered vlan unit.

dhalperi (commented):
> @dhalperi reviewed 6 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9952)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #9964, opened 2026-05-26
Junos: don't crash on community regex with dk.brics-reserved chars

dhalperi:
The community regex risk-analysis path constructs dk.brics RegExp objects
from user-supplied Java-regex syntax. Chars like `<`, `>`, `~`, `&`, `@`,
`#`, and `"` are literals in Java regex but are reserved in dk.brics
syntax (numeric ranges, complement, intersection, etc.). A Junos config
with `community FOO members 12345<?:1` thus crashed the parser with
`IllegalArgumentException: expected '>'`, propagating up to a
`BatfishParseException` that caused the entire snapshot to fail to load
even in lenient mode.

Escape these chars before feeding the user's regex to dk.brics so they
are treated as literals — matching what java.util.regex does at runtime.
The risk-analysis warnings now fire on the surrounding regex parts
(e.g., `12345<?:1` warns about unintended longer matches such as
`12345:10`) rather than crashing.

Verified on vJunos 25.4R1.12 that `set policy-options community FOO
members "12345<?:1"` is accepted by the CLI and passes commit check, so
Batfish must handle such configs gracefully.

----

Prompt:
```
A config file is purported to crash Batfish's junos parser. Create a
minimal repro, then use red/green TDD to fix it.
```

---
**Stack**:
- #9964 ⬅
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/RegexCommunityMember.java
@@ -8,6 +8,7 @@ import dk.brics.automaton.Automaton;
 import dk.brics.automaton.RegExp;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 import org.batfish.datamodel.bgp.community.Community;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9964)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9966, opened 2026-05-26
PaloAlto: resolve address-object refs in BGP peer/local-address and static-route nexthop

### LINKED ISSUE #8934: PaloAlto - parsing error when using address object as routing peer or nexthop

gh-ajinks:
## Summary

Resolves PAN-OS address objects used as BGP `peer-address`, BGP `local-address ip`, and static-route nexthop `ip-address` to their host IP. Before this, only literal `ip`/`prefix` values parsed in these positions, so a config referencing an address object there silently lost the peer/route.

Carries forward @natl-set's work in #9777 (which covered `peer-address` and static-route nexthop), rebased onto current master, with two additions:

- **`local-address ip` support.** `bgppgp_la_ip` now accepts `interface_address_or_reference`, mirroring `peer-address` — the last BGP address position that didn't resolve object references.
- **Mask-agnostic host resolution.** `ip-netmask` objects are commonly defined with their interface-subnet mask (e.g. `10.0.0.1/29`) but used as a single host elsewhere. Host use now takes the object's configured IP regardless of mask (`10.0.0.1/29` → `10.0.0.1`). A bare network/broadcast address, `ip-range`, or `fqdn` object still has no single host IP and is rejected with a warning.

#9777 has been inactive since March; opening this so the full set is testable together. Original authorship preserved via `Co-authored-by`.

Fixes #8934.

## Test plan

- [x] Extended `address-object-nexthop` testbed with a `/29`-masked object as `local-address ip`; asserts it resolves to the host IP and records a `BGP_PEER_LOCAL_ADDRESS_IP` reference
- [x] `address-object-nexthop-warnings` still warns on a network-address object
- [x] PaloAlto grammar + representation tests, `ref_tests`, and google-java-format pass
- [x] Validated against production PAN-OS configs (PA firewalls peering to Arista via `local-address` objects): affected BGP sessions now model as established


gh-ajinks:
Adding some real-world validation. I ran this branch against a production network snapshot — multi-site PAN-OS firewalls peering eBGP to Arista switches — with our config-preprocessing workarounds disabled, so Batfish parses the raw address-object references directly.

**Before:** every PA↔switch eBGP session whose PA side used an address object for `local-address ip` came back `NOT_COMPATIBLE` in `bgpSessionStatus`. The local address never resolved (`Local_IP` was null, with no interface to bind to), even though the real devices have the sessions established. The objects in question are `ip-netmask` objects carrying the interface's `/29`, referenced as `local-address ip <obj>`.

**With this branch:**
- The previously-broken PA↔switch sessions all move to `ESTABLISHED`; those were the only sessions whose status changed.
- Interface IPs defined via object references resolve as well — no interfaces dropped from the model (the #8934 case).
- No regressions elsewhere.

The masked-host case is the real-world driver: PAN-OS configs routinely define these objects with the interface-subnet mask (e.g. `10.0.0.1/29`), and a `/32`-only guard would reject them even though only the host IP (`10.0.0.1`) is used. This branch resolves them correctly.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9966)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #9969, opened 2026-05-27
Junos: drop "dead-with-bare-reject" risky warning

dhalperi:
The warning fired on patterns like `then community add RED; then reject`,
claiming the mutating action had no effect on the propagated route. In
practice this is a false positive: mutations applied before reject are
still observable in received-route data and other contexts where users
inspect what a policy did, so flagging them as risky just adds noise.

Drops the warning, the supporting helpers in PsThens
(appendBareRejectWarningIfMutating, mutatingLabels, isMutatingAction,
actionLabel), the corresponding ConfigurationBuilder branch, the unit
tests, and the FC-MUTATE-THEN-REJECT / FC-REJECT-THEN-MUTATE policies
in the juniper-ps-then-conflicts testconfig that only existed to drive
this warning. Other dead-config warnings (next-term suppresses
accept/reject, default-action with bare terminator, last-wins
overwrites, community conflicts, dedup) are unchanged.

----

Prompt:
```
I reviewed with users and the "has no effect on the propagated route
when then reject is also present" risky warning is false positive / not
useful. As one example, you can see the communities that are added in
received-route data, a common use case.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9969)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9973, opened 2026-06-01
caches: migrate Guava caches to Caffeine, ban Guava cache imports

dhalperi:
Guava's own javadoc recommends Caffeine as the successor to its caching
API (better performance, fewer bugs). Convert all Guava caches to
Caffeine and forbid future com.google.common.cache imports via
checkstyle.

Most conversions are mechanical: CacheBuilder.newBuilder() ->
Caffeine.newBuilder(), .build(CacheLoader.from(f)) -> .build(f), and
.getUnchecked(k) -> .get(k).

Two call sites needed adjustment:
- Batfish.loadDataPlane used Guava's get(key, Callable), which throws
ExecutionException. Caffeine's get(key, Function) cannot throw checked
exceptions, so the loader's IOException is wrapped in
UncheckedIOException and the now-dead try/catch is removed.
- CommunitySet's cache key is retyped from Set<Community> to
Set<? extends Community> to match the of(...) parameter. Under Guava
the looser getIfPresent(Object) hid the mismatch; Caffeine's typed
getIfPresent(K) made it explicit. The retype removes two casts at the
call sites, leaving one in the loader where an assert documents that
only ImmutableSet keys are ever inserted.

IpSpaceToBDD stays on Guava: its loader recurses into the same cache
(forbidden by Caffeine's ConcurrentHashMap.computeIfAbsent) and it
relies on a synchronous removalListener to free BDDs on the calling
thread, since BDDFactory is not threadsafe. It is exempted from the
import ban via a checkstyle SuppressionFilter.

----

Prompt:
```
It seems that Guava caches are slow and even they recommend we switch to
Caffeine on their wiki. Let's do that conversion globally, and then also
forbid importing Guava caches in findbugs or pmd or whatever we use for
this.
```

Follow-ups during the task: the recommendation is in Guava's CacheBuilder
javadoc, not their wiki. Per review, do not convert the most complicated
BDD cache (IpSpaceToBDD) -- leave it on Guava. And type the CommunitySet
cache on Set<? extends Community> rather than casting at the call sites.

---
**Stack**:
- #9973 ⬅
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/main/Batfish.java
+++ b/projects/batfish/src/main/java/org/batfish/main/Batfish.java
@@ -31,9 +31,9 @@ import static org.batfish.specifier.LocationInfoUtils.computeLocationInfo;
 import static org.batfish.vendor.check_point_management.parsing.CheckpointManagementParser.parseCheckpointManagementData;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.github.benmanes.caffeine.cache.Cache;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
-import com.google.common.cache.Cache;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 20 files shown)


dhalperi (commented):
> @dhalperi reviewed 20 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9973)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9975, opened 2026-06-06
BGP: honor advertise-inactive on iBGP sessions; add Cisco suppress-inactive

dhalperi:
BgpSessionProperties.from gated advertiseInactive on
SessionType.isEbgp, so the flag could never hold for an iBGP session.
That restriction dates to the first BGP advertisement implementation in
2017 ("Only applicable to eBGP sessions") and was carried through every
refactor since without re-examination. Real devices do not gate the
behavior on session type: FRR advertises best paths regardless of
iBGP/eBGP, and Junos advertise-inactive is honored on whichever session
it is configured (the only documented exception is route-reflector VRF
routes). Remove the isEbgp conjunct so advertiseInactive is taken from
the sender's IPv4-unicast capabilities for both session types.

Cisco IOS and ASA advertise inactive routes by default and suppress that
with "bgp suppress-inactive", which Batfish did not parse -- IOS/ASA
conversion hardcoded advertiseInactive=true. Parse the command into a
nullable suppressInactive flag on the peer-group hierarchy (inherited
like other per-neighbor settings, default applied at conversion as
advertiseInactive = !suppressInactive).

IOS-XR had a parser for "bgp advertise-inactive" whose extracted value
conversion ignored, and IOS-XR has no command to suppress inactive
advertisement, so the parsed value could only ever be the default. Delete
that dead chain (lexer token, grammar rule, extractor handler, and the
representation field) and keep the hardcoded model default.

Tests cover the IOS/ASA suppress-inactive grammar, the session-property
layer for iBGP, and an end-to-end iBGP dataplane scenario (an inactive
eBGP-learned route reaches an iBGP peer only when not suppressed) that
fails if the isEbgp gate is restored.

----

Prompt:
```
I've been doing a lot of work on BGP and I discovered an issue: we refuse
to let advertiseInactive hold for an iBGP session. Can you look at git
history and see why? This is in BgpSessionProperties.java
```

Followups: research vendor behavior (FRR/Junos/Cisco) without consulting
Batfish; add suppress-inactive support on Cisco IOS and ASA and any other
vendor with an equivalent feature, remove the isEbgp gate, and run the
full test suite to surface breakage; name the config-derived property
suppressInactive to match the command; keep the IOS flag nullable and
apply the default at conversion after inheritance; delete the dead IOS-XR
advertise-inactive parser; add an end-to-end iBGP dataplane test.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/CiscoLexer.g4
@@ -5455,6 +5455,8 @@ SUPPRESS_ARP: 'suppress-arp';
 
 SUPPRESS_FIB_PENDING: 'suppress-fib-pending';
 
+SUPPRESS_INACTIVE: 'suppress-inactive';
+
 SUPPRESS_MAP: 'suppress-map';
 
 SUPPRESSED: 'suppressed';
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco/Cisco_bgp.g4
@@ -202,6 +202,7 @@ bgp_rb_stanza
    | bgp_redistribute_internal_rb_stanza
    | bgp_refresh_rb_stanza_null
    | bgp_regexp_rb_stanza_null
+   | bgp_suppress_inactive_rb_stanza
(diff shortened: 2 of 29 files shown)


dhalperi (commented):
> @dhalperi reviewed 29 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9975)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9977, opened 2026-06-06
sros: extract Nokia SR-OS configs into a typed feature model

dhalperi:
Build the SR-OS extraction stage (P4) on top of the P3 canonical
statement stream: a navigable tree, a preprocessor for the two
mechanisms deferred from the grammar, and a typed feature model.

SrosConfigurationBuilder now builds an SrosStatementTree from the same
word stream that produces the P3 canonical statement list, so the brace,
flat /configure, and mixed input forms all yield an identical tree.
Every word is one level of the tree, so a leaf's value is the single
child key under the leaf node and a `policy [ a b ]` leaf-list is the
ordered children of the policy node (order preserved for ordered-by-user
lists).

SrosPreprocessor runs the mechanisms P1 characterized and P3 deferred,
on the tree, before extraction: apply-groups expansion (graft the
matching subtree of each groups group onto the branch that applied it,
with local-wins, first-listed-group-wins, apply-groups-exclude, regex
list-keys like interface "<to-.*>", and a convergence loop) and
delete/- edits (remove the named subtree, silent no-op when absent).
The groups definition container is pruned afterward.

SrosFeatureExtractor walks the preprocessed tree into the typed
representation: hardware (Card/Mda, Port), routers (Router,
RouterInterface), BGP (BgpProcess, BgpGroup, BgpNeighbor), and routing
policy (PrefixList with composite prefix+type entries, PolicyStatement
with numbered entries and a default-action). List keys follow YANG.
Subtrees that are valid config but not control-plane-relevant
(system security/ssh/user-params, persistent-indices) are intentionally
left unread, with no warnings; the one system leaf that matters,
system name, becomes the hostname.

SrosExtractionTest asserts the full captured r1 feature model extracts
with zero parse, red-flag, and unimplemented warnings, that the
unmodeled subtrees are skipped silently, and that the preprocessor
handles apply-groups (incl. regex keys + exclude + local-wins) and
delete edits. Updates docs/parsing/vendors/sros.md with an
"Extraction (P4)" section.

----

Prompt:
```
complete P4 (SR-OS extraction) to its Done-when: extraction populates the
SR-OS VendorConfiguration for every characterized feature, extraction unit tests
green, no unexpected todo/warn — following ORCHESTRATION/ROADMAP. Use a clean subagent as the judge and make sure to re-judge any time we make changes
```

---
**Stack**:
- #9981
- #9980
- #9979
- #9978
- #9977 ⬅
---

### DIFF
--- a/docs/parsing/vendors/sros.md
+++ b/docs/parsing/vendors/sros.md
@@ -84,23 +84,6 @@ Why this over the flatten pipeline:
 This mirrors the *spirit* of the Junos pipeline (one canonical statement form fed
 to extraction) while avoiding the text round-trip.
 
-### Deferred to extraction (P4+)
-
-The following were characterized in P1 and are deferred to the extraction phase,
-where feature-specific modeling needs them; they operate on the canonical
-statement tree this parser produces:
-
-- Incremental edit verbs (`delete`/`-`/`~`/`insert before|after`/`replace … with`/
-  `copy`/`rename`) — applied in order as a preprocessor stage.
-- `apply-groups` config-group inheritance — Batfish must expand it (configs ship
-  unexpanded): regex/wildcard key match, first-listed-wins precedence,
-  `apply-groups-exclude`.
-- Per-list ordering from YANG `ordered-by`; absent leaf ⇒ YANG default. BGP
-  group→neighbor inheritance resolved in extraction.
-
-As of P3 the builder records the canonical statements and extracts the system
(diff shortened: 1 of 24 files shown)


dhalperi (commented):
> @dhalperi reviewed 24 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9977)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9978, opened 2026-06-06
sros: convert Nokia SR-OS configs to the vendor-independent model

dhalperi:
Build the SR-OS conversion stage (P5): SrosConfiguration
.toVendorIndependentConfigurations() now produces a real Configuration
from the typed P4 feature model. Conversion logic lives in the new
SrosConversions.

Each router instance becomes a VRF (Base -> the Batfish default VRF).
Router interfaces convert to VI interfaces (port-less system/loopback ->
LOOPBACK, port-bound -> PHYSICAL) with their ipv4 primary address. The
hardware tree (cards/MDAs/ports) does not map to the VI model, so it is
left unconverted with a red-flag warning rather than dropped silently.

BGP group->neighbor inheritance is resolved here (P4 recorded only the
neighbor's group leafref): a neighbor's peer-as and import/export policy
lists come from its own config if set, else from its group. Each
neighbor becomes a BgpActivePeerConfig keyed by peer IP, with local-as
from the instance autonomous-system and local-ip/router-id derived from
the system interface address when unset.

eBGP default-reject is implemented at conversion via the Junos-style
generated-peer-policy idiom. Each policy-statement becomes a chainable
RoutingPolicy whose entry actions are correct both as a subroutine
(ReturnTrue/False under CALL_EXPR_CONTEXT) and standalone
(ExitAccept/Reject); next-policy -> FallThrough, next-entry/absent
default-action -> fall through. Each peer gets a generated import and
export policy that SetDefaultPolicy to a shared accept-all (iBGP) or
reject-all (eBGP) backstop, then evaluates a FirstMatchChain of the
peer's named policies. So an eBGP peer with no accepting policy rejects
both directions ("no policy" != "permit all"), while iBGP defaults to
accept.

Adds DeviceModel.NOKIA_SROS_UNSPECIFIED.

SrosConversionTest parses the captured r1 config through the full
pipeline and asserts the VI Configuration shape, interface types and
addresses, the prefix-list -> RouteFilterList, group->neighbor
inheritance, behavioral eBGP default-reject on the generated peer
policies, and the hardware-not-converted warning. Updates
docs/parsing/vendors/sros.md with a "Conversion (P5)" section.

----

Prompt:
```
complete P5 (SR-OS conversion) to its Done-when: SrosConfiguration.toVendorIndependentConfigurations() produces a vendor-independent Configuration from the typed P4 model; conversion unit tests green; significant-but-unmodeled
constructs emit warnings, not silent drops — following ORCHESTRATION/ROADMAP. Resolve BGP group→neighbor inheritance and eBGP default-reject at conversion. Use a clean subagent as the judge and re-judge any time we make changes.
```

---
**Stack**:
- #9981
- #9980
- #9979
- #9978 ⬅
- #9977
---


dhalperi:

PR #9979, opened 2026-06-06
sros: fix BGP local-ip, origin, and connected-route modeling

dhalperi:
Three conversion corrections surfaced by the first lab-validation slice
(P5-V) against the captured srsim+cEOS eBGP snapshot:

- BGP peer local-ip: leave it unset instead of forcing the system
address. SR-OS auto-selects the source address per peer; for a
directly-connected eBGP peer that is the egress interface address, so
Batfish must resolve it from the topology. Forcing the system address
put the local IP off the peering /31 and the session never
established.

- Advertised local routes carry origin IGP. SR-OS originates
locally-sourced routes (the system/connected prefixes a peer learns
via an export policy) with origin igp, like Junos; the generated
per-peer export policy now sets origin IGP for non-BGP routes rather
than leaving Batfish's redistribution default of incomplete. Verified
on the cEOS peer (it learns the system prefix with as-path "65001 i").

- Suppress the local /32 host route. SR-OS installs the connected
subnet route for an interface address but not a local /32 for the
interface's own IP; interface address metadata with
generateLocalRoute=false stops Batfish synthesizing one, so the main
RIB matches the device route-table.

Confirmed the BGP export model is correct as-is (main-RIB export, not
BGP-RIB origination): the device's operational BGP table holds only
learned routes (plaintext "Total Remote Rts: 1"; only the learned route
has in-rtm true in the state tree), so setExportBgpFromBgpRib stays
false. SrosConversionTest updated to assert the three behaviors.

----

Prompt:
```
Complete P5-V, using a clean agent to judge, rejudging after any change.
```

---
**Stack**:
- #9981
- #9980
- #9979 ⬅
- #9978
- #9977
---

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConfiguration.java
@@ -93,7 +93,11 @@ public final class SrosConfiguration extends VendorConfiguration {
     c.setDeviceModel(DeviceModel.NOKIA_SROS_UNSPECIFIED);
     c.setDefaultCrossZoneAction(LineAction.PERMIT);
     c.setDefaultInboundAction(LineAction.PERMIT);
-    // SR-OS, like Junos, runs the BGP export pipeline from the main RIB.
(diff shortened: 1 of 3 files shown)


dhalperi:

PR #9980, opened 2026-06-06
sros: model the port/router-interface split so L1 topology drives L3

dhalperi:
SR-OS separates the physical port (e.g. 1/1/c1/1) from the L3
router-interface (e.g. to-r2) that binds it -- the same physical/logical
split as Junos (ge-0/0/0 vs unit ge-0/0/0.0). Conversion previously
collapsed a port-bound interface into a single PHYSICAL interface that
held the IP, with no port interface and no dependency. That breaks
Layer-1 topology: collected SR-OS labs name the L1 endpoint by the
physical port path (1/1/c1/1), not the router-interface, so the L1 edge
resolved to INVALID_INTERFACE and was silently dropped. The cross-vendor
eBGP adjacency then formed only via the same-subnet fallback in
HybridL3Adjacencies -- accidental, since it holds only because the cEOS
side is never node1 of a resolved L1 edge.

Model it the way Batfish models Junos: a port-bound router-interface
becomes a LOGICAL interface holding the address with a BIND dependency on
an addressless PHYSICAL interface named by the port path (the L1
endpoint); port-less interfaces (system, loopbacks) stay LOOPBACK.
PointToPointComputer walks the BIND dependency from to-r2 to its port, so
the user L1 edge between ports yields the L3 edge between the router
interfaces. The port's admin-state is honored on the PHYSICAL interface.

New SrosPostProcessingTest loads the captured sros_ceos_ebgp lab (SR-OS r1
+ cEOS r2 + the user L1 topology) through the full pipeline including
postProcessSnapshot and asserts: every r1 interface active with the
expected type and the to-r2 -> 1/1/c1/1 BIND dependency; the user L1 edge
(named by the port) survives into the active logical L1; the L3 edge
r1:to-r2 <-> r2:Ethernet1 forms; and the eBGP session is established.

----

Prompt:
```
Complete P6 including P6-V, using a clean agent to judge, rejudging after any change.
```

---
**Stack**:
- #9981
- #9980 ⬅
- #9979
- #9978
- #9977
---

### DIFF
--- a/docs/parsing/vendors/sros.md
+++ b/docs/parsing/vendors/sros.md
@@ -165,15 +165,27 @@ main routing instance — maps to the Batfish default VRF (`"default"`); any oth
 instance maps to a VRF of the same name. Interfaces and the BGP process are
 attached to their instance's VRF.
 
-### Interfaces
+### Interfaces (the port / router-interface split)
 
-A router interface with no port binding (`system`, loopbacks) is modeled as
-`InterfaceType.LOOPBACK`; a port-bound interface is `InterfaceType.PHYSICAL`. The
-`ipv4 primary {address, prefix-length}` pair becomes the interface's
-`ConcreteInterfaceAddress`. The hardware tree (cards/MDAs/ports) does **not** map
-to the VI model — Batfish derives interfaces from the router instance, not the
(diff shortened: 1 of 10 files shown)


dhalperi:

PR #6840, opened 2021-04-03
IOS-XR grammar remove lexer state variables

arifogel:
- remove all lexer state variables
- simplify surviving tokens
- remove non-xr rules and tokens broken by removal of state variables
- use modes for most of community set lexing

PR #6798, opened 2021-03-29
IOS: delete _enableIpAddress predicate

dhalperi:
It's never changed from true

PR #9984, opened 2026-06-09
ASA: remove lexer state-variable predicates

dhalperi:
Delete the _enableIpAddress, _enableDec, _enableAclNum, _enableIpv6Address,
and _enableRegex lexer state variables and their semantic predicates,
restoring ANTLR's per-mode lexer DFA cache. Mirrors the IOS-XR cleanup in
batfish/batfish#6840 (and #6798 for _enableIpAddress).

ANTLR disables a lexer mode's DFA cache when any reachable predicate remains
in that mode, forcing full ATN closure on every token instead of cached DFA
edges. DEFAULT_MODE - where access-list bodies are lexed - carried
predicates on UINT8/16/32, DEC, IP_ADDRESS/PREFIX, IPV6_ADDRESS/PREFIX, the
VARIABLE rule, and REGEX. Because the cache is disabled mode-wide as long as
any one predicate survives, removing them individually yields no speedup;
they must all go. On an access-list-heavy config of ~179K lines (~5M
tokens), this cuts cold lex+parse from ~123s to ~1.4s (lexing alone ~122s to
~0.6s).

Changes:
- ACCESS_LIST, ARP, COMMANDER_ADDRESS, COMMUNITY, FIREWALL, VRF, EXTENDED,
  STANDARD, RULE become plain keyword tokens (drop their flag-setting
  actions). INTERFACE keeps its mode push, drops the flag.
- IP_ADDRESS/PREFIX, IPV6_ADDRESS/PREFIX, UINT8/16/32, DEC, NEWLINE become
  predicate-free. VARIABLE is simplified and moved below the IP/IPv6 tokens
  so maximal-munch + rule order disambiguate, matching IOS-XR. VARIABLE
  keeps the colon-excluding character set the predicated rule used by
  default, so IPv6 addresses tokenize as IPV6_ADDRESS and "keyword:value"
  trailers (e.g. cryptochecksum lines) still split into separate tokens.
- Delete the REGEX token: it has no parser consumer (dead Frankenparser
  vestige).
- Remove ACL_NUM and its range-classification (ACL_NUM_STANDARD/EXTENDED/
  APPLETALK/IPX/MAC/...). On ASA a numeric ACL ID is just a name: the ASA
  does not enforce the IOS 1-99/100-199 ranges, and standard vs extended
  comes from the explicit keyword. Numeric names still parse as
  variable_aclname. Delete the IOS-only numbered/appletalk/ipx-sap/
  protocol-type-code/numbered-mac ACL stanzas.
- s_no_access_list_extended/standard collapse into one keyword-free
  s_no_access_list removing from both ACL maps.
- Delete _enableCommunityListNum (declared, never referenced) and the
  now-empty printStateVariables override in AsaBaseLexer.

Verified behavior-preserving on the motivating config: the serialized
vendor-independent model is byte-identical (same SHA-256) and the parse
warnings and parse status are unchanged before and after. The full unit and
reference test suites pass with no ref-file changes.

----

Prompt:
```
A large Cisco ASA config (~179K lines, almost entirely access-list lines)
takes a very long time to parse. Profile it and improve parsing
performance.
```

Profiling (async-profiler via tools/profiling) showed ~100% of parse time in
LexerATNSimulator.closure: the state-variable predicates were defeating the
lexer DFA cache. The fix follows dhalperi's established playbook for these
grammars - make tokens predicate-free, rely on lexer modes and parser-level
disambiguation, and delete vendor-inapplicable Frankenparser alternatives -
using IOS-XR #6840 as the template.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_acl.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_acl.g4
@@ -67,22 +67,6 @@ access_list_mac_range
    )
 ;
 
-appletalk_access_list_null_tail
-:
-   action = access_list_action
-   (
-      (
-         CABLE_RANGE ~NEWLINE*
-      )
-      | OTHER_ACCESS
-   )? NEWLINE
-;
-
-appletalk_access_list_stanza
-:
-   ACCESS_LIST name = ACL_NUM_APPLETALK appletalk_access_list_null_tail
-;
-
 aruba_access_list_action
 :
    action = access_list_action
@@ -267,10 +251,6 @@ extended_access_list_stanza
    (
       (
          IP ACCESS_LIST EXTENDED name = variable_aclname
-      )
-      |
-      (
-         ACCESS_LIST num = ACL_NUM_EXTENDED
       )
       |
       (
@@ -628,16 +608,6 @@ ipv6_prefix_list_tail
    )* NEWLINE
 ;
 
-ipx_sap_access_list_null_tail
-:
-   action = access_list_action null_rest_of_line
-;
-
-ipx_sap_access_list_stanza
-:
-   ACCESS_LIST name = ACL_NUM_IPX_SAP ipx_sap_access_list_null_tail
-;
-
 irs_stanza
 :
    bandwidth_irs_stanza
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9984)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9985, opened 2026-06-09
Transformation: serialize deep orElse chains iteratively

dhalperi:
A Juniper NAT rule-set is converted into a chain of Transformations
linked via orElse, one link per rule (NatRuleSet#rulesTransformation).
A rule-set with thousands of rules produces a chain thousands of links
deep. Transformation used default Java serialization, which recurses
once per link, so initializing a snapshot with such a config threw
StackOverflowError while deserializing the vendor-independent
configurations (during completion-metadata computation).

Add a writeReplace/readResolve serialization proxy that walks the orElse
spine iteratively into flat lists on write and rebuilds it iteratively
on read, matching the existing Prefix.SerializedForm pattern. Chains of
any depth now round-trip without recursion. andThen sub-trees are not
part of the spine and are not deep in practice, so they are left to
normal serialization.

----

Prompt:
```
A Juniper SRX config with a NAT rule-set containing thousands of rules
crashed Batfish during snapshot initialization. Investigate and fix it.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/transformation/Transformation.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/transformation/Transformation.java
@@ -8,7 +8,10 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
+import java.io.ObjectStreamException;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9985)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9981, opened 2026-06-06
sros: document P7 RIB validation in the vendor guide

dhalperi:
Add the "RIB validation (P7)" section to the SR-OS vendor guide, the
doc-as-deliverable for the P7 phase. P7 validates r1's computed main RIB
and BGP RIB against the device's operational state (info json /state
router "Base" route-table / bgp rib), the SR-OS-side counterpart to what
the cEOS validator does for r2.

The section records the SR-OS modeling facts the comparison rests on,
since they constrain conversion: a learned eBGP route has admin distance
170 (not 20); the bgp rib local-rib over-lists owner=local connected
routes (in-rtm=false) that BGP can advertise via an export policy but
does not originate, so the validator compares only the owner==bgp learned
subset; and SR-OS advertises from the main RIB (Junos-like), so
conversion keeps setExportBgpFromBgpRib(false).

The functional validation code (the route-table and bgp-rib JSON parsers,
the SrosValidator RIB methods, and their unit tests) was front-loaded in
the sibling lab-validation repo at P5-V and has been active and green
since; pytest lab_tests/ --labname=sros_ceos_ebgp is 13/13 with no r1 RIB
check sickbayed. This commit is the batfish-side documentation only.

----

Prompt:
```
Complete P7 including any verification steps, using a clean agent to judge, rejudging after any change.
```

---
**Stack**:
- #9981 ⬅
- #9980
- #9979
- #9978
- #9977
---
⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #9976, opened 2026-06-06
sros: parse Nokia SR-OS configs (lexer + grammar + extractor scaffold)

dhalperi:
Add the SR-OS parser under the vendor-scoped path
org.batfish.vendor.sros (grammar + representation), and route detected
NOKIA_SROS configs to it in ParseVendorConfigurationJob (removing
NOKIA_SROS from UNIMPLEMENTED_FORMATS). The captured SR-SIM lab config
now parses with zero FATAL warnings.

Architecture: a single hierarchical ANTLR grammar handles all three
input forms SR-OS operators produce — the brace form from `admin show
configuration`, the absolute-path flat form (`/configure ...`, the
Junos-`set` analog), and a mix of the two in one file. A flat line is
just a leaf statement whose first word starts with `/`; the
SrosConfigurationBuilder normalizes every leaf to one canonical
absolute-path string (rewriting a leading `/configure` to `configure`),
so the three forms yield the identical model. This resolves the P1
flatten-vs-hierarchical open question in favor of the
hierarchical-canonical approach rather than a flatjuniper-style
flatten-to-set-lines pass: the mixed-form equivalence is correct by
construction, true source line numbers are preserved for warnings (no
FlattenerLineMap round-trip), and it is the fewest moving parts for the
zero-FATAL parse gate. Edit verbs and apply-groups expansion are
deferred to extraction (P4), where they operate on this canonical
statement tree.

SrosGrammarTest covers the gate and the mixed-form acceptance
requirement: the captured r1 config parses cleanly with nothing
unrecognized, and pure-brace, pure-flat, and mixed inputs describing
the same configuration produce the identical canonical statement list.
Updates docs/parsing/vendors/sros.md with a "Parser (P3)" section
recording the decision and rationale.

----

Prompt:
```
Complete P3 (the Nokia SR-OS parser: lexer + grammar) to its Done-when
condition: the captured r1 config parses via the SR-OS grammar with
zero FATAL parse warnings and bazel test //...sros...grammar... is
green, following ORCHESTRATION.md, ROADMAP, and the other tracking
files; resolve the flatten-vs-hierarchical open question at P3 start
and write the mixed brace+flat acceptance test. Ensure accurate updates
to all tracking data.
```

---
**Stack**:
- #9981
- #9980
- #9979
- #9978
- #9977
- #9976 ⬅
---
⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #9986, opened 2026-06-09
sros: convert and post-process Nokia SR-OS, with structure/provenance tracking

dhalperi:
Continue Nokia SR-OS support after parsing (#9976) and extraction (#9977),
taking it from a parsed config through to a fully converted, post-processed
vendor-independent model. Consolidates what had been per-commit spr PRs
(#9978–#9981, now closed) into one branch, and folds in the review feedback and
doc improvements that followed.

Conversion (P5) builds a `Configuration` from the typed P4 model: each router
instance becomes a VRF, prefix-lists become RouteFilterLists, policy-statements
become chainable RoutingPolicies, and BGP is realized with group→neighbor
inheritance and eBGP default-reject (the Junos-style generated-peer-policy
chain). Hardware that does not map to the VI model warns rather than dropping
silently. Three conversion corrections came out of validating against real
device state: eBGP local-ip is left unset (resolved from the connected
interface), advertised local routes get origin igp, and the spurious local /32
host route is suppressed.

Post-processing (P6) hinges on the SR-OS port / router-interface split: an L3
router-interface becomes a LOGICAL interface with a BIND dependency on an
addressless PHYSICAL port named by the port path (the Junos physical/unit
split), so a user Layer-1 topology — which names the port — drives the L3
adjacency. The vendor guide also records the SR-OS RIB modeling facts the P7
lab-validation slice depends on (learned eBGP preference 170; the `bgp rib
local-rib` over-lists `owner=local` connected routes with `in-rtm=false`; SR-OS
advertises from the main RIB, so `setExportBgpFromBgpRib(false)`).

Extraction gained source-provenance tracking: each statement's
`ParserRuleContext` is carried on its tree node, so extraction emits line-stamped
`ParseWarning`s for malformed/out-of-range values and records structure
definitions and references (powering the `definedStructures`,
`undefinedReferences`, and unused-structure questions). This is the Juniper
approach adapted to a non-flattened hierarchical vendor — no `FlattenerLineMap`,
since the single grammar parses brace/flat/mixed directly — and it lets the
`annotate` tool render SR-OS warnings inline with a `#` header. Values are parsed
with `tryParse`/in-space helpers (range-checked against YANG bounds) instead of
throwing `parse` calls, and unrecognized enumerated values warn on their line via
a shared `toEnum` helper rather than dropping silently.

Review feedback (#9986) moved BGP group→neighbor inheritance off inline-in-
conversion onto the model (`BgpNeighbor.inheritFrom`, the NX-OS `doInherit`
pattern), modeled the BGP peer `type` (internal/external) so eBGP/iBGP follows an
explicit type when set, added a warning for prefix-list match types whose bounds
are not yet modeled, and tidied terminology (the SR-OS `preference` keyword in
vendor code, VI "admin distance" only at the conversion boundary). Every SR-OS
test config got a `system name` so hostnames are deterministic.

Finally, the vendor-adding docs gained a production-readiness checklist
(`tutorials/adding_vendor_support.md`), an expanded structure-tracking pattern
(`parsing/implementation_guide.md`), and a "production-readiness concerns"
section (`extraction/README.md`) — the capabilities a new vendor needs beyond a
happy-path parser, with the SR-OS guide as the worked example.

Testing: SR-OS grammar/extraction/conversion/post-processing unit tests + PMD,
the `annotate` test, and detector/parse-job regression all pass. The
lab-validation suite (separate repo) runs the captured SR-OS↔cEOS eBGP snapshot
through this pipeline green — interfaces, main RIB, and BGP RIB on the SR-OS
side, with the fully-modeled cEOS side as an independent cross-vendor oracle.


dhalperi (commented):
> @dhalperi reviewed 28 files and all commit messages, and made 13 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9986)* status: all files reviewed, 13 unresolved discussions.

___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 200 at r5](https://reviewable.io/reviews/batfish/batfish/9986#-OueRw_w235XbqDMCQiw:-OueRw_w235XbqDMCQix:b-uiqj4y) ([raw file](https://github.com/batfish/batfish/blob/3f1cbdf34e145a41d96dd9a35f1079f0915afdd0/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L200)):*
> ```Java
>       if (primary != null) {
>         iface.setPrimaryAddress(
>             parseIp(
> ```

prefer `toIp`, `toPrefix`, etc.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 448 at r5](https://reviewable.io/reviews/batfish/batfish/9986#-OueS0ky7JVFU8Rym8dM:-OueS0ky7JVFU8Rym8dN:bc61dyl) ([raw file](https://github.com/batfish/batfish/blob/3f1cbdf34e145a41d96dd9a35f1079f0915afdd0/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L448)):*
> ```Java
>   }
> 
>   private @Nullable Ip parseIp(
> ```

prefer toIp
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 461 at r5](https://reviewable.io/reviews/batfish/batfish/9986#-OueS3FV0Jo_ksTIzY3g:-OueS3FV0Jo_ksTIzY3h:b46pghs) ([raw file](https://github.com/batfish/batfish/blob/3f1cbdf34e145a41d96dd9a35f1079f0915afdd0/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L461)):*
> ```Java
>   }
> 
>   private @Nullable Prefix parsePrefix(@Nullable String text, @Nullable SrosStatementTree ctxNode) {
> ```

prefert toPrefix.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 466 at r5](https://reviewable.io/reviews/batfish/batfish/9986#-OueS5ltBXyVvb9_L4Yg:-OueS5ltBXyVvb9_L4Yh:bwh1zcp) ([raw file](https://github.com/batfish/batfish/blob/3f1cbdf34e145a41d96dd9a35f1079f0915afdd0/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L466)):*
> ```Java
>     }
>     try {
>       return Prefix.parse(text);
> ```

isn't there like a Prefix.tryParse? and similar for Ip
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 544 at r5](https://reviewable.io/reviews/batfish/batfish/9986#-OueSDJn3rKDRqh9dyUy:-OueSDJn3rKDRqh9dyUz:b-ufb5ck) ([raw file](https://github.com/batfish/batfish/blob/3f1cbdf34e145a41d96dd9a35f1079f0915afdd0/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L544)):*
> ```Java
>    * silently lost.
>    */
>   private void warn(@Nullable SrosStatementTree ctxNode, String message) {
> ```

I don't like this being nullable. Why do we do it?

Why not have a separate helper for the redFlag case?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 54 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OuePpaqDkAIa2wcuKO3:-OuePpaqDkAIa2wcuKO4:b87rs9u) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L54)):*
> ```Java
>  *       it maps to the Batfish default VRF. Any other router instance maps to a VRF of the same
>  *       name.
>  *   <li><b>BGP {@code group} → {@code neighbor} inheritance.</b> A neighbor inherits {@code
> ```

Comment is very specific. We'll add more propreties later.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 58 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OuePs626axxzAPp5cjU:-OuePs626axxzAPp5cjV:b-wy4wtp) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L58)):*
> ```Java
>  *       does not configure them directly (per-neighbor config wins). This is resolved here, not in
>  *       extraction (P4 only recorded the {@code group} leafref).
>  *   <li><b>eBGP default-reject.</b> SR-OS drops eBGP routes in both directions unless an explicit
> ```

Comment is very specific. Does this really belong at the class level or on specific functions?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 69 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OuePiTH9GeVZt9SOpLd:-OuePiTH9GeVZt9SOpLe:b-ffb247) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L69)):*
> ```Java
> 
>   /**
>    * The default administrative distance Batfish assigns BGP routes. SR-OS uses route {@code
> ```

Is it preference or admin disance? If preference is the vendor-specific (VS) keyword, prefer that and keep the VI terms only in the actual conversion code.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 101 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OueQ8NL9EQuaMIRA1az:-OueQ8NL9EQuaMIRA1b-:bswi8e9) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L101)):*
> ```Java
>             case EXACT -> SubRange.singleton(len);
>             case LONGER -> new SubRange(len + 1, Prefix.MAX_PREFIX_LENGTH);
>             case THROUGH, RANGE, TO, ADDRESS_MASK ->
> ```

we need warnings for these!
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 306 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OueQjM3-cul0iaoQ9cU:-OueQjM3-cul0iaoQ9cV:b-ff74ut) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L306)):*
> ```Java
> 
>     // peer-as: per-neighbor wins, else inherited from the group.
>     Long peerAs = neighbor.getPeerAs();
> ```

implement the nx-os style inheritance pattern for BGP groups, not this per-property inheritance inline with conversion. I bet the group also inherits from the bgp process and the global bgp config too, right?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 311 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OueQy9xAWgXo7Jfy3VM:-OueQy9xAWgXo7Jfy3VN:b-womdbg) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L311)):*
> ```Java
>     }
>     if (peerAs == null) {
>       w.redFlagf(
> ```

is there a version of type internal (junos) where you don't need this because it's based on the tpe?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 354 at r1](https://reviewable.io/reviews/batfish/batfish/9986#-OueR3BL5pNBgwS33KjS:-OueR3BL5pNBgwS33KjT:b3zndzz) ([raw file](https://github.com/batfish/batfish/blob/b4d5e143885504f9f3e0e80e46b8bea67be4200a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L354)):*
> ```Java
>    * otherwise the group's list.
>    */
>   private static @Nonnull List<String> inheritedPolicies(
> ```

same comment about inheritance here.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 340 at r2](https://reviewable.io/reviews/batfish/batfish/9986#-OueRJhDC9Pz6NjcfJi7:-OueRJhDC9Pz6NjcfJi8:bovabz1) ([raw file](https://github.com/batfish/batfish/blob/184504ba0ea79a892950bcc2d2299d55d66b4ad0/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L340)):*
> ```Java
>     String exportPolicyName =
>         generatedBgpPeerExportPolicyName(router.getName(), neighbor.getIpAddress());
>     buildPeerPolicy(importPolicyName, importPolicies, ebgp, /* isExport= */ false, c);
> ```

don't do this style of commenting the args.

dhalperi:
Addressed all 13 comments in `sros: address review feedback` (825db36):

**BGP inheritance (SrosConversions:306, :354)** — moved group→neighbor inheritance off the inline per-property logic into `BgpNeighbor.inheritFrom` on the model (resolved at extraction, NX-OS `doInherit`-style); conversion now reads a fully-populated neighbor. Dropped the `inheritedPolicies` helper. SR-OS BGP nesting is group + neighbor (no separate session/global template levels in the modeled subset), so inheritance is the single group→neighbor hop.

**`type internal/external` (SrosConversions:311)** — modeled the `peer-type` leaf. eBGP/iBGP now follows the explicit type when set (Junos-style), falling back to peer-as vs local-as otherwise.

**preference vs admin distance (SrosConversions:69)** — `preference` is the SR-OS keyword; renamed to `DEFAULT_BGP_PREFERENCE` and kept the VI "admin distance" terms only at the conversion boundary.

**prefix-list THROUGH/RANGE/TO/ADDRESS_MASK warnings (SrosConversions:101)** — these now emit a warning that the filter is over-approximated as "this length or longer", instead of widening silently.

**class-level comments too specific (SrosConversions:54, :58)** — trimmed to the router-instance→VRF mapping; inheritance/default-reject details moved onto the functions.

**arg-commenting style (SrosConversions:340)** — removed the `/* isExport= */` style.

**toIp/toPrefix + tryParse (SrosFeatureExtractor:200, :448, :461, :466)** — renamed `parseIp`/`parsePrefix` → `toIp`/`toPrefix` and switched to `Ip.tryParse`/`Prefix.tryParse`.

**nullable `warn` (SrosFeatureExtractor:544)** — `warn` now takes a non-null value node; a value node always carries a source context by construction, so the nullable-context red-flag fallback is gone.

Also gave every SR-OS test config a `system name` so hostnames are deterministic, and refreshed `docs/parsing/vendors/sros.md` to match. Full gate green (SR-OS unit + PMD, annotate, detector/parse-job regression, lab 13/13).

dhalperi (commented):
> @dhalperi reviewed 24 files and all commit messages, and resolved 13 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9986)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9986)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

Issue #9991, opened 2026-06-09
SR-OS VPRN: model MPLS L3VPN (vpn-ipv4) inter-PE route import


dhalperi:
SR-OS VPRN: model MPLS L3VPN (vpn-ipv4) inter-PE route import

**Context.** The `sros_vprn_bgp` lab-validation rung is a real MPLS L3VPN
between two SR-SIM PEs: OSPF+LDP underlay, MP-BGP `vpn-ipv4` between the PE
system loopbacks, and a VPRN `red` on each PE with RD `65000:1`/`65000:2` and a
shared `vrf-target target:65000:1`. On the device, each PE's VPRN `red` RIB
imports the remote PE's CE loopback (e.g. pe1 `red` learns `172.16.2.1/32` via
MP-BGP with the route-target, resolved over an LDP tunnel) — confirmed live on
SR-SIM 26.3.R1.

**What is modeled.** SR-OS extraction captures `bgp-ipvpn mpls`
route-distinguisher + `vrf-target` import/export route-targets into the vendor
representation (`BgpIpvpn`), and conversion sets the route-distinguisher on the
VI VRF (`Vrf.setRouteDistinguisher`). Config-format, interfaces, and the BGP RIB
validate green.

**What is NOT modeled (this issue).** The Batfish VI datamodel has no `vpn-ipv4`
(VPNv4) BGP address family, and its cross-VRF leaking (`VrfLeakConfig` /
`BgpVrfLeakConfig`) is intra-node and policy-driven, not RT-driven L3VPN over
MP-BGP between nodes. So PE-to-PE vpn-ipv4 route exchange with route-target
import is not reproducible: the VPRN's RT-imported routes are absent from the
computed main RIB. The two `test_main_rib_routes` checks for pe1/pe2 in the
`sros_vprn_bgp` lab are sickbayed to this issue.

**Scope to close.** Add a VPNv4 address family to the VI BGP model (or an
RT-driven inter-VRF leak), wire SR-OS `bgp-ipvpn` + `vrf-target` conversion to
it, and unsickbay the `sros_vprn_bgp` main-RIB checks. The collected ground
truth is staged at `lab-validation/collected-rungs/sros_vprn_bgp/`.

PR #9988, opened 2026-06-09
sros: P8 complexity-ladder rungs (static, OSPF, iBGP, route reflection, policy, redistribution, VPRN, misconfig)

dhalperi:
Adds the **P8 complexity-ladder** rungs for Nokia SR-OS on top of the merged
#9986 base (parse → extract → convert → post-process → RIB validation). Each
rung was modeled against a live SR-SIM 26.3.R1 lab and validated route-for-route
against device ground truth in the sibling `lab-validation` repo (9 SR-OS labs,
123 lab tests, no sickbays).

The 7 commits are independent and meant to be reviewed one at a time:

1. **sros: model static routes (L1)** — `static-routes route … next-hop/blackhole`
   → VI StaticRoute; only `admin-state enable` routes install.
2. **sros: scope iBGP default-accept to BGP routes (L3)** — iBGP default-accept
   propagates BGP routes only, not connected/static; fixes a connected-/31 leak.
3. **sros: model policy set-clauses and prefix-list match bounds (L5)** —
   metric/as-path-prepend/community set; prefix-list through/range length windows
   (removes the prior over-approximation for those types).
4. **docs: P8 ladder L1/L3/L5/L8** — vendor-guide section.
5. **sros: model OSPF, route reflection, redistribution, and VPRN (L2/L4/L6/L7)**
   — OspfProcess/area/interface-settings (admin 10, 100 Gbps ref-bw); RR
   cluster-id + next-hop-self; `from protocol` policy match (fixes a TRUE-guard
   leak); connected=IGP/redistributed=incomplete origin + MED-0 export;
   `service vprn` → same-named VRF.
6. **docs: P8 ladder L2/L4/L6/L7** — vendor-guide section.
7. **docs: L8 interface_up lab_builder false-negative** — records a collection
   tooling caveat.

Known modeled-but-unexercised gaps (documented in the vendor guide, no test
fails): ISIS; multi-area / external (E1/E2) OSPF; VPRN with BGP + RD/RT;
aggregate/redundant interfaces; `to`/`address-mask` prefix-list bounds (still
over-approximated + warned).

The matching `lab-validation` snapshots/validator changes are on that repo's
`sros-p8` branch (separate PR to follow).

Per-commit testing: SR-OS Java suite + PMD green; `fix_java_format` + checkstyle
+ buildifier clean. Lab validation green for all 9 SR-OS labs.

PR #9992, opened 2026-06-09
sros: ECMP, prefix-list to/address-mask, IS-IS, communities, VPRN RD (P9 follow-ups)

### LINKED ISSUE #9989: SR-OS: model ECMP (multiple next-hops) for static routes
SR-OS static routes support multiple `next-hop <ip>` entries under one `route
<prefix> route-type unicast` (ECMP). The current model (`StaticRoute`) holds a
single next-hop IP, so conversion takes only the first next-hop and extraction
emits a warning ("static route ... has N next-hops (ECMP); only the first is
modeled"). This under-models ECMP static routes.

Follow-up to PR batfish/batfish#9988 (P8 ladder). Confirm the device behavior in
a lab (build a static route with two next-hops, verify both install as ECMP in
the route-table) and then model all next-hops (one VI `StaticRoute` per
next-hop, or the appropriate multi-next-hop representation).

Code: `SrosFeatureExtractor.extractStaticRoutes` (the ECMP warning),
`SrosConversions.convertStaticRoutes`.

### LINKED ISSUE #9990: SR-OS: model prefix-list 'to' and 'address-mask' match-type bounds
SR-OS prefix-list match types `to` and `address-mask` are not fully modeled.
Conversion (`SrosConversions.toRouteFilterList`) over-approximates them as "this
prefix length or longer" and emits a red-flag warning, so the resulting
RouteFilterList is broader than the device's.

- `to`: matches from the prefix's own length through a configured `to-prefix`
  upper bound. The `to-prefix` leaf is not captured at extraction.
- `address-mask`: matches by an address+mask pair, not a contiguous length
  window; needs its own modeling.

(`through` and `range` are now fully modeled via through-length /
start-length+end-length; `exact` and `longer` were already exact.)

Follow-up to PR batfish/batfish#9988 (P8 ladder, L5 BGP policy depth). Capture
the `to-prefix`/address-mask bounds at extraction and build the exact length
window in conversion, removing the over-approximation warning for these types.

Code: `SrosConversions.toRouteFilterList` / `overApproximate`,
`SrosFeatureExtractor.extractPolicyOptions` (prefix-list entry bounds),
`PrefixListEntry`.

dhalperi:
SR-OS follow-up rungs on top of the merged P8 ladder (#9988): closes the two
deferred review bugs and adds three more features, each modeled against a fresh
live SR-SIM 26.3.R1 lab and validated route-for-route in the sibling
lab-validation repo. Commits are independent and meant to be reviewed in order.

1. **sros: model ECMP static routes and prefix-list to/address-mask bounds**
   — #9989: a static route with multiple next-hops emits one VI StaticRoute per
   next-hop (equal-preference legs install as ECMP; lower-preference wins).
   #9990: `type to` is a length-range match `[base-length .. to-prefix-length]`
   within the base network (one exact line per ancestor length, so off-path
   siblings don't match); contiguous `address-mask` is an exact match,
   non-contiguous is over-approximated + warned. Both confirmed on device.

2. **sros: parse empty same-line blocks (`{ }`)** — the device renders an
   empty-body list entry inline (`to-prefix X { }`), which the grammar rejected
   with a FATAL. Prerequisite for the to/address-mask extraction above.

3. **sros: model IS-IS (single-area, levels)** — IsisProcess
   (system-id/area/level/interfaces) → VI IsisProcess (NET, levels, per-interface
   point-to-point/passive). Adds NOKIA_SROS IS-IS admin distances to
   RoutingProtocol (L1 15 / L2 18 internal, confirmed on device).

4. **sros: extract VPRN bgp-ipvpn RD/route-targets; convert RD to VRF** — extracts
   `bgp-ipvpn mpls` route-distinguisher + vrf-target and sets the RD on the VI
   VRF. Inter-PE vpn-ipv4 L3VPN import is not modeled (no VI VPNv4 AF; cross-VRF
   leak is intra-node only) — tracked in #9991.

5. **docs** — SR-OS vendor guide section for all of the above.

Fixes #9989. Fixes #9990. For #9991.

SR-OS Java suite + PMD + checkstyle + buildifier clean. Validated green (or
sickbayed to #9991) across the new labs plus no regression on the 9 existing
SR-OS labs.

### DIFF
--- a/docs/parsing/vendors/sros.md
+++ b/docs/parsing/vendors/sros.md
@@ -502,3 +502,84 @@ the `show router interface` line) trips on the `Up/Down` column header, so it
 reports the (genuinely up) link interface as down. It is a collection-time
 precondition, not a lab test, so it does not affect the green result; the lab is
 collected regardless and validates 13/13.
+
+## Follow-up rungs (P9): ECMP, prefix-list bounds, communities, IS-IS, L3VPN
+
+A second pass closed the two deferred review bugs and added three more rungs,
+each modeled against a fresh live SR-SIM 26.3.R1 collection.
+
+### Empty same-line block `{ }` (grammar)
+
+The device renders a list entry with no body inline as `<words> { }` (open
+brace, optional whitespace, close brace, no newline) — e.g. the `to-prefix`es
+of a `prefix-list ... type to`, `mask-pattern`s, or a community `member`. The
+`block` rule required `OPEN_BRACE NEWLINE+ ...` and rejected `{ }` with a FATAL
+"extraneous input '}' expecting NEWLINE", corrupting any empty-body list entry;
+the rule now also accepts `OPEN_BRACE CLOSE_BRACE`. Lesson banked: test fixtures
+must mirror device-rendered output (inline empty blocks), not idealized
+hand-written multi-line form — the bug was masked by a `{ \n }` fixture and
+exposed only by a device-collected config.
+
+### ECMP static routes (`sros_ecmp`, batfish/batfish#9989)
+
+A `static-routes route <prefix>` with multiple `next-hop` entries is an ECMP
+route. Extraction now emits one VI `StaticRoute` per next-hop (each with its own
+admin-state/metric/preference) instead of taking the first and warning;
+Batfish's best-preference selection installs the equal-preference legs as ECMP
+and drops worse ones. Device-confirmed: two equal-preference next-hops both
+install (route-table `nexthop` list has two entries); two unequal-preference
+next-hops install only the lower-preference leg.
+
+### prefix-list `to` / `address-mask` bounds (batfish/batfish#9990)
+
+Both were over-approximated as "this length or longer" and warned. Device
+probing settled the semantics. `type to` is a **length-range match**: a route
+matches iff it is contained in the base network and its length is in
+`[base-length .. to-prefix-length]` — the to-prefix supplies the upper length
+bound (the sibling of `through`), and must be nested in the base (commit rejects
+a non-nested to-prefix: `PLCY #1001 ... does not share same most significant
+bits`). It is modeled as one exact `RouteFilterLine` per ancestor length of each
+to-prefix, so off-path siblings do not match (a single base-prefix line with a
+length range would wrongly match them). `address-mask` with a contiguous mask
+equal to the base length is an exact match; a non-contiguous mask cannot be a
+length `SubRange` and is over-approximated with a warning.
+
+### BGP communities + complex attributes (`sros_bgp_comm_in`)
+
+The cEOS oracle (r2) advertises three loopbacks to r1 tagged with standard
+communities, MED, and AS-path prepend; r1 imports all, so its BGP RIB carries
+non-empty attributes validated on both sides. Finding: **SR-OS renders
+well-known communities symbolically** (`no-export`), so the validator
+canonicalizes them to numeric (`no-export` → `65535:65281`, etc.) to match
+Batfish, raising on an unknown symbolic name rather than masking. Also: for a
+BGP route the device's main-RIB metric is the IGP cost to the next-hop (0 for a
+directly-connected peer) while Batfish carries the MED into the main-RIB metric,
+so the validator compares the BGP MED on the BGP RIB and skips the main-RIB
+metric for BGP-protocol routes.
+
+### IS-IS single area (`sros_isis`)
+
+`router "<name>" isis <instance>` is parsed into a typed `IsisProcess`
+(system-id, area-address, level-capability, per-interface interface-type/passive)
+and converted to a VI `IsisProcess`: the NET is the first area-address +
+system-id + an `00` N-selector, a passive interface advertises its subnet but
+forms no adjacency (VI `PASSIVE` mode), and per-level interface settings carry
+the point-to-point flag. SR-OS IS-IS route preferences (admin distances) were
+added to the `NOKIA_SROS` `RoutingProtocol` defaults — L1 15 / L2 18 internal,
+L1 160 / L2 165 external — confirmed on device (an L2 internal route installs at
+preference 18, default interface metric 10). Validated with a two-SR-SIM L2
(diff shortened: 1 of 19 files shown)


dhalperi (commented):
> @dhalperi reviewed 19 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9992)* status: all files reviewed, 5 unresolved discussions.

___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 354 at r1](https://reviewable.io/reviews/batfish/batfish/9992#-OuilCzT26RsdZRlZ553:-OuilCzT26RsdZRlZ554:bsta7lk) ([raw file](https://github.com/batfish/batfish/blob/d914addece34d7d1a8b7d1bb378bad5634d9657c/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L354)):*
> ```Java
>         // https://github.com/batfish/batfish/issues/9989.
>         for (Map.Entry<String, SrosStatementTree> nhe : nextHop.getChildren().entrySet()) {
>           SrosStatementTree nhEntry = nhe.getValue();
> ```

why isn't the code shared, and jus tnext-hop be different by nhip vs blackhole?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 549 at r3](https://reviewable.io/reviews/batfish/batfish/9992#-OuimIeW53w2h7LMsGtv:-OuimIeW53w2h7LMsGtw:b-9t5u6l) ([raw file](https://github.com/batfish/batfish/blob/e9fd03f671f18222676c94e45e01c7a2c4e5f1ee/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L549)):*
> ```Java
>       instance = Integer.parseInt(e.getKey());
>     } catch (NumberFormatException ex) {
>       instance = 0;
> ```

why do we fallback to 0 vs warn and skip or something else?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 554 at r3](https://reviewable.io/reviews/batfish/batfish/9992#-OuimQF5BucmdBMK6lgm:-OuimQF5BucmdBMK6lgn:b-jgy6ea) ([raw file](https://github.com/batfish/batfish/blob/e9fd03f671f18222676c94e45e01c7a2c4e5f1ee/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L554)):*
> ```Java
>     IsisProcess proc = new IsisProcess(instance);
>     Boolean adminUp = toEnum(isisNode.getChild("admin-state"), ADMIN_STATE, "admin-state");
>     proc.setAdminStateEnable(!Boolean.FALSE.equals(adminUp));
> ```

what is this? If the point is "not explicitly false, might be null if not configured" -- The more idiomatic version if `firstNonNull(adminUp. Boolean.FALSE)` 
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 573 at r3](https://reviewable.io/reviews/batfish/batfish/9992#-OuimdsR0l6KYNm5qauN:-OuimdsR0l6KYNm5qauO:bop6j9y) ([raw file](https://github.com/batfish/batfish/blob/e9fd03f671f18222676c94e45e01c7a2c4e5f1ee/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L573)):*
> ```Java
>             toEnum(ifNode.getChild("interface-type"), ISIS_INTERFACE_TYPE, "isis interface-type"));
>         Boolean passive = toBoolean(ifNode.getChild("passive"), "isis interface passive");
>         isisIf.setPassive(Boolean.TRUE.equals(passive));
> ```

similar comment here and everywhere we do this Boolean.CONST.equals() as a skirt for unconfigured boolean values.
___
*[`projects/batfish/src/test/java/org/batfish/vendor/sros/grammar/SrosConversionTest.java` line 243 at r1](https://reviewable.io/reviews/batfish/batfish/9992#-OuilrPm8bsD_1uB5TPI:-OuilrPm8bsD_1uB5TPJ:b1d3fst) ([raw file](https://github.com/batfish/batfish/blob/d914addece34d7d1a8b7d1bb378bad5634d9657c/projects/batfish/src/test/java/org/batfish/vendor/sros/grammar/SrosConversionTest.java#L243)):*
> ```Java
>     assertNotNull(toList);
>     // On-path ancestors of 10.20.0.0/20 (lengths 16..20):
>     assertTrue(toList.permits(Prefix.parse("10.20.0.0/16")));
> ```

is there no matchers for this?

dhalperi:
Addressed all five in `ed6c266e64` ("sros: address review feedback"):

1. **Static-route shared code (line 354):** extracted a single `toStaticRoute(prefix, leg, nextHopKey)` helper used by both legs — the only difference is now `nextHopKey == null` → blackhole vs. set next-hop IP. admin-state/metric/preference are common.
2. **IGP instance fallback to 0 (line 549):** now `toInstanceId(...)` warns ("ospf/isis instance '<key>' is not an integer; skipping") and returns null → the process is skipped, instead of silently defaulting to 0. Applied to both OSPF and IS-IS.
3 & 4. **`Boolean.CONST.equals()` idiom (lines 554, 573):** replaced everywhere (static admin-state, ospf/isis process admin-state, isis interface passive) with `firstNonNull(value, default)` — default FALSE for static admin-state/passive, TRUE for the IGP-process admin-state.
5. **RouteFilterList matchers (test line 243):** converted the prefix-list assertions to `assertThat(rfl, permits(...))` / `rejects(...)` from `RouteFilterListMatchers` (also updated the pre-existing system-pfx/lo-range/host-range assertions in the same test).

dhalperi (commented):
> @dhalperi reviewed 2 files and all commit messages, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9992)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9993, opened 2026-06-10
sros: aggregate routes, LAG, multi-area OSPF, and richer policy

dhalperi:
Four more SR-OS feature areas on top of the merged P9 work (#9992), each modeled
against a live SR-SIM 26.3.R1 lab and validated route-for-route in the sibling
lab-validation repo.

**Aggregate routes** — `router "Base" aggregates aggregate <prefix>` → a discard
`GeneratedRoute` at admin 130. Batfish's generated-route activation matches SR-OS
(installed only with a contributing more-specific); device-confirmed that the
peer learns the /16 aggregate but not the /24 contributors under `summary-only`.

**LAG** — `lag "<name>"` → an `AGGREGATED` interface with `AGGREGATE` member
dependencies and channel-group members; the members' `channelGroup` is set so the
logical-L1 computation collapses the per-member edges into one `lag`↔`lag` edge
and the LAG comes up via a BIND dependency rather than an LACP failure. (Device
note: `lacp mode active` requires a mandatory `administrative-key`.)

**Multi-area OSPF** — no conversion change needed (the existing area loop handles
N areas / ABR); added tests proving a two-area config → two VI `OspfArea`s.
(OSPF `export-policy` redistribution into OSPF did not originate AS-external LSAs
on SR-SIM 26.3.R1, so E1/E2 is left unmodeled and documented.)

**Policy depth** — `from community` / `from as-path` matching (named community /
as-path lists; adds an `AsPathList` model), and the actions `local-preference`,
`metric add`, and `origin`.

Each feature has extraction + conversion unit tests; the comprehensive policy and
multi-area cases are tested behaviorally. SR-OS Java suite + PMD + checkstyle +
buildifier clean. Validated green across the three new lab-validation labs
(sros_aggregate, sros_ospf_multiarea, sros_policy_depth) plus no regression on the
14 existing SR-OS labs. Vendor guide updated.

### DIFF
--- a/docs/parsing/vendors/sros.md
+++ b/docs/parsing/vendors/sros.md
@@ -583,3 +583,62 @@ from the computed main RIB; those two main-RIB checks are sickbayed to
 batfish/batfish#9991. Everything else (config/interfaces/BGP RIB/RD-on-VRF)
 validates green. Device-confirmed: BGP-VPN learned route preference 170,
 resolved over an LDP tunnel; VPRN `red` stays separate from Base.
+
+## Follow-up rungs (P10): aggregate routes, LAG, multi-area OSPF, policy depth
+
+A third pass added four more feature areas, each modeled against a live SR-SIM
+26.3.R1 lab.
+
+### Aggregate routes (`sros_aggregate`)
+
+`router "Base" aggregates aggregate <prefix>` converts to a discard {@code
+GeneratedRoute} at admin distance 130 (the device's aggregate preference).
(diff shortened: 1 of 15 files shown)


dhalperi (commented):
> @dhalperi reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9993)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9997, opened 2026-06-10
sros: derive IS-IS system-id, honor OSPF preference and IS-IS interface metric

dhalperi:
Three underlay-model gaps surfaced by the Nokia service_config_qrg lab (the
combined 4-PE/2-P topology from github.com/nokia/sros-docs-lab), all confirmed
against live SR-SIM 26.3.R1 ground truth:

- IS-IS system-id auto-derivation. The Nokia configs set no explicit
  `system-id`, so SR OS derives it from the system interface IPv4 address
  (each octet zero-padded to three digits, regrouped as XXXX.XXXX.XXXX, e.g.
  10.10.10.10 -> 0100.1001.0010). Conversion previously skipped IS-IS entirely
  when system-id was absent; now it derives one when the system interface has
  an address.

- OSPF `preference`. Extracted and used as the internal-route admin distance
  (was hardcoded to the SR-OS default 10). The lab raises OSPF preference to 20
  so IS-IS (preference 18) is the preferred IGP; without this the wrong IGP
  populated the RIB.

- IS-IS interface `metric`. Extracted from both the per-interface and per-level
  (`level <N> metric`) forms and applied to the VI interface level settings
  (default 10 when unset). The lab uses `level 2 metric 100`, so without this
  the computed IS-IS route metrics were 10/20 instead of 100/200.

With all three, the six PE/P configs parse with zero FATAL warnings and the
computed main-RIB IS-IS routes match the device route-for-route (admin 18,
metric 100 one-hop / 200 two-hop).

----

Prompt:
```
My Nokia contact tells me they have pre-canned labs that run in containerlab.
Let's use this as our next steps on the complexity ladder. Check these out,
assess whether they have interesting content for us, and then go ahead and run
the full ec2->collect->validate->batfish-fixes-(if appropriate) cycle for them.
Probably, first step is to categorize each lab and organize them in increasing
complexity and flag whether they rely on features not in the Batfish VI model
yet. Prefer lower complexity first, and full Batfish support is better than pure
extraction!

We have several configuration examples here:
https://github.com/nokia/sros-docs-lab based on this guide:
https://documentation.nokia.com/sr/26-3/7750-sr/titles/services-config-quick-ref.html.
They all run in Containerlab so I hope that makes it easy for you.
```

The Nokia repo turned out to be a single combined "kitchen-sink" topology (4 PE
+ 2 P SR-SIM + 4 Linux CE) stacking the full SP service suite on one IGP/MPLS
underlay, not a graded ladder. Per the "full support > extraction, lower
complexity first" steer, this commit models the routed underlay the new lab
exercises (the service overlay parses cleanly and is sickbay'd on the
lab-validation side).

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java
@@ -581,6 +581,9 @@ public final class SrosFeatureExtractor {
   /** OSPF interface {@code metric} (cost): YANG {@code uint32 range "1..65535"}. */
   private static final IntegerSpace OSPF_METRIC_SPACE = IntegerSpace.of(new SubRange(1, 65535));
 
+  /** IS-IS interface {@code metric}: YANG wide-metric {@code uint32 range "1..16777215"}. */
+  private static final IntegerSpace ISIS_METRIC_SPACE = IntegerSpace.of(new SubRange(1, 16777215));
+
   /** OSPF {@code interface-type} enumeration (the modeled subset). */
   private static final Map<String, OspfAreaInterface.InterfaceType> OSPF_INTERFACE_TYPE =
       ImmutableMap.of(
@@ -624,6 +627,13 @@ public final class SrosFeatureExtractor {
     Boolean adminUp = toEnum(ospfNode.getChild("admin-state"), ADMIN_STATE, "admin-state");
     // admin-state defaults to enable for an OSPF/IS-IS process: enabled unless explicitly disable.
     proc.setAdminStateEnable(firstNonNull(adminUp, Boolean.TRUE));
+    // OSPF route preference (admin distance) for internal routes; SR-OS default 10.
+    toIntegerInSpace(
+            singleValue(ospfNode, "preference"),
+            singleValueNode(ospfNode, "preference"),
+            ROUTE_PREFERENCE_SPACE,
+            "ospf preference")
+        .ifPresent(proc::setPreference);
 
     SrosStatementTree areaList = ospfNode.getChild("area");
     if (areaList != null) {
@@ -694,6 +704,25 @@ public final class SrosFeatureExtractor {
             toEnum(ifNode.getChild("interface-type"), ISIS_INTERFACE_TYPE, "isis interface-type"));
         Boolean passive = toBoolean(ifNode.getChild("passive"), "isis interface passive");
         isisIf.setPassive(firstNonNull(passive, Boolean.FALSE));
+        // Metric is configured per level (`interface "<name>" level <N> metric <M>`); SR-OS has no
+        // interface-wide metric leaf. Extract each level's metric independently; the default 10
+        // applies per level in conversion when unset.
+        SrosStatementTree ifLevelList = ifNode.getChild("level");
+        if (ifLevelList != null) {
+          for (int lvl : new int[] {1, 2}) {
+            SrosStatementTree levelNode = ifLevelList.getChild(Integer.toString(lvl));
+            if (levelNode == null) {
+              continue;
+            }
+            int metricLevel = lvl;
+            toIntegerInSpace(
+                    singleValue(levelNode, "metric"),
+                    singleValueNode(levelNode, "metric"),
+                    ISIS_METRIC_SPACE,
+                    "isis interface metric")
+                .ifPresent(m -> isisIf.setMetric(metricLevel, m));
+          }
+        }
         proc.getInterfaces().put(ifName, isisIf);
       }
     }
--- a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/IsisProcessInterface.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/IsisProcessInterface.java
@@ -44,7 +44,26 @@ public final class IsisProcessInterface implements Serializable {
     _interfaceType = interfaceType;
   }
 
+  /**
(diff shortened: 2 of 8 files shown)


dhalperi (commented):
> @dhalperi reviewed 8 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9997)* status: all files reviewed, 4 unresolved discussions.

___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 630 at r1](https://reviewable.io/reviews/batfish/batfish/9997#-Ounw-nCB3HIO7HIbdam:-Ounw-nCB3HIO7HIbdan:b-xwvm5c) ([raw file](https://github.com/batfish/batfish/blob/83c1ea257d98d96736a60bbdc22c1c06c6a37529/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L630)):*
> ```Java
>     // admin-state defaults to enable for an OSPF/IS-IS process: enabled unless explicitly disable.
>     proc.setAdminStateEnable(firstNonNull(adminUp, Boolean.TRUE));
>     // OSPF route preference (admin distance) for internal routes; SR-OS default 10. The lab raises
> ```

don't refer to the lab in VI model! This is just standard extraction for a proeprty we were previously failing to extract.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 706 at r1](https://reviewable.io/reviews/batfish/batfish/9997#-OunxpEXEH53jVe_xnJI:-OunxpEXEH53jVe_xnJJ:b-4rheia) ([raw file](https://github.com/batfish/batfish/blob/83c1ea257d98d96736a60bbdc22c1c06c6a37529/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L706)):*
> ```Java
>         isisIf.setInterfaceType(
>             toEnum(ifNode.getChild("interface-type"), ISIS_INTERFACE_TYPE, "isis interface-type"));
>         Boolean passive = toBoolean(ifNode.getChild("passive"), "isis interface passive");
> ```

do we support IS-IS preference override? I assume it exists.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java` line 718 at r1](https://reviewable.io/reviews/batfish/batfish/9997#-OunxmdF7D5Kx1tsKOe_:-OunxmdF7D5Kx1tsKOea:boo1se) ([raw file](https://github.com/batfish/batfish/blob/83c1ea257d98d96736a60bbdc22c1c06c6a37529/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosFeatureExtractor.java#L718)):*
> ```Java
>           if (level2 != null) {
>             metricVal = singleValue(level2, "metric");
>             metricNode = singleValueNode(level2, "metric");
> ```

presumably this level2 metric doesn't apply to level1, right? How are we modeling this?
___
*[`projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java` line 1099 at r1](https://reviewable.io/reviews/batfish/batfish/9997#-OunyxLZ7-g7aEsQboPh:-OunyxLZ7-g7aEsQboPi:b-lhg0sh) ([raw file](https://github.com/batfish/batfish/blob/83c1ea257d98d96736a60bbdc22c1c06c6a37529/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java#L1099)):*
> ```Java
>     String digits =
>         String.format(
>             "%03d%03d%03d%03d",
> ```

why not put the periods in here instead of doing the substring stuff below?

dhalperi:
Thanks — addressed all four (pushed as an amend, `c652340`):

1. **OSPF preference comment referring to the lab** — fixed. Dropped the lab reference; it now just reads "OSPF route preference (admin distance) for internal routes; SR-OS default 10." It's standard extraction of a property we were previously dropping.

2. **IS-IS preference override** — there isn't one to support. Checked `nokia-conf-router-isis.yang`: the top-level IS-IS instance has no route-`preference` leaf (unlike OSPF, which is exactly why the lab raises OSPF's to make IS-IS win). The only `preference` leaves under IS-IS are `multi-homed-prefix` LFA backup preference and a per-level *external* route preference — neither is the internal-route admin distance. So SR-OS IS-IS internal admin distance is fixed (L1 15 / L2 18), which is what we already convert. Nothing to extract here.

3. **level-2 metric vs level-1** — good catch, this was sloppy and I fixed it properly. SR-OS has **no** interface-wide metric leaf; metric is configured per level (`interface "<name>" level <N> metric <M>`). The old code read only the level-2 metric and applied it to *both* L1 and L2 settings. Now the extractor reads each level's metric independently (`IsisProcessInterface.getMetric(level)`), and conversion applies each to its own level, defaulting to 10 per level when unset. Added `testIsisPerLevelMetric` (level 1 metric 30 + level 2 metric 100 → distinct per-level costs) to lock it in.

4. **periods in the format string for system-id** — can't put them directly in the `%03d` format because the digit grouping doesn't align with octet boundaries: each octet is 3 digits but the system-id groups are 4 digits (`010|010|010|010` → `0100.1001.0010`), so a period would fall mid-octet. I kept the format-to-12-digits + slice approach since that's the part that handles the regrouping; open to a cleaner idiom if you have one in mind.

dhalperi (commented):
> @dhalperi reviewed 5 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9997)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9998, opened 2026-06-11
sros: report silently-ignored statements in the annotate tool

dhalperi:
The annotate tool prints a comment above each config line it parses but
intentionally does not model ("SILENTLY IGNORED"). Grammar-driven
vendors detect these at parse time: any ANTLR rule whose name ends in
`_null` is flagged by SilentSyntaxListener. SR-OS has no such rules — its
grammar is generic (every statement is a word sequence), and the
"do we model this?" decision lives in SrosFeatureExtractor, which reads
only the specific tree paths it knows. So the annotate tool's
SILENTLY IGNORED output was always empty for SR-OS.

Track which SrosStatementTree nodes the extractor reads, then report the
rest. The tracking is automatic: the tree's own read accessors (getChild,
getChildren, hasChildren) record the nodes they touch into a tree-shared
VisitRecorder, so the extractor needs no visit-tracking code of its own.
The recorder is inactive during the build and the SrosPreprocessor walk
(whose accesses must not count as a "read") and is activated by
SrosControlPlaneExtractor only for the extraction pass, and only when
printParseTree is set — so the normal pipeline and reference tests are
unchanged. After extraction, SrosSilentSyntax sweeps the tree and emits
one element per maximal unvisited subtree: a multi-line brace block
collapses to one comment at its header line, a single-line leaf is
reported in place.

Verified against five real lab captures: each flags the unmodeled
`system security` block (one comment, not per-leaf) and the top-level
`persistent-indices` block, and the prefix-bounds lab additionally flags
an unmodeled `family` leaf nested inside a modeled BGP group — with no
false positives on modeled config.

----

Prompt:
```
The annotate tool also prints out lines that are silently ignored. Usually, it
does this by inspecting the ANTLR4 extrator. For SR-OS., however, we don't have
one. Can we develop a way to track which tree leaves have been visited, so that
annotate tool keeps this functionality?
```

Per follow-up direction, the marking is fully automatic in the
SrosStatementTree accessors (no intentional mark() calls in the
extractor), and the tool was run on real lab configs to confirm it works
before writing tests.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/sros/grammar/SrosControlPlaneExtractor.java
@@ -53,7 +53,21 @@ public final class SrosControlPlaneExtractor implements ControlPlaneExtractor {
     // populate the typed feature model from it.
     SrosStatementTree root = cb.getTree();
     SrosPreprocessor.preprocess(root, _w);
+
+    // For the annotate tool, track which tree nodes the feature extractor reads (it has no
+    // visit-tracking code of its own — see SrosStatementTree), then report the rest as silently
+    // ignored. SR-OS has no `_null` grammar rules, so this replaces the parse-time mechanism that
+    // grammar-driven vendors use. Tracking starts after preprocessing so that the preprocessor's
+    // tree walk does not count as "read".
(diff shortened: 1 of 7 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9998)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #9999, opened 2026-06-11
sros: qualify VPRN interface names with their VRF

### LINKED ISSUE #203: SR-OS: VPRN interface with reused name across VRFs is dropped (modeled in only one VRF)
## Summary

On Nokia SR-OS, the same interface name can be configured in multiple
`service vprn` instances (each VPRN is a separate VRF). Batfish's SR-OS model
keys interfaces by name within a node, so when a name is reused across two
VPRNs only one interface survives — the other is silently dropped (no parse
warning). The device has the interface present in both VRFs with distinct
addresses, so `test_interface_properties` fails for the affected nodes.

## Repro (sros_services lab)

pe1 configures `to-cea` in both the RED and BLUE VPRNs:

```
/configure service vprn "BLUE" interface "to-cea" ipv4 primary address 192.168.40.254 prefix-length 24
/configure service vprn "RED"  interface "to-cea" ipv4 primary address 192.168.30.254 prefix-length 24
```

The device (`info json /state service vprn "<name>" interface *`) reports
`to-cea` in *both* VPRNs (RED 192.168.30.254/24, BLUE 192.168.40.254/24).
Batfish `interfaceProperties` for pe1 returns a single `to-cea`, in VRF BLUE
(the first-configured); the RED instance is absent. pe3 shows the same with
`to-cez` reused across RED/BLUE.

## Expected

Batfish should model both interfaces, each scoped to its VPRN's VRF — i.e.
interface identity in the SR-OS model needs to be (VRF, name), not name alone.

## Notes

Surfaced when the lab_builder collector was changed to discover VPRN names
dynamically (batfish/lab-validation#202) and the sros_services snapshot was
refreshed with real per-VPRN interface state; previously the VPRN interface
captures were empty `{}`, so the validator's per-VRF interface comparison was
never actually exercised on pe1/pe3 (the lab README's "per-VRF interfaces →
modeled" claim was untested). `test_interface_properties` for pe1/pe3 is
sickbay'd against this issue.

dhalperi:
SR-OS scopes interface names per router instance, so the same name (e.g.
to-cea) can be configured in two different VPRNs, each its own VRF. The
VI model keys interfaces by name within a node, so a name reused across
two VPRNs collided and only the first-configured copy survived; the
other was silently dropped.

Name every non-Base (VPRN) interface <vrf>.<name> in the VI model so
both copies survive under distinct keys; Base-instance interfaces keep
their bare names. The qualification is applied uniformly to all VPRN
interfaces (not only on a detected collision) so the VI name is a
deterministic function of the VPRN and interface alone. The OSPF and
IS-IS interface lookups use the same qualified name.

Surfaced by the lab-validation sros_services lab, where pe1 reuses
to-cea across the RED and BLUE VPRNs (pe3: to-cez) with distinct
addresses; the device reports the interface in both VRFs but Batfish
modeled only one.

Fixes batfish/lab-validation#203.

----

Prompt:
```
Let's fix the interface duplicating naming across VPRN bug you just
filed, and make that part of the lab pass. (See recent PR to
lab-validation). I think the idiomatic fix is to name both interfaces
<vrf>.<iface> when collisions happen
```

Refined in conversation: name all non-Base VPRN interfaces <vrf>.<iface>
unconditionally (not only on collision), for a deterministic VI name.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/sros/representation/SrosConversions.java
@@ -457,13 +457,17 @@ public final class SrosConversions {
    *       its physical port, so an L1 edge between ports yields the L3 edge between the router
    *       interfaces (and the logical interface shares the port's fate).
    * </ul>
+   *
+   * <p>A non-Base (VPRN) interface's VI name is qualified with its VRF — see {@link
+   * #viInterfaceName} — because SR-OS scopes interface names per router instance and Batfish keys
+   * interfaces by name per node.
    */
   static void convertInterfaces(SrosConfiguration vc, Router router, Configuration c, Vrf vrf) {
     for (RouterInterface ri : router.getInterfaces().values()) {
       String portPath = ri.getPort();
       Interface.Builder ib =
           Interface.builder()
-              .setName(ri.getName())
+              .setName(viInterfaceName(router, ri.getName()))
               .setOwner(c)
               .setVrf(vrf)
               // SR-OS router interfaces have no shutdown leaf in the modeled subset; they are up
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/9999)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10001, opened 2026-06-17
BgpTopologyUtils: allow source IP inference without a FIB

zphw:
The `sourceFib` is nullable as `UseConstantIp.getPotentialSourceIps()` does not need a FIB. The `sourceFib != null` guard prevented local IP resolution for peers using this strategy (Junos-style devices with default-address-selection), causing it to skip session establishment when FIBs are unavailable.

The fix removed the `sourceFib != null` check. If the `SourceIpInference` is a `UseConstantIp`, it returns the answer directly. If it's `InferFromFib` (needs a routing table), `InferFromFib.getPotentialSourceIps` already gracefully handles null FIB by returning an empty set, which is the same behavior as before.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10001)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on zphw).

PR #10008, opened 2026-06-23
Junos: warn when then add-path send-count lacks group add-path send

dhalperi:
Junos ignores a policy's `then add-path send-count N` unless add-path
send is also enabled at the BGP group/neighbor level (`family inet
unicast add-path send path-count N`).

This parses `then add-path send-count` (previously unparsed) and models
it as `PsThenAddPathSendCount`. It is a no-op in the vendor-independent
model, since Batfish does not yet model the additional-paths count (same
gap as the group-level directive). During BGP conversion, neighbors
without effective add-path send get a risky warning for each export
policy using the directive, naming the policy, term, count, and
neighbor.

Follow-up: also handle policies reached indirectly through `from policy`
subroutine calls, not just those listed directly in `export`.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -2663,6 +2663,8 @@ SELF_PING_DURATION: 'self-ping-duration';
 
 SEND: 'send';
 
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10008)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10011, opened 2026-06-25
BGP: model tcp_l3mdev_accept via SessionVrfScope (accept from any VRF)

### LINKED ISSUE #10010: Model tcp_l3mdev_accept: BGP sessions accepted regardless of ingress VRF (FRR/SONiC/Cumulus)
## Problem

Linux-based network OSes (FRR, Cumulus, SONiC) run BGP over a single global
listening socket in the default VRF and rely on the kernel sysctl
`net.ipv4.tcp_l3mdev_accept=1` to let that socket accept connections that
arrive on *any* VRF. SONiC ships `tcp_l3mdev_accept=1` (and
`udp_l3mdev_accept=1`) by default; FRR's vrf-lite backend requires it for VRF
BGP to come up.

Batfish models BGP acceptance as strictly per-ingress-VRF. As a result, it
reports legitimate VRF-BGP sessions on these platforms as not establishing:
the reachability check in `canEstablishBgpSession` requires the reverse flow's
ingress VRF to equal the listener's session VRF, and listener registration is
keyed to a single VRF. There is no way to express "this listener accepts a
connection regardless of which VRF it arrived on."

## Background: what the flag actually does

`tcp_l3mdev_accept` relaxes **socket selection**, not forwarding/local
delivery. The two layers:

1. Routing/local-delivery layer (unchanged by the flag). When an interface is
   enslaved to a VRF, its local /32 host route lives in that VRF's table. For a
   SYN to be locally delivered at all, the destination /32 must be deliverable
   in the VRF the packet physically arrives on — its *native* VRF. The flag
   does not change this.
2. Socket layer (what the flag changes). Given the packet was locally
   delivered on VRF A, which listening socket accepts it? Default (`=0`): only
   a socket bound to A (`SO_BINDTODEVICE`); an unbound/global socket sees only
   default-VRF traffic. With `=1`: the unbound global socket accepts the
   A-arriving connection and pins the resulting child socket to the ingress
   VRF A (`newsk->sk_bound_dev_if = ir_iif`).

Consequence for modeling: we do NOT need to fabricate or leak a /32 into the
BGP instance's VRF. The /32 already exists in its native VRF (and Batfish's
per-VRF IP ownership already encodes this). The only thing that needs to
change is decoupling "the VRF the connection arrived on" from "the VRF the BGP
instance is configured in" at acceptance time.

Reply traffic egresses the ingress VRF (the kernel pins the child socket to
`ir_iif`), which is already what `canEstablishBgpSession` sources via
`reverseFlow.getIngressVrf()`.

Protocol asymmetry to keep in mind: TCP and UDP default to 0, but
`raw_l3mdev_accept` defaults to 1. UDP does not get automatic
reply-out-ingress-VRF behavior. This issue scopes to TCP/BGP.

## Proposed model

Generalize the existing `BgpPeerConfig.sessionVrf` (currently a nullable
`String`) into a sealed tri-state describing which ingress VRF(s) the
listener's socket will accept from:

```java
sealed interface SessionVrfScope {
  record OwnVrf()             // was: sessionVrf == null
  record SpecificVrf(String)  // was: sessionVrf == non-null (Junos forwarding-context)
  record AnyVrf()             // new: tcp_l3mdev_accept on the listener's node
}
```

The three states are mutually-exclusive answers to the same question, so a
sealed type unifies them. The unification assumes l3mdev (global socket) and a
specific Junos `forwarding-context` never co-occur on one peer, which holds:
forwarding-context is Junos-specific and l3mdev is the FRR/SONiC global-socket
design.

`AnyVrf` only relaxes socket selection. Forwarding still does the real work:
in the `checkReachability=true` path the end-to-end forward+reverse traceroute
must still succeed, which means the listener's local IP still had to be locally
delivered in some (native) VRF.

### Initiation vs. acceptance asymmetry

`AnyVrf` is a listener-side concept only. `tcp_l3mdev_accept` relaxes
*acceptance* of inbound connections; it does not change origination. When a
device opens an outbound BGP connection the kernel still routes the SYN out of
exactly one VRF, so the forward flow always has a single concrete ingress VRF.
The scope resolves per role:

| Scope          | As listener (accepts)   | As initiator (forward flow origin) |
|----------------|-------------------------|------------------------------------|
| `OwnVrf`       | own VRF only            | own VRF                            |
| `SpecificVrf(v)` | `v` only              | `v`                                |
| `AnyVrf`       | any VRF                 | own/config VRF (collapses to OwnVrf) |

So the forward-flow construction at
`BgpTopologyUtils.java:594` (`setIngressVrf(firstNonNull(initiator.getSessionVrf(),
initiatorId.getVrfName()))`) needs no semantic relaxation — it just resolves the
initiator's scope to a concrete origin VRF, with `AnyVrf` falling back to the
config VRF. `AnyVrf` only fans out at listener registration (`:205-209`) and
only relaxes the ingress-VRF equality clause on accept (`:626-629`).

This asymmetry is forced by physics, not a configuration degree of freedom:
the kernel flag is one bit, and `OwnVrf`/`SpecificVrf` are symmetric (same VRF
for both roles) while only `AnyVrf` differs by role. So a single unified type
is correct — splitting into separate listener/initiator fields would make
unrealizable states representable (e.g. `listen=SpecificVrf(blue),
initiate=AnyVrf`). Rather than encode the role behavior at each call site, put
it on the type as resolver methods, so the "AnyVrf collapses to own on
initiate" rule lives in exactly one tested place:

```java
sealed interface SessionVrfScope {
  /** Concrete VRF an outbound (initiated) connection is sourced from. */
  String originVrf(String configVrf);                // AnyVrf -> configVrf

  /** VRFs this listener registers under to accept inbound connections. */
  Set<String> listenVrfs(String configVrf, Set<String> allNodeVrfs);  // AnyVrf -> allNodeVrfs

  /** Whether an inbound connection arriving on ingressVrf is acceptable. */
  boolean acceptsIngressVrf(String ingressVrf, String configVrf);     // AnyVrf -> always true

  record OwnVrf() implements SessionVrfScope { ... }
  record SpecificVrf(String vrf) implements SessionVrfScope { ... }

  /**
   * tcp_l3mdev_accept on the listener's node: the global socket accepts
   * connections arriving on any VRF. As an initiator there is no "originate
   * from any" — the outbound socket is bound to one VRF — so origination
   * collapses to the peer's own/config VRF.
   */
  record AnyVrf() implements SessionVrfScope { ... }
}
```

Call sites then use the methods rather than pattern-matching the raw type:
`:594` calls `originVrf(...)`, listener registration `:205-209` calls
`listenVrfs(...)`, and the accept check `:626-629` calls
`acceptsIngressVrf(...)`.

Initiation is unambiguous: FRR vrf-lite bgpd binds the outbound socket to the
VRF device (`SO_BINDTODEVICE`) before connecting, so the egress VRF is pinned by
the BGP instance's configured VRF — the remote/peer address is routed only
within that already-chosen VRF, so dual default routes across VRFs do not create
an origination race. (The "wrong VRF" pathology in FRR #8641 is a separate,
accept-side, transient-startup bug and is out of scope.)

## Affected code

- `projects/common/src/main/java/org/batfish/datamodel/BgpPeerConfig.java:296-303`
  — replace the `sessionVrf` String with `SessionVrfScope`. Ripple effects to
  account for:
  - **JSON serialization.** `PROP_SESSION_VRF` currently serializes a bare
    string/null and `null` means own-VRF. Need a forward encoding for all three
    that keeps old readers and reference files working: e.g. keep emitting
    `null` for `OwnVrf` and the bare VRF name for `SpecificVrf`, and add a
    sentinel (e.g. `"*"`) or a structured object for `AnyVrf`. This is a
    ref-test-affecting decision — call it out explicitly rather than "preserve
    JSON compat."
  - **Builder/constructor churn.** `Builder.setSessionVrf(@Nullable String)`
    (`:524`) and the constructor param (`:125`) are `String`-typed and called
    from many vendor extraction sites. Either migrate callers or keep a
    `String` overload mapping null->`OwnVrf` / non-null->`SpecificVrf` for
    source compatibility.
  - **Value semantics.** `_sessionVrf` participates in `equals` (`:331`),
    `hashCode` (`:354`), and `toString` (`:377`); update for the new type.
- `projects/common/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java:205-209`
  — receiver registration: `OwnVrf`/`SpecificVrf` register under one VRF as
  today; `AnyVrf` registers the listener under all VRFs on the node so the
  initiator's owner lookup (`:261-278`) finds it from the native VRF.
- `projects/common/src/main/java/org/batfish/datamodel/bgp/BgpTopologyUtils.java:624-639`
  — in `canEstablishBgpSession`, drop the
  `reverseFlow.getIngressVrf().equals(listenerSessionVrf)` clause for `AnyVrf`,
  keeping the ingress-node match and the `getCheckLocalIpOnAccept()` + localIp
  guard (the "dst is an IP we actually own" check that stays honest).
- `projects/question/src/main/java/org/batfish/question/bgpsessionstatus/BgpSessionAnswererUtils.java:69`
  — the `INVALID_LOCAL_IP` check here is on the **active peer as initiator**: it
  verifies the initiator's `localIp` is owned in its *source* VRF. Because an
  `AnyVrf` initiator originates from its config VRF (`originVrf` collapses to
  config VRF), this check stays per-VRF and needs **no** relaxation. The
  `AnyVrf` effect on the config-only compatibility result is carried entirely by
  topology construction (the `HALF_OPEN` / `adjacentNodes` outcome at `:77`),
  which consumes the relaxed registration/matching above. Listed here only to
  document that it is intentionally left unchanged.
- `BgpSessionStatusAnswerer` / `BgpSessionCompatibilityAnswerer` — render
  `AnyVrf` for the `COL_SESSION_VRF` column.

The data-plane ACCEPTED disposition (`vrfAcceptsIp` /
`TracerouteEngineImplContext.interfaceAcceptingIp`) should stay per-VRF — it
faithfully models local delivery and the flag does not relax it.

## Extraction (separate, later)

No Batfish vendor currently parses kernel sysctls. The flag value lives in
`/etc/sysctl.d/` (FRR/Cumulus) or `90-sonic.conf` (SONiC), which Batfish does
not ingest. Pragmatic first step: default `AnyVrf` for SONiC (matches what it
always ships) and for FRR vrf-lite VRF BGP, deferring real sysctl-file
extraction.

## Suggested phasing

1. `SessionVrfScope` type + BGP topology relaxation + SONiC/FRR defaults. High
   value, low blast radius; fixes "VRF BGP sessions don't come up."
2. Sysctl-file extraction for fidelity beyond the SONiC/FRR default.

### Tests

The existing `BgpTopologyUtilsTest.testInitTopologySessionVrf` exercises
`SpecificVrf`-equivalent behavior only; new coverage is needed for `AnyVrf`:
- resolver-method unit tests on `SessionVrfScope` (`originVrf` collapses
  `AnyVrf` to config VRF; `listenVrfs` returns all node VRFs for `AnyVrf`;
  `acceptsIngressVrf` is always true for `AnyVrf`);
- listener with `AnyVrf` registers under all node VRFs and is matched by an
  initiator whose peer address is owned in a non-config VRF;
- session accepted when the reverse-flow ingress VRF differs from the config
  VRF (`checkReachability=true`), while the localIp guard still rejects a
  non-owned dst;
- an `AnyVrf` initiator still sources the forward flow from its config VRF;
- JSON round-trip for all three scopes, plus a back-compat test that a legacy
  null/string `sessionVrf` deserializes to `OwnVrf`/`SpecificVrf`.

dhalperi:
Linux-based NOSes (FRR, Cumulus, SONiC) run BGP over a listening socket
that is not bound to a particular VRF and rely on the kernel sysctl
tcp_l3mdev_accept=1 to accept connections arriving on any VRF. Batfish
modeled BGP acceptance as strictly per-ingress-VRF, so it reported
legitimate VRF-BGP sessions on these platforms as not establishing.

Generalize BgpPeerConfig's nullable String sessionVrf into a sealed
SessionVrfScope tri-state:

  OwnVrf        - the peer's own (configured) VRF (was: null)
  SpecificVrf   - a specific VRF, e.g. Junos forwarding-context (was: non-null)
  AnyVrf        - accept from any VRF (tcp_l3mdev_accept)

The three states are mutually-exclusive answers to "which VRF is this
peer's session in", so a sealed type unifies them. Role behavior lives on
the type as resolver methods (originVrf/listenVrfs/acceptsIngressVrf), so
call sites in BgpTopologyUtils use the methods rather than pattern-matching:

  - originVrf: outbound connections source from the config VRF; AnyVrf
    collapses to the config VRF (the outbound socket is bound to one VRF).
  - listenVrfs: AnyVrf registers the listener under every VRF on the node
    so an initiator whose peer address is owned in a non-config VRF finds
    it; others register under one VRF.
  - acceptsIngressVrf: AnyVrf accepts any ingress VRF; others require equality.

AnyVrf only relaxes socket selection. The checkReachability path still
requires the end-to-end forward+reverse traceroute to succeed, so local
delivery stays per-VRF and the localIp guard is unchanged.

Default AnyVrf for all FRR/Cumulus/SONiC peers (SONiC delegates to the
FRR conversion). The listening socket is not bound to a particular VRF,
so a peer accepts a connection regardless of the VRF the BGP process is
configured in: a default-VRF process still accepts a connection whose
local IP is delivered on a non-default VRF. Junos forwarding-context now
produces SpecificVrf. Sysctl-file extraction for finer fidelity is
deferred.

OwnVrf serializes as null (omitted), so existing peer JSON is unchanged;
SpecificVrf and AnyVrf serialize as structured objects. The vimodel ref
gains AnyVrf objects on the affected peers.

Fixes batfish/batfish#10010.

----

Prompt:
```
Implement https://github.com/batfish/batfish/issues/10010
```

During implementation the user confirmed: encode the scopes as
structured JSON objects (not a "*" sentinel), migrate all callers to a
scope-typed setter (no String overload), and that polluting reference
test outputs with the new objects is acceptable.

### DIFF
--- a/projects/allinone/src/test/java/org/batfish/e2e/bgp/forwardingcontext/BgpForwardingContextTest.java
+++ b/projects/allinone/src/test/java/org/batfish/e2e/bgp/forwardingcontext/BgpForwardingContextTest.java
@@ -11,6 +11,7 @@ import static org.hamcrest.Matchers.nullValue;
 import java.io.IOException;
 import org.batfish.common.NetworkSnapshot;
 import org.batfish.datamodel.answers.Schema;
+import org.batfish.datamodel.bgp.SessionVrfScope.SpecificVrf;
 import org.batfish.datamodel.questions.BgpSessionStatus;
 import org.batfish.datamodel.table.TableAnswerElement;
 import org.batfish.main.Batfish;
@@ -107,6 +108,6 @@ public class BgpForwardingContextTest {
             .iterator()
             .next()
             .getSessionVrf(),
-        equalTo("default"));
+        equalTo(new SpecificVrf("default")));
   }
 }
--- a/projects/batfish/src/main/java/org/batfish/representation/cumulus_nclu/CumulusConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/cumulus_nclu/CumulusConversions.java
@@ -89,6 +89,7 @@ import org.batfish.datamodel.bgp.Ipv4UnicastAddressFamily;
 import org.batfish.datamodel.bgp.Layer2VniConfig;
 import org.batfish.datamodel.bgp.Layer3VniConfig;
 import org.batfish.datamodel.bgp.RouteDistinguisher;
+import org.batfish.datamodel.bgp.SessionVrfScope.AnyVrf;
 import org.batfish.datamodel.bgp.community.Community;
 import org.batfish.datamodel.bgp.community.ExtendedCommunity;
 import org.batfish.datamodel.ospf.OspfArea;
@@ -492,6 +493,9 @@ public final class CumulusConversions {
     @Nullable
     RoutingPolicy importRoutingPolicy = computeBgpNeighborImportRoutingPolicy(c, neighbor, bgpVrf);
 
+    // Cumulus runs BGP over a listening socket that is not bound to a particular VRF and relies on
+    // the kernel sysctl tcp_l3mdev_accept=1 to accept connections arriving on any VRF.
+    peerConfigBuilder.setSessionVrfScope(AnyVrf.instance());
     peerConfigBuilder
         .setBgpProcess(newProc)
         .setClusterId(inferClusterId(bgpVrf, newProc.getRouterId(), neighbor, localAs))
--- a/projects/batfish/src/main/java/org/batfish/representation/frr/FrrConversions.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/frr/FrrConversions.java
@@ -88,6 +88,7 @@ import org.batfish.datamodel.bgp.Ipv4UnicastAddressFamily;
 import org.batfish.datamodel.bgp.Layer2VniConfig;
 import org.batfish.datamodel.bgp.Layer3VniConfig;
 import org.batfish.datamodel.bgp.RouteDistinguisher;
+import org.batfish.datamodel.bgp.SessionVrfScope.AnyVrf;
 import org.batfish.datamodel.bgp.community.Community;
 import org.batfish.datamodel.bgp.community.ExtendedCommunity;
 import org.batfish.datamodel.ospf.OspfAddresses;
@@ -697,6 +698,9 @@ public final class FrrConversions {
     @Nullable
     BgpNeighborIpv4UnicastAddressFamily ipv4u =
         bgpVrf.getIpv4UnicastConfiguration(neighbor.getName());
+    // FRR/Cumulus/SONiC run BGP over a listening socket that is not bound to a particular VRF and
+    // rely on the kernel sysctl tcp_l3mdev_accept=1 to accept connections arriving on any VRF.
+    peerConfigBuilder.setSessionVrfScope(AnyVrf.instance());
     peerConfigBuilder
         .setBgpProcess(newProc)
         .setClusterId(inferClusterId(bgpVrf, newProc.getRouterId(), neighbor, localAs))
--- a/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/juniper/JuniperConfiguration.java
@@ -139,6 +139,7 @@ import org.batfish.datamodel.bgp.EvpnAddressFamily;
 import org.batfish.datamodel.bgp.Ipv4UnicastAddressFamily;
 import org.batfish.datamodel.bgp.Layer3VniConfig;
 import org.batfish.datamodel.bgp.RouteDistinguisher;
+import org.batfish.datamodel.bgp.SessionVrfScope.SpecificVrf;
 import org.batfish.datamodel.bgp.community.Community;
 import org.batfish.datamodel.bgp.community.ExtendedCommunity;
 import org.batfish.datamodel.dataplane.rib.RibId;
@@ -876,10 +877,11 @@ public final class JuniperConfiguration extends VendorConfiguration {
       // forwarding-context: resolve the VRF from which the BGP TCP session is sourced
       String forwardingContext = ig.getForwardingContext();
       if (forwardingContext != null) {
-        neighbor.setSessionVrf(
-            forwardingContext.equals(DEFAULT_ROUTING_INSTANCE_NAME)
-                ? Configuration.DEFAULT_VRF_NAME
-                : forwardingContext);
+        neighbor.setSessionVrfScope(
+            new SpecificVrf(
+                forwardingContext.equals(DEFAULT_ROUTING_INSTANCE_NAME)
+                    ? Configuration.DEFAULT_VRF_NAME
+                    : forwardingContext));
       }
       neighbor.setBgpProcess(proc);
       neighbor.setIpv4UnicastAddressFamily(
--- a/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/grammar/flatjuniper/FlatJuniperGrammarTest.java
@@ -347,6 +347,8 @@ import org.batfish.datamodel.bgp.AddressFamily;
 import org.batfish.datamodel.bgp.AddressFamilyCapabilities;
 import org.batfish.datamodel.bgp.BgpConfederation;
 import org.batfish.datamodel.bgp.RouteDistinguisher;
+import org.batfish.datamodel.bgp.SessionVrfScope.OwnVrf;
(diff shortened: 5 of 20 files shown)


dhalperi (commented):
> @dhalperi reviewed 21 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10011)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10011)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10012, opened 2026-06-25
pre-commit: run buildifier.check in CI

dhalperi:
The buildifier.check target existed in //BUILD.bazel but no workflow
invoked it, so Bazel file formatting was never enforced in CI. Add a
Buildifier step to the existing format job, which already configures
Bazel caching. Also reorder loads in projects/allinone/BUILD.bazel,
which the check flags.

----

Prompt:
```
Is buildifer running in github CI? If not, add it
```

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -42,6 +42,8 @@ jobs:
             ${{runner.os}}-bazel-format-
       - name: Java Formatting
         run: bazel run //tools/format:format.check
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10012)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10013, opened 2026-06-26
Junos: allow '>' in config object names

dhalperi:
Junos permits '>' in config object names, e.g. MPLS LSP names like
"WASH->ATLA". The lexer's F_NameChar fragment did not include '>', so
such a name failed to lex as a single NAME token and the entire
enclosing block (its "to", "admin-group exclude", etc.) was dropped as
unrecognized. Add '>' to F_NameChar.

This cannot conflict with F_Wildcard, which is anchored by a leading
'<' (F_WildcardInner: '<' ~'>'* '>') and self-terminates at its first
'>'; the WILDCARD alternative is also declared before NAME in every
lexer mode. The only existing '>' token, GREATER_THAN, is referenced by
no parser rule.

----

Prompt:
```
I was told that there are issues parsing ~/oss/internet2 configurations.
Figure out what the issues are and recommend fixes
```

Then, after diagnosing the issues and confirming the wildcard analysis:
implement the first fix (allow '>' in F_NameChar) with a regression test
that also guards against wildcard interference.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -3675,6 +3675,8 @@ F_NameChar
   | ','
   | '.'
   | ':'
+  // Junos allows '>' in names, e.g. LSP name "WASH->ATLA".
+  | '>'
 ;
 
 // Any number of newlines, allowing whitespace in between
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10013)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10013)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10016, opened 2026-06-26
Junos: parse and apply routing-options static defaults

dhalperi:
routing-options static defaults { ... } was unrecognized. It sets
baseline attributes (preference, metric, tag, community, etc.) that
every static route in the same "static {}" block inherits unless it
overrides them per-field. Inheritance is scoped to the enclosing block:
defaults do not cascade between inet.0 and inet6.0, nor from the master
instance into routing-instances (verified on vJunos-router 25.4R1.12).

Parse it as ros_defaults (DEFAULTS rosr_common), reusing the per-route
attribute rule so the defaults block accepts exactly what a route can,
mirroring the existing aggregate/generated roa_defaults/rog_defaults.
Each RIB owns a v4 and v6 "defaults" StaticRoute holding the block's
attributes.

Model inheritance on the route itself: a static route holds a pointer to
its RIB's defaults route (set at construction), and its getters return
the route's own value if set, else the inherited value, else the Junos
default. There is no separate inheritance pass and no field copying, so
both conversion and the BGP static-route community setters see inherited
values for free, and a route's own unset/explicit distinction is
preserved. StaticRoute's metric/distance become nullable internally;
getMetric/getDistance still return the Junos defaults (0 and 5) when
neither the route nor its defaults set a value.

For batfish/batfish#10017.

----

Prompt:
```
Implement the next fix for parsing ~/oss/internet2 Junos configs: add
support for "routing-options static defaults". Actually implement it in
extraction and test it (inheritance into the static routes), not just
parse-and-ignore.
```

Follow-up: refined the model so routes resolve "defaults" inheritance
lazily through a parent pointer in their getters rather than an explicit
inherit-and-copy step, and made the testconfig mirror the
junos_static_route_defaults lab (v6 inet6.0 scope, routing-instance
scope with RED/BLUE, preference-only override).

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_routing_instances.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_routing_instances.g4
@@ -332,6 +332,7 @@ ro_static
    STATIC
    (
       apply
+      | ros_defaults
       | ros_rib_group
       | ros_route4
    )
@@ -342,9 +343,17 @@ ro6_static
    STATIC
    (
       apply
+      | ros_defaults
       | ros_rib_group
       | ros_route6
    )
+;
+
+// Default attributes inherited by all static routes in this RIB. The attribute set matches that of
+// an individual static route; an empty "defaults {}" block matches via rosr_common -> apply.
+ros_defaults
+:
+   DEFAULTS rosr_common
 ;
 
 ro_validation
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/ConfigurationBuilder.java
@@ -715,6 +715,7 @@ import org.batfish.grammar.flatjuniper.FlatJuniperParser.Rib_nameContext;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Riv_communityContext;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Riv_exportContext;
 import org.batfish.grammar.flatjuniper.FlatJuniperParser.Riv_importContext;
+import org.batfish.grammar.flatjuniper.FlatJuniperParser.Ro6_staticContext;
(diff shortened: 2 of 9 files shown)


dhalperi (commented):
> @dhalperi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10016)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on dhalperi).

___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/10016#-Ow46KDsBgq1K3ciE17h:-Ow46KDsBgq1K3ciE17i:b1w0ol9) (no related file):*
pending some lab validation to understand VS semantics

dhalperi (commented):
> @dhalperi reviewed 9 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10016)* status: all files reviewed, 1 unresolved discussion.

___
*[`projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRouteV4.java` line 10 at r2](https://reviewable.io/reviews/batfish/batfish/10016#-Ow5cCNhB8D9ID9c4-4e:-Ow5cCNhB8D9ID9c4-4f:b4i1n8w) ([raw file](https://github.com/batfish/batfish/blob/ed454b8a8fdc11c226e29b94eb08c9870c0b9158/projects/batfish/src/main/java/org/batfish/representation/juniper/StaticRouteV4.java#L10)):*
> ```Java
>   private Prefix _prefix;
> 
>   public StaticRouteV4(Prefix prefix) {
> ```

get rid of this one and just inline it in the 2 places it's needed.

dhalperi:
Removed the single-arg `StaticRouteV4(Prefix)`/`StaticRouteV6(Prefix6)` constructors; the 2-arg form is now used directly at both call sites (the RIB's static-route defaults objects pass `null`). Pushed as e53ddd0.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10016)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

Issue #10018, opened 2026-06-26
Junos extended-community literal with type octet >= 0x80 fails to parse (signed-byte arithmetic)


dhalperi:
## Problem

A Junos extended-community literal whose leading type octet has its high bit set — e.g. `members 65000:672277L:36867` (type octet `0xFD`) — fails literal parsing in `ExtendedCommunity.parse` and silently falls back to being treated as a regex community member. In a `community add` action this produces the convert warning:

```
FATAL: '<NAME>' community contains no non-wildcard members in an add action
```

and the community is never added to the model.

## Technical Details

In `ExtendedCommunity.parse` (`projects/common/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java`), a bare-integer first field is split into type/subtype octets via signed-byte casts:

```java
subTypeByte = (byte) (intVal & 0xFF);          // 0xE8 -> -24
typeByte    = (byte) ((intVal & 0xFF00) >> 8); // 0xFD -> -3
...
return of((int) typeByte << 8 | (int) subTypeByte, gaLong, laLong);
```

When `typeByte`/`subTypeByte` are >= 0x80 they become negative; sign extension makes the combined `type` argument negative, and `of()` rejects it via `checkArgument(type >= 0 && type <= 0xFFFF, ...)`. `tryParse` catches the exception and the caller falls back to a regex member. Type octets < 0x80 (e.g. `target:` = 0x02) are unaffected, which is why ordinary route targets parse fine.

## Observed In

`junos_ext_community_literal` (lab-validation). Confirmed on vJunos-router 25.4R1.12, which accepts the literal and renders it as `unknown type 0xffe8:0xa4215:0x9003` (4-byte global administrator `0xa4215` = 672277 forced by the `L` suffix, 2-byte local administrator `0x9003` = 36867).

## Symptoms

- `community add` / `community set` of such a literal is dropped (FATAL convert warning).
- Any policy matching the community misbehaves because the literal was parsed as a wildcard regex.

## Next Steps

Use unsigned/masked arithmetic when assembling the type/subtype, and decide whether experimental/unknown extended-community types (high-bit-set) should be modeled as opaque values rather than rejected.

PR #10020, opened 2026-06-26
Extended community: accept any type octet; fix signed-byte parsing

dhalperi:
A generic extended-community literal whose type octet has the high bit
set -- e.g. Junos "65000:672277L:36867" (type octet 0xFD) -- failed to
parse and was silently dropped. Two causes:

- parse() and of() assembled the type via signed-byte arithmetic, so a
  type or subtype octet >= 0x80 sign-extended to a negative int and was
  rejected. Mask each byte to its unsigned value.
- the constructor's VALID_TYPES allowlist rejected any type it did not
  specifically recognize. Extended communities observed on real devices
  may use experimental or vendor-specific types; model them as opaque
  values and let the type-specific predicates (isRouteTarget,
  isRouteOrigin, etc.) return false for unrecognized types. Only the
  single-byte type/subtype range is enforced now.

The Junos flatjuniper path previously fell back to a regex community
member for such literals, producing a FATAL convert warning on
"community add". It now parses as a literal extended community.

For batfish/batfish#10018.

----

Prompt:
```
now fix batfish (~/batfish/batfish), re-test the lab (rerunning the
server of course), and send a batfish PR
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/bgp/community/ExtendedCommunity.java
@@ -5,12 +5,10 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.Ints;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10020)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10022, opened 2026-06-27
Junos: allow interface rewrite-rules exp without protocol

dhalperi:
class-of-service interfaces X unit Y rewrite-rules exp <name> (no
protocol clause) was unrecognized: scosiiu_exp_rw required PROTOCOL. The
Juniper CLI reference lists protocol in the canonical syntax, but Junos
accepts a bare "exp rewrite-name" and defaults the protocol. Make the
protocol clause optional. Extend class-of-service-comprehensive.

----

Prompt:
```
Continue clearing internet2 Junos parse warnings in separate orthogonal
PRs. Next: interface class-of-service rewrite-rules "exp <name>" without
a protocol clause.
```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
@@ -444,7 +444,9 @@ scosiiu_dscp_ipv6_rw
 
 scosiiu_exp_rw
 :
-    EXP name = junos_name PROTOCOL proto = scos_protocol_type
+    // The protocol clause is optional in practice: Junos accepts a bare "exp rewrite-name", even
+    // though the CLI reference lists protocol in the canonical syntax.
+    EXP name = junos_name (PROTOCOL proto = scos_protocol_type)?
 ;
 
 scos_protocol_type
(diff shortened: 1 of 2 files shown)


projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
```diff
@@ -444,7 +444,9 @@ scosiiu_dscp_ipv6_rw
 
 scosiiu_exp_rw
 :
-    EXP name = junos_name PROTOCOL proto = scos_protocol_type
+    // The protocol clause is optional: Junos accepts a bare "exp rewrite-name" and defaults the
+    // protocol, though the CLI reference lists protocol in the canonical syntax.
```
dhalperi: what is the default?

projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
```diff
@@ -444,7 +444,9 @@ scosiiu_dscp_ipv6_rw
 
 scosiiu_exp_rw
 :
-    EXP name = junos_name PROTOCOL proto = scos_protocol_type
+    // The protocol clause is optional: Junos accepts a bare "exp rewrite-name" and defaults the
```
dhalperi: is this really a default protocol or that ther'es an entire hierarchical block here, and any suffix is optional?

dhalperi:
Good questions. Two separate things:

1. **Optional protocol vs. hierarchical block:** The interface-unit form is a single statement, `exp (rewrite-name | default) protocol protocol-types;` (CLI reference, [edit class-of-service interfaces … unit … rewrite-rules]) — not a nested block. The only optional part is the trailing `protocol <type>` clause. The Internet2 configs commit it bare (`exp ion_exp_rewrite-rule;`), which the device accepts, so I made just that clause optional.

2. **Default protocol:** I don't have a documented value for what protocol applies when the clause is omitted, so I've reworded the comment to drop the 'defaults the protocol' claim — it now just states the clause is optional in practice. We parse-and-ignore the rewrite-rule reference here regardless, so the omitted protocol has no modeling impact.

dhalperi (commented):
> @dhalperi reviewed 2 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10022)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10023, opened 2026-06-27
Junos: parse host-outbound-traffic ieee-802.1 block

dhalperi:
class-of-service host-outbound-traffic ieee-802.1 { default <value>;
rewrite-rules; override-firewall; } was unrecognized: only
forwarding-class was accepted. The Juniper CLI reference documents the
ieee-802.1 block (and its default/rewrite-rules/override-firewall
options) for marking Routing Engine outbound traffic.

Add the ieee-802.1 alternative under scos_host_outbound_traffic and an
OVERRIDE_FIREWALL token. ieee-802.1 normally pushes a name lexer mode for
its classifier/rewrite-rule reference; under host-outbound-traffic it is
a keyword block, so suppress the mode push in that context. Extend
class-of-service-comprehensive.

----

Prompt:
```
Continue clearing internet2 Junos parse warnings in separate orthogonal
PRs. Next: class-of-service host-outbound-traffic ieee-802.1. Ground the
syntax in the Juniper docs.
```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
@@ -285,12 +285,40 @@ scos_host_outbound_traffic
     HOST_OUTBOUND_TRAFFIC
     (
         scoshob_forwarding_class
+        | scoshob_ieee_802_1
     )
 ;
 
 scoshob_forwarding_class
 :
     FORWARDING_CLASS name = junos_name
+;
+
+// IEEE 802.1p marking of host (Routing Engine) outbound traffic.
+scoshob_ieee_802_1
+:
+    IEEE_802_1
+    (
+        scoshobi_default
+        | scoshobi_override_firewall
+        | scoshobi_rewrite_rules
+    )
+;
+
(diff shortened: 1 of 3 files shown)


projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniper_class_of_service.g4
```diff
@@ -293,6 +294,32 @@ scoshob_forwarding_class
...
+    (
+        scoshobi_default
+        | scoshobi_override_firewall
+        | scoshobi_rewrite_rules
+    )
+;
+
+scoshobi_default
+:
+    DEFAULT value = dec
```
dhalperi: dec? no range more specific than this?

dhalperi:
Updated: the `default` value now uses `ieee_802_1_code_point_or_alias` (3-bit code point or alias, 000–111) per the CLI reference, instead of `dec`. `DEFAULT` pushes the code-point lexer mode when it follows `ieee-802.1`.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10023)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10034, opened 2026-06-27
Junos: parse and track ieee-802.1 rewrite-rules import

dhalperi:
class-of-service rewrite-rules ieee-802.1 <name> accepts "import
<rule>" (commonly "import default"), like the dscp/exp variants. Accept
it and track the imported rule as a built-in reference.

----

Prompt:
```
let's do the c-o-s one and then do the next-hop one
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10034)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10035, opened 2026-06-27
Junos: lex next-hop in route-filter inline then

dhalperi:
A route-filter can carry an inline action, e.g.
"route-filter 0.0.0.0/0 exact next-hop self". The match type (exact,
longer, orlonger) precedes "next-hop", so the lexer pushed M_Interface
and lexed "self" as an interface name, failing the parse. Suppress the
M_Interface push after those match-type tokens, like the existing THEN
case.

----

Prompt:
```
let's do the c-o-s one and then do the next-hop one
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10035)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10040, opened 2026-06-28
BGP: drop non-transitive extended communities on eBGP export

### LINKED ISSUE #10019: Non-transitive extended communities are not stripped on eBGP advertisement
## Problem

Batfish retains non-transitive extended communities when a route is advertised over an eBGP session. Real devices drop them at the AS boundary (the Transitive bit in the extended-community type octet, [RFC 4360 §2](https://www.rfc-editor.org/rfc/rfc4360#section-2): high-order bit `0x40`, value 1 = non-transitive across ASes). This leaves extra communities on routes received across eBGP in the Batfish model versus the device.

## Technical Details

An extended community's high-order type octet carries the Transitive bit (`0x40`; 1 = non-transitive). `ExtendedCommunity.isTransitive()` already exposes this. However, the eBGP export path in `BgpProtocolHelper` only clears non-transitive *attributes* such as weight, local-preference, and tag -- it does not filter non-transitive extended communities out of the community set. As a result they leak across AS boundaries in the data plane model.

## Observed In

`junos_ext_community_literal` (lab-validation). The sender (AS 65001) attaches a generic extended-community literal `65000:672277L:36867` whose type octet `0xFD` has the Transitive bit set (non-transitive). On real vJunos-router 25.4R1.12 the device drops this community when advertising to the receiver (AS 65002). Batfish keeps it: the receiver's `10.10.1.0/24` shows `65000:672277L:36867` in the BGP RIB where the device shows no community. Transitive route targets on sibling routes (`514:672277L:36867`, `2:65000:36867`) correctly cross unchanged on both.

## Symptoms

- BGP route validation mismatches on community sets for routes that carry non-transitive extended communities across eBGP.

## Next Steps

On eBGP export (outside a confederation), drop extended communities whose type marks them non-transitive, consistent with how weight / local-preference / tag are already cleared in `BgpProtocolHelper`.

dhalperi:
An extended community's type octet carries the Transitive bit (RFC 4360
section 2: 0x40, value 1 = non-transitive across ASes). Real devices
drop such communities at the AS boundary, but the eBGP export path in
BgpProtocolHelper only cleared non-transitive attributes (weight,
local-preference, tag) -- it left non-transitive extended communities in
the community set, so they leaked across eBGP in the data plane model.

Filter them out on true eBGP export (outside a confederation), mirroring
the existing local-pref / MED clearing. Standard and large communities,
and transitive extended communities (including route targets), are
unaffected.

Fixes batfish/batfish#10019.

----

Prompt:
```
address https://github.com/batfish/batfish/issues/10019

design an elegant and efficient solution and, when done, send a PR
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/protocols/BgpProtocolHelper.java
@@ -41,6 +41,8 @@ import org.batfish.datamodel.bgp.AllowRemoteAsOutMode;
 import org.batfish.datamodel.bgp.BgpAggregate;
 import org.batfish.datamodel.bgp.BgpTopologyUtils.ConfedSessionType;
 import org.batfish.datamodel.bgp.EvpnAddressFamily;
+import org.batfish.datamodel.bgp.community.Community;
+import org.batfish.datamodel.bgp.community.ExtendedCommunity;
 import org.batfish.datamodel.bgp.community.StandardCommunity;
 import org.batfish.datamodel.route.nh.NextHop;
 import org.batfish.datamodel.route.nh.NextHopDiscard;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10040)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10046, opened 2026-06-29
BDDRoute: include communities and tracks in hashCode

dhalperi:
hashCode() omitted the community atomic predicates and tracks. When a
BDDRoute is used as a symbolic route function, the scalar attribute
integers (prefix, local-pref, MED, ...) are frequently left as their free
identity functions, so communities are the axis distinguishing most
distinct route functions. Omitting them collapsed distinct routes into a
handful of hash buckets, treeifying HashMaps keyed on BDDRoute and turning
every insert into a chain of full structural equals() calls.

equals() already compares both fields, so this only improves bucket
distribution; it does not change which routes are considered equal.

---
**Stack**:
- #10048
- #10047
- #10046 ⬅
---

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java
@@ -785,7 +785,14 @@ public final class BDDRoute implements IDeepCopy<BDDRoute> {
 
   @Override
   public int hashCode() {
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10046)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10053, opened 2026-07-06
tools: disable implicit __init__.py creation for py_binary targets

dhalperi:
rules_python warns that legacy_create_init defaults to true and will
be flipped to false in a future release. Neither generate_pan_apps
nor analyze_profile is part of a Python package tree, so no
__init__.py files are needed.

----

Prompt:
```
Run bazel build //... and fix all the warnings about implicit
__init__.py creation in this repo (not the ones where python rules
are used from external places)
```

### DIFF
--- a/tools/BUILD.bazel
+++ b/tools/BUILD.bazel
@@ -47,6 +47,7 @@ java_binary(
 py_binary(
     name = "generate_pan_apps",
     srcs = ["generate_pan_apps.py"],
+    legacy_create_init = 0,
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10053)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10054, opened 2026-07-06
BDDInteger: add firstBitsEqual for symbolic subnet comparison

dhalperi:
Add firstBitsEqual(BDDInteger other, int numBits): the set of
assignments under which two symbolic integers of equal width agree on
their first (most-significant) numBits bits, ignoring the rest. For an
IP-valued integer this is "the two addresses are in the same /numBits
subnet". numBits == width reduces to eq, which now delegates to it;
numBits == 0 is unconstrained.

Tested exhaustively over all (a, b) pairs and bit-lengths, and against
eq at the boundary widths.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10054)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10057, opened 2026-07-07
BDDRoute: fix copy constructor to not alias COMMUNITIES/TRACKS

dhalperi:
The copy constructor gave every scalar attribute an independently
owned copy (MutableBDDInteger re-id()'s each bit), but cloned the
COMMUNITIES and TRACKS arrays with a plain array clone: each cell of
the copy was the same BDD object as the corresponding cell of the
route it was copied from. Copy each cell via id() instead, so every
field of a copy is independently owned.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10057)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10059, opened 2026-07-08
minesweeper: BDDRoute support for relational image computation

dhalperi:
Add infrastructure for treating a BDDRoute as one side of a relation
rather than only a function of implicit input variables:

- BDDRouteFactory builds a factory's canonical BDDRoute together
  with any number of additional fresh, mutually-interleaved blocks,
  all in a single constructor call -- interleaving requires that the
  factory hold exactly one previously-built BDDRoute and nothing
  else, and building the canonical route and its extra blocks as
  separate steps would leave a window in which that precondition
  could be silently violated (e.g. by other code allocating factory
  variables in between, or by building the extra blocks twice).
  Blocks share the destination prefix/prefix-length variables, since
  no route policy can ever change a route's destination; comparing
  two blocks bit-by-bit stays linear in route width only if each bit
  sits next to its counterpart in the variable order, so the extra
  blocks are interleaved with the canonical route rather than
  appended contiguously.
- BDDRouteFactory can also allocate a block of caller-defined shared
  bits (e.g. tracking which snapshot a route's underlying config
  came from), at the very front of the variable order, ahead of even
  prefix/prefix-length. Unlike the extra blocks, these are shared the
  same way prefix/prefix-length is: one copy that every block refers
  to identically, since this state is a property of the route rather
  than of which symbolic variable block currently represents it.
- A new TransferBDD(BDDFactory, BDDRoute, ConfigAtomicPredicates)
  constructor lets a caller that also needs BDDRouteFactory's extra
  blocks hand in the same canonical route TransferBDD operates over,
  instead of TransferBDD building a second, un-interleaved one.
- support()/mutableSupport() expose a route's BDD variable set, the
  latter excluding the shared prefix/prefix-length variables (needed
  when quantifying out one block's variables must not erase another
  block's copy of the same shared variables).
- equalsRelation(other)/equalsRelationOn(other, fields) build the BDD
  relation "this route's attributes equal other's," in whole or
  restricted to a subset of fields; the latter leaves every other
  field completely unmentioned rather than adding a trivial equality
  term, so it composes cleanly with a caller that only cares about
  the fields actually written.
- Attribute (a per-attribute strategy enum) and the Field sealed
  interface (Scalar/Community/Track) give per-attribute, and for
  COMMUNITIES/TRACKS per-index, granularity for naming which piece of
  a route a computation touches, via fieldSupport(On)/
  augmentPairingOn/touchedFields.

### DIFF
--- a/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
+++ b/projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
@@ -1,0 +1,174 @@
+package org.batfish.minesweeper.bdd;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.assertThrows;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import java.util.Arrays;
+import net.sf.javabdd.BDD;
+import net.sf.javabdd.BDDFactory;
+import net.sf.javabdd.JFactory;
+import org.batfish.minesweeper.CommunityVar;
+import org.batfish.minesweeper.ConfigAtomicPredicates;
+import org.junit.Test;
+
+/** Tests for {@link BDDRouteFactory}. */
+public class BDDRouteFactoryTest {
+  private static final ConfigAtomicPredicates CONFIG_APS =
+      new ConfigAtomicPredicates(
+          ImmutableList.of(),
+          ImmutableSet.of(CommunityVar.from("30:30"), CommunityVar.from("40:40")),
+          ImmutableSet.of());
+
+  @Test
+  public void testZeroRoutesRejected() {
+    assertThrows(IllegalArgumentException.class, () -> new BDDRouteFactory(CONFIG_APS, 0));
+  }
+
+  @Test
+  public void testSingleRouteIsCanonical() {
+    // With a single route requested, identityRoute(0) must be the same as building a BDDRoute
+    // directly against a fresh factory for the same aps.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 1);
+    BDDRoute direct = new BDDRoute(routeFactory.getFactory(), CONFIG_APS);
+    assertThat(routeFactory.identityRoute(0), equalTo(direct));
+  }
+
+  @Test
+  public void testRoutesAreMutuallyInterleaved() {
+    // The routes must be pairwise interleaved, not contiguous: each of the first route's
+    // non-prefix variables' level must be immediately followed, in the factory's variable order,
+    // by its counterpart in the second and then in the third.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 3);
+    BDDRoute first = routeFactory.identityRoute(0);
+    BDDRoute second = routeFactory.identityRoute(1);
+    BDDRoute third = routeFactory.identityRoute(2);
+
+    int[] firstLevels = sortedLevels(first.mutableSupport());
+    int[] secondLevels = sortedLevels(second.mutableSupport());
+    int[] thirdLevels = sortedLevels(third.mutableSupport());
+    assertThat(secondLevels.length, equalTo(firstLevels.length));
+    assertThat(thirdLevels.length, equalTo(firstLevels.length));
+    for (int i = 0; i < firstLevels.length; i++) {
+      assertThat(secondLevels[i], equalTo(firstLevels[i] + 1));
+      assertThat(thirdLevels[i], equalTo(firstLevels[i] + 2));
+    }
+
+    // A blown-up (non-interleaved) relation over hundreds of bits would have astronomically more
+    // nodes than this; a generous but finite bound catches a regression without being brittle.
+    BDD firstVsSecond = first.equalsRelation(second);
+    BDD firstVsThird = first.equalsRelation(third);
+    BDD secondVsThird = second.equalsRelation(third);
+    assertThat(firstVsSecond.nodeCount(), lessThan(10_000));
+    assertThat(firstVsThird.nodeCount(), lessThan(10_000));
+    assertThat(secondVsThird.nodeCount(), lessThan(10_000));
+
+    assertThat(routeFactory.identityRoutes(), equalTo(ImmutableList.of(first, second, third)));
+  }
+
+  /**
+   * The factory-order levels of every BDD variable in {@code support}'s support, sorted ascending.
+   */
+  private static int[] sortedLevels(BDD support) {
+    int[] vars = support.scanSet();
+    BDDFactory factory = support.getFactory();
+    int[] levels = new int[vars.length];
+    for (int i = 0; i < vars.length; i++) {
+      levels[i] = factory.var2Level(vars[i]);
+    }
+    Arrays.sort(levels);
+    return levels;
+  }
+
+  @Test
+  public void testExplicitFactoryWithNoSharedBits() {
+    // The (BDDFactory, ConfigAtomicPredicates, int) overload -- a caller-supplied factory, no
+    // shared bits -- must build the same shape as passing numSharedBits=0 explicitly.
+    BDDFactory factory = JFactory.init(100000, 10000);
+    BDDRouteFactory routeFactory = new BDDRouteFactory(factory, CONFIG_APS, 3);
+    assertThat(routeFactory.getFactory(), equalTo(factory));
+    assertThat(routeFactory.numSharedBits(), equalTo(0));
+    assertThat(routeFactory.identityRoutes().size(), equalTo(3));
+  }
+
+  @Test
+  public void testTransferBDDCanReuseIdentityRoute() {
+    // A caller that needs both a TransferBDD and other routes interleaved with its canonical route
+    // must build the canonical route once, via BDDRouteFactory, and hand it to the
+    // TransferBDD(factory, originalRoute, aps) overload -- which stores it as-is -- rather than to
+    // TransferBDD(factory, aps), which builds its OWN second canonical route that
+    // BDDRouteFactory's other routes are not interleaved with (a hazard distinct from object
+    // equality: for the same aps, that second route happens to allocate the identical variable
+    // range and so is .equals() to routeFactory's, since BDDRoute equality is purely structural --
+    // it is the interleaving, not the identity, that TransferBDD(factory, aps) would get wrong).
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 2);
+    TransferBDD tbdd =
+        new TransferBDD(routeFactory.getFactory(), routeFactory.identityRoute(0), CONFIG_APS);
+    assertThat(tbdd.getOriginalRoute(), equalTo(routeFactory.identityRoute(0)));
+  }
+
+  @Test
+  public void testMutableRouteIsIndependentCopy() {
+    // mutableRoute(i) must be a genuinely independent deep copy: mutating it must not affect
+    // identityRoute(i) or any other mutableRoute(i) call's result.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 1);
+    BDDRoute identity = routeFactory.identityRoute(0);
+    BDDRoute mutableA = routeFactory.mutableRoute(0);
+    BDDRoute mutableB = routeFactory.mutableRoute(0);
+    assertThat(mutableA, equalTo(identity));
+
+    mutableA.getLocalPref().setValue(100);
+    assertThat(mutableA, not(equalTo(identity)));
+    assertThat(mutableB, equalTo(identity));
+    assertThat(routeFactory.identityRoute(0), equalTo(identity));
+  }
+
+  @Test
+  public void testSharedBitsAreReservedAndDisjointFromEveryRoute() {
+    // numSharedBits() only reserves indices [0, numSharedBits()) -- it's the caller's job to build
+    // something over them, e.g. a MutableBDDInteger. Those reserved variables sit at the very front
+    // of the var order; the shared prefix/prefix-length variables (which support() includes, and
+    // every route shares the SAME instance of, unlike its own non-prefix variables) come right
+    // after, so every route's support -- not just its own non-prefix portion -- has its lowest
+    // (top) level exactly at numSharedBits(), never lower: the reserved range is never touched.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 2, 4);
+    assertThat(routeFactory.numSharedBits(), equalTo(4));
+
+    assertThat(routeFactory.identityRoute(0).support().level(), equalTo(4));
+    assertThat(routeFactory.identityRoute(1).support().level(), equalTo(4));
+  }
+
+  @Test
+  public void testSharedBitsDoNotPerturbInterleaving() {
+    // Requesting shared bits must not change the interleaving/width math for the routes: each of
+    // the first route's non-prefix variables' level must still be immediately followed by its
+    // counterpart in the second, exactly as without shared bits.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 2, 4);
+    BDDRoute first = routeFactory.identityRoute(0);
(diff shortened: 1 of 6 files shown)


projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+    BDDRoute first = routeFactory.identityBlock(0);
+    BDDRoute second = routeFactory.identityBlock(1);
+
+    BDD originVsFirst = origin.equalsRelation(first);
+    BDD originVsSecond = origin.equalsRelation(second);
+    BDD firstVsSecond = first.equalsRelation(second);
+
+    // A blown-up (non-interleaved) relation over hundreds of bits would have astronomically more
+    // nodes than this; a generous but finite bound catches a regression without being brittle.
+    assertThat(originVsFirst.nodeCount(), lessThan(10_000));
```
dhalperi: can we actually test directly the interleaving?

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Test
+  public void testTransferBDDCanReuseOriginalRoute() {
+    // A caller that needs both a TransferBDD and fresh blocks interleaved with its canonical route
+    // must build the canonical route once, via BDDRouteFactory, and hand it to TransferBDD rather
+    // than letting TransferBDD build its own (which would allocate a second, un-interleaved copy).
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 1);
+    TransferBDD tbdd =
+        new TransferBDD(routeFactory.getFactory(), routeFactory.originalRoute(), CONFIG_APS);
+    assertThat(tbdd.getOriginalRoute(), equalTo(routeFactory.originalRoute()));
```
dhalperi: I don't know what this is testing.

dhalperi: Do we want a second assertion that transferBDD only passing the BDDFactory would NOT be equal?

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+    // numSharedBits() only reserves indices [0, numSharedBits()) -- it's the caller's job to build
+    // something over them, e.g. a MutableBDDInteger. Those reserved variables sit at the very front
+    // of the var order and are never touched by the interleaving reorder, so every block's own
+    // variables have a strictly higher level; a block's support cube's top level (the lowest-level
+    // variable it mentions) being >= numSharedBits() confirms the block never touches the reserved
+    // range.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 2, 4);
+    assertThat(routeFactory.numSharedBits(), equalTo(4));
+
+    assertThat(routeFactory.originalRoute().support().level(), greaterThanOrEqualTo(4));
```
dhalperi: can't we test the exact values? should be 4,5,6 I think?

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+    // Requesting shared bits must not change the interleaving/width math for the route blocks:
+    // relating two blocks should still stay linear in the route width.
+    BDDRouteFactory routeFactory = new BDDRouteFactory(CONFIG_APS, 2, 4);
+    BDDRoute origin = routeFactory.originalRoute();
+    BDDRoute first = routeFactory.identityBlock(0);
+    BDDRoute second = routeFactory.identityBlock(1);
+
+    BDD originVsFirst = origin.equalsRelation(first);
+    BDD firstVsSecond = first.equalsRelation(second);
+    assertThat(originVsFirst.nodeCount(), lessThan(10_000));
```
dhalperi: similar comment

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java
```diff
@@ -420,4 +425,175 @@ public void testCopyConstructorFieldsAreIndependentlyOwned() {
...
+                3,
+                4,
+                5,
+                6,
+                7,
+                2,
+                ImmutableList.of(),
+                /* prefixStartIdx= */ 0,
+                /* startIdx= */ 37));
+  }
```
dhalperi, line 465: also assert that 0, 38 does not throw.

dhalperi: aka just call it, and comment , I think

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java
```diff
@@ -420,4 +425,175 @@ public void testCopyConstructorFieldsAreIndependentlyOwned() {
...
+
+  @Test
+  public void testEqualsRelationDoesNotLeak() {
+    BDDFactory factory = JFactory.init(10000, 1000);
+    BDDRoute route = new BDDRoute(factory, 3, 4, 5, 6, 7, 2, ImmutableList.of());
+    BDDRoute other = new BDDRoute(route);
+    long baseline = factory.numOutstandingBDDs();
+
+    BDD relation = route.equalsRelation(other);
+    relation.free();
```
dhalperi, line 489: can we mutate other so that they are different?

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
dhalperi: (Claude) Good idea — replaced the node-count heuristic with an exact check: origin's non-prefix variable levels, sorted, must have first's levels equal to origin's + 1 and second's equal to origin's + 2 (immediately following, per-bit). This directly tests the interleaving invariant rather than approximating it via a size bound.

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
dhalperi: (Claude) Rewrote and clarified: this test exists to confirm that a caller who already has a BDDRouteFactory and wants a TransferBDD must pass routeFactory.originalRoute() into the TransferBDD(factory, originalRoute, aps) overload rather than calling TransferBDD(factory, aps), which would build its own second canonical route un-interleaved with the factory's other blocks. I tried adding a negative assertion per the next comment but it actually revealed something more interesting -- see that reply.

dhalperi: (Claude) I tried this and it actually failed -- but not because the code is wrong. TransferBDD(factory, aps) allocates a second canonical route starting at the same index 0 as routeFactory's (since routeFactory here has numSharedBits=0), and BDDRoute.equals() is purely structural (delegates down to BDD node-index equality), so the two happen to be .equals() even though they are two distinct objects with no real relationship -- the second route's variables aren't interleaved with the extra blocks routeFactory built, which is the actual hazard, not object identity/equality. So a not(equalTo(...)) assertion here would be asserting something false. I documented this reasoning in a comment on the (unchanged) positive assertion instead of adding a misleading negative one.

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
dhalperi: (Claude) Tested the exact value -- it's not 4/5/6 though. support() includes the shared prefix/prefix-length variables (which every block shares as the literal same instance), and the prefix block sits immediately after numSharedBits() in the var order, untouched by the later interleaving reorder of the non-prefix bits. So origin's, block(0)'s, and block(1)'s support() all have the SAME top level: exactly numSharedBits() (4), not 4/5/6. Tightened the assertion to equalTo(4) for all three and explained why in the comment.

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteFactoryTest.java
dhalperi: (Claude) Same fix as testBlocksAreMutuallyInterleaved above: replaced the node-count heuristic with an exact per-bit level check (first/second's levels equal origin's + 1/+2, confirming true interleaving even with shared bits present).

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java
dhalperi: (Claude) Added testStartIdxAtEndOfPrefixBlockAccepted right after this test: calls the constructor with startIdx=38 (the boundary immediately at the end of the prefix/prefix-length block -- one past the rejected 37) and expects it to just succeed, with a comment explaining why 38 is the boundary value.

dhalperi: (Claude) Done -- the new test just calls new BDDRoute(...) with startIdx=38 (no assertThrows wrapper needed, since a plain call that doesn't throw already fails the test if it does) plus a one-line comment on why 38 is the boundary.

projects/minesweeper/src/test/java/org/batfish/minesweeper/bdd/BDDRouteTest.java
dhalperi: (Claude) Done -- mutated other's localPref and tunnelEncapsulationAttribute (via setValue) before computing the relation, with a comment noting this also exercises the TUNNEL_ENCAPSULATION_ATTRIBUTE .allDifferences(...).notEq() path specifically, which is the one that actually had the leak this test guards against.

millstein (commented):
> @millstein reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10059)* status: 3 of 5 files reviewed, 9 unresolved discussions (waiting on dhalperi).

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 898 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6TF-m0tUPkHKr5GDh:-Ox6TF-m0tUPkHKr5GDi:b-yzy7ry) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L898)):*
> ```Java
>       @Override
>       public BDD eq(BDDRoute route, BDDRoute other) {
>         return route._originType.getInteger().eq(other._originType.getInteger());
> ```

Could add an `eq` method to BDDDomain since you use the above pattern several times here.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java` line 117 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6cbxU1ms-hYDybc0A:-Ox6cbxU1ms-hYDybc0B:b2ybrh8) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java#L117)):*
> ```Java
>    * shared-by-reference block, not a copy -- mutate a {@link BDDRoute#deepCopy} of it instead.
>    */
>   public BDDRoute identityBlock(int i) {
> ```

The term "block" here and below was confusing to me.  I think you mean that to refer to some full set of BDDRoute variables.  So maybe routeVars, symbolicRoute, something like that instead?
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java` line 126 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6dUeC0YBUoU1wSBUY:-Ox6dUeC0YBUoU1wSBUZ:bb1j1rc) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java#L126)):*
> ```Java
>    */
>   public ImmutableList<BDDRoute> identityBlocks() {
>     return _extraBlocks;
> ```

So these BDDRoutes (and all of their attribute MutableBDDIntegers etc.) should never be mutated, is that the intent?

dhalperi:
(Claude) Addressed all three:

1. Added `BDDDomain#eq(BDDDomain<T>)` and switched `Attribute.PROTOCOL_HISTORY`/`ORIGIN_TYPE`/`OSPF_METRIC`/`NEXT_HOP_INTERFACES`/`PEER_ADDRESS`/`SOURCE_VRFS`'s `eq()` overrides to use it instead of `.getInteger().eq(other.getInteger())`.

2. Renamed `identityBlock(i)`/`identityBlocks()` to `identityRoute(i)`/`identityRoutes()`. While doing this, realized `originalRoute()` had no actual mechanical distinction from the other routes -- it was just the one built first, at a fixed index -- so folded it in too: the constructor now takes `numRoutes` (total, must be >= 1, enforced via `checkArgument`) instead of `numExtraBlocks`, and every route is uniformly `identityRoute(i)` for `i` in `[0, numRoutes)`. Index 0 is still the one to hand to `TransferBDD`'s `(factory, originalRoute, aps)` overload, now by documented convention rather than a distinctly-named accessor.

3. Yes, that's the intent -- added `mutableRoute(i)`, which returns `identityRoute(i).deepCopy()`, so callers who need to mutate get an explicit, independent copy instead of being tempted to mutate the shared identity route directly.

dhalperi (commented):
> @dhalperi reviewed 5 files and all commit messages, made 3 comments, resolved 9 discussions, and dismissed @millstein from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10059)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java` line 898 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6TF-m0tUPkHKr5GDh:-Ox6xNWADd5g_9mx3-XQ:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRoute.java#L898)):*

Could add an `eq` method to BDDDomain since you use the above pattern several times here.
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java` line 117 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6cbxU1ms-hYDybc0A:-Ox6xcwj3yZ9zoUZwhRJ:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java#L117)):*

The term "block" here and below was confusing to me.  I think you mean that to refer to some full set of BDDRoute variables.  So maybe routeVars, symbolicRoute, something like that instead?
</blockquote></details>

Done.
___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java` line 126 at r2](https://reviewable.io/reviews/batfish/batfish/10059#-Ox6dUeC0YBUoU1wSBUY:-Ox6xdjsCCaIHQ5E_A6M:b-896fix) ([raw file](https://github.com/batfish/batfish/blob/a3fd945747a362c967946daeb9b04fc7879ef5ee/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/BDDRouteFactory.java#L126)):*

So these BDDRoutes (and all of their attribute MutableBDDIntegers etc.) should never be mutated, is that the intent?
</blockquote></details>

Done.

PR #10063, opened 2026-07-09
Support SpecialCasesRendering in minesweeper community regex analysis

zphw:
FRR creates CommunityMatchRegex with a SpecialCasesRendering (wrapping ColonSeparatedRendering as fallback). The var collector previously rejected anything that wasn't exactly ColonSeparatedRendering.

Fixes:
1. Accept any rendering whose effective fallback is colon-separated.
2. For special-cased communities (i.e., Community -> String), resolve them to exact CommunityVars when the regex matches their rendered name. This is necessary because strings like "no-export" falls outside the `<num>:<num>` and would otherwise be invisible to the analysis.
3. Exclude special-cased communities whose colon form matches the regex but whose rendered name does not.  For example, regex [0-9]+:[0-9]+ would incorrectly match NO_EXPORT because 65535:65281 matches, even though FRR renders it as "no-export" which won't match.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java
@@ -125,9 +125,18 @@ public class CommunityMatchExprToBDD implements CommunityMatchExprVisitor<BDD, A
 
   @Override
   public BDD visitCommunityMatchRegex(CommunityMatchRegex communityMatchRegex, Arg arg) {
-    return CommunitySetMatchExprToBDD.communityVarsToBDD(
-        communityMatchRegex.accept(new CommunityMatchExprVarCollector(), arg.getConfiguration()),
-        arg);
+    Set<CommunityVar> allVars =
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> @dhalperi reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10063)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on zphw).

___
*[`projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java` line 136 at r3](https://reviewable.io/reviews/batfish/batfish/10063#-OxD3kR58Ba0HLA_La_N:-OxD3kR58Ba0HLA_La_O:bdlrs4b) ([raw file](https://github.com/batfish/batfish/blob/b43e1e6964dee5b030f8d93d6ad1a55797dde99a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/CommunityMatchExprToBDD.java#L136)):*
> ```Java
>         CommunityMatchExprVarCollector.regexExcludedSpecialCases(
>             communityMatchRegex.getRegex(), communityMatchRegex.getCommunityRendering());
>     if (!excluded.isEmpty()) {
> ```

Ongoing discussion about this... we're writing a dog :)

PR #10064, opened 2026-07-09
juniper: hand-write AsPathRegex and GroupWildcard, drop parboiled

dhalperi:
Replace the two parboiled-based Juniper regex converters with hand-written
recursive-descent parsers in the parent juniper package, and delete the
org.batfish.representation.juniper.parboiled subpackage. The produced Java
regex strings are unchanged except for one bug fix (below); behavior is
pinned by the existing AsPathRegex and GroupWildcard tests.

These grammars are small (~13-30 rules), self-contained parse-and-transform
converters with no autocomplete needs, so recursive descent is a direct
replacement that removes a parboiled dependency and makes the code
coverage-instrumentable and debuggable.

Fix a latent bug in the process: the former parboiled Op_Exact rule matched a
bare "{n}" repetition operator but pushed nothing, silently dropping the
count (so "1234{2}" matched a single 1234). The new parser applies the count
like the other repetition operators, and a new AsPathRegexTest case pins the
exact-repetition behavior.

Part of removing parboiled from Batfish entirely.

----

Prompt:
```
keep going on the rewrite and remove parboiled from Batfish entirely
without any loss of functionality
```


dhalperi (commented):
> @dhalperi partially reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10064)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10065, opened 2026-07-09
check_point: hand-write service-other match parser, drop parboiled

dhalperi:
Replace the parboiled-based ServiceOtherMatchExpr with a hand-written
recursive-descent parser implementing the same PEG. Ordered choice is
realized by saving and restoring the input position on a failed
alternative. The AST it produces is unchanged; behavior is pinned by the
existing ServiceOtherMatchExprTest (including the per-rule test entry point,
which now takes a Function<ServiceOtherMatchExpr, AstNode> in place of a
parboiled Rule).

The surrounding AST node classes in this package were already hand-written
and did not depend on parboiled; only the parser did. With the parser
converted, the package no longer references parboiled, so its coverage
exclusion is removed from .bazelrc. That exclusion also named the (now
deleted) juniper parboiled package, so the whole instrumentation_filter line
is dropped: both parboiled parsers it guarded are gone, and coverage now
instruments the replacements without crashing.

Part of removing parboiled from Batfish entirely.

----

Prompt:
```
keep going on the rewrite and remove parboiled from Batfish entirely
without any loss of functionality
```


dhalperi (commented):
> @dhalperi partially reviewed 53 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10065)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

Issue #10062, opened 2026-07-09
IOS-XR: 'rd' and 'extcommunity-set rt' with 4-byte (asplain) ASN fail to parse; 2-byte 'rd' under router bgp vrf reports unsupported


btarno:
Cisco IOS-XR configurations using 4-byte ASNs (RFC 6793, asplain notation) in route distinguishers and RT extcommunity sets produce parse warnings, which degrades VRF modeling and creates false positives in `undefinedReferences` (an `extcommunity-set rt` whose body fails to parse is later reported as undefined at its `route-policy` reference site).

## Repro

Single-file snapshot, config below (synthetic; 4200000001 is a private-range 4-byte ASN):

```
hostname xr-rd-repro
vrf CUSTOMER-A
 address-family ipv4 unicast
  import route-target
   4200000001:100
  !
  export route-target
   4200000001:100
  !
 !
!
extcommunity-set rt ECM_TWO_BYTE_ASN_LARGE_VALUE
  65000:11311111
end-set
!
extcommunity-set rt ECM_FOUR_BYTE_ASN
  4200000001:100
end-set
!
router bgp 4200000001
 vrf CUSTOMER-A
  rd 4200000001:100
 !
 vrf CUSTOMER-B
  rd 65000:100
 !
!
```

`parseWarning` output (aggregateDuplicates=False):

```
line 17: '4200000001:100'    | This syntax is unrecognized              | ctx: [extcommunity_set_rt s_extcommunity_set stanza cisco_xr_configuration]
line 18: 'end-set'           | This syntax is unrecognized              | ctx: [cisco_xr_configuration]
line 22: 'rd 4200000001:100' | This syntax is unrecognized              | ctx: [route_distinguisher bgp_vrf_rd vrf_block_rb_stanza router_bgp_stanza ...]
line 25: 'rd 65000:100'      | This feature is not currently supported  | ctx: [bgp_vrf_rd vrf_block_rb_stanza router_bgp_stanza ...]
```

## Observed

1. `rd <4-byte-asn>:<nn>` under `router bgp / vrf` fails the grammar entirely (unrecognized).
2. `extcommunity-set rt` entries with a 4-byte ASN admin field fail the grammar, and the failure cascades: the following `end-set` is also flagged, the set is not defined, and any `route-policy` referencing it shows up in `undefinedReferences` even though the operator's config is valid.
[...]

PR #10066, opened 2026-07-09
xr: fix lexing of 32-bit ASNs after 'rd', extract rd under router bgp vrf

btarno:
Partially addresses #10062 (the `rd` half; the `extcommunity-set rt` local-administrator half is a separate, larger change and I'll follow up on the issue with corrected analysis).

## Problem

The `M_Rd` lexer mode declares:

```
M_Rd_UINT32: F_Uint16 -> type(UINT16);
```

so an asplain 32-bit ASN administrator after `rd` can never tokenize. `rd 4200000001:100` shatters into a run of UINT16 fragments (`'42000' '0' '0' '0' '0' '1'`) and the whole line is reported as unrecognized, even though the `route_distinguisher` parser rule already accepts a 32-bit ASN via `bgp_asn`. Every VRF on a device whose backbone ASN is above 65535 hits this.

Separately, `rd` under `router bgp / vrf` was `todo()`, so even the 2-byte form reported "This feature is not currently supported" and nothing was extracted.

## Fix

* `M_Rd_UINT32` now uses `F_Uint32` and emits `UINT32` (one-line copy-paste fix).
* `exitBgp_vrf_rd` extracts the `RouteDistinguisher` onto the vendor-model `Vrf`, mirroring what `CiscoControlPlaneExtractor` already does for IOS. Handles type 0 (2-byte ASN admin, 4-byte value), type 1 (IP admin), and type 2 (4-byte asplain or asdot ASN admin, 2-byte value); out-of-range combinations produce a conversion warning instead of a crash.

## Testing

* New `GitHub10062Test` + `gh10062` testconfig covering all four RD administrator forms (2-byte ASN, 4-byte asplain, asdot, IP address), asserting the extracted `RouteDistinguisher` values.
* `//projects/batfish/src/test/java/org/batfish/grammar/cisco_xr:tests` passes; `bazel test //projects/...` run locally with only two pre-existing environment-dependent failures in `projects/common` (container runs as root, so the mkdir-permission test cannot fail as intended); `//tools/format:format.check` and `tools/run_checkstyle.sh` clean.
* Verified against a real ~1900-line IOS-XR 25.2.2 production config: the `rd` parse warnings disappear and the VRF RDs extract correctly.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXrLexer.g4
@@ -8046,7 +8046,7 @@ M_Rd_COLON: ':' -> type(COLON);
 M_Rd_IP_ADDRESS: F_IpAddress -> type(IP_ADDRESS);
 M_Rd_PERIOD: '.' -> type(PERIOD);
 M_Rd_UINT16: F_Uint16 -> type(UINT16);
-M_Rd_UINT32: F_Uint16 -> type(UINT16);
+M_Rd_UINT32: F_Uint32 -> type(UINT32);
 M_Rd_NEWLINE: F_Newline -> type(NEWLINE), popMode;
 M_Rd_WS: F_Whitespace+ -> channel(HIDDEN);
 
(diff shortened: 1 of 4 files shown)


btarno:
Ready for review. Let me know if you see any issues.

btarno:
addressed the codecov concerns.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10066)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on btarno).

dhalperi:
A great PR! Thanks for the contribution!

PR #10069, opened 2026-07-10
bazel: target Java 21 source level

dhalperi:
Batfish's minimum supported runtime is now JDK 21, so raise the source
language level from 17 to 21 to allow use of Java 18-21 language
features. Bump tool_java_runtime_version to match; the build already
runs on remotejdk_25.

Raising the tool runtime to JDK 21 makes PMD run on a JDK where
ExecutorService implements AutoCloseable (Java 19+), so its CloseResource
rule now flags ExecutorService locals. Both flagged spots were real
leaks on some path:

- SnapshotMetadataMgrTest called awaitTermination without ever calling
  shutdown, so the executor and its thread were never reclaimed. Wrap it
  in try-with-resources, whose close() shuts down and awaits termination.
- BatfishJobExecutor shut the pool down in a finally, but the finally
  only spanned the result-collection loop; the pool was created earlier
  and the intervening setup (shuffle, submit, stats init) could throw and
  leak it. Widen the try/finally to cover everything after pool creation.
  Keep shutdown() rather than try-with-resources: close() blocks awaiting
  termination, but the error path intentionally propagates immediately
  without waiting for in-flight jobs, so CloseResource is suppressed here.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,8 +1,8 @@
 # Test targets are required to depend on our junit rather than using the one provided by Bazel
 build --explicit_java_test_deps
-build --java_language_version=17
+build --java_language_version=21
 build --java_runtime_version=remotejdk_25
-build --tool_java_runtime_version=17
+build --tool_java_runtime_version=21
 
 # Error Prone warnings to enforce
 build --javacopt "-Xep:WildcardImport:ERROR"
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10069)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10069)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10069)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10074, opened 2026-07-14
TransferBDD: build symbolic routes over the analysis's canonical route

dhalperi:
Several places in the symbolic route analysis built a fresh
new BDDRoute(factory, configAtomicPredicates) at the default variable
positions and then related it against routes/constraints the analysis
produces, which are expressed over the _originalRoute the TransferBDD was
constructed with. The two agree only when _originalRoute is itself at the
default positions; a caller that supplies an _originalRoute elsewhere --
e.g. one built by BDDRouteFactory interleaved with other routes, or offset
past reserved leading variables -- gets a variable mismatch, yielding wrong
results or large BDD blowups. With the default _originalRoute the two are
identical, so this is a no-op for existing callers.

Fixed the sites where the analysis's canonical route is directly in scope:
- TransferBDD.computePaths seeds from _originalRoute rather than a fresh
  route.
- TransferBDDUtils.makeRoutePairing pairs with getOriginalRoute() rather
  than a fresh route.
- SearchRoutePoliciesAnswerer / CompareRoutePoliciesUtils build the
  input-constraint / well-formedness route from getOriginalRoute().

For the decoder helpers that are reached from other pipelines and only had
a ConfigAtomicPredicates (ModelGeneration.satAssignmentTo*/constraintsToModel
and CompareRoutePoliciesUtils.relevantAttributesFor), added an overload that
takes the canonical identityRoute explicitly; the existing
ConfigAtomicPredicates overload builds the default-position route and
delegates, so current callers are unchanged while a non-default caller can
now decode correctly.

Adds a regression test that runs computePaths with an originalRoute placed
past reserved shared bits and checks the accept path's output route is over
that route's variables.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/bdd/ModelGeneration.java
@@ -237,12 +237,22 @@ public class ModelGeneration {
    */
   public static AbstractRoute satAssignmentToInputRoute(
       BDD model, ConfigAtomicPredicates configAPs) {
-    BDDRoute bddRoute = new BDDRoute(model.getFactory(), configAPs);
-    RoutingProtocol p = bddRoute.getProtocolHistory().satAssignmentToValue(model);
+    return satAssignmentToInputRoute(model, new BDDRoute(model.getFactory(), configAPs), configAPs);
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10074)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10075, opened 2026-07-14
Model interface-level IS-IS enable/passive (Arista EOS + Cisco)

btarno:
## Problem

Interface-level `isis enable` and `isis passive` parse but are silently dropped by the extractors, so the interface IS-IS mode is never set. The vendor-independent conversion only builds `IsisInterfaceSettings` when the mode is set, so Batfish cannot compute IS-IS adjacencies for interfaces that enable IS-IS at the interface level. This affects **Arista EOS** in particular, which enables IS-IS per-interface with `isis enable <instance>`.

## Changes

**Arista (`vendor/arista`):**
- Grammar: `isis enable` on EOS takes an IS-IS **instance name** (`isis enable CORE`), not a number. The rule was `ENABLE num = dec`, which fails to parse real configs. Changed to `ENABLE name = variable` (the same capture used for other instance-name references, e.g. `router isis <name>`). Verified against the Arista EOS User Manual (`isis enable <instance>`).
- Extractor: added `exitIf_isis_enable` (sets `IsisInterfaceMode.ACTIVE`) and `exitIf_isis_passive` (sets `PASSIVE`), mirroring the existing `exitIf_isis_metric`.

**Cisco (`grammar/cisco`, shared IOS path):**
- Extractor: added the same two handlers. `if_isis_enable` / `if_isis_passive` parsed but had no exit handler.

No conversion changes are needed in either vendor: the existing interface → VI IS-IS conversion already reads `getIsisInterfaceMode()` and builds `IsisInterfaceSettings` once the mode is set.

## Tests

- Arista: new `arista-isis` test config + `testIsisInterfaceEnable` — asserts an `isis enable` interface and an `isis passive` interface get IS-IS settings, and an unconfigured interface does not.
- Cisco: extended the `ios-isis` config + `testIosIsisConfigOnInterface` with `isis enable` and `isis passive` interfaces.

Both changes verified by reverting the extractor with the tests in place (tests fail), then restoring (tests pass). Full `cisco:tests` and `arista/grammar:tests` suites pass; `format.check` and checkstyle clean.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/arista/grammar/Legacy_interface.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/arista/grammar/Legacy_interface.g4
@@ -711,7 +711,7 @@ if_isis_circuit_type
 
 if_isis_enable
 :
-   ENABLE num = dec NEWLINE
+   ENABLE name = variable NEWLINE
 ;
 
 if_isis_hello_interval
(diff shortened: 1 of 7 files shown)


btarno:
Let me know if you see any issues with this.  Thanks guys!

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10075)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on btarno).

dhalperi:
Thanks for the fix!

PR #10077, opened 2026-07-14
palo_alto: split catch-all _null rules into per-token rules

dhalperi:
Rewrite many-token catch-all `_null` rules of the form

    ss_null: ( ADMIN_ROLE | AUTHENTICATION_PROFILE | ... ) null_rest_of_line;

into one idiomatic `<prefix>_<token>_null` leaf rule per token, e.g.

    ss_admin_role_null: ADMIN_ROLE null_rest_of_line;

and expand every parent reference to `ss_null` into the corresponding list
of per-token alternatives. This restores single-token advancement / LL(1)
structure and the `<prefix>_next_token` naming convention described in
docs/parsing/parser_rule_conventions.md; the catch-all form hid many
distinct first-tokens behind a single rule.

The split preserves the set of accepted tokens and the parse-tree shape
(only leaf rule names change), so behavior is unchanged. Scope is limited
to simple catch-alls whose alternatives are each a single token with no
leading `NO?`.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

(with other variants, like null_filler, etc.)

This many-token-null-rule is no longer acceptable. It violates our rules.

First, confirm that I'm writing by reading docs/. Read about LL(1), rule
naming, when to use _null, etc.

Second, if so, develop a dynamic workflow (or similar use of subagents at
your discretion) to rewrite all of these, vendor by vendor, in an
idiomatic way. In the example: bgp_sometoken_null: SOMETOKEN
null_rest_of_line;, getting rid of the catch-all bgp_null.
```

Punt on the `NO?`-prefixed catch-alls; do one vendor per commit.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoParser.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoParser.g4
@@ -74,14 +74,17 @@ newline
    NEWLINE
 ;
 
-s_null
+s_log_collector_null
 :
-    (
-        LOG_COLLECTOR
-        | LOG_COLLECTOR_GROUP
-        | MGT_CONFIG
-    )
-    null_rest_of_line
+   LOG_COLLECTOR null_rest_of_line
+;
+s_log_collector_group_null
+:
+   LOG_COLLECTOR_GROUP null_rest_of_line
+;
+s_mgt_config_null
+:
+   MGT_CONFIG null_rest_of_line
 ;
 
 /*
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10079, opened 2026-07-14
cisco_asa: split catch-all _null rules into per-token rules

dhalperi:
Rewrite catch-all _null rules into per-token leaf rules. Each token
that shared a many-token _null rule now has its own
<prefix>_<token>_null: TOKEN null_rest_of_line; leaf. Leaf names are
matched to their sibling alternatives' prefix and merged alphabetically
among the pre-existing parent alternatives without reordering those
alternatives. Inline/quantified references are wrapped in a
parse-tree-neutral group. This restores LL(1) and the naming
conventions described in docs/parsing. The parse tree shape is
unchanged; only leaf rule names change. Only simple catch-alls are
handled; NO?-prefixed, leading-token, and ambiguous cases are punted.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_cable.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_cable.g4
@@ -30,10 +30,21 @@ c_load_balance
 :
    LOAD_BALANCE
    (
-      clb_docsis_group
+      clb_d20_ggrp_default_null
+      | clb_d30_ggrp_default_null
+      | clb_docsis30_enable_null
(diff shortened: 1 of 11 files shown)


dhalperi (commented):
> @dhalperi reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10079)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10080, opened 2026-07-14
cisco_xr: split catch-all _null rules into per-token rules

dhalperi:
Rewrite many-token catch-all _null rules into one idiomatic
<prefix>_<token>_null leaf rule per token, delete the catch-all, and
expand every parent reference. Leaf prefixes are matched to sibling
alternatives and merged alphabetically among the parent's pre-existing
alternatives without reordering them. Inline/quantified references (e.g.
X_null*) are wrapped in a parenthesized group ( a | b )* that is a
parse-tree no-op, preserving any quantifier. Restores LL(1) and the
<prefix>_next_token naming convention per docs/parsing.

The parse tree shape is unchanged; only leaf rule names change (verified by
the pure rule-name substitution in the parsing ref test). Only simple
catch-alls (every alternative a single token) are handled; NO?-prefixed and
leading-token cases are punted for later.

Also updates the one extractor listener that referenced a removed rule
(ror_routing_instance_null, now split into ..._ospf_null / ..._rip_null);
both remain todo() no-ops.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10080)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10081, opened 2026-07-14
Arista: split catch-all _null rules into per-token rules

dhalperi:
Rewrite the catch-all `_null` rules that matched many tokens under a
single rule into per-token leaf rules. Each leaf takes a prefix matched
to its sibling alternatives (usually the catch-all's own prefix) and is
interleaved alphabetically into the parent alternation. Inline,
quantified, and sequence references to the catch-all are wrapped in a
parse-tree-neutral `( a | b )` group so the parse tree shape is
unchanged -- only leaf rule names change. This restores LL(1) structure
and idiomatic naming per docs/parsing.

Only simple catch-alls are converted. NO?-prefixed and leading-token
cases (e.g. ip_nat_null) are punted for a later change.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10081)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10082, opened 2026-07-14
cisco_nxos: split catch-all _null rules into per-token rules

dhalperi:
Rewrite the many-token catch-all _null rules into per-token leaf rules:
each catch-all alternative becomes its own rule (prefix matched to the
catch-all's siblings, merged alphabetically into the parent), and the
catch-all rule is deleted. Inline and quantified references (e.g.
role_null*, pmquc_null*, and NO pmnqc_no_null) are wrapped in a
parse-tree-neutral ( a | b ) group so the tree shape is preserved.

This restores LL(1)/naming conventions per docs/parsing. The parse tree
shape is unchanged; only leaf rule names change. Only simple catch-alls
are handled; NO?-prefixed cases and the colliding no_null (whose split
names collide with the existing no_ip_null) are punted.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10082)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10083, opened 2026-07-14
a10: split catch-all _null rules into per-token rules

dhalperi:
Rewrite the many-token catch-all `_null` rules into one idiomatic
per-token leaf rule each (prefix matched to sibling alternatives,
merged alphabetically into the parent alternation), delete the
catch-all, and wrap inline references in a parse-tree-neutral group.
This restores LL(1) single-token advancement and the
<prefix>_<token>_null naming convention described in docs/parsing.

The parse tree shape is unchanged aside from the leaf rule names; only
simple catch-alls (single uppercase token per alternative, tail
null_rest_of_line/null_filler) are handled. NO?-prefixed cases are
punted.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/a10/grammar/A10_router_bgp.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/a10/grammar/A10_router_bgp.g4
@@ -55,23 +55,33 @@ bgp_neighbor: ip_address;
 srbn
 :
    srbn_activate
+   | srbn_capability_null
    | srbn_description
+   | srbn_fall_over_null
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10083)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10084, opened 2026-07-14
flatvyos: split catch-all _null rules into per-token rules

dhalperi:
Rewrite the many-token catch-all `_null` rules (bnt_null, it_null,
st_null, iket_null) into one idiomatic `<prefix>_<token>_null` leaf per
token, delete the catch-alls, and expand every parent reference into the
corresponding list of leaf rules.

This restores LL(1) single-token advancement and the `<prefix>_next_token`
naming convention per docs/parsing/parser_rule_conventions.md and
docs/parsing/README.md. The parse tree shape is unchanged; only leaf rule
names change. Only simple catch-alls (single UPPERCASE token per
alternative, tail `null_rest_of_line`/`null_filler`) are handled;
NO?-prefixed cases are punted.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatvyos/FlatVyosParser.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatvyos/FlatVyosParser.g4
@@ -27,8 +27,16 @@ s_system
 
 s_system_tail
 :
-   st_host_name
-   | st_null
+   st_config_management_null
+   | st_console_null
+   | st_flow_accounting_null
+   | st_host_name
+   | st_login_null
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10084)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10085, opened 2026-07-14
f5_bigip_imish: split catch-all _null rules into per-token rules

dhalperi:
Rewrite the many-token catch-all _null rules into per-token leaf rules,
matching prefixes to their sibling alternatives and merging them
alphabetically into the parent. Inline references are wrapped in a
parse-tree-neutral group (NO rb_no_null -> NO ( rb_no_bgp_null |
rb_no_max_paths_null )). Restores LL(1) and the naming convention per
docs/parsing. The parse tree shape is unchanged; only leaf rule names
change. Only simple catch-alls are handled here; NO?-prefixed cases are
punted.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10085)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10086, opened 2026-07-14
Junos: split catch-all _null rules into per-token rules

dhalperi:
Rewrite many-token catch-all `_null` rules of the form

    x_null: ( A | B | C ) null_filler;

into one idiomatic per-token leaf rule each (`x_a_null: A null_filler;` ...),
delete the catch-all, and merge the leaves into the parent's alternative
list. This restores single-token advancement / LL(1) structure and the
`<prefix>_next_token` naming convention described in
docs/parsing/parser_rule_conventions.md; the catch-all form hid many distinct
first-tokens behind a single rule.

Leaf rule names take the prefix of their sibling alternatives, and are merged
alphabetically among the pre-existing alternatives without reordering those
(reordering could change ALL(*) match precedence). Inline/quantified/sequence
references are wrapped in a parenthesized group, which is a parse-tree no-op.

The split preserves the set of accepted tokens and the parse-tree shape (only
leaf rule names change), so behavior is unchanged. Scope is limited to simple
catch-alls whose alternatives are each a single token with no leading `NO?`.

----

Prompt:
```

bgp:
   bgp_abc
   | bgp_def
   | bgp_null
;

bgp_null:
   (SOMETOKEN
   | OTHER_TOKEN
   | OTHER_TOKEN
   ...
   ) null_rest_of_line;

no longer acceptable. Rewrite all of these, vendor by vendor, in an idiomatic
way: bgp_sometoken_null: SOMETOKEN null_rest_of_line;, getting rid of the
catch-all bgp_null. Punt on the NO?-prefixed cases; one vendor per commit; fan
out across vendors in parallel worktrees. Leaf rule names should match sibling
alternatives, and be interleaved alphabetically into the parent.
```


dhalperi (commented):
> @dhalperi reviewed 15 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10086)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10086)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10088, opened 2026-07-15
Arista: parse `ipv6 unicast-routing`

btarno:
Arista EOS `ipv6 unicast-routing` (the global IPv6 forwarding enable) is currently flagged as an unrecognized line. It is a global routing statement, and its sibling `ip routing` already parses via `s_ip_routing`, so this models it the same way.

Changes:
- Add the `UNICAST_ROUTING` lexer token (placed before `UNICAST` so maximal munch matches the longer literal).
- Add an `ipv6_unicast_routing` parser rule under `s_ipv6`, accepting the bare form and the `vrf <name>` form.
- Accepted, not converted, matching how `ip routing` is handled today.

Before this change, feeding a config with `ipv6 unicast-routing` produced `no viable alternative at input 'ipv6 unicast-routing'` (the `unicast-routing` string lexed as a bare identifier).

Adds a test fixture (`arista_ipv6_unicast_routing`) and a test asserting the config parses with no parse warnings. `//projects/batfish/src/test/java/org/batfish/vendor/arista/grammar:tests` passes; checkstyle and google-java-format are clean.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/arista/grammar/AristaLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/arista/grammar/AristaLexer.g4
@@ -4678,6 +4678,8 @@ UNAUTHORIZED: 'unauthorized';
 
 UNAUTHORIZED_DEVICE_PROFILE: 'unauthorized-device-profile';
 
+UNICAST_ROUTING: 'unicast-routing';
+
 UNICAST: 'unicast';
 
 UNIDIRECTIONAL: 'unidirectional';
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10088)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on btarno).

dhalperi:
Out of curiosity, does this cause any problems or you're just cleaning warnings? :)

btarno:
Cleaning warnings mostly.  I hope it's not a problem.  If so, let me know and I can step back on that stuff.

dhalperi:
It's fine, as long as they're real and from your configs :)

PR #10091, opened 2026-07-15
TestrigText: support filtering config files in loadTestrig

zphw:
Added a `loadTestrig` overload that takes a `Predicate<String>` applied to filenames under `configs/`, so tests can load a subset of a snapshot's configuration files. The existing `loadTestrig(dir)` behavior is unchanged (it delegates with an always-true filter).


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10091)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on zphw).

PR #10093, opened 2026-07-16
Arista: test summary-only aggregate suppression in a VRF

btarno:
Adds a dataplane regression test for `aggregate-address <prefix> summary-only` configured in a `vrf` stanza under `router bgp`.

Motivation: a production change on a pair of Arista border leafs applied `aggregate-address 0.0.0.0/0 summary-only advertise-only` inside a VRF. summary-only suppressed every contributing more-specific VRF-wide, collapsing a downstream router's received routes from 35 prefixes to 1 and causing a brief outage. Current master models this correctly (the downstream neighbor's computed BGP RIB shows only the aggregate), but existing coverage (`testBgpAggregateWithLearnedSuppressedRoutes`) only exercises the default VRF, an asplain AS, and a non-default aggregate prefix. This test locks in the VRF-context behavior with the exact shape of that change:

- sessions under `router bgp <asdot AS>` / `vrf <name>` (asdot `65166.10010`; one peer references it in asdot, the other in asplain)
- a `0.0.0.0/0` aggregate, so every learned route is a contributor
- `advertise-only` present (accepted and extracted; local-RIB-install semantics remain a conversion TODO and are not asserted here)
- contributors include a /27 and a covered /32 host route

Verified the test fails if `summary-only` is removed from the snapshot config.

### DIFF
--- a/projects/batfish/src/test/java/org/batfish/vendor/arista/grammar/AristaGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/vendor/arista/grammar/AristaGrammarTest.java
@@ -1880,6 +1880,47 @@ public class AristaGrammarTest {
   }
 
   @Test
+  public void testBgpAggregateVrfSummaryOnly() throws IOException {
+    /*
+     * Snapshot contains c1, c2, and c3. All BGP sessions on c2 live in vrf vxlan_public under
+     * router bgp with an asdot AS (65166.10010 = 4270728986; c1 peers with it in asdot notation,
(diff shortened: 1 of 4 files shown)


dhalperi:
FYI I was inspired by you to create a Lab for these two scenarios, and found a few issues: https://github.com/batfish/lab-validation/pull/220

btarno:
oh, nice catch!

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10093)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on btarno).

PR #10098, opened 2026-07-16
Arista: model aggregate AS_PATH as contributors' common leading sequence

### LINKED ISSUE #10094: Arista EOS: locally-generated aggregate AS_PATH omits contributor AS
## Problem
On Arista EOS, a locally-generated BGP aggregate carries an AS_PATH built
from its contributors' AS numbers. Batfish models the aggregate with an
empty AS_PATH. Lab validation detects the difference on both the
aggregating router and every downstream neighbor, and it changes route
selection on the originating AS.

## Technical Details
Configure `aggregate-address 0.0.0.0/0` (in a VRF or the default VRF) on a
router in AS X that has learned contributors from a neighbor in AS Y. On
the device, `show ip bgp` reports the aggregate with `asPathType: Local`
and `asPath: "Y ?"` — i.e., the contributor AS Y is present. Batfish's BGP
RIB shows the aggregate with an empty AS_PATH.

Because the aggregate then inherits the contributor AS, the AS-path loop
check on the contributor router (AS Y) rejects the re-advertised aggregate:
its own ASN is in the path. Batfish, modeling an empty AS_PATH, sees no
loop and installs the aggregate in both the BGP RIB and the main RIB on
that router.

## Observed In
`eos_bgp_agg_vrf_summary_only` and `eos_bgp_agg_vrf` (companion labs for
batfish/batfish#10093).

## Symptoms
- `test_bgp_rib_routes` on the aggregating router: aggregate as_path `(Y,)`
  on device vs `()` in Batfish.
- `test_bgp_rib_routes` on downstream neighbors: received aggregate carries
  `(X, Y)` on device vs `(X,)` in Batfish.
- `test_bgp_rib_routes` / `test_main_rib_routes` on the contributor router
  (AS Y): device has no aggregate (rejected as an AS-path loop); Batfish
  installs `0.0.0.0/0` from the aggregating router.

## Configuration Patterns
`aggregate-address <prefix>` on EOS where the contributors were learned via
eBGP, and the aggregate is re-advertised toward the contributor's AS.

## Next Steps
Model the AS_PATH of a locally-generated EOS aggregate to include the
contributor AS numbers (matching device behavior), which also makes the
AS-path loop check reject the aggregate on the contributor router.

dhalperi:
On Arista EOS, a locally-generated BGP aggregate carries an AS_PATH
derived from its contributors: the longest common leading AS_SEQUENCE,
with the divergent tail dropped. Batfish modeled the aggregate with an
empty AS_PATH, which changed route selection and defeated the AS-path
loop check on upstream routers, so they wrongly installed the aggregate.

Ground truth captured on cEOS 4.36.0.1F (no as-set): two contributors
"65000 65001 ?" and "65000 65003 ?" produce aggregate "65000 ?" with
atomic-aggregate. EOS does not build an AS_SET union; it keeps only the
shared leading sequence. This matches RFC 4271 aggregation without
as-set, and is consistent with the single-origin case (whole shared path
kept, nothing dropped).

Add an AsPathMode enum to the vendor-neutral BgpAggregate (NONE default,
COMMON_SEQUENCE). NONE preserves the empty-AS_PATH behavior of
Cisco/NX-OS/XR; AristaConversions sets COMMON_SEQUENCE. In the dataplane,
initBgpAggregateRoutes now collects the AS_PATHs of all activating
contributors (rather than stopping at the first) and BgpProtocolHelper
.toBgpv4Route assigns the derived AS_PATH. The new
AsPath.aggregateContributors helper computes the common leading
AS_SEQUENCE; it is factored around commonLeadingAsns so an as-set variant
can reuse the same traversal. The existing on-import loop check then
rejects the aggregate on the contributor/transit ASes with no additional
code.

Recompute-per-round already reflects contributor-set changes: AS_PATH is
part of Bgpv4Route identity, so a changed aggregate produces a real
withdraw/add delta and quiesces once contributors stabilize. No
dependency tracking needed.

Not modeled (no bearing on this issue, deferred): the as-set keyword's
AS_SET form (needs a device-observed lab) and the ATOMIC_AGGREGATE /
AGGREGATOR attributes (non-transitive, large ref-test churn).

Fixes batfish/batfish#10094.

----

Prompt:
```
Read batfish/batfish#10094 ... I don't think we've taken this on before
in Batfish, can you think about it?
```

Built a divergent-contributor lab (eos_bgp_agg_as_path) to capture how
EOS combines contributors on different AS_PATHs, then planned and
implemented the fix scoped to the common-leading-sequence behavior the
lab proved.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -31,6 +31,7 @@ import com.google.common.collect.MultimapBuilder;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
 import com.google.common.graph.ValueGraph;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -57,6 +58,7 @@ import org.apache.logging.log4j.Logger;
 import org.batfish.datamodel.AbstractRoute;
 import org.batfish.datamodel.AbstractRouteDecorator;
 import org.batfish.datamodel.AnnotatedRoute;
+import org.batfish.datamodel.AsPath;
 import org.batfish.datamodel.BgpActivePeerConfig;
 import org.batfish.datamodel.BgpAdvertisement;
 import org.batfish.datamodel.BgpAdvertisement.BgpAdvertisementType;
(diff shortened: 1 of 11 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10098)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10099, opened 2026-07-17
feat(cisco): extract Cisco NX-OS NTP authentication

Murt2005:
## Summary

Add support for `ntp authenticate`, `ntp authentication-key`, `ntp trusted-key`, and `ntp server <> key` for determining whether a NX-OS NTP server is configured for authentication.

## Changes

- Add `TRUSTED_KEY` lexer token (NX-OS already had `AUTHENTICATION_KEY`, `KEY`, and `MD5` tokens with the correct mode predicates, so no other lexer changes were needed)

- Add `ntp_trusted_key` grammar rule and wire it into `s_ntp` (the `ntp_authenticate`, `ntp_authentication_key`, and `ntps_key` rules already existed

- Add a `_key` field to the existing `NtpServer.java` VS model class and create `NtpAuthenticationKey.java`

- Add `_ntpAuthenticate`, `_ntpAuthenticationKeys`, and `_ntpTrustedKeys` fields to `CiscoNxosConfiguration.java` with accessors, plus an `isNtpServerAuthenticated` helper (`_ntpServers` already existed). Trusted keys are stored as a `Set<Integer>` rather than an `IntegerSpace`, since NX-OS `trusted-key` takes exactly one key ID per line with no range support

- Add `exit*` methods in `CiscoNxosControlPlaneExtractor.java` to extract NTP authentication attributes into the VS model, with the server key captured in the existing `ntps_key` rule via `exitNtps_key`

## Testing

- Extend the existing `nxos_ntp` testconfig and add a new `nxos_ntp_no_authenticate` testconfig

- Add unit tests confirming extraction works correctly, that keys are validated against the `1-65535` (uint16) range, and that a server is only reported authenticated when `ntp authenticate` is enabled

- All tests pass locally

- Regenerated `tests/parsing-tests/unit-tests.ref` via `./tools/update_refs.sh` (a one-line parse-tree rendering change for the `nxos_ntp` config due to regenerating the parser after the grammar change)

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
@@ -2460,6 +2460,8 @@ TRUST: 'trust';
 
 TRUSTED: 'trusted';
 
+TRUSTED_KEY: 'trusted-key';
+
 TRUSTPOINT: 'trustpoint';
 
 TTL: 'ttl';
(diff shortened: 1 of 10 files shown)


dhalperi (commented):
> @dhalperi reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10099)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10100, opened 2026-07-17
Arista: exclude non-BGP contributors from aggregate AS_PATH

### LINKED ISSUE #10094: Arista EOS: locally-generated aggregate AS_PATH omits contributor AS

dhalperi:
The common-leading-sequence AS_PATH added in batfish/batfish#10098 was
computed over every activating contributor, including non-BGP main-RIB
routes. For an aggregate like 0.0.0.0/0 (or any prefix covering a
connected/static route), those non-BGP contributors carry an empty
AS_PATH, which collapsed the common leading sequence to empty even when
all BGP contributors shared a leading AS.

On the device the aggregate is formed from BGP-table routes only;
connected routes are not in the BGP table and contribute no AS_PATH.
Restrict the AS_PATH computation to BGP contributors while still letting
any contributor (BGP or not) activate the aggregate. Locally-originated
BGP contributors (e.g. redistributed statics) still carry an empty
AS_PATH and correctly collapse the sequence to empty.

Regression test covers a 0.0.0.0/0 aggregate generated from the main RIB
with a connected contributor alongside a BGP contributor: the connected
route no longer zeroes the aggregate's AS_PATH.

Fixes batfish/batfish#10094 follow-up surfaced by the eos_bgp_agg_vrf
lab-validation labs (default-route aggregate).

----

Prompt:
```
221 is not green, figure out why?
```

The lab-validation PR failed after #10098 merged: the eos_bgp_agg_vrf
labs use a 0.0.0.0/0 advertise-only aggregate, whose contributor set
includes the device's connected interfaces. Those non-BGP contributors
zeroed the computed AS_PATH, so the aggregate still diverged from the
device.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -78,6 +78,7 @@ import org.batfish.datamodel.EvpnType3Route;
 import org.batfish.datamodel.EvpnType5Route;
 import org.batfish.datamodel.GeneratedRoute;
 import org.batfish.datamodel.GenericRibReadOnly;
+import org.batfish.datamodel.HasReadableAsPath;
 import org.batfish.datamodel.Ip;
 import org.batfish.datamodel.KernelRoute;
 import org.batfish.datamodel.MultipathEquivalentAsPathMatchMode;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10100)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10101, opened 2026-07-17
Hierarchy: allocate apply-groups node collections lazily

dhalperi:
Every HierarchyNode eagerly allocated three LinkedHashMaps (children plus
the literal/wildcard child indexes), a LinkedHashSet of applied groups, and
a HashSet of except groups in its constructor, and re-allocated the three
maps on every resetChildren() during group inheritance. For large configs
with many apply-groups, the hierarchy tree holds a large number of nodes --
most of them leaves or nodes with no applied/except groups -- and the tree
is rebuilt as groups are inherited, so these per-node empty collections
dominated allocation during preprocessing.

Allocate all five collections lazily (a null field means empty) and route
access through accessors. Flattened output is byte-for-byte unchanged.

On two large group-heavy Junos configs, full flatten+parse+extract
allocation dropped ~32% and ~39% (roughly 3-4 GB per parse), with a
smaller wall-clock improvement.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10101)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10102, opened 2026-07-18
Hierarchy: skip error-recovery when re-parsing generated set-lines

dhalperi:
When applying apply-groups and apply-path, the preprocessor generates
set-lines from an already-parsed hierarchy tree and re-parses each one with
a fresh FlatJuniperCombinedParser. These generated lines are always valid,
but the parser was constructed with unrecognized-line recovery enabled, so
each one installed the recovery error strategy (which splits the whole
input text with a freshly compiled regex), the recovery-mode parser ATN
simulator, and the lexer recovery strategy -- all unused overhead for a
single valid line.

Disable unrecognized-line recovery for these parsers. Parser and lexer
errors still throw via the error listeners, so a malformed generated line
would still fail loudly. Flattened output is byte-for-byte unchanged.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
@@ -785,7 +785,11 @@ final class Hierarchy {
       return new GrammarSettings() {
         @Override
         public boolean getDisableUnrecognized() {
-          return false;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10102)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10103, opened 2026-07-18
Hierarchy: cache wildcard match results during group inheritance

dhalperi:
Applying apply-groups and apply-path tests each group wildcard against the
literal names present in the tree, via HierarchyWildcardNode.matches. Every
non-trivial match allocated a fresh regex Matcher, and during inheritance
the same wildcard is tested against the same literals repeatedly, so this
was the dominant allocation in preprocessing large group-heavy configs.

Cache match results per wildcard node, keyed by literal text. matches() is
a pure function of the wildcard's immutable pattern and the text, so
caching cannot change results. The cache is allocated lazily, and the bare
"<*>" wildcard (which matches anything) is special-cased to skip both the
regex and the cache. Flattened output is byte-for-byte unchanged.

On several large group-heavy Junos configs this reduced full
flatten+parse+extract allocation by roughly 12%.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/flatjuniper/Hierarchy.java
@@ -731,6 +731,8 @@ final class Hierarchy {
 
       private final String _wildcard;
       private final Pattern _wildcardPattern;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10103)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10107, opened 2026-07-20
comparePeerGroupPolicies: skip peers with no group

### LINKED ISSUE #10106: comparePeerGroupPolicies throws AssertionError on EVPN topologies (AutoAs not supported by CRP)
**Environment**
    Batfish: 2025.7.7
    pybatfish: 2025.7.7.2423
    Topology: EVPN-VXLAN fabric with spine-border route reflectors (Cisco NX-OS style configs)
**Steps to reproduce**
1、Build a snapshot containing a BGP route reflector: a route-reflector-client under a
template peer-policy EVPN-POLICY on the spine-border nodes (e.g. Spine-Border1/2).
2、Build a second (changed) snapshot.
3、Run:
    bf.q.comparePeerGroupPolicies().answer(reference_snapshot="base")

**Observed behavior**
 java.lang.AssertionError
    at org.batfish.minesweeper.question.comparepeergrouppolicies.ComparePeerGroupPoliciesUtils.lambda$findDifferenceCandidates$0(ComparePeerGroupPoliciesUtils.java:137)
    at org.batfish.minesweeper.question.comparepeergrouppolicies.ComparePeerGroupPoliciesUtils.findDifferenceCandidates(ComparePeerGroupPoliciesUtils.java:140)
    at org.batfish.minesweeper.question.comparepeergrouppolicies.ComparePeerGroupPoliciesUtils.getDifferencesStream(ComparePeerGroupPoliciesUtils.java:41)
    at org.batfish.minesweeper.question.comparepeergrouppolicies.ComparePeerGroupPoliciesAnswerer.answerDiff(ComparePeerGroupPoliciesAnswerer.java:31)

**Root cause**
1、route-reflector-client (or the RR role implied by EVPN nv overlay evpn / retain route-target all)
causes Batfish to auto-generate BGP_PEER_IMPORT_POLICY_EVPN, which contains a symbolic
AutoAs (neighbor-AS placeholder).
2、The CRP engine (minesweeper) does not support AutoAs, so the assertion in
findDifferenceCandidates fails.
3、The same AutoAs limitation also breaks searchRoutePolicies when it is not filtered:
bf.q.searchRoutePolicies(policy="BGP_PEER_IMPORT_POLICY_EVPN") crashes identically, while
bf.q.searchRoutePolicies(policies="SET-NH-VIP,DIRECT-TO-EVPN", nodes=...) (avoiding the
auto policy) works fine.

**Why this is hard to work around**
1、comparePeerGroupPolicies is a zero-parameter differential question (variables=[]);
there is no nodes= / policies= filter to exclude RR nodes, so any topology containing
an RR cannot use this question at all.
2、Removing only the route-reflector-client line is not enough — the RR import policy is still
inferred from the EVPN RR role, so the question still crashes.
**Workarounds (not fixes)**
1、Build a separate snapshot pair that excludes every route-reflector node (e.g. only leaf nodes).
2、Or build a pure IPv4 BGP topology without EVPN.
3、Verify peer-policy text changes with a config-level diff or definedStructures instead.
**Suggested fix**
1、Handle/ignore symbolic AutoAs inside ComparePeerGroupPoliciesUtils (e.g. skip or
substitute the placeholder during CRP comparison), or
3、Allow nodes= filtering so RR nodes can be excluded from the comparison.

dhalperi:
findDifferenceCandidates matched peers across snapshots by peer group
name and asserted the group was non-null. BgpPeerConfig.getGroup() is
nullable: a directly-configured neighbor that does not inherit a peer
template (common on Cisco NX-OS, including EVPN route-reflector nodes)
has a null group. Any snapshot containing such a peer crashed the
question with an AssertionError, and because the question takes no
node/policy parameters there was no way to exclude the offending nodes.

This question compares policies per peer group, so a peer with no group
has no group identity to match or deduplicate on. Skip such peers
instead of asserting. Grouped peers on the same device are still
compared. Policy changes on group-less peers are not flagged by this
question.

Also guard the IPv4 unicast address-family dereferences: a peer whose
group is set but that exchanges only EVPN routes has no IPv4 unicast
address family, which would otherwise NPE.

Fixes batfish/batfish#10106.

----

Prompt:
```
Investigate https://github.com/batfish/batfish/issues/10106
```

Investigation found the crash was a null peer group (line 137 assert),
not the AutoAs limitation the issue title proposed. After weighing
whether to skip group-less peers or compare them by peer identity, the
decision was to skip them, keeping the fix aligned with the question's
per-peer-group semantics.

### DIFF
--- a/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesUtils.java
+++ b/projects/minesweeper/src/main/java/org/batfish/minesweeper/question/comparepeergrouppolicies/ComparePeerGroupPoliciesUtils.java
@@ -15,6 +15,7 @@ import org.batfish.datamodel.BgpPeerConfig;
 import org.batfish.datamodel.BgpProcess;
 import org.batfish.datamodel.Configuration;
 import org.batfish.datamodel.Vrf;
+import org.batfish.datamodel.bgp.Ipv4UnicastAddressFamily;
 import org.batfish.datamodel.questions.BgpRoute;
 import org.batfish.datamodel.routing_policy.Environment;
 import org.batfish.minesweeper.question.compareroutepolicies.CompareRoutePoliciesAnswerer;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10107)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10109, opened 2026-07-20
comparePeerGroupPolicies: add nodes specifier

dhalperi:
The question took no parameters, so it always compared every node in
both snapshots. On a network containing a node that cannot be analyzed
(e.g. a BGP route reflector whose generated EVPN import policy uses
AutoAs, which minesweeper does not yet support), there was no way to
exclude that node from the comparison.

Add an optional `nodes` node specifier, matching the convention used by
compareRoutePolicies and bgpPeerConfiguration. When set, the comparison
is restricted to nodes the specifier resolves to in both snapshots;
unset, it defaults to all nodes, preserving existing behavior.

For batfish/batfish#10106.

----

Prompt:
```
Now also add a nodespecifier to the question
```

Followed up by choosing the specifier arguments to match existing BGP
and route-policy questions: a single optional `nodes` nodeSpec, resolved
via SpecifierFactories.getNodeSpecifierOrDefault with an all-nodes
default.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10109)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10110, opened 2026-07-20
checkstyle: drive via bazel target, mirror format.check

dhalperi:
Resolve checkstyle through @maven and run it from a java_binary on
bazel's hermetic JDK, replacing the tools/run_checkstyle.sh that
fetched the jar from GitHub with wget and shelled out to system
`java`. run_checkstyle.sh now invokes //tools/checkstyle:checkstyle.check,
and the pre-commit GitHub workflow's checkstyle job builds and runs
that target instead of doing its own setup-java plus jar caching,
matching the existing format job.

This lets checkstyle run in environments without network access to
GitHub, since the jar is resolved via Maven and cached like every
other dependency.

----

Prompt:
```
Make checkstyle a bazel-native target mirroring
//tools/format:format.check so the jar resolves via @maven (and works
in environments without network access to GitHub) rather than being
downloaded at runtime.
```

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -60,17 +60,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - name: Checkstyle cache
+      - name: Bazelisk cache
         uses: actions/cache@v6
         with:
-          path: "~/.cache/checkstyle"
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10110)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

Issue #10113, opened 2026-07-21
Arista EVPN: Issue multiple evpn import/export route-targets as collections instead of single string/value


demiurges-tech:
## Problem statement
Hi,
In Arista EVPN configurations, a single VRF can define multiple route-target import evpn and multiple route-target export evpn entries.
In Batfish, the current behavior appears to effectively retain/consume only one value per direction (import or export) in key parts of the model/conversion path.

### simplified example
configuration on network device
``` arista
vrf PRODUCTION
  route-target import evpn 65000:100
  route-target import evpn 65000:200
  route-target export evpn 65000:100
  route-target export evpn 65000:200
```

Question submitted via pybatfish
```python
result = bf.q.evpnL3VniProperties().answer().frame()
result
//Node   VRF 	               VNI 	Route_Distinguisher 	Import_Route_Target 	Export_Route_Target
//nod1   PRODUCTION            XXX   XXXXXX                   65000:100                 65000:100
//The other RTs are missing
```

### Questions
- Is there a recommended workaround today for accurately modeling EVPN VRFs ARISTA with multiple import/export RTs?
- Is there any planned roadmap item to evolve these fields to collections (List/Set) instead of single string/value?

Thank you for your feedback.

dhalperi:
Thanks for the report — confirmed and largely fixed.

We reproduced this end-to-end with a two-node Arista cEOS EVPN lab where a VRF declares multiple `route-target import/export evpn` lines (batfish/lab-validation#225). The root cause was that Batfish's Arista model stored a single import and single export target per VRF and overwrote it on each line, so only the last value per direction survived — a VRF whose matching target wasn't listed last failed to import/export the affected EVPN routes.

Fixes now on `master`:

- **batfish/batfish#10115** — the Arista model now retains all import and export EVPN route-targets per VRF. The BGPv4→EVPN export leak attaches every export target, and the EVPN→BGPv4 import leak accepts a route carrying any import target. Cross-VRF EVPN routes now import/export correctly with multiple RTs.
- **batfish/batfish#10116** — a related bug found while validating the fix: the EVPN address family keyed its send-extended-community capability off the wrong flag, so `send-community extended` (without plain `send-community`) stripped route targets on EVPN export. Fixed to match the IPv4 family.

Lab validation (batfish/lab-validation#225, re-triaged in batfish/lab-validation#226) confirms the imported cross-VRF routes now reach the RIB against the real device's state.

[...]

PR #10115, opened 2026-07-21
Arista: retain all EVPN import/export route-targets per VRF

dhalperi:
An Arista VRF may declare several `route-target import evpn` and
`route-target export evpn` lines. AristaBgpVrf stored a single import
and single export target, and the extractor overwrote on each line, so
only the last value per direction survived. A VRF whose matching target
was not listed last failed to import or export the affected EVPN routes.

Store the import and export route targets as insertion-ordered sets and
accumulate them in the extractor. In conversion, the BGPv4->EVPN export
leak now attaches all export targets, and the EVPN->BGPv4 import leak
policy accepts a route carrying any of the import targets rather than
only one.

The vendor-independent Layer3VniConfig still models a single import and
export target, so evpnL3VniProperties continues to report the first of
each; widening that model and the question output is left to a
follow-up. Functional import/export leaking uses the full sets.

For batfish/batfish#10113.

----

Prompt:
```
Now scope the Batfish fixes

Yes let's do PR1. Let's skip L2 widening.

When we do PR 2, let's also rename the columns (Route_Targets) and make
sure that in the answer they're sorted (but probably not in the internal
storage)
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/arista/grammar/AristaControlPlaneExtractor.java
@@ -3640,10 +3640,10 @@ public class AristaControlPlaneExtractor extends AristaParserBaseListener
       return;
     }
     if (ctx.EXPORT() != null) {
-      _currentAristaBgpVrf.setExportRouteTarget(toRouteTarget(ctx.rt));
+      _currentAristaBgpVrf.addExportRouteTarget(toRouteTarget(ctx.rt));
     }
     if (ctx.IMPORT() != null) {
-      _currentAristaBgpVrf.setImportRouteTarget(toRouteTarget(ctx.rt));
+      _currentAristaBgpVrf.addImportRouteTarget(toRouteTarget(ctx.rt));
     }
   }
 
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> @dhalperi partially reviewed 6 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10115)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion.

___
*[`projects/batfish/src/main/java/org/batfish/vendor/arista/representation/eos/AristaBgpVrf.java` line 91 at r1](https://reviewable.io/reviews/batfish/batfish/10115#-Oy4d1-67JFKMiOAEGrM:-Oy4d1-67JFKMiOAEGrN:bfb86hq) ([raw file](https://github.com/batfish/batfish/blob/436d64da98e1597449b0ce73810506f79383a0ee/projects/batfish/src/main/java/org/batfish/vendor/arista/representation/eos/AristaBgpVrf.java#L91)):*
> ```Java
>     // Insertion order is preserved so that the modeled route targets follow
>     // the order they appear in the configuration.
>     _exportRouteTargets = new LinkedHashSet<>(0);
> ```

why? Does order matter? If not, just make it a regular hashset I think.

dhalperi:
Good call — order doesn't matter. The export leak attaches the RTs as a `Set` and the import policy is a `CommunityMatchAny`, both order-insensitive; the only order-sensitive spot is the deferred metadata reporting (arbitrary pick, and PR2 will sort in the answer anyway). Switched to `HashSet` and updated the javadoc/comment/test accordingly. Force-pushed.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10115)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10116, opened 2026-07-21
Arista: honor `send-community extended` on EVPN neighbors

dhalperi:
The EVPN address family set its send-extended-community capability from
the neighbor's send-community flag rather than its send-extended-community
flag. A neighbor configured with `send-community extended` (but not plain
`send-community`) therefore did not advertise extended communities on the
EVPN address family, stripping route targets from exported EVPN routes.
The IPv4 unicast family already used the correct flag.

Use getSendExtendedCommunity() for the EVPN family, matching IPv4.

----

Prompt:
```
While re-triaging the eos_evpn_multiple_rts lab against Batfish, EVPN
route targets were being stripped on export even though the config used
`send-community extended`. Fix the Arista EVPN address-family conversion
to honor send-community extended, and add a regression test.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10116)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10117, opened 2026-07-21
EVPN VNI configs: model multiple import/export route-targets

dhalperi:
VniConfig held a single export route target and a single import route
target pattern, so a VNI configured with several route targets in a
direction surfaced only one. Widen both to sorted sets across Layer2 and
Layer3 VNI configs.

Changes:
- VniConfig: _routeTarget/_importRouteTarget become sorted sets
  routeTargets/importRouteTargets. JSON properties go plural. Builders
  keep single-value convenience setters (setRouteTarget /
  setImportRouteTarget) alongside the new plural setters, so single-RT
  vendor conversions are unchanged.
- Layer2 Type-3 origination (BgpRoutingProcess.initEvpnType3Route) now
  attaches all export route targets to the route's community set,
  matching how the Layer3 BGPv4-to-EVPN leak already behaves.
- Arista: L2 vlan/vlan-aware-bundle extraction accumulates route targets
  (AristaBgpVlanBase, extractor) instead of overwriting, and conversion
  passes the full sets; L3 conversion passes the full sets (the vendor
  model already retained them).
- evpnL3VniProperties: Import_Route_Target / Export_Route_Target columns
  become Import_Route_Targets / Export_Route_Targets, typed
  list-of-string, emitting all values sorted.

For batfish/batfish#10113.

----

Prompt:
```
Implement PR2 for batfish/batfish#10113: widen the vendor-independent
VniConfig model so EVPN VNIs can carry multiple import and export route
targets, covering both Layer2 and Layer3 (L2 has the same
multiple-route-target bug and its export RT is used for Type-3
origination). Rename the evpnL3VniProperties columns to the plural
Route_Targets and emit the values sorted in the answer, keeping internal
storage a plain set.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -891,7 +891,7 @@ final class BgpRoutingProcess implements RoutingProcess<BgpTopology, BgpRoute<?,
               EvpnType3Route route =
                   initEvpnType3Route(
                       l2Vni,
-                      vniConfig.getRouteTarget(),
+                      vniConfig.getRouteTargets(),
                       vniConfig.getRouteDistinguisher(),
                       _process.getRouterId());
               initializationBuilder.from(_evpnType3Rib.mergeRouteGetDelta(route));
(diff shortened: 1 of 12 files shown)


dhalperi (commented):
> @dhalperi reviewed 13 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10117)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10118, opened 2026-07-22
feat(cisco): extract Cisco NX-OS syslog server and logfile attributes

Murt2005:
## Summary

Add extraction for the NX-OS `logging server` and `logging logfile` commands to capture syslog server attributes (severity level, port, facility, VRF, and whether the secure/TLS option is enabled) and local logfile attributes (name, severity, size, and persistent threshold).

## Changes

- Add `LOGFILE` and `PERSISTENT` lexer tokens (`LOGFILE` pushes `M_Word` so the filename lexes as a `WORD`). NX-OS already had `SERVER`, `PORT`, `SECURE`, `FACILITY` (and all facility names), `USE_VRF`, `SIZE`, and `THRESHOLD` tokens, so no other lexer changes were needed

- Add the `logging_logfile` grammar rule with `ll_size` and `ll_persistent` child rules and wire it into `s_logging`. The `logging_server` rule already parsed `level`, `port`, `secure`, `facility`, and `use-vrf`, but the extractor discarded them; add a label to the facility alternative in `ls_facility` so it can be read

- Add `_severityLevel`, `_port`, `_facility`, `_useVrf`, and `_secure` fields to the existing `LoggingServer.java` VS model class and create `LoggingLogfile.java` (name, severity, size, persistent threshold). `secure` is stored as a boolean only — the transport protocol is not modeled since it is not stated explicitly in the device config

- Add a `_loggingLogfile` field to `CiscoNxosConfiguration.java` with accessors (`_loggingServers` already existed). A single logfile is retained since a later `logging logfile` command overrides the previous one. Add a `LOGGING_SERVER_USE_VRF` structure usage so `use-vrf` records a VRF reference (matching the existing AAA `use-vrf` pattern)

- Add `exitLogging_logfile` and populate `exitLogging_server` in `CiscoNxosControlPlaneExtractor.java`. Severity (`0-7`), port (`1-65535`), logfile size (`4096-4194304`), and persistent threshold (`0-99`) are range-validated, with out-of-range values warned and dropped. Logfile size is parsed as a long to avoid overflow on a malformed uint32

## Testing

- Extend the existing `nxos_logging` testconfig with severity/port/facility/use-vrf/secure variations across servers and a `logging logfile` line configuring both size and persistent threshold

- Add unit tests confirming extraction works correctly, that the attributes are validated against their documented ranges, and a reference test confirming `logging server ... use-vrf` records a VRF structure reference

- All tests pass locally

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
@@ -1258,6 +1258,11 @@ LOG_BUFFER: 'log-buffer';
 
 LOG_NEIGHBOR_CHANGES: 'log-neighbor-changes';
 
+LOGFILE
+:
+  'logfile' -> pushMode ( M_Word )
+;
+
 LOGGING: 'logging';
 
 LOGIN: 'login';
@@ -1758,6 +1763,8 @@ PERIODIC: 'periodic';
 PERMIT: 'permit';
 
 PERMIT_ALL: 'permit-all';
+
+PERSISTENT: 'persistent';
 
 PIM: 'pim';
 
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10118)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10123, opened 2026-07-24
feat(cisco): extract ASA syslog logging attributes

Murt2005:
## Summary

Add extraction for the Cisco ASA `logging host`, `logging facility`, `logging trap`, `logging buffered`, and `logging buffer-size` commands to capture per-host syslog server attributes (interface, IPv4/IPv6 address, transport protocol, and port) and global logging attributes (facility, trap severity, and internal log buffer size/severity). This is captured in new vendor-specific `cisco_asa` model classes; the shared `vendor_family.cisco.Logging` model is intentionally left untouched.

## Changes

- Add `EMBLEM`, `SECURE`, and `REFERENCE_IDENTITY` lexer tokens. ASA already had `HOST`, `FACILITY`, `TRAP`, `BUFFER_SIZE`, `TCP`/`UDP`, `PORT`, `FORWARD_SLASH`, and all severity-name tokens, so no other lexer changes were needed.

- Rewrite the `logging_host` grammar rule to real ASA syntax (`logging host interface_name syslog_ip [tcp[/port] | udp[/port]] [format emblem] [secure [reference-identity name]]`). Because the lexer folds the slash form (`tcp/1500`) into a single `VARIABLE` token the transport/port pair is captured via a `logging_host_transport` subrule (bare `tcp`/`udp` keyword or the combined token) and split in the extractor.

- Promote `logging_facility` and `logging_buffer_size` from ignored `_null` rules to value-capturing rules and wire them into `logging_common`.

- Add new VS model classes in `representation/cisco_asa`: `LoggingHost.java` (interface name and host in the constructor, host stored as a `String` so both IPv4 and IPv6 servers are supported, plus transport and port), `SyslogTransportProtocol.java` (enum `TCP`/`UDP` carrying its default port), and `Logging.java` (host map keyed by server IP literal, plus global facility, trap severity, and buffer size/severity). Trap and buffer severities are each captured by name and numeric level to mirror the shared model. Transport defaults to `UDP`; port defaults to `514` (UDP) / `1470` (TCP), sourced from the enum rather than the extractor.

- Add a `_logging` field to `AsaConfiguration.java` with a `getAsaLogging()` accessor (named to distinguish it from the shared `getCf().getLogging()`). The existing shared-model host write is preserved so the vendor-independent `loggingServers` behavior is unchanged.

- Populate `exitLogging_host` and add `exitLogging_facility` and `exitLogging_buffer_size` in `AsaControlPlaneExtractor.java`. The host port is range-validated (`1025-65535`), with out-of-range or non-numeric values warned and the host dropped; port parsing uses `Ints.tryParse` to avoid overflow on a malformed value. Facility and severities are captured via the existing `toLoggingSeverity`/`toLoggingSeverityNum` helpers, which canonicalize either the numeric or named form. IPv6 hosts are accepted and modeled the same as IPv4.

## Testing

- Extend the new `asa-logging` testconfig with `logging host` variations across servers (tcp/port, udp/port, bare host, bare `tcp`, bare `udp`, and an IPv6 server), plus `logging facility`, `logging trap`, `logging buffered`, and `logging buffer-size` lines. Add an `asa-logging-bad-port` testconfig for the out-of-range case.

- Add unit tests confirming extraction works correctly: per-host interface/transport/port including the correct default ports, IPv6 modeling, and the global facility/trap/buffer attributes by both name and level and a negative test confirming an out-of-range port is warned and the host is dropped.

- All tests pass locally.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_logging.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_asa/AsaLegacy_logging.g4
@@ -65,7 +65,7 @@ logging_common
    | logging_archive
    | logging_asdm_buffer_size_null
    | logging_asdm_null
-   | logging_buffer_size_null
+   | logging_buffer_size
    | logging_buffered
    | logging_cmts_null
    | logging_console
@@ -77,7 +77,7 @@ logging_common
    | logging_esm_null
    | logging_event_null
    | logging_events_null
-   | logging_facility_null
+   | logging_facility
    | logging_format
    | logging_history_null
    | logging_host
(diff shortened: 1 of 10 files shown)


dhalperi (commented):
> @dhalperi reviewed 10 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10123)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10124, opened 2026-07-24
Replace Jettison with Jackson and drop the dependency

dhalperi:
Jettison was used only for its lenient, org.json-style mutable tree API in
two places: question-template preprocessing and CLI parameter parsing.

Question preprocessing (Question.java), QuestionHelper, Client,
coordinator Main, and CoordinatorClient now build and traverse Jackson
JsonNode/ObjectNode trees via BatfishObjectMapper. These paths always saw
strict JSON, so the mapping is direct.

The CLI parameter path is the exception: it historically accepted
Jettison's relaxed grammar (=/=> and : key separators, ;/, pair
separators, unquoted keys, single-quoted and bare values,
case-insensitive true/false/null, octal/hex numbers, trailing commas,
and comments). To preserve that behavior without Jettison, add
LenientJsonParser, a focused port of Jettison's
JSONTokener/JSONObject/JSONArray grammar that emits Jackson nodes.
Client.parseParams is its only caller. LenientJsonParserTest covers the
non-standard forms; existing unit and reference tests cover the strict
paths.

Removes org.codehaus.jettison:jettison from MODULE.bazel and the six
BUILD.bazel deps, and re-pins.

----

Prompt:
```
Remind me why we use Jettison in this project? Can we replace it with
something else, like Jackson?
```

Follow-up: after confirming Jettison was used only for its lenient tree
API, the instruction was to do the full migration to Jackson and drop the
dependency. For the CLI parameter path, which relied on Jettison's
lenient parser, the chosen approach was to add a small lenient normalizer
that accepts the historical CLI grammar and emits strict Jackson nodes,
covered by reference tests plus new unit tests.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -50,7 +50,6 @@ BATFISH_MAVEN_ARTIFACTS = [
     "org.apache.logging.log4j:log4j-api:2.26.1",
     "org.apache.logging.log4j:log4j-core:2.26.1",
     "org.apache.logging.log4j:log4j-slf4j-impl:2.26.1",
-    "org.codehaus.jettison:jettison:1.5.7",
     "org.glassfish.grizzly:grizzly-framework:2.4.4",
     "org.glassfish.grizzly:grizzly-http-server:2.4.4",
     "org.glassfish.jersey.containers:jersey-container-grizzly2-http:2.48",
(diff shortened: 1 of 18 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10124)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10124)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10125, opened 2026-07-25
feat(cisco): parse NXOS ip domain-lookup and DNS source-interface

Murt2005:
## Summary

Add extraction for the Cisco NX-OS `ip domain-lookup`, `ip dns source-interface`, and the per-name-server `source-interface` forms to capture DNS-resolver state: whether name lookup is enabled, the global DNS source interface(s) (with optional VRF), and the source interface used to reach each configured name server. This is captured in the vendor-specific `cisco_nxos` model; interface references are recorded so DNS source interfaces are not reported as unused.

## Changes

- Guard the `DNS` lexer token so it only pushes `M_Words` when it does not follow `IP`. This lets `ip dns ...` and `no ip dns ...` tokenize normally (so `source-interface`, interface names, and `vrf` are real tokens) while preserving the existing  behavior for `dns <hostname>` inside an `ip sla` block.

- Add the `ip_dns` grammar rule (`DNS ip_dns_source_interface`) and `ip_dns_source_interface` (`source-interface <intf> [vrf <vrf>]`) under `s_ip`. Widen `ip_name_server` from `(use-vrf <vrf>)?` to `(source-interface <intf> | use-vrf <vrf>)*` so a name-server line may carry a source interface, a fallback VRF, or both in any order. Promote `ip_domain_lookup` and `no_ip_domain_lookup` from ignored `_null` rules to value-capturing rules (strict `NEWLINE`, since the command takes no arguments) and wire them into `s_ip`.

- Add a new VS model class in `representation/cisco_nxos`: `DnsSourceInterface.java` (source interface plus optional VRF for the global `ip dns source-interface` command). Extend `NameServer.java` with an optional `sourceInterface` field (nullable, included in `equals`/`hashCode`/`toString`) alongside the existing `useVrf`.

- Add DNS state to `CiscoNxosConfiguration.java`: a tri-state `Boolean _dnsLookupEnabled` (null = unconfigured), an ordered `_dnsSourceInterfaces` list with a dedup-on-add setter, and a `getAllDnsSourceInterfaces()` accessor that merges both source-interface forms (global `ip dns source-interface` and per-name-server `source-interface`).

- Add `IP_DNS_SOURCE_INTERFACE` and `IP_NAME_SERVER_SOURCE_INTERFACE` structure usages. Populate `exitIp_domain_lookup`/`exitNo_ip_domain_lookup`, `exitIp_dns_source_interface`, and extend `exitIp_name_server` in `CiscoNxosControlPlaneExtractor.java`. Both source-interface forms record an `INTERFACE` structure reference (matching the existing `logging source-interface` pattern), so interfaces used only as a DNS source are counted as referenced. Name servers configured at global scope remain under the default VRF with `use-vrf` stored as a fallback-resolution attribute rather than as VRF ownership.

## Testing

- Add three `cisco_nxos` testconfigs: `nxos_ip_dns` (domain-lookup enabled, mixed `use-vrf` and per-name-server `source-interface` name servers, and multiple global `ip dns source-interface` lines), `nxos_ip_dns_multi_vrf` (a multi-VRF aggregation-router shape with per-VRF and global name servers, one `ip dns source-interface` per VRF, and per-name-server `source-interface` combined with `use-vrf` on two global lines), and `nxos_ip_domain_lookup_disabled` (`no ip domain-lookup`).

- Add unit tests confirming extraction works correctly: the tri-state lookup flag (enabled/disabled/unset), ordered global source interfaces with VRF, name servers under the correct VRF with their `use-vrf` and `source-interface` attributes, the merge of both source-interface forms via `getAllDnsSourceInterfaces()`, and a reference test asserting each DNS source interface carries the expected referrer count.

- All tests pass locally, formatting (`google-java-format`) passes, and the full `cisco_nxos` grammar suite runs clean.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosLexer.g4
@@ -578,7 +578,16 @@ DISTRIBUTE_LIST: 'distribute-list';
 
 DNS
 :
-  'dns' -> pushMode(M_Words)
+  'dns'
+  {
+    // In "[no] ip dns ..." the preceding token is IP, and we want the remainder of the line to be
+    // tokenized normally (e.g. source-interface, interface names, vrf). Everywhere else (notably
+    // the "dns <hostname>" line inside an "ip sla" block), push M_Words so arbitrary text such as
+    // hostnames is consumed as words.
+    if (lastTokenType() != IP) {
+      pushMode(M_Words);
+    }
+  }
 ;
 
 DNSIX: 'dnsix';
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosParser.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/cisco_nxos/grammar/CiscoNxosParser.g4
@@ -216,8 +216,9 @@ s_ip
     | ip_as_path_access_list
     | ip_community_list
     | ip_dhcp
+    | ip_dns
     | ip_domain_list_null
-    | ip_domain_lookup_null
+    | ip_domain_lookup
     | ip_domain_name
     | ip_name_server
     | ip_pim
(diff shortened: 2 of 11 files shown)


dhalperi (commented):
> @dhalperi reviewed 11 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10125)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10073, opened 2026-07-13
ios-xr: test route-target extraction with 4-byte asplain ASN (#10062)

btarno:
Adds regression coverage for the extcommunity/route-target half of #10062.

A 4-byte asplain ASN administrator in `import`/`export route-target` under
`router bgp / vrf / address-family ipv4 unicast` now parses cleanly and is
extracted onto the VRF address-family route-target import/export sets. This
complements the `rd` fix in #10066 (which handled the route-distinguisher half).

New test `GitHub10062Test.testVrfRouteTargetExtraction` + `testconfigs/gh10062rt`:
asserts the import set contains `65000:100` and `4200000001:200`, and the export
set contains `4200000001:300`, as `ExtendedCommunity.target(asn, value)`.

Verified: `bazel test //projects/batfish/src/test/java/org/batfish/grammar/cisco_xr:tests` passes.

PR #10135, opened 2026-07-30
cisco_xr: accept 4-byte local administrator in extcommunity-set rt

### LINKED ISSUE #10062: IOS-XR: 'rd' and 'extcommunity-set rt' with 4-byte (asplain) ASN fail to parse; 2-byte 'rd' under router bgp vrf reports unsupported

dhalperi:
An RFC 4360 type-0 route target has a 2-byte global administrator and a
4-byte local administrator. `extcommunity_set_rt_elem_colon` accepted
only the type-2 shape (`elem_32 COLON elem_16`), so a set element like
`65000:11311111` failed to parse. Adds the `elem_16 COLON elem_32`
alternative, mirroring `ecgalal4_colon` added for IOS in a4f5c8e59a.
The two alternatives are ambiguous when both halves fit in 16 bits;
ANTLR resolves to the existing rule, which extracts the same value.

Both administrators are now modeled as 32-bit range expressions, since
only one of them may be 4 bytes wide and the type is derived from the
values by ExtendedCommunity#target. This widens
ExtendedCommunityLocalAdministratorMatch and the local-administrator
half of RouteTargetExtendedCommunityExpr from Int to Long, matching
their global-administrator counterparts, and drops a narrowing cast in
CommunityMatchExprEvaluator that made a local administrator above 2^31
compare as negative.

Fixes batfish/batfish#10062. The `rd` half was fixed in #10066 and the
vrf address-family route-target half was covered in #10073.

----

Prompt:
```
Investigate https://github.com/batfish/batfish/issues/10062 -- where did we get to?

Look at other vendors. How did we handle that there?

Make the fix using TDD
```

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_extcommunity_set.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/cisco_xr/CiscoXr_extcommunity_set.g4
@@ -45,6 +45,7 @@ extcommunity_set_rt_elem
 :
    extcommunity_set_rt_elem_as_dot_colon
    | extcommunity_set_rt_elem_colon
+   | extcommunity_set_rt_elem_colon_la32
 ;
 
 extcommunity_set_rt_elem_as_dot_colon
@@ -53,9 +54,19 @@ extcommunity_set_rt_elem_as_dot_colon
   extcommunity_set_rt_elem_16 COLON low = extcommunity_set_rt_elem_16
 ;
 
+// RFC 4360 type 2: 4-byte global administrator, 2-byte local administrator.
 extcommunity_set_rt_elem_colon
 :
   high = extcommunity_set_rt_elem_32 COLON low = extcommunity_set_rt_elem_16
+;
+
+// RFC 4360 type 0: 2-byte global administrator, 4-byte local administrator.
+// Ambiguous with extcommunity_set_rt_elem_colon when both halves fit in 16
+// bits; ANTLR resolves to that rule, which extracts to the same value. Only
+// one administrator may be 4 bytes, so neither rule accepts two 32-bit halves.
+extcommunity_set_rt_elem_colon_la32
+:
+  high = extcommunity_set_rt_elem_16 COLON low = extcommunity_set_rt_elem_32
 ;
 
 extcommunity_set_rt_elem_16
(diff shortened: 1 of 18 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10135)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10139, opened 2026-07-31
Junos: parse bare vstp interface entries

dhalperi:
Allow "set protocols vstp interface <name>" without requiring a
trailing mode, edge, priority, or no-root-port substatement.

Add the bare form to xstp-interface-validation and assert it parses
without warnings.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10139)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10140, opened 2026-07-31
claude: add a skill for reviewing vendor parsing/extraction PRs

dhalperi:
The skill reviews PRs touching grammar, extraction, representation, or
conversion code. It is read-only: it reads both revisions with git
show/git grep rather than checking out or building the PR, since that
would run a contributor's build rules and tests locally and CI already
runs the suite. Findings come from tracing which rule matches a line,
what the listener writes, and whether anything reads the field, cited
at file:line per revision, then rank by whether the bad value reaches
the vendor-independent model. Running a probe is an escalation needing
approval.

PR text, fixtures, and fetched pages are treated as evidence rather
than instructions, and docs/ is read from the merge base because a PR
can edit it.

Both files name file and directory shapes to search rather than one
vendor's classes, since layout, the extractor-vs-ConfigurationBuilder
split, and grammar-vs-model directory names all vary.
references/probe-harness.md holds the probe mechanics.

Also gitignore Claude Code's local settings and state directories so
project config can be shared while permission allowlists and
transcripts stay local.

----

Prompt:
```
Use this to start a repo local code review skill. It should be able to
produce both your and my insights on this PR, it should include reading
or pulling the important parts of the docs/ about idiomatic grammar,
parsing, extraction, converstion, reference tracking. It should include
using local or fetching CLI reference manuals as appropriate, etc.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10140)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10151, opened 2026-08-19
InterfaceDependencies: don't crash if aggregate dependency doesn't exist

frazew:
Hi!

This PR is my attempt at fixing an unexpected crash wile trying to parse the following minimal invalid configuration:
```
!RANCID-CONTENT-TYPE: cisco-xr
!
hostname bundle-ether-l3-member
!
interface Bundle-Ether1
  ipv4 address 10.0.0.1 255.255.255.254
interface HundredGigE0/0/0/1
  bundle id 1 mode active
  ipv4 address 10.0.0.1 255.255.255.254
```

```
Cannot invoke "org.batfish.datamodel.Interface.getOwner()" because "iface" is null
java.lang.NullPointerException: Cannot invoke "org.batfish.datamodel.Interface.getOwner()" because "iface" is null
	at org.batfish.datamodel.interface_dependency.InterfaceDependencies.getL1Neighbors(InterfaceDependencies.java:232)
	at org.batfish.datamodel.interface_dependency.InterfaceDependencies.getL1Neighbor(InterfaceDependencies.java:267)
	at org.batfish.datamodel.interface_dependency.InterfaceDependencies.lambda$initialize$3(InterfaceDependencies.java:163)
```

The root cause is that `InterfaceDependencies` wrongly assumes that the interface dependency actually exists on its `getL1Neighbor` calls. That seems wrong, especially given the `Interface.Dependency` docstring: `Therefore callers may not assume that the interface named by getInterfaceName() actually exists.`.

This fix instead first checks that the aggregate members actually exist in the configuration before doing anything else: if any member doesn't exist, then that must mean the configuration was invalid anyway so the bundle should just be shutdown in my opinion.

EDIT: to be clear because it took me some time to understand where exactly the member interface got removed, that's done (and rightly so) by `ConvertConfigurationJob.verifyInterfaces` 

I've struggled a bit with where to put the test, I _think_ the place I put it here looks fine, but obviously feel free to tell me.

As a final note: I've seldom written production-grade Java, so if anything looks weird I'll be glad to receive feedback! :grin:  In any case, thank you for maintaining and keeping batfish open!

### DIFF
--- a/projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
+++ b/projects/batfish/src/test/java/org/batfish/main/BatfishTest.java
@@ -27,6 +27,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertEquals;
@@ -72,6 +73,7 @@ import org.batfish.datamodel.Bgpv4Route;
 import org.batfish.datamodel.Configuration;
 import org.batfish.datamodel.ConfigurationFormat;
 import org.batfish.datamodel.Edge;
+import org.batfish.datamodel.InactiveReason;
 import org.batfish.datamodel.Interface;
 import org.batfish.datamodel.Interface.Dependency;
 import org.batfish.datamodel.Interface.DependencyType;
@@ -703,6 +705,32 @@ public class BatfishTest {
         name -> assertThat(c1.getAllInterfaces().get(name).getActive(), equalTo(true)));
     inactiveIfaces.forEach(
         name -> assertThat(c1.getAllInterfaces().get(name).getActive(), equalTo(false)));
+  }
+
+  @Test
+  public void testPostProcessInterfaceDependenciesAggregateMissingChild() throws IOException {
+    String hostname = "bundle-ether-l3-member";
+
+    SortedMap<String, byte[]> configurationsBytes =
+        ImmutableSortedMap.of(
+            hostname, readResourceBytes("org/batfish/main/testconfigs/" + hostname));
+    Batfish batfish =
+        BatfishTestUtils.getBatfishFromTestrigText(
+            TestrigText.builder().setConfigurationBytes(configurationsBytes).build(), _folder);
+
+    Map<String, Configuration> configurations = batfish.loadConfigurations(batfish.getSnapshot());
+    Configuration c1v = configurations.get(hostname);
+
+    assertThat(c1v.getAllInterfaces(), not(hasKey("HundredGigE0/0/0/1")));
+    assertThat(c1v.getAllInterfaces().get("Bundle-Ether1").getActive(), equalTo(false));
+    assertThat(
+        c1v.getAllInterfaces().get("Bundle-Ether1").getInactiveReason(),
+        equalTo(InactiveReason.INVALID));
+    assertThat(
+        c1v.getAllInterfaces().get("Bundle-Ether1").getDependencies().stream()
+            .map(Dependency::getInterfaceName)
+            .collect(ImmutableSet.toImmutableSet()),
+        equalTo(ImmutableSet.of("HundredGigE0/0/0/1")));
   }
 
   @Test
--- a/projects/batfish/src/test/resources/org/batfish/main/testconfigs/bundle-ether-l3-member
+++ b/projects/batfish/src/test/resources/org/batfish/main/testconfigs/bundle-ether-l3-member
@@ -1,0 +1,9 @@
+!RANCID-CONTENT-TYPE: cisco-xr
+!
+hostname bundle-ether-l3-member
+!
+interface Bundle-Ether1
+  ipv4 address 10.0.0.1 255.255.255.254
+interface HundredGigE0/0/0/1
+  bundle id 1 mode active
+  ipv4 address 10.0.0.1 255.255.255.254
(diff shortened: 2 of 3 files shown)


dhalperi:
Thanks for digging into this — and for the nice minimal repro, that made it easy to trace.

I think the crash is real but the root cause is a bit further upstream than InterfaceDependencies. Quick walkthrough of what happens to your config:

A bundle member can't have L3 config. Batfish models that on purpose  ConvertConfigurationJob.verifyInterfaces deletes the member outright:

  ```java
  // ConvertConfigurationJob.java:960
  if (!i.getAllAddresses().isEmpty()) {
    w.redFlagf("Interface %s is a member of %s interface %s but it has L3 addresses", ...);
    c.getAllInterfaces().remove(name);
  ```

Ideally, we'd handle this in XR itself... what actually happens on your production device?

(XR sets channelGroup for bundle members over in CiscoXrConfiguration.java:1293, so it matches.) Bundle-Ether1 sticks around still holding an AGGREGATE dependency on the interface that just got deleted, and that's the dangling dep you're NPEing on.

The thing is, `verifyInterfaces` already has code specifically meant to prevent that — verifyDependencies at line 982 strips AGGREGATE deps that point at interfaces which don't exist. It just doesn't run late enough. It's a single pass over a snapshot of the key set, and getAllInterfaces() is a TreeMap, so Bundle-Ether1 gets visited before HundredGigE0/0/0/1. At that point the member is still there, so the dep survives; the member is removed a few iterations later and nothing goes back to fix up Bundle-Ether1. Fun consequence: rename the member to something that sorts before B and the crash goes away.

So I'd rather fix it there — either iterate the removals to a fixpoint, or just re-run the dependency cleanup after the removal pass — so the "no VI interface depends on a nonexistent interface" invariant actually holds. Nice side effect: once the stale dep is gone, Bundle-Ether1 has zero AGGREGATE deps and the existing NO_MEMBERS branch in InterfaceDependencies (line 105) shuts it down for you. Same end state you're after, through the path that was already built for it.

Two other reasons I'd prefer that over patching InterfaceDependencies:

  - There's a second allIfaces.get(dep.getInterfaceName()) fed straight into getL1Neighbor at line 132 (the neighbors.size() == 2 virtual-portchannel check) with the same latent NPE. Your early break happens to be reached first so it hides that one, but it doesn't fix it.
  - Deactivating the whole aggregate as soon as one member is missing changes behavior for the multi-member case. Today a missing member maps to NON_EXISTENT_INTERFACE in the dep graph and just counts as one inactive member, so the bundle only goes down when all members are down — see the intent spelled out in InterfaceDependenciesTest.testGetInterfacesToDeactivate_missing_aggregate ("one of the AGGREGATE deps is present (and active), so don't deactivate"). That test doesn't actually fail with your patch, since it builds a PHYSICAL interface and your branch is under AGGREGATED/REDUNDANT, but a bundle with 3 members where one was dropped would now go down entirely.

If it turns out we do want defensive handling in InterfaceDependencies too (the Dependency javadoc does warn callers not to assume existence, so that's defensible), I'd do it as a null-guard on the member lookups rather than a deactivation — filter the missing deps out before calling getL1Neighbor, and let the existing graph machinery decide the bundle's fate.

Minor stuff, only if you end up keeping code in this file: INVALID isn't used anywhere in InterfaceDependencies today (it's a vendor-conversion reason), so NO_MEMBERS/NO_ACTIVE_MEMBERS would fit the local vocabulary better. And on where the test goes — for the ConvertConfigurationJob fix, a unit test there or in InterfaceDependenciesTest in projects/common will be less brittle than the end-to-end one in BatfishTest, since that one is asserting on XR conversion behavior as a side effect. Your repro config is still worth keeping as an XR-level test somewhere.

None of this is Java-style feedback, for what it's worth — the code reads fine. Happy to look at a revision.

frazew:
thanks for the quick review!

A few things before I change anything:
- I went with force-deactivating the bundle because that seemed like the sensible option to me: that configuration is straight up rejected by an actual IOS-XR device (since it's indeed invalid) so I thought "breaking" the reconstructed config would make that very clear
- `Your early break happens to be reached first so it hides that one, but it doesn't fix it.` that was actually very much intentional, specifically to handle all cases :grin: 
- I indeed agree that handling null values inside `InterfaceDependencies` would still make sense just to avoid catastrophic crashes like that in case any other code path leads to this, I'll check how to do that
 
Anyway since that configuration would be fully rejected by a real router I feel like batfish should maybe completely mark the host with a parsing/reconstruction error? (which I realize is not what this PR does, I didn't end up implementing that for some reason):
```
RP/0/RP0/CPU0:aa(config)#interface GigabitEthernet0/0/0/0
RP/0/RP0/CPU0:aa(config)#ipv4 address 10.0.0.1 255.255.255.254
RP/0/RP0/CPU0:aa(config)#interface Bundle-Ether 1
RP/0/RP0/CPU0:aa(config-if)#ipv4 address 10.0.0.1 255.255.255.254
RP/0/RP0/CPU0:aa(config)#interface GigabitEthernet 0/0/0/0
RP/0/RP0/CPU0:aa(config-if)#bundle id 1
RP/0/RP0/CPU0:aa(config-if)#exit
RP/0/RP0/CPU0:aa(config)#exit
Uncommitted changes found, commit them before exiting(yes/no/cancel)? [cancel]:yes

% Failed to commit one or more configuration items during a pseudo-atomic operation. All changes made have been reverted. Please issue 'show configuration failed [inheritance]' from this session to view the errors
RP/0/RP0/CPU0:aa(config)#show configuration failed
Wed Aug 19 12:42:27.540 UTC
!! SEMANTIC ERRORS: This configuration was rejected by
!! the system due to semantic errors. The individual
!! errors with each failed configuration command can be
!! found below.


interface GigabitEthernet0/0/0/0
 bundle id 1 mode on
!!% Bundle membership configuration conflicts with underlying interface configuration (either already existing or also present in the same commit). Conflicting: /ipv4/io/address/primary
!
end

```

Since I feel like that last question kind of decides which direction to take, I'll wait for your input before making any other change, thank again!

EDIT: I realize I'm maybe too focused on this behavior with IOS-XR here, since probably other vendors handle that differently and may silently accept the configuration while leaving the interface out of the bundle. But also: I initially looked at implementing that only in `CiscoXrConfiguration` but I felt like that would maybe duplicated the logic from `ConvertConfigurationJob.verifyInterfaces`

PR #10153, opened 2026-08-21
docs: record that the cisco grammar is for Cisco IOS only

dhalperi:
New docs/parsing/vendors/cisco_ios.md explains that the Cisco
"Frankenparser" is being unwound into an IOS(-XE)-specific vendor, so
the ARUBAOS, CADANT, FORCE10, and FOUNDRY configs still routed through
it work largely by accident. Narrowing a parser rule or a value range
to match the IOS command reference is therefore acceptable even when
those four lose input that used to parse. The review-vendor-pr skill
now says not to report that lost coverage as a finding.

----

Prompt:
```
Changes that improve IOS fidelity at the cost of non-Cisco-IOS variants such as Aruba or Cadant are fine. Maybe we shuold record that explicitly in the repo docs.
```

Later in the same task: keep the skill's advice specific to IOS rather
than stating a generic multi-format rule, and trim the doc to the
minimum.

### DIFF
--- a/.claude/skills/review-vendor-pr/SKILL.md
+++ b/.claude/skills/review-vendor-pr/SKILL.md
@@ -233,6 +233,12 @@ independently-written CLIs; an existing `representation/cisco_asa` class is not
 evidence about IOS. A sibling PR for another OS tells you what a reviewer accepted
 there, not what this device parses — check this platform's own book.
 
+**In the `cisco` grammar, IOS fidelity wins.** ARUBAOS, CADANT, FORCE10, and FOUNDRY
+configs are still routed through this parser, but it is deliberately being narrowed to
+Cisco IOS(-XE), and where those four work it is largely by accident. Do not raise input
+they lose as a finding, and do not go looking for their manuals. See
+[`docs/parsing/vendors/cisco_ios.md`](../../../docs/parsing/vendors/cisco_ios.md).
+
 ## 3. Trace by reading, not by running
 
 For each hypothesis, answer it by reading both revisions. The question is always
--- a/docs/parsing/README.md
+++ b/docs/parsing/README.md
@@ -17,6 +17,7 @@ then processed by an [extractor](../extraction/README.md).
 
 Different vendors have unique configuration formats and parsing requirements. For vendor-specific guidance, see:
 
+- [Cisco IOS](vendors/cisco_ios.md) - why the `cisco` grammar is IOS-only, and what that means for the formats still routed through it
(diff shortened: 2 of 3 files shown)


dhalperi (commented):
> @dhalperi reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10153)* status: all files reviewed, 2 unresolved discussions.

___
*[`docs/parsing/vendors/cisco_ios.md` line 5 at r1](https://reviewable.io/reviews/batfish/batfish/10153#-P-aaofM8uKgC2xqjaf7:-P-aaofM8uKgC2xqjaf8:btdtsuc) ([raw file](https://github.com/batfish/batfish/blob/99db3d45e4d1ddbd86b45680ae72f9a3d5b52e94/docs/parsing/vendors/cisco_ios.md#L5)):*
> ```Markdown
> ## Cisco IOS fidelity takes precedence in the `cisco` grammar
> 
> The `Cisco_*.g4` grammars, `CiscoControlPlaneExtractor`, and the `representation/cisco` model serve
> ```

be more specific/negative.

A few vendors that copied Cisco syntax may produce useful results with the IOS parser. At one point in time, we tried to maintain a single Cisco "Frankenparser" but this proved a maintenance and correctness burden. We are now unwinding that by making the Cisco vendor specific to Cisco IOS(-XE). Batfish still passes configs from Aruba, Cadant, Force10, and Foundry through the IOS parser, but if anything works it's largely by accident and expected to get worse of time.
___
*[`docs/parsing/vendors/cisco_ios.md` line 19 at r1](https://reviewable.io/reviews/batfish/batfish/10153#-P-abGU45k9wjMK1BniQ:-P-abGU45k9wjMK1BniR:b228yug) ([raw file](https://github.com/batfish/batfish/blob/99db3d45e4d1ddbd86b45680ae72f9a3d5b52e94/docs/parsing/vendors/cisco_ios.md#L19)):*
> ```Markdown
> manual to justify the change.
> 
> Only lost coverage is licensed. Silently extracting a wrong value is still a defect, whatever the
> ```

What does this mean?

dhalperi (commented):
> @dhalperi reviewed 3 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10153)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10156, opened 2026-08-25
PaloAlto: model IKE gateways and IPsec tunnels

### LINKED ISSUE #8633: PANO: ipsec configuration is unrecognized
**Describe the bug and expected behavior**
ipsec configuration is not recognized .

**Runnable example**
```py
TXT = """
set network interface tunnel units tunnel.150 ip 169.254.231.26/30
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 auto-key ike-gateway ike-tunnel-1 
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 auto-key ipsec-crypto-profile ipsec-phase2
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 tunnel-monitor enable yes
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 tunnel-monitor destination-ip 169.254.231.25
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 tunnel-monitor tunnel-monitor-profile default
set network tunnel ipsec ipsec-ca1-tgw-tunnel-1 tunnel-interface tunnel.150
"""
bf = Session()
bf.set_network("test")
bf.init_snapshot_from_text(txt, snapshot_name="pan_testing", overwrite=True)
bf.q.initIssues().answer()

        File_Name                  Status File_Format            Nodes
0  configs/config  PARTIALLY_UNRECOGNIZED   PALO_ALTO  ['pan_testing']


  Nodes          Source_Lines           Type                      Details                                          Line_Text                                     Parser_Context
0  None  [configs/config:[9]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
1  None  [configs/config:[6]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
2  None  [configs/config:[8]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
3  None  [configs/config:[4]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
4  None  [configs/config:[7]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
5  None  [configs/config:[5]]  Parse warning  This syntax is unrecognized  set network tunnel ipsec ipsec-ca1-tgw-tunnel-...  [s_network statement_config_devices set_line_c...
6  None  [configs/config:[3]]  Parse warning  This syntax is unrecognized  set network interface tunnel units tunnel.150 ...                 [set_line palo_alto_configuration]

```

**Additional context**

gh-ajinks:
## What

Adds PAN-OS IKE gateway and IPsec tunnel support. Fixes #8633.

Before this change `set network tunnel ipsec ...` had no grammar rule at all and `sn_ike_gateway` was `GATEWAY null_rest_of_line`, so the whole subtree was discarded. A PAN-OS device could never be modeled as an IPsec endpoint, which means any VPN peer opposite one — including AWS VPN connections — reports `MISSING_END_POINT` and no tunnel edge is ever built.

Three commits: the feature, the response to review, and a fix for pre-existing `dh-group` parse gaps that the phase-1 work made consequential.

## Syntax provenance

Every construct here is either observed in `show config running` output from four production PAN-OS firewalls (30 IKE gateways), or confirmed verbatim in the PAN-OS 11.1 CLI reference. The observed set covers `peer-address ip`, `local-address interface|ip`, `pre-shared-key`, `certificate`, `protocol ikev2` with `version ikev2`, `auto-key`, `ike-crypto-profile`, `ipsec-crypto-profile` and `tunnel-interface`. The rest — `ipsec-mode <tunnel|transport>`, `local-address floating-ip`, `peer-address fqdn|dynamic`, `protocol version ikev2-preferred`, `ah authentication`, and the `dh-group` value set — comes from the CLI reference and is quoted in the relevant commit.

## Design notes

**Phase 1 proposals are per gateway, and per algorithm and dh group.** A PAN-OS `ike-crypto-profile` carries no authentication method — that lives on the gateway — so proposals cannot be shared between gateways: a certificate-authenticated gateway would otherwise advertise `PRE_SHARED_KEYS` while its key says `RSA_PUB_KEY`, and `IkePhase1Proposal.isCompatibleWith` compares the auth method. A profile also lists several encryption algorithms and several dh groups while a proposal holds one of each, so the proposals are the cross product and the policy offers them all, as Cisco does with transform-sets.

**Phase 2 proposals and policies are per tunnel.** `ipsec-mode` is configured on the tunnel while a crypto profile is shared between tunnels, and `IpsecPhase2Proposal` equality includes the encapsulation mode — so without per-tunnel emission a transport-mode tunnel would falsely match a tunnel-mode peer and falsely fail against a peer correctly modeled as transport.

**Pre-shared keys.** PAN-OS only ever exports the key encrypted, so the ciphertext is carried through as the key hash and typed `PRE_SHARED_KEY_ENCRYPTED`, as Cisco treats `crypto isakmp key 6`. The value must be present at all: `negotiateIkePhase1Key` treats a key with no value as a peer with nothing configured to authenticate with.

**`protocol version` selects the crypto profile.** PAN-OS allows a different profile per IKE version. In `ikev2-preferred` mode, or when the config does not say, both are offered with IKEv2 first — and the ambiguous case is reported, since the CLI reference documents the three values without documenting a default.

**Unmodeled but valid configuration is reported, not silently dropped.** `peer-address dynamic` and `fqdn`, HA `local-address floating-ip`, and `dh-group no-pfs` are all parsed and then warned about, rather than appearing as unresolved addresses or as tunnels that mysteriously fail to negotiate.

**Parse safety.** Sub-keys that are documented but not modeled get a named `*_null` rule per keyword, added to the parent's alternatives, per `docs/parsing/implementation_guide.md`. That keeps them out of the warning stream without a catch-all absorbing genuine typos, so the grammar still records exactly what it drops. Both `ike gateway` and `tunnel ipsec` also accept a bare keyword with no name. The keyword set at each node was taken from tab completion on a PAN-OS 11.1.13-h3 device rather than from the CLI reference, which does not enumerate them.

## Testing

`PaloAltoGrammarTest` covers extraction, conversion, authentication type, structure definitions and undefined references, warning text, and an end-to-end `initIpsecTopology` + `retainCompatibleTunnelEdges` assertion that an edge actually forms — which is the assertion that catches a phase 1 key with no value. The fixture exercises the same profile on both IKE versions, a multi-algorithm profile, a bracketed dh-group list, `no-pfs`, AH, transport mode, cert auth, subinterface and floating-ip local addresses, address-object peer addresses both defined and undefined, and tunnels with a missing, dangling or absent crypto profile.

Green locally: the PAN-OS suite, both parsing corpora, the Cisco, Juniper, AWS and question suites, checkstyle, and `//tools/format:format.check`.

## Known limitations

Post-quantum `dh-group` values (`kyber-*`, `bike-*`, `frodokem-*`, `hqc-*`, `classic-mceliece-*`, `ntruprime-*`) and the `ake` subtree are documented in 11.1 and not modeled. They are parse errors recovered at the newline, not silent fall-throughs, so they surface as parse warnings — left as follow-up rather than adding values and rules this PR does not otherwise need.

`hash non-auth` is a documented `ike-crypto-profile` value that is still missing; it has no obvious `IkeHashingAlgorithm` counterpart, so it needs a modelling decision rather than a grammar one. Also follow-up.

IKEv1 `exchange-mode`, `peer-id`/`local-id`, `proxy-id`, `lifesize` and `authentication-multiple` are parsed and ignored via named `*_null` rules.


dhalperi (commented):
> @dhalperi+AGNT reviewed 10 files and all commit messages, and made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10156)* status: 0 of 10 files reviewed, 10 unresolved discussions (waiting on gh-ajinks).

___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSqnN8sAKQsQo-RQm:-P-rSqnN8sAKQsQo-RQn:b-ct6rt2) (no related file):*
Read-only review against `r1`, grounded in the PAN-OS 11.1 Web Interface Help (IKE Gateway General / Advanced Options tabs, IPSec Tunnel General tab) and the PAN-OS 11.1 CLI networking cheat sheet. I did not check out or build the branch; CI is green.

The direction is right and the design notes are mostly well-reasoned. Two things block, and both are cheap to fix.

**1. Master moved under this branch, and the PR no longer does what it says.** It is based on `cba1873338`, which predates batfish/batfish#10157 and batfish/batfish#10160. `toIkePhase1Key` never sets `keyHash`, and master's `negotiateIkePhase1Key` now returns early unless both keys have a value — so no session negotiates, no tunnel edge is built, and the `MISSING_END_POINT` this PR is meant to fix persists. The grammar throws the PSK away (`snikega_pre_shared_key: PRE_SHARED_KEY null_rest_of_line`), so the fix is to capture the encrypted key and store it as the hash, which is what `enterCis_key` does for `crypto isakmp key 6`. Details inline.

Please rebase before the next revision — the Follow-up section is stale now that #10157 has landed.

**2. `ikev1` and `ikev2` are conflated.** The web help states the two profiles "can differ", but the grammar folds them into one alternation writing one field, and `protocol version` is discarded, so a gateway with both blocks silently uses whichever came last. That feeds the phase 1 proposal's DH group / encryption / hash, which is exactly what `isCompatibleWith` compares.

Non-blocking, roughly in order of how likely they are to bite: an unresolvable `ike-crypto-profile` produces an empty proposal list with no warning (relevant because PAN-OS's shipped `default` profile never appears in an export, which would also make it a new undefined reference); the tests never assert an edge actually forms, which is why 1 passed CI; `Warnings suppressed` suppresses nothing and duplicates address-object warnings; `peer-address dynamic`/`fqdn` and `local-address floating-ip` red-flag on valid config; `ipsec-mode transport` is modeled as tunnel; new lexer tokens are out of alphabetical order; and 11 of the new null guards are unreachable.

One question worth a device check before merge: the General tab documents Interface as required on an IKE gateway, which would make the `local-address ip` with no `interface` case — and the whole `buildIpToInterfaceName` index built for it — unreachable. The `GW-EXPLICIT` fixture relies on that form.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 170 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSf2NEEZebBCxVF2K:-P-rSf2NEEZebBCxVF2L:bqtqzom) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L170)):*
> ```ANTLR
> ;
> 
> snikeg_peer_address
> ```

Only one of the three documented peer address types is modeled. PAN-OS 11.1 IKE Gateway General tab, Peer IP Address Type:

- *"Dynamic—Select this option if the peer IP address or FQDN value is unknown."*
- *"IP—Enter Peer Address as an IPv4 or IPv6 address or an address object that is an IPv4 or IPv6 address."*
- *"FQDN—Enter Peer Address as an FQDN or an address object that uses an FQDN."*

`peer-address dynamic` isn't a parse error — this alternative dies at the token after `PEER_ADDRESS` and prediction falls to `sn_ike_gateway`'s trailing `null_rest_of_line` — but `getPeerAddress()` stays null, so `toIpsecPeerConfig` red-flags `No peer-address resolved` and drops the tunnel. That's a conversion warning on entirely valid config. Dynamic is what `IpsecDynamicPeerConfig` exists for; FQDN is genuinely unresolvable, but it deserves its own message rather than being reported as unresolved.

Same shape one rule up in `snikeg_local_address`: `local-address floating-ip` (HA active/active) falls to the outer catch-all and leaves the gateway with no local IP.

Follow-up is fine for the modeling, but the spurious red flag is worth avoiding now.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 175 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSTVI2rEtH61DWZvz:-P-rSTVI2rEtH61DWZw-:b-d387pg) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L175)):*
> ```ANTLR
> ;
> 
> snikeg_protocol
> ```

`ikev1` and `ikev2` are folded into one alternation here, and `exitSnikegp_ike_crypto_profile` writes both to the single `IkeGateway._ikeCryptoProfile`, so a gateway configuring both keeps whichever line came last. `protocol version` isn't modeled at all.

PAN-OS 11.1 web help, IKE Gateway → Advanced Options, on the IKE Crypto Profile field: *"The profiles selected for IKEv1 and IKEv2 can differ."* The General tab's Version offers *"IKEv1 only mode, IKEv2 only mode, or IKEv2 preferred mode."*

So for:

```
set network ike gateway GW protocol ikev1 ike-crypto-profile CP-V1
set network ike gateway GW protocol ikev2 ike-crypto-profile CP-V2
set network ike gateway GW protocol version ikev1
```

`CP-V2` wins and `version ikev1` falls to `null_rest_of_line`. The phase 1 proposal's DH group, encryption, hash and lifetime then come from the wrong profile — and those are exactly the four fields `IkePhase1Proposal.isCompatibleWith` compares, so the session can negotiate where the device wouldn't, or the reverse.

Splitting this into `snikegp_ikev1` / `snikegp_ikev2` child rules fixes the modeling and the single-token-advancement violation at once (docs/parsing/parser_rule_conventions.md — `PROTOCOL` then `IKEV1`/`IKEV2` is two tokens in one rule). Keeping the profile per version, plus the selected version, is what lets conversion pick correctly.

No fixture covers a gateway with both blocks today.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4` line 98 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSjwz6wNJS71oEnqf:-P-rSjwz6wNJS71oEnqg:b8d8xur) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4#L98)):*
> ```ANTLR
> AUTHENTICATION_PROFILE: 'authentication-profile';
> 
> AUTO_KEY: 'auto-key';
> ```

This file is alphabetical, and three of the four insertion points break it:

- `AUTO_KEY` lands between `AUTHENTICATION_PROFILE` and `AUTHENTICATION_TYPE`
- `IKE_CRYPTO_PROFILE` / `IKE_GATEWAY` / `IKEV1` / `IKEV2` land before `IKE_CRYPTO_PROFILES`
- `IPSEC` and `IPSEC_CRYPTO_PROFILE` land after `IPSEC_CRYPTO_PROFILES`

Lexing is unaffected — all six strings are distinct, and maximal munch keeps `ike-crypto-profiles`/`ipsec-crypto-profiles` on their plural tokens — so this is purely the sorted-file invariant.

Separately on naming, per docs/parsing/parser_rule_conventions.md's child-rule pattern `<parent_prefix><local_extension>_next_token`: `snikeg_la_interface` and `snikeg_la_ip` sit under `snikeg_local_address`, so they'd be `snikegla_*`.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3481 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSW9M8sE6DE-_HiLr:-P-rSW9M8sE6DE-_HiLs:bd36mun) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3481)):*
> ```Java
>       ikeP1Keys.put(String.format("~IKE_PHASE1_KEY_%s~", gateway.getName()), key);
> 
>       CryptoProfile cp = ikeProfiles.get(gateway.getIkeCryptoProfile());
> ```

When this lookup misses, the gateway still gets an `IkePhase1Policy` and its tunnel still gets a peer config — just with an empty proposal list and no warning. `getMatchingIkeP1Proposal` then iterates nothing and returns null, so the endpoint looks configured but can never negotiate. Every other failure path in `toIpsecPeerConfig` red-flags; this one is silent. A `_w.redFlagf` here would help.

Beyond a typo'd or template-inherited name, I think the common trigger is PAN-OS's shipped profiles. The Advanced Options help says *"Select an existing profile, keep the default profile, or create a new profile"*, and the IPSec Tunnel General tab says *"Select an existing profile or keep the default profile"* — so a gateway left alone emits `ike-crypto-profile default` with no matching `set network ike crypto-profiles ike-crypto-profiles default` line. Since this PR adds the first-ever references to `IKE_CRYPTO_PROFILE` (base had `markConcreteStructure` with zero references), that would also be a new undefined-reference false positive.

I couldn't find a doc page that names the predefined profiles or states they're omitted from `show config running`, so treat that half as unconfirmed — a real export would settle it. Worth checking before merge, since if it's right it fires on most real PAN-OS configs.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3510 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSQbi1A-AMc_sU31D:-P-rSQbi1A-AMc_sU31E:b63exff) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3510)):*
> ```Java
>    * empty key rather than failing. Certificate-authenticated gateways use RSA instead.
>    */
>   private @Nonnull IkePhase1Key toIkePhase1Key(IkeGateway gateway) {
> ```

After a rebase onto current master, no PAN-OS tunnel will negotiate, so as it stands this doesn't yet accomplish what it sets out to.

This branch is based on `cba1873338`, which predates both batfish/batfish#10157 and batfish/batfish#10160. Master now opens `negotiateIkePhase1Key` with:

```java
if (initiatorKey.getKeyHash() == null || responderKey.getKeyHash() == null) {
  return;
}
```

`toIkePhase1Key` sets `keyType` and nothing else, so `keyHash` is null and that guard fires. From there `getIpsecSession` returns before `negotiateIpsecP2`, `getNegotiatedIpsecP2Proposal()` is null, and `retainCompatibleTunnelEdges` skips the edge — same `MISSING_END_POINT` as before the change.

It can't be fixed here alone: `snikega_pre_shared_key: PRE_SHARED_KEY null_rest_of_line` discards the key value, so there is nothing to put in `keyHash`. Please extract the value from `authentication pre-shared-key key <value>` and set it.

The description cites `crypto isakmp key 6` as precedent, but that path keeps the ciphertext *as* the key hash — `enterCis_key` does `String psk = ctx.key.getText()` and stores it unhashed for encType 6. Doing the same here is what makes the encrypted-PSK case work at all.

Worth rebasing regardless: the Follow-up note is now stale, since batfish/batfish#10157 landed and an encrypted/unencrypted PSK pair does negotiate to an empty key.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3542 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rShYN3tw8nBY8XLsF:-P-rShYN3tw8nBY8XLsG:byxsmwy) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3542)):*
> ```Java
>         .ifPresent(algos -> proposal.setEncryptionAlgorithm(algos.get(0)));
>     proposal.setProtocols(ImmutableSortedSet.of(IpsecProtocol.ESP));
>     proposal.setIpsecEncapsulationMode(IpsecEncapsulationMode.TUNNEL);
> ```

`ipsec-mode transport` is silently modeled as tunnel mode. The PAN-OS 11.1 CLI networking cheat sheet documents `set network tunnel ipsec <name> ipsec-mode [tunnel | transport]`, and the IPSec Tunnel General tab advanced options say *"Select Tunnel mode to encrypt the entire packet including the header"* versus Transport.

Since `IpsecPhase2Proposal.equals` is what `getMatchingIpsecP2Proposal` uses, a transport-mode PAN-OS tunnel will falsely match a tunnel-mode peer, and falsely fail against a peer correctly modeled as `TRANSPORT`.

Not asking for it here, but a note for whoever picks it up: `ipsec-mode` is per tunnel while this proposal is keyed by crypto-profile name, so it can't just be threaded through — the proposal has to become per tunnel, or per profile+mode pair. Worth adding to the follow-up list alongside the ESP-only and first-encryption-algorithm notes.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3560 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rScJv819OMwFM5Sy1:-P-rScJv819OMwFM5Sy2:blpso10) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3560)):*
> ```Java
>    * address-resolution warnings.
>    */
>   private @Nonnull Map<Ip, String> buildIpToInterfaceName(Vsys vsys) {
> ```

I'm not convinced this index is reachable. It exists for a gateway whose `local-address` has an `ip` and no `interface`, but PAN-OS 11.1's IKE Gateway General tab documents Interface as the required field — *"Specify the outgoing firewall interface to the VPN tunnel"* — with Local IP Address as the optional refinement, *"Select or enter the IP address for the local interface that is the endpoint of the tunnel"*. If that holds, `set network ike gateway GW local-address interface ...` is always present, `gateway.getLocalInterface() != null` always wins, and `ipToInterfaceName.get(localIp)` is dead code.

The `GW-EXPLICIT` fixture omits `local-address interface`, and `testIpsecTunnelConversion` asserts `explicit.getSourceInterface()` off exactly that input — so the test justifying the index rests on config I don't think a device emits.

Could you check a real `show config running` from a device with an IKE gateway? If `local-address` can't omit `interface`, this whole index plus its per-conversion sweep over every interface and unit address can go.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3562 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rS_t71hqnQ6mOFQIs:-P-rS_t71hqnQ6mOFQIt:b-d4jkj8) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3562)):*
> ```Java
>   private @Nonnull Map<Ip, String> buildIpToInterfaceName(Vsys vsys) {
>     Map<Ip, String> index = new HashMap<>();
>     Warnings suppressed = new Warnings();
> ```

This doesn't suppress anything. `interfaceAddressToConcreteInterfaceAddress` uses its `w` argument only to recurse into `_panorama`/`_shared`; the actual warning site is `addrObject.toConcreteInterfaceAddress(_w)`, on the field. So the sweep still warns through `_w`, and any config with an IPsec tunnel now emits `Could not convert ... AddressObject ...` a second time for every interface address object that isn't concrete — once from interface conversion, once from here.

Either fix the helper to honor its `w` parameter (and drop the parameter if it isn't going to be used), or drop `suppressed` and the comment claiming suppression.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java` line 2205 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSm8kBK0wyRZb3-KJ:-P-rSm8kBK0wyRZb3-KK:bfhn7oi) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2205)):*
> ```Java
> 
>   @Override
>   public void exitSnikega_pre_shared_key(Snikega_pre_shared_keyContext ctx) {
> ```

`_currentIkeGateway` can't be null here. `sn_ike_gateway` is `GATEWAY ( name = variable ( snikeg_authentication | ... ) )?`, so a `snikeg_*` sub-rule only matches after `name` did, which means `enterSn_ike_gateway`'s `ctx.name == null` early return didn't run. Same argument for `sntun_ipsec` / `_currentIpsecTunnel`.

That's 11 dead branches across the new listeners, and they account for most of what codecov reports as 21 partials in `PaloAltoConfiguration.java` and 10 in this file. If one of them is guarding a path I've missed, a comment naming it would help; otherwise they can go.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java` line 468 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSYVV7xaCqEmNmNma:-P-rSYVV7xaCqEmNmNmb:bvwohfy) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java#L468)):*
> ```Java
> 
>   @Test
>   public void testIpsecTunnelConversion() {
> ```

These tests stop at "the peer config exists" — key sets, field values, key types. All of that holds with `keyHash == null`, which is why the missing key value passes CI green.

The stated goal is that a peer opposite a PAN-OS device stops reporting `MISSING_END_POINT` and a tunnel edge gets built; nothing asserts that. A two-node snapshot run through `IpsecUtil.initIpsecTopology` + `retainCompatibleTunnelEdges`, or a `toEdgeSet` assertion, would cover it — and would have caught the key issue.

Two fixture gaps worth closing at the same time: no gateway carries both `protocol ikev1 ike-crypto-profile` and `protocol ikev2 ike-crypto-profile`, and nothing covers `peer-address dynamic` or `peer-address fqdn`.

gh-ajinks:
Thanks — this was a genuinely useful review, and two of your points were things I had wrong rather than just incomplete.

Rebased onto master, now three commits: the original feature, the response to this review, and a third for `dh-group` parse gaps that the phase-1 work made consequential.

**Both blocking items**

1. **Key hash never set.** Correct, and the PR was a no-op as it stood. #10157 and #10160 landed after I opened this, and `negotiateIkePhase1Key` now returns unless both keys have a value. I capture the encrypted pre-shared key (new `KEY` token, `snikegapsk_key`) and set it as the hash, as `enterCis_key` does for `key 6`. Certificate gateways carry their certificate configuration and are typed `RSA_PUB_KEY`. My commit message previously justified using `null_rest_of_line` by claiming the ciphertext lexes as multiple tokens — that was wrong, `VARIABLE` accepts every character in a base64 key. The rule stays for quoted keys; the claim is gone.

2. **ikev1/ikev2 conflated.** Fixed: separate profile fields per version, a `Version` enum, and `protocol version` modeled. `ikev2-preferred` and the unset case offer both profiles with IKEv2 first, and the ambiguous case warns — the CLI reference documents the three values without documenting a default, so the preference order is an assumption and says so.

**Your device question — you were right.** I checked `show config running` on four production firewalls: 30 IKE gateways, every one configures `local-address interface`, none uses `ip` alone. So `buildIpToInterfaceName` was unreachable and the fixture that justified it was config a device does not emit. Both are gone; the source interface now comes from `local-address interface`, with a warning when absent.

**Non-blocking, all addressed**

- Unresolvable `ike-crypto-profile` now warns, and distinguishes "none configured" from "none for the selected version".
- `peer-address dynamic`/`fqdn` and `local-address floating-ip` are modeled and reported as unmodeled rather than as unresolved addresses. `floating-ip` warns even when an interface also resolves, since the modeled source address is then wrong.
- `Warnings suppressed` is gone with the index.
- Tests now assert an edge forms, via `initIpsecTopology` + `retainCompatibleTunnelEdges` — that is what would have caught the missing key value.
- Fixture covers both IKE versions configured together, dynamic, fqdn, floating-ip, and a gateway with no profile.
- The 11 unreachable null guards are gone; the two `ctx.name == null` guards in the `enter` methods are reachable and stay.
- Lexer tokens are in alphabetical position.

**`ipsec-mode`** — you said you were not asking for it here, but the phase-1 rework made the shape obvious, so it is in: phase 2 proposals and policies are per tunnel rather than per crypto profile, since `IpsecPhase2Proposal` equality includes the encapsulation mode. The same commit implements AH and one proposal per encryption algorithm, which were in my own follow-up list.

**Third commit, beyond your review.** Reading the 11.1 CLI reference to confirm the syntax above turned up two pre-existing `cp_dh_group` gaps: the ike form is documented as a bracketed list, and the ipsec form includes `no-pfs`, `group15`, `group16` and `group21` — all parse errors before. `no-pfs` also produced tunnels that could never negotiate, silently, since `negotiateIpsecP2` needs a non-empty PFS intersection. Both fixed, and `no-pfs` now warns. Happy to split that into its own PR if you would rather keep this one scoped to #8633.

Post-quantum `dh-group` values and the `ake` subtree are documented in 11.1 and still unmodeled; they fall through to `null_rest_of_line`.

dhalperi (commented):
> @dhalperi+AGNT reviewed 10 files and all commit messages, made 15 comments, and resolved 9 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10156)* status: 0 of 11 files reviewed, 5 unresolved discussions (waiting on gh-ajinks).

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 170 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSf2NEEZebBCxVF2K:-P-uLC8G8d89HZxjBEMS:bm9fz65) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L170)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

Only one of the three documented peer address types is modeled. PAN-OS 11.1 IKE Gateway General tab, Peer IP Address Type:

- *"Dynamic—Select this option if the peer IP address or FQDN value is unknown."*
- *"IP—Enter Peer Address as an IPv4 or IPv6 address or an address object that is an IPv4 or IPv6 address."*
- *"FQDN—Enter Peer Address as an FQDN or an address object that uses an FQDN."*

`peer-address dynamic` isn't a parse error — this alternative dies at the token after `PEER_ADDRESS` and prediction falls to `sn_ike_gateway`'s trailing `null_rest_of_line` — but `getPeerAddress()` stays null, so `toIpsecPeerConfig` red-flags `No peer-address resolved` and drops the tunnel. That's a conversion warning on entirely valid config. Dynamic is what `IpsecDynamicPeerConfig` exists for; FQDN is genuinely unresolvable, but it deserves its own message rather than being reported as unresolved.

Same shape one rule up in `snikeg_local_address`: `local-address floating-ip` (HA active/active) falls to the outer catch-all and leaves the gateway with no local IP.

Follow-up is fine for the modeling, but the spurious red flag is worth avoiding now.
</blockquote></details>

Fixed. All three peer address types are modeled with their own rules plus a `null_rest_of_line` fallback, `local-address floating-ip` too, and the warnings now name what is unmodeled instead of claiming the address failed to resolve. The mutual-exclusion handling (clearing `_peerAddress` when a later line sets `dynamic`/`fqdn`, and clearing `_localAddress`/`_localFloatingIp` for each other) is the right call for flattened input where order is not guaranteed.

`GW-HA` covering `floating-ip` alongside an interface — where a local IP does resolve but is the wrong one — is a case I did not think to ask for.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 175 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSTVI2rEtH61DWZvz:-P-uL2-n6hR-NtravH3G:b1txsfh) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L175)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

`ikev1` and `ikev2` are folded into one alternation here, and `exitSnikegp_ike_crypto_profile` writes both to the single `IkeGateway._ikeCryptoProfile`, so a gateway configuring both keeps whichever line came last. `protocol version` isn't modeled at all.

PAN-OS 11.1 web help, IKE Gateway → Advanced Options, on the IKE Crypto Profile field: *"The profiles selected for IKEv1 and IKEv2 can differ."* The General tab's Version offers *"IKEv1 only mode, IKEv2 only mode, or IKEv2 preferred mode."*

So for:

```
set network ike gateway GW protocol ikev1 ike-crypto-profile CP-V1
set network ike gateway GW protocol ikev2 ike-crypto-profile CP-V2
set network ike gateway GW protocol version ikev1
```

`CP-V2` wins and `version ikev1` falls to `null_rest_of_line`. The phase 1 proposal's DH group, encryption, hash and lifetime then come from the wrong profile — and those are exactly the four fields `IkePhase1Proposal.isCompatibleWith` compares, so the session can negotiate where the device wouldn't, or the reverse.

Splitting this into `snikegp_ikev1` / `snikegp_ikev2` child rules fixes the modeling and the single-token-advancement violation at once (docs/parsing/parser_rule_conventions.md — `PROTOCOL` then `IKEV1`/`IKEV2` is two tokens in one rule). Keeping the profile per version, plus the selected version, is what lets conversion pick correctly.

No fixture covers a gateway with both blocks today.
</blockquote></details>

Fixed, and more thoroughly than I asked for. Separate `snikegp_ikev1`/`snikegp_ikev2` rules with per-version fields, `protocol version` modeled, and `getEffectiveIkeCryptoProfiles()` selecting by version with IKEv2-first for `ikev2-preferred`. `GW-V1` and `GW-WRONGVER` cover the two cases that were previously wrong, and the dedupe on the preferred path is a good catch I had not thought about.

Agreed on not inventing a default for an absent `protocol version` — I could not find one documented either, and warning is the right call. One residual on the ambiguity warning, raised separately.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4` line 98 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSjwz6wNJS71oEnqf:-P-uLGAu30TUWLI5LmSE:bxymd03) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAltoLexer.g4#L98)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

This file is alphabetical, and three of the four insertion points break it:

- `AUTO_KEY` lands between `AUTHENTICATION_PROFILE` and `AUTHENTICATION_TYPE`
- `IKE_CRYPTO_PROFILE` / `IKE_GATEWAY` / `IKEV1` / `IKEV2` land before `IKE_CRYPTO_PROFILES`
- `IPSEC` and `IPSEC_CRYPTO_PROFILE` land after `IPSEC_CRYPTO_PROFILES`

Lexing is unaffected — all six strings are distinct, and maximal munch keeps `ike-crypto-profiles`/`ipsec-crypto-profiles` on their plural tokens — so this is purely the sorted-file invariant.

Separately on naming, per docs/parsing/parser_rule_conventions.md's child-rule pattern `<parent_prefix><local_extension>_next_token`: `snikeg_la_interface` and `snikeg_la_ip` sit under `snikeg_local_address`, so they'd be `snikegla_*`.
</blockquote></details>

Ordering fixed in all three places, and the tokens added since (`AH`, `FLOATING_IP`, `GROUP15`/`16`/`21`, `IPSEC_MODE`, `KEY`, `NO_PFS`, `VERSION`) are all in the right slots.

The child-rule naming is not addressed — `snikeg_la_interface`, `snikeg_la_ip` and the new `snikeg_la_floating_ip` are under `snikeg_local_address`, so by the convention they would be `snikegla_*`. Not worth another revision on its own; fold it in if you touch these rules again.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java` line 2205 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSm8kBK0wyRZb3-KJ:-P-uLGeN9Q6Tn_K02jHj:bvhk668) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2205)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

`_currentIkeGateway` can't be null here. `sn_ike_gateway` is `GATEWAY ( name = variable ( snikeg_authentication | ... ) )?`, so a `snikeg_*` sub-rule only matches after `name` did, which means `enterSn_ike_gateway`'s `ctx.name == null` early return didn't run. Same argument for `sntun_ipsec` / `_currentIpsecTunnel`.

That's 11 dead branches across the new listeners, and they account for most of what codecov reports as 21 partials in `PaloAltoConfiguration.java` and 10 in this file. If one of them is guarding a path I've missed, a comment naming it would help; otherwise they can go.
</blockquote></details>

Fixed — all 11 removed, and the listeners added since (`exitSntuni_ipsec_mode`, `exitSnicpi_ah`, `exitSnicpi_esp`, `exitSnikegapsk_key`, the `snikegpa_*` and `snikegpv*` handlers) follow the same rule and dereference the current object directly. Consistent, and the grammar backs it in each case.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3481 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSW9M8sE6DE-_HiLr:-P-uL4vxA12R80docG7Z:b-sajvr1) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3481)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

When this lookup misses, the gateway still gets an `IkePhase1Policy` and its tunnel still gets a peer config — just with an empty proposal list and no warning. `getMatchingIkeP1Proposal` then iterates nothing and returns null, so the endpoint looks configured but can never negotiate. Every other failure path in `toIpsecPeerConfig` red-flags; this one is silent. A `_w.redFlagf` here would help.

Beyond a typo'd or template-inherited name, I think the common trigger is PAN-OS's shipped profiles. The Advanced Options help says *"Select an existing profile, keep the default profile, or create a new profile"*, and the IPSec Tunnel General tab says *"Select an existing profile or keep the default profile"* — so a gateway left alone emits `ike-crypto-profile default` with no matching `set network ike crypto-profiles ike-crypto-profiles default` line. Since this PR adds the first-ever references to `IKE_CRYPTO_PROFILE` (base had `markConcreteStructure` with zero references), that would also be a new undefined-reference false positive.

I couldn't find a doc page that names the predefined profiles or states they're omitted from `show config running`, so treat that half as unconfirmed — a real export would settle it. Worth checking before merge, since if it's right it fires on most real PAN-OS configs.
</blockquote></details>

The silent case is fixed — three distinct warnings now (`Cannot find ike-crypto-profile`, `No ike-crypto-profile configured`, and the at-version variant), all asserted. Nice touch distinguishing "no profile anywhere" from "a profile exists, just not for the version in effect".

The predefined-`default` half is left as a TODO on the ipsec path, which I am fine with as follow-up now that it warns instead of failing silently. Note the reference-tracking half is still open: `ike-crypto-profile default` will report an undefined reference, and this PR adds the first references to that structure type, so it is a new false positive rather than a pre-existing one. Worth carrying into the same follow-up.

Also confirmed no NPE on `gateway.getVersion()` in the at-version branch: `getEffectiveIkeCryptoProfiles()` only returns empty with a profile configured when the version is explicitly IKEv1 or IKEv2, so it cannot be null there. Subtle enough that it took me a minute.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3510 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSQbi1A-AMc_sU31D:-P-uL0nl5UJndTIKYXPd:b9di3vk) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3510)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

After a rebase onto current master, no PAN-OS tunnel will negotiate, so as it stands this doesn't yet accomplish what it sets out to.

This branch is based on `cba1873338`, which predates both batfish/batfish#10157 and batfish/batfish#10160. Master now opens `negotiateIkePhase1Key` with:

```java
if (initiatorKey.getKeyHash() == null || responderKey.getKeyHash() == null) {
  return;
}
```

`toIkePhase1Key` sets `keyType` and nothing else, so `keyHash` is null and that guard fires. From there `getIpsecSession` returns before `negotiateIpsecP2`, `getNegotiatedIpsecP2Proposal()` is null, and `retainCompatibleTunnelEdges` skips the edge — same `MISSING_END_POINT` as before the change.

It can't be fixed here alone: `snikega_pre_shared_key: PRE_SHARED_KEY null_rest_of_line` discards the key value, so there is nothing to put in `keyHash`. Please extract the value from `authentication pre-shared-key key <value>` and set it.

The description cites `crypto isakmp key 6` as precedent, but that path keeps the ciphertext *as* the key hash — `enterCis_key` does `String psk = ctx.key.getText()` and stores it unhashed for encType 6. Doing the same here is what makes the encrypted-PSK case work at all.

Worth rebasing regardless: the Follow-up note is now stale, since batfish/batfish#10157 landed and an encrypted/unencrypted PSK pair does negotiate to an empty key.
</blockquote></details>

Fixed, and verified end to end. `snikegapsk_key` captures the value, `exitSnikegapsk_key` stores the ciphertext as the hash, and `toIkePhase1Key` carries it through. Rebased onto `656ebd7102`, so the guard is live, and `testIpsecTunnelTopology` now asserts an edge actually forms against a peer — which is the assertion that would fail if the hash went missing again.

Good detail that `authentication pre-shared-key` with no `key` line still leaves the hash null: that is a gateway with nothing configured to authenticate with, which is exactly what #10160 wants to reject.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3542 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rShYN3tw8nBY8XLsF:-P-uLDZQAYEWJgswL8w0:b-viibtm) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3542)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

`ipsec-mode transport` is silently modeled as tunnel mode. The PAN-OS 11.1 CLI networking cheat sheet documents `set network tunnel ipsec <name> ipsec-mode [tunnel | transport]`, and the IPSec Tunnel General tab advanced options say *"Select Tunnel mode to encrypt the entire packet including the header"* versus Transport.

Since `IpsecPhase2Proposal.equals` is what `getMatchingIpsecP2Proposal` uses, a transport-mode PAN-OS tunnel will falsely match a tunnel-mode peer, and falsely fail against a peer correctly modeled as `TRANSPORT`.

Not asking for it here, but a note for whoever picks it up: `ipsec-mode` is per tunnel while this proposal is keyed by crypto-profile name, so it can't just be threaded through — the proposal has to become per tunnel, or per profile+mode pair. Worth adding to the follow-up list alongside the ESP-only and first-encryption-algorithm notes.
</blockquote></details>

Fixed, including the structural part I flagged as the hard bit: phase 2 proposals and policies are now emitted per tunnel rather than per crypto profile, so `ipsec-mode` genuinely lands on the proposal that `IpsecPhase2Proposal.equals` compares. AH came along too. `TUN-MULTI` covers transport mode and `TUN-AH` covers the protocol.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3560 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rScJv819OMwFM5Sy1:-P-uL9ay2go7k4zj8aWA:bl4avkx) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3560)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

I'm not convinced this index is reachable. It exists for a gateway whose `local-address` has an `ip` and no `interface`, but PAN-OS 11.1's IKE Gateway General tab documents Interface as the required field — *"Specify the outgoing firewall interface to the VPN tunnel"* — with Local IP Address as the optional refinement, *"Select or enter the IP address for the local interface that is the endpoint of the tunnel"*. If that holds, `set network ike gateway GW local-address interface ...` is always present, `gateway.getLocalInterface() != null` always wins, and `ipToInterfaceName.get(localIp)` is dead code.

The `GW-EXPLICIT` fixture omits `local-address interface`, and `testIpsecTunnelConversion` asserts `explicit.getSourceInterface()` off exactly that input — so the test justifying the index rests on config I don't think a device emits.

Could you check a real `show config running` from a device with an IKE gateway? If `local-address` can't omit `interface`, this whole index plus its per-conversion sweep over every interface and unit address can go.
</blockquote></details>

Thanks for actually checking the devices — 30 gateways across four firewalls all configuring `local-address interface` settles it. The index is gone and `sourceInterface` now comes straight from the gateway interface, with a warning when it is absent. The `GW-SUBIF` fixture keeping both `local-address interface` and `local-address ip` is the realistic shape, which is better than what it replaced.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3562 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rS_t71hqnQ6mOFQIs:-P-uL8ou6qThazq-Bvxl:b-b0csvq) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3562)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

This doesn't suppress anything. `interfaceAddressToConcreteInterfaceAddress` uses its `w` argument only to recurse into `_panorama`/`_shared`; the actual warning site is `addrObject.toConcreteInterfaceAddress(_w)`, on the field. So the sweep still warns through `_w`, and any config with an IPsec tunnel now emits `Could not convert ... AddressObject ...` a second time for every interface address object that isn't concrete — once from interface conversion, once from here.

Either fix the helper to honor its `w` parameter (and drop the parameter if it isn't going to be used), or drop `suppressed` and the comment claiming suppression.
</blockquote></details>

Resolved by deleting `buildIpToInterfaceName` outright, which takes the misleading `Warnings` object with it. Note the underlying bug in `interfaceAddressToConcreteInterfaceAddress` — it ignores its `w` parameter and warns on `_w` — is still there for other callers. Not this PR's problem now that nothing here depends on it, but it is a trap for the next caller that tries to suppress.
___
*[`projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java` line 468 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSYVV7xaCqEmNmNma:-P-uL6Ch6q1CdrNOx3KP:b-krxcxf) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java#L468)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

These tests stop at "the peer config exists" — key sets, field values, key types. All of that holds with `keyHash == null`, which is why the missing key value passes CI green.

The stated goal is that a peer opposite a PAN-OS device stops reporting `MISSING_END_POINT` and a tunnel edge gets built; nothing asserts that. A two-node snapshot run through `IpsecUtil.initIpsecTopology` + `retainCompatibleTunnelEdges`, or a `toEdgeSet` assertion, would cover it — and would have caught the key issue.

Two fixture gaps worth closing at the same time: no gateway carries both `protocol ikev1 ike-crypto-profile` and `protocol ikev2 ike-crypto-profile`, and nothing covers `peer-address dynamic` or `peer-address fqdn`.
</blockquote></details>

Fixed. `testIpsecTunnelTopology` builds a peer and runs `initIpsecTopology` + `retainCompatibleTunnelEdges`, asserting the edge is non-empty — exactly the gap. Having the peer's policy reference its own proposal name rather than the PA's is the right way to write it; a test that shared names would have passed for the wrong reason.

Both fixture gaps are closed too: `GW-V1`/`GW-WRONGVER`/`GW-AMBIGUOUS`/`GW-SAMEPROF` for the version cases, `GW-DYNAMIC`/`GW-FQDN` for the peer types.
___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLkkg1ujxNY4yGD4n:-P-uLkkg1ujxNY4yGD4o:bnr9l8b) (no related file):*
Reviewed `r2`. All ten items from `r1` are addressed, several more thoroughly than I asked. Rebased onto `656ebd7102`, so the key-value guard is actually exercised, and CI is green.

Worth calling out what went beyond the feedback: the ikev1/ikev2 split came with `protocol version` modeling and a dedupe on the preferred path; `ipsec-mode` was fixed structurally by moving phase 2 emission per tunnel rather than patching a value through; the `local-address ip` index was removed on the strength of an actual check against 30 gateways on four firewalls; and `testIpsecTunnelTopology` asserts the thing the feature is for. The generated-name collision warning and the address-object reference marking weren't asked for at all. The updated description is accurate about provenance, which makes it much easier to review the parts I can't check against a device.

One thing to settle before merge: an ipsec-crypto-profile with **no** `dh-group` line leaves `pfsKeyGroups` empty, so the tunnel silently can't negotiate — the same failure the `no-pfs` warning exists to prevent, and your own `IPSEC-AH`/`TUN-AH` fixture is in that state. Either it needs the equivalent warning, or the fixture should carry a DH group.

Two smaller ones: the ambiguous-version warning fires even when both versions name the same profile, where the assumption changes nothing (`GW-SAMEPROF`); and `snicp_ike_crypto_profiles` has no `null_rest_of_line`, so the description's claim that post-quantum `dh-group` values and the `ake` subtree "fall through to `null_rest_of_line`" isn't true — they're parse warnings. Since the description becomes the squash-merged message, fix one or the other. Plus a `CryptoProfile.equals` hygiene note.

Nothing blocking. Happy to sign off once the `dh-group`-absent case is decided either way.
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 413 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLTYXDV98OoIktpTQ:-P-uLTYXDV98OoIktpTR:br6qfkh) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L413)):*
> ```ANTLR
> ;
> 
> snicp_ike_crypto_profiles
> ```

This rule has no `null_rest_of_line` alternative, which makes two claims in the PR description wrong.

"Post-quantum `dh-group` values ... and the `ake` subtree are documented in 11.1 and not modeled; they fall through to `null_rest_of_line`." They don't. `dh-group kyber-768` enters `cp_dh_group`, fails at the value, and there is nothing above it to catch the line — `snicp_ike_crypto_profiles`, `sn_ike_crypto_profiles`, `sn_ike`, `s_network` and `statement_config_devices` all lack a catch-all. `ake ...` has no matching alternative at all. Both are parse errors recovered at the newline, i.e. a parse warning per line, not a silent fall-through.

Low severity in itself — a warning on config Batfish doesn't model is the acceptable outcome, and it's pre-existing. But the description becomes the squash-merged commit message, so it shouldn't assert parse safety the grammar doesn't provide. Either add `| null_rest_of_line` here (which would make the claim true and is a one-line change) or reword it.

Two adjacent gaps in the same rule, since this revision is already about accepting the documented value set on crypto profiles:

- `cp_hash` has no `non-auth`. The IKE Crypto profile help lists Authentication as "sha512, sha384, sha256, sha1, md5, non-auth" and says that with an AES-GCM encryption algorithm "you must select the Authentication setting non-auth" — and `cp_encryption_algo` already accepts `aes-128-gcm`/`aes-256-gcm`. So a GCM IKE profile cannot parse its mandatory hash line today.
- `snicp_ipsec_crypto_profiles` has the same missing fallback, so the documented `lifesize` field is a parse error too.

Both are pre-existing and fine as follow-up; flagging them because you're in this rule and citing this doc page anyway.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/IkeGateway.java` line 141 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLPdL7NAuxi-2b1jP:-P-uLPdL7NAuxi-2b1jQ:b-744amn) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/java/org/batfish/representation/palo_alto/IkeGateway.java#L141)):*
> ```Java
>    * profile for both, so the choice of preference order is an assumption rather than the config.
>    */
>   public boolean hasAmbiguousIkeVersion() {
> ```

This fires when both versions name the *same* profile, where the preference order changes nothing. `getEffectiveIkeCryptoProfiles()` dedupes, so `GW-SAMEPROF` produces exactly one proposal either way — and `testIpsecTunnelConversion` asserts precisely that — yet it still gets warned that Batfish is "assuming IKEv2 is preferred".

The same profile on both versions is what the GUI writes when you don't differentiate them, so this is the common case, not the exotic one. Adding `&& !Objects.equals(_ikeV1CryptoProfile, _ikeV2CryptoProfile)` limits the warning to gateways where the assumption actually decides something, and the `GW-SAMEPROF` fixture is already in place to assert the warning is absent.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java` line 68 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLdmhCmcMm8174WSl:-P-uLdmhCmcMm8174WSm:b-1eaj37) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java#L68)):*
> ```Java
> 
>   @Override
>   public boolean equals(Object o) {
> ```

`equals` and `hashCode` were updated for the `_dhGroup` → `_dhGroups` rename but the two fields added in the same commit, `_protocol` and `_noPfs`, are missing from both. `toString` got `dh-groups` but not those either.

No user-visible effect today — `getCryptoProfileOrCreate` looks up by name and type with an explicit filter rather than by `equals`, and nothing else compares profiles — so this is hygiene rather than a defect. Worth closing while the reason is fresh: an ESP profile and an otherwise-identical AH profile currently compare equal, which is a surprising thing for the next caller to discover.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3676 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLKTm2w-WkoPtqpKL:-P-uLKTm2w-WkoPtqpKM:b-d4zh0r) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3676)):*
> ```Java
>     }
>     if (!cp.getDhGroups().isEmpty()) {
>       policy.setPfsKeyGroups(ImmutableSortedSet.copyOf(cp.getDhGroups()));
> ```

A profile with no `dh-group` line at all lands here with `getDhGroups()` empty, so `setPfsKeyGroups` is never called and the policy keeps its default empty set. `negotiateIpsecP2` requires a non-empty intersection, so the tunnel can never negotiate — the same outcome as `no-pfs`, but with no warning.

Your own fixture hits it: `IPSEC-AH` (line 25) configures only `ah authentication sha256`, and `TUN-AH` uses it. `testIpsecTunnelConversion` asserts the proposal's protocol is AH, but nothing asserts the tunnel can form an edge — and it can't.

The PAN-OS 11.1 IPSec Crypto profile help states no default for DH Group (it only advises picking the highest-numbered group), so applying `group2` silently would be inventing one. Warning is the right move, and it should read the same way the `no-pfs` case does.

So either the config form is reachable and this needs a warning, or DH Group is effectively mandatory and `IPSEC-AH` should carry one — either way something here should change. Worth checking against the same firewalls: does any ipsec-crypto-profile in those exports omit `dh-group`?

dhalperi (commented):
> @dhalperi reviewed 2 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10156)* status: 2 of 11 files reviewed, 8 unresolved discussions (waiting on dhalperi+AGNT and gh-ajinks).

___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java` line 3481 at r1](https://reviewable.io/reviews/batfish/batfish/10156#-P-rSW9M8sE6DE-_HiLr:-P-uqxqxBCMzUw7vbnXk:b4hf7pg) ([raw file](https://github.com/batfish/batfish/blob/028369f86188f07c267c8d5f46287f996f436065/projects/batfish/src/main/java/org/batfish/representation/palo_alto/PaloAltoConfiguration.java#L3481)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

The silent case is fixed — three distinct warnings now (`Cannot find ike-crypto-profile`, `No ike-crypto-profile configured`, and the at-version variant), all asserted. Nice touch distinguishing "no profile anywhere" from "a profile exists, just not for the version in effect".

The predefined-`default` half is left as a TODO on the ipsec path, which I am fine with as follow-up now that it warns instead of failing silently. Note the reference-tracking half is still open: `ike-crypto-profile default` will report an undefined reference, and this PR adds the first references to that structure type, so it is a new false positive rather than a pre-existing one. Worth carrying into the same follow-up.

Also confirmed no NPE on `gateway.getVersion()` in the at-version branch: `getEffectiveIkeCryptoProfiles()` only returns empty with a profile configured when the version is explicitly IKEv1 or IKEv2, so it cannot be null there. Subtle enough that it took me a minute.
</blockquote></details>

I don't want undefined references for built-in structures.. we have a standard pattern for handling this that you should employ here. Check out `referenceService`
___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 269 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-upxgi6nDnXsx_gttS:-P-upxgi6nDnXsx_gttT:b-niy1sn) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L269)):*
> ```ANTLR
> snikegp_version
> :
>     VERSION version = variable
> ```

normally we'd have these as lexer tokens. Why not just add the new token as an alternative? IKEV2_PREFERRED. 

I guess the big issue is putting `| null_rest_of_line` everywhere -- do we need to do that? I prefer to only have named null rules (1 token then null_rest_of_line) in alternatives list.
___
*[`projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java` line 2326 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uqWHl9Lk4YK7IJZ6w:-P-uqWHl9Lk4YK7IJZ6x:b-rmqct1) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/java/org/batfish/grammar/palo_alto/PaloAltoConfigurationBuilder.java#L2326)):*
> ```Java
>     switch (version) {
>       case "ikev1" -> _currentIkeGateway.setVersion(IkeGateway.Version.IKEV1);
>       case "ikev2" -> _currentIkeGateway.setVersion(IkeGateway.Version.IKEV2);
> ```

see parser comments.
___
*[`projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java` line 41 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uqYuz7RK67Qqa1DoU:-P-uqYuz7RK67Qqa1DoV:b-67z46i) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java#L41)):*
> ```Java
> 
>   /** ESP or AH; only meaningful for {@link Type#IPSEC}. Defaults to ESP. */
>   private @Nonnull IpsecProtocol _protocol = IpsecProtocol.ESP;
> ```

default in conversion, not in the extractor. That is, `firstNonNull(_protocol, ESP)` so you can distinguish between unset-and-default and explicitly-set. This often matters in inheritance.

dhalperi (commented):
> @dhalperi+AGNT made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10156)* status: 2 of 11 files reviewed, 8 unresolved discussions (waiting on dhalperi and gh-ajinks).

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4` line 413 at r2](https://reviewable.io/reviews/batfish/batfish/10156#-P-uLTYXDV98OoIktpTQ:-P-ytmLj3K06qFYaU8H4:bq1xa52) ([raw file](https://github.com/batfish/batfish/blob/38fde1741628c21bb955c70d485bfde868731526/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto/PaloAlto_network.g4#L413)):*
<details><summary><i>Previously, dhalperi+AGNT (Agent AGNT for Dan Halperin) wrote…</i></summary><blockquote>

This rule has no `null_rest_of_line` alternative, which makes two claims in the PR description wrong.

"Post-quantum `dh-group` values ... and the `ake` subtree are documented in 11.1 and not modeled; they fall through to `null_rest_of_line`." They don't. `dh-group kyber-768` enters `cp_dh_group`, fails at the value, and there is nothing above it to catch the line — `snicp_ike_crypto_profiles`, `sn_ike_crypto_profiles`, `sn_ike`, `s_network` and `statement_config_devices` all lack a catch-all. `ake ...` has no matching alternative at all. Both are parse errors recovered at the newline, i.e. a parse warning per line, not a silent fall-through.

Low severity in itself — a warning on config Batfish doesn't model is the acceptable outcome, and it's pre-existing. But the description becomes the squash-merged commit message, so it shouldn't assert parse safety the grammar doesn't provide. Either add `| null_rest_of_line` here (which would make the claim true and is a one-line change) or reword it.

Two adjacent gaps in the same rule, since this revision is already about accepting the documented value set on crypto profiles:

- `cp_hash` has no `non-auth`. The IKE Crypto profile help lists Authentication as "sha512, sha384, sha256, sha1, md5, non-auth" and says that with an AES-GCM encryption algorithm "you must select the Authentication setting non-auth" — and `cp_encryption_algo` already accepts `aes-128-gcm`/`aes-256-gcm`. So a GCM IKE profile cannot parse its mandatory hash line today.
- `snicp_ipsec_crypto_profiles` has the same missing fallback, so the documented `lifesize` field is a parse error too.

Both are pre-existing and fine as follow-up; flagging them because you're in this rule and citing this doc page anyway.
</blockquote></details>

Correcting the fix I suggested: adding `| null_rest_of_line` to this rule is the wrong remedy and I shouldn't have proposed it. `docs/parsing/implementation_guide.md` §"When to Use the `_null` Suffix" gives the sanctioned shape — one named rule per keyword, added to the parent's alternatives — and this grammar already followed it: 30 named `*_null` rules against 3 bare `| null_rest_of_line` alternatives at the merge base. A bare catch-all absorbs any unrecognized sub-key including typos, so the grammar stops recording what is dropped. Same point applied across the revision in [the blocking comment on `snikegp_version`](https://reviewable.io/reviews/batfish/batfish/10156#-P-upxgi6nDnXsx_gttS).

The finding itself stands: `dh-group kyber-768` and the `ake` subtree are parse errors recovered at the newline, not fall-throughs, so the description's "they fall through to `null_rest_of_line`" is inaccurate.

Two ways to make it accurate, in order of preference:

1. Reword the description to say they produce parse warnings. No code change, and honest about the state.
2. If you want them parsed, add named rules — a `snicp_ike_ake_null` alternative, and the post-quantum values either as their own `cp_dh_group_value` alternatives or a named null rule — rather than a catch-all.

The same correction applies to the `lifesize` half: a named `cp_lifesize_null` rather than a fallback on `snicp_ipsec_crypto_profiles`. The `cp_hash` / `non-auth` gap is unaffected — that's a missing documented value, not a null-rule question.

gh-ajinks:
Thanks for the correction on the null rules — and for going back to it unprompted.

I'd already moved off the catch-alls before your last comment, having read `bgp_null` in `PaloAlto_bgp.g4`, but I had the granularity wrong: I'd bundled the keywords at each node into one rule per parent. `docs/parsing/implementation_guide.md` is explicit that it's one named rule per keyword, so that's what it is now — 28 named `*_null` rules, no catch-alls anywhere in the file. Your point about typos is the part I'd missed; a bundled rule still swallows more than it names.

I took the keyword set at each of the 12 nodes from tab completion on a device rather than the CLI reference, which does not enumerate them. Removing a catch-all turns any sub-key I failed to list into a parse error on valid config, so this had to be exhaustive rather than plausible. That also turned up `authentication-multiple` on `ike-crypto-profiles` and `lifesize`, both documented and both parse errors on master today — `cp_authentication_multiple_null` and `cp_lifesize_null` now.

**Description.** Reworded, option 1. You're right that "fall through to `null_rest_of_line`" was wrong — they're parse errors recovered at the newline. It now says so. I also listed `hash non-auth` there rather than leaving it unmentioned.

**`hash non-auth`.** Left as follow-up. There's no `IkeHashingAlgorithm` that means "no integrity", so it's a question of whether to add an enum value or warn and skip, and that seemed like your call rather than something to decide inside this PR.

**Built-in structures.** Done via `IKE_CRYPTO_PROFILE_OR_NONE` / `IPSEC_CRYPTO_PROFILE_OR_NONE` following `referenceService`. One thing worth flagging: the false positive you predicted does not reproduce on the four PAN-OS 11.1 firewalls I have. All of them export the predefined profiles as full definition blocks in `show config running` — `default`, `Suite-B-GCM-128`, `Suite-B-GCM-256`, with an older one shipping only `default` — so every reference already resolves and there are no undefined references to report. The pattern still looks right for a config that references a predefined profile without exporting it, but it's latent rather than live, so treat the change as belt-and-braces rather than a fix for something observable.

**`protocol version`.** Now `IKEV1 | IKEV2 | IKEV2_PREFERRED` as tokens, with the listener switching on them instead of comparing strings.

**`CryptoProfile._protocol`.** Default moved to conversion via `firstNonNull`, extractor leaves it null.

**`snikeg_la_*`.** Renamed to `snikegla_*`, since I was in these rules anyway.

The fixture gains a `GW-NULLS`/`TUN-NULLS` pair covering every new null rule plus `ikev2-preferred`, `lifesize` and `authentication-multiple`. Green locally: PAN-OS suite, both parsing corpora, AWS and common-util suites, checkstyle, `format.check`.

One judgement call worth surfacing: `snikeg_protocol_common_null` now silently ignores `protocol-common`, which real configs use for `nat-traversal`, `fragmentation` and `dpd`. That's genuinely unmodeled, so ignoring is right, but it does mean it stops showing up as an unrecognized line. Say the word if you'd rather it stayed visible.

dhalperi (commented):
> @dhalperi reviewed 11 files and all commit messages, resolved 8 discussions, and dismissed @dhalperi+AGNT from a discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10156)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #10157, opened 2026-08-25
IPsec: treat a mixed encrypted/plaintext pre-shared key pair as unverifiable

gh-ajinks:
## What

`negotiateIkePhase1Key` returns early when the initiator's and responder's `IkeKeyType` differ. When one side's config carries an encrypted pre-shared key and the other's carries the plaintext value, there is no negotiated phase 1 key, so `getIpsecSession` never runs phase 2 and `retainCompatibleTunnelEdges` drops the edge.

## Why this is an inconsistency rather than a new policy

Batfish already treats an unreadable pre-shared key as *unverifiable*, not as wrong. `IpsecUtilTest.testNegotiatedIkePhase1EncryptedPsk` asserts that two `PRE_SHARED_KEY_ENCRYPTED` keys with deliberately different hashes (`key1` vs `key2`) negotiate successfully, and the main-code comment says as much: "encrypted PSKs will not be equal and there is no common negotiated key so creating an empty negotiated key".

A mixed pair is epistemically identical — one side's value is unreadable, so the configs cannot prove the keys differ — but it fails a key-type equality check before reaching that logic. This change extends the existing rule to the mixed case, leaving the negotiated hash unset so nothing asserts the keys match.

## What this changes for users

`ipsecSessionStatus` moves from `IKE_PHASE1_KEY_MISMATCH` to `IPSEC_SESSION_ESTABLISHED` for these pairs. That is a deliberate trade and I think the more honest answer: in the mixed case Batfish never had evidence the keys differ — one side is a raw ciphertext, the other a salted hash — so the old verdict was a false negative presented as a check. Every configuration in which the keys are provably different is a same-type pair, and the `UNENCRYPTED`/`UNENCRYPTED` path still enforces hash equality unchanged.

## Who is affected

Vendors disagree about whether the key is recoverable from a config:

| Vendor | Key type |
|---|---|
| Cisco IOS / XR / ASA / Arista | `ENCRYPTED` for `key 6`, `UNENCRYPTED` for `key 0`; keyring-derived keys are always `UNENCRYPTED` |
| Juniper | `UNENCRYPTED` — `$9$` values are decrypted, then hashed |
| VyOS, AWS | `UNENCRYPTED` |

So a Cisco device using `crypto isakmp key 6` against an AWS VPN connection forms no tunnel today. That case is real but uncommon, since it needs type-6 encryption enabled.

In fairness, the broader trigger is my own #10156, which adds PAN-OS IPsec modeling: PAN-OS has no plaintext pre-shared key form at all, so once that lands every PAN-OS↔non-PAN-OS tunnel would hit this. The two changes are independent — this branch is cut from master, contains none of the PAN-OS work, and stands on the Cisco case alone — but I would rather flag the connection than have it discovered. Reviewing them in either order is fine.

## Testing

`IpsecUtilTest`: the mixed pair in both directions, RSA against both pre-shared key flavours in both directions (the guard is a conjunction over both operands), and an end-to-end `getIpsecSession` test asserting a negotiated phase 2 proposal, which is what `retainCompatibleTunnelEdges` actually filters on.

`common/util` 77 tests, plus the Cisco, AWS, PAN-OS and question suites all pass unchanged — no existing test depended on the previous outcome.

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/util/IpsecUtil.java
+++ b/projects/common/src/main/java/org/batfish/common/util/IpsecUtil.java
@@ -252,7 +252,9 @@ public class IpsecUtil {
 
   /**
    * Negotiates key for IKE phase 1 and sets it in the provided ipsecSessionBuilder. Negotiated IKE
-   * P1 key in the ipsecSessionBuilder will be null if no valid key was detected or negotiated
+   * P1 key in the ipsecSessionBuilder will be null if no valid key was detected or negotiated. If
+   * the two key types differ but both are pre-shared keys, one side's value is unreadable, so an
+   * empty key is set rather than none.
    */
   @VisibleForTesting
   static void negotiateIkePhase1Key(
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10157)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #10159, opened 2026-08-25
Junos: parse cRPD eth interfaces, syslog archive option lists, and scrubbed names

dhalperi:
Three unrecognized-syntax parse warnings on a cRPD route server:

- "eth" is now an interface media type, so cRPD's Linux netdev names
  (eth0, eth1.100) lex as interface ids. Extraction, conversion, and
  INTERFACE reference tracking already covered these once the name
  lexes.
- "system syslog file <f> archive" accepts a list of options rather than
  at most one, e.g. "archive size 1m files 5 binary-data".
- F_NameChar accepts a scrubbing marker, so a name that embeds one
  lexes, e.g. the certificate name "local-cert-%CENSORED%". Config
  management left the marker inside the name in every pre-deployment
  rendering of this config; only the rendering read back off the device
  carried the original name.

----

Prompt:
```
Adding parsing so that there are no "this syntax is unrecognized" parse
warnings in this config.
```

Follow-ups during the task: implement extraction, conversion, reference
tracking, and todos as applicable, not just parsing; and drop support
for anything that would not show up in a config read back off a device.
The config supplied was a pre-deployment rendering, so each pattern was
checked against the device-collected rendering too. That ruled out
speculative support for archive option lists that lead with a
no-argument flag.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/flatjuniper/FlatJuniperLexer.g4
@@ -3774,6 +3774,7 @@ F_InterfaceMediaType
    'em' |
    'es' |
    'et' |
+   'eth' | // Linux netdev name, used by cRPD
    'fab' |
    'fe' |
    'fti' |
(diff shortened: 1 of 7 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10159)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10160, opened 2026-08-25
IPsec: require a key value before negotiating an IKE phase 1 key

dhalperi:
batfish/batfish#10157 keyed the mixed encrypted/plaintext pre-shared key
case on the key types alone, so a peer with no key configured at all
negotiated as if its key were merely unreadable: an IOS crypto keyring
with no pre-shared-key statement converts to an unencrypted key with a
null hash. That null hash also threw when comparing two unencrypted
keys. negotiateIkePhase1Key now returns early unless both keys have a
value, and the three same-family cases are one if/else chain.

----

Prompt:
```
Author the fixup changes identified while reviewing
batfish/batfish#10157: guard the mixed encrypted/plaintext pre-shared
key case on the key value being present, and fix the pre-existing NPE
when an unencrypted key has no hash.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/util/IpsecUtil.java
+++ b/projects/common/src/main/java/org/batfish/common/util/IpsecUtil.java
@@ -252,38 +252,46 @@ public class IpsecUtil {
 
   /**
    * Negotiates key for IKE phase 1 and sets it in the provided ipsecSessionBuilder. Negotiated IKE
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10160)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10144, opened 2026-08-06
feat(fastpath): add initial Fastpath support

Murt2005:
## Summary
  
  Add initial Batfish support for **Broadcom FASTPATH**, a line-oriented, IOS-like
  NOS found on merchant-silicon switches. This PR walks the vendor through the
  pipeline stages: characterize → detect → parse → extract → convert, with an initial
  modeling scope of the management plane only: hostname, DNS client, SNTP, and logging
  (syslog). All other lines parse gracefully (silently ignored where operational, or
  reported as unrecognized), so real configs are `PARTIALLY_UNRECOGNIZED` instead of
  failing.
  
  ## FASTPATH characterization
  
  - **Config format(s):** a single line-oriented running-config (one statement per
    line, space-separated, IOS-like). There is no alternate flat/hierarchical form and
    no `set`-style path form, so no flatten/preprocess pass is needed. `set prompt "..."`
    is a legacy (Quanta-era) *alternative to* `hostname`, not a separate config form.
  - **Comment / banner style:** configs open with a `!Current Configuration:` /
    `!System Software Version "x.y.z"` header; whole-line comments start with `!`. This
    makes FASTPATH resemble Cisco IOS at a glance, which drives detector ordering (below).
  - **Quoting:** the CLI accepts unquoted string values, while running-configs render
    them quoted (e.g. `hostname "sw1"`, `ip domain name "example.com"`, and even quoted
    IPs like `sntp server "10.0.0.2"`). Both quoted and unquoted forms are accepted.
  - **Inheritance / templates:** none (no apply-groups / peer-groups / profiles), so no
    inheritance-resolution pass is required before conversion.
  - **Mutation & identity quirks:** negation via `no ...` (e.g. `no ip domain lookup`,
    `no logging console`). No UUID/rename-on-edit and no numeric-ID/name aliasing.
  - **Structured show/state output:** not used by this PR (management-plane modeling
    only); relevant only to future device validation.
  - **OS versions:** a single polyglot grammar handles the observed syntax across real
    FASTPATH configurations spanning software versions ~1.2.0.x–3.4.3.x; no version
    branching needed.
  
  ## Changes
  
  - **Detect:** `ConfigurationFormat.FASTPATH`, `DeviceModel.FASTPATH_UNSPECIFIED`, and
    `VendorConfigurationFormatDetector.checkFastPath`, keyed on FASTPATH-specific tells
    (`!System Software Version` / `!Current SNTP Synchronized Time:` banner,
[...]

PR #10161, opened 2026-08-25
feat(fastpath): add tacacs-server parsing and extraction

Murt2005:
## Summary
  
  Add FastPath extraction for the `tacacs-server` command to capture TACACS+ resolver state: which servers are configured along with their port, priority, and timeout, plus the global source-interface, timeout, and whether the shared key is encrypted. This is captured in a new vendor-specific `representation` Tacacs model, feeding server hosts and the source-interface to the vendor-independent tacacsServers` and `tacacsSourceInterface` fields.
  
  ## Changes
  
  - Add lexer tokens `ENCRYPTED`, `EXIT`, `KEY`, `KEYSTRING`, `PRIORITY`, and `TACACS_SERVER`, plus an `M_Key` mode so key material is matched and dropped rather than retained. Extend the `HOST` mode-push predicate to fire after `TACACS_SERVER` as well as `LOGGING`, reusing the existing `M_HostValue` mode so hosts accept quoted strings, bare IPs, and hostnames.
  
  - Add the `s_tacacs_server` grammar rule with global `ts_*` sub-rules (`key`, `source-interface`, `timeout`, and a parsed-but-ignored `ts_keystring_null`) and per-host `tsh_*` sub-rules for the `tacacs-server host <host>` … `exit` submode (`key`, `port`, `priority`, `timeout`, and `tsh_keystring_null`).
  
  - Add two VS model classes in `representation`: `TacacsServer` (a single host with nullable `port`/`priority`/`timeout` and a `keyEncrypted` flag) and `Tacacs` (an insertion-ordered host map keyed by host text, plus the global `sourceInterface`, `timeout`, and `keyEncrypted`).
  
  - Add TACACS state to `FastpathConfiguration.java`: a `Tacacs` field with a `getTacacs()` accessor, and `convertTacacsServers`/`convertTacacsSourceInterface` populating the VI `Configuration`.
  
  - Populate handlers in `FastpathConfigurationBuilder.java`: `enterTs_host`/`exitTs_host` track the current server via `computeIfAbsent`, so repeated blocks for one host merge; the `key` handlers record only `ENCRYPTED() != null`; global and per-host `timeout` are validated against `1-30` via `toIntegerInSpace`, so out-of-range values are warned and dropped rather than stored.
  
  ## Testing
  
  - Add a `fastpath_tacacs` testconfig covering all three host forms, both key variants, and the `keystring` no-op lines.
 
  - Add `testTacacsParsesWithoutWarnings`, `testTacacsExtraction` (per-server port/priority/timeout/keyEncrypted plus globals), and `testTacacs` (VI `getTacacsServers()`/`getTacacsSourceInterface()`) to `FastpathGrammarTest`.
  
  - All tests pass locally and no reference tests are affected.
  ````

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
@@ -47,6 +47,8 @@ HOST
 :
   'host'
   {
+    // `logging host <host>` / `tacacs-server host <host>`: the host (quoted, IP, or bare word)
+    // follows in M_HostValue. `ip host ...` is consumed as a null rest-of-line in DEFAULT mode.
     if (lastTokenType() == LOGGING || lastTokenType() == TACACS_SERVER) {
       pushMode(M_HostValue);
     }
@@ -379,12 +381,12 @@ M_Remainder_NEWLINE
   F_Newline -> type ( NEWLINE ) , popMode
 ;
 
-// Host value for `sntp server` / `logging host`: a quoted string, a bare IP, or a bare word
-// (hostname). The keyword alternatives that can appear in this position instead of a host are
-// recognized explicitly so they route to their own parser rules: `status` (sntp operational
-// leakage) and `reconfigure`/`remove` (logging host maintenance). Emits exactly one token then
-// returns to DEFAULT so any trailing tokens (priority/version/port, address-type, severity) lex
-// normally.
+// Host value for `sntp server` / `logging host` / `tacacs-server host`: a quoted string, a bare IP,
+// or a bare word (hostname). The keyword alternatives that can appear in this position instead of
+// a host are recognized explicitly so they route to their own parser rules: `status` (sntp
+// operational leakage) and `reconfigure`/`remove` (logging host maintenance). Emits exactly one
+// token then returns to DEFAULT so any trailing tokens (priority/version/port, address-type,
+// severity) lex normally.
 mode M_HostValue;
 
 M_HostValue_STATUS
@@ -432,6 +434,11 @@ mode M_Key;
 M_Key_ENCRYPTED
 :
   'encrypted' -> type ( ENCRYPTED )
+;
+
+M_Key_DOUBLE_QUOTE
+:
+  '"' -> type ( DOUBLE_QUOTE ) , mode ( M_DoubleQuote )
 ;
 
 M_Key_WORD
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
@@ -37,15 +37,17 @@ EMAIL
 
 EMERGENCY: 'emergency';
 
+ENCRYPTED: 'encrypted';
+
 ERROR: 'error';
+
+EXIT: 'exit';
 
 HOST
 :
   'host'
   {
-    // `logging host <host>`: the host (quoted, IP, or bare word) follows in M_HostValue.
-    // `ip host ...` is consumed as a null rest-of-line in DEFAULT mode.
-    if (lastTokenType() == LOGGING) {
+    if (lastTokenType() == LOGGING || lastTokenType() == TACACS_SERVER) {
       pushMode(M_HostValue);
     }
   }
@@ -63,6 +65,13 @@ IP: 'ip';
 IPV4: 'ipv4';
 
 IPV6: 'ipv6';
+
+KEY
+:
+  'key' -> pushMode ( M_Key )
+;
+
+KEYSTRING: 'keystring';
 
 LIST: 'list';
 
@@ -97,6 +106,8 @@ POLL_INTERVAL: 'poll-interval';
 POLL_RETRY: 'poll-retry';
 
 PORT: 'port';
+
+PRIORITY: 'priority';
 
 PROMPT
 :
@@ -135,6 +146,8 @@ STATUS
 ;
 
 SYSLOG: 'syslog';
+
+TACACS_SERVER: 'tacacs-server';
 
 TIMEOUT: 'timeout';
 
@@ -410,6 +423,28 @@ M_HostValue_WS
 ;
 
 M_HostValue_NEWLINE
+:
+  F_Newline -> type ( NEWLINE ) , popMode
+;
+
+mode M_Key;
+
+M_Key_ENCRYPTED
+:
+  'encrypted' -> type ( ENCRYPTED )
+;
+
+M_Key_WORD
+:
+  F_Word -> type ( WORD ) , popMode
+;
+
+M_Key_WS
+:
+  F_Whitespace+ -> channel ( HIDDEN )
+;
+
+M_Key_NEWLINE
 :
   F_Newline -> type ( NEWLINE ) , popMode
 ;
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4
@@ -393,7 +393,10 @@ tsh_block
 
 tsh_key
 :
-  KEY ENCRYPTED? WORD NEWLINE
+  KEY
+  (
+    ENCRYPTED? word
+  )? NEWLINE
 ;
 
 tsh_keystring_null
@@ -423,7 +426,10 @@ ts_source_interface
 
 ts_key
 :
-  KEY ENCRYPTED? WORD NEWLINE
+  KEY
+  (
+    ENCRYPTED? word
+  )? NEWLINE
 ;
 
 ts_keystring_null
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4
@@ -21,6 +21,7 @@ statement
   | s_no
   | s_set
   | s_sntp
+  | s_tacacs_server
 ;
 
 s_hostname
@@ -360,4 +361,77 @@ logging_severity_keyword
   | INFO
   | NOTICE
   | WARNING
+;
+
+s_tacacs_server
+:
+  TACACS_SERVER
+  (
+    ts_host
+    | ts_key
+    | ts_keystring_null
+    | ts_source_interface
+    | ts_timeout
+  )
+;
+
+ts_host
+:
+  HOST host_value NEWLINE tsh_block
+;
+
+tsh_block
+:
+  (
+    tsh_key
+    | tsh_keystring_null
+    | tsh_port
+    | tsh_priority
+    | tsh_timeout
+  )* EXIT NEWLINE
+;
+
+tsh_key
+:
+  KEY ENCRYPTED? WORD NEWLINE
+;
+
+tsh_keystring_null
+:
+  KEYSTRING null_rest_of_line
+;
+
+tsh_port
+:
+  PORT port = uint16 NEWLINE
+;
+
+tsh_priority
+:
+  PRIORITY priority = uint16 NEWLINE
+;
+
+tsh_timeout
+:
+  TIMEOUT timeout = uint8 NEWLINE
+;
+
+ts_source_interface
+:
+  SOURCE_INTERFACE iface = interface_name NEWLINE
+;
+
+ts_key
+:
+  KEY ENCRYPTED? WORD NEWLINE
+;
+
+ts_keystring_null
+:
+  KEYSTRING null_rest_of_line
+;
+
+ts_timeout
+:
+  TIMEOUT timeout = uint8 NEWLINE
 ;
--- a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java
@@ -148,11 +148,13 @@ public final class FastpathConfigurationBuilder extends FastpathParserBaseListen
 
   @Override
   public void exitTsh_port(Tsh_portContext ctx) {
+    // grammar enforces range (0-65535), so no toIntegerInSpace range check needed
     _currentTacacsServer.setPort(Integer.parseInt(ctx.port.getText()));
   }
 
   @Override
   public void exitTsh_priority(Tsh_priorityContext ctx) {
+    // grammar enforces range (0-65535), so no toIntegerInSpace range check needed
     _currentTacacsServer.setPriority(Integer.parseInt(ctx.priority.getText()));
   }
 
--- a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java
@@ -43,11 +43,20 @@ import org.batfish.vendor.fastpath.grammar.FastpathParser.Sntp_source_interfaceC
 import org.batfish.vendor.fastpath.grammar.FastpathParser.Sntpc_modeContext;
 import org.batfish.vendor.fastpath.grammar.FastpathParser.Sntpc_portContext;
 import org.batfish.vendor.fastpath.grammar.FastpathParser.Ss_hostContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Ts_hostContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Ts_keyContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Ts_source_interfaceContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Ts_timeoutContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Tsh_keyContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Tsh_portContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Tsh_priorityContext;
+import org.batfish.vendor.fastpath.grammar.FastpathParser.Tsh_timeoutContext;
 import org.batfish.vendor.fastpath.grammar.FastpathParser.WordContext;
 import org.batfish.vendor.fastpath.representation.FastpathConfiguration;
 import org.batfish.vendor.fastpath.representation.LoggingBuffered;
 import org.batfish.vendor.fastpath.representation.LoggingServer;
 import org.batfish.vendor.fastpath.representation.Sntp;
+import org.batfish.vendor.fastpath.representation.TacacsServer;
 
 /** Populates a {@link FastpathConfiguration} by walking a FastPath parse tree. */
 public final class FastpathConfigurationBuilder extends FastpathParserBaseListener
@@ -63,6 +72,8 @@ public final class FastpathConfigurationBuilder extends FastpathParserBaseListen
   private static final IntegerSpace SNTP_CLIENT_PORT_RANGE =
       IntegerSpace.of(Range.closed(1, 65535));
 
+  private static final IntegerSpace TACACS_TIMEOUT_RANGE = IntegerSpace.of(Range.closed(1, 30));
+
   public FastpathConfigurationBuilder(
       FastpathCombinedParser parser,
       String text,
@@ -75,6 +86,14 @@ public final class FastpathConfigurationBuilder extends FastpathParserBaseListen
     _c = configuration;
     _silentSyntax = silentSyntax;
   }
+
+  private final @Nonnull FastpathConfiguration _c;
+  private final @Nonnull FastpathCombinedParser _parser;
+  private final @Nonnull SilentSyntaxCollection _silentSyntax;
+  private final @Nonnull String _text;
+  private final @Nonnull Warnings _w;
+
+  private @Nullable TacacsServer _currentTacacsServer;
 
   @Override
   public void exitS_hostname(S_hostnameContext ctx) {
@@ -114,6 +133,54 @@ public final class FastpathConfigurationBuilder extends FastpathParserBaseListen
   @Override
   public void exitSs_host(Ss_hostContext ctx) {
     _c.getSntp().addServer(toString(ctx.host_value()));
+  }
+
+  @Override
+  public void enterTs_host(Ts_hostContext ctx) {
+    _currentTacacsServer =
+        _c.getTacacs().getServers().computeIfAbsent(toString(ctx.host_value()), TacacsServer::new);
+  }
+
+  @Override
+  public void exitTs_host(Ts_hostContext ctx) {
+    _currentTacacsServer = null;
+  }
+
+  @Override
+  public void exitTsh_port(Tsh_portContext ctx) {
+    _currentTacacsServer.setPort(Integer.parseInt(ctx.port.getText()));
+  }
+
+  @Override
+  public void exitTsh_priority(Tsh_priorityContext ctx) {
+    _currentTacacsServer.setPriority(Integer.parseInt(ctx.priority.getText()));
+  }
+
+  @Override
+  public void exitTsh_timeout(Tsh_timeoutContext ctx) {
+    toIntegerInSpace(ctx, ctx.timeout, TACACS_TIMEOUT_RANGE, "tacacs-server host timeout")
+        .ifPresent(_currentTacacsServer::setTimeout);
+  }
+
+  @Override
+  public void exitTsh_key(Tsh_keyContext ctx) {
+    _currentTacacsServer.setKeyEncrypted(ctx.ENCRYPTED() != null);
+  }
+
+  @Override
+  public void exitTs_key(Ts_keyContext ctx) {
+    _c.getTacacs().setKeyEncrypted(ctx.ENCRYPTED() != null);
+  }
+
+  @Override
+  public void exitTs_timeout(Ts_timeoutContext ctx) {
+    toIntegerInSpace(ctx, ctx.timeout, TACACS_TIMEOUT_RANGE, "tacacs-server timeout")
+        .ifPresent(_c.getTacacs()::setTimeout);
+  }
+
+  @Override
+  public void exitTs_source_interface(Ts_source_interfaceContext ctx) {
+    _c.getTacacs().setSourceInterface(toInterfaceName(ctx.iface));
   }
 
   @Override
@@ -349,14 +416,4 @@ public final class FastpathConfigurationBuilder extends FastpathParserBaseListen
   public @Nonnull Warnings getWarnings() {
     return _w;
   }
-
-  private final @Nonnull FastpathConfiguration _c;
-
-  private final @Nonnull FastpathCombinedParser _parser;
-
-  private final @Nonnull SilentSyntaxCollection _silentSyntax;
-
-  private final @Nonnull String _text;
-
-  private final @Nonnull Warnings _w;
 }
--- a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/FastpathConfiguration.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/FastpathConfiguration.java
@@ -25,6 +25,7 @@ public final class FastpathConfiguration extends VendorConfiguration {
   private final Dns _dns;
   private final Sntp _sntp;
   private final Logging _logging;
+  private final Tacacs _tacacs;
 
   private transient Configuration _c;
 
@@ -32,6 +33,7 @@ public final class FastpathConfiguration extends VendorConfiguration {
     _dns = new Dns();
     _sntp = new Sntp();
     _logging = new Logging();
+    _tacacs = new Tacacs();
   }
 
   @Override
@@ -63,6 +65,10 @@ public final class FastpathConfiguration extends VendorConfiguration {
     return _logging;
   }
 
+  public @Nonnull Tacacs getTacacs() {
+    return _tacacs;
+  }
+
   private @Nonnull Configuration toVendorIndependentConfiguration() {
     _c = new Configuration(_hostname, ConfigurationFormat.FASTPATH);
     _c.setHumanName(_rawHostname);
@@ -81,6 +87,8 @@ public final class FastpathConfiguration extends VendorConfiguration {
     convertNtpSourceInterface();
     convertLoggingServers();
     convertLoggingSourceInterface();
+    convertTacacsServers();
+    convertTacacsSourceInterface();
 
     return _c;
   }
@@ -118,6 +126,16 @@ public final class FastpathConfiguration extends VendorConfiguration {
   private void convertLoggingSourceInterface() {
     if (_logging.getSourceInterface() != null) {
       _c.setLoggingSourceInterface(_logging.getSourceInterface());
+    }
+  }
+
+  private void convertTacacsServers() {
+    _c.setTacacsServers(ImmutableSet.copyOf(_tacacs.getServers().keySet()));
+  }
+
+  private void convertTacacsSourceInterface() {
+    if (_tacacs.getSourceInterface() != null) {
+      _c.setTacacsSourceInterface(_tacacs.getSourceInterface());
     }
   }
 
--- a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/Tacacs.java
+++ b/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/Tacacs.java
@@ -12,7 +12,7 @@ public final class Tacacs implements Serializable {
   private final @Nonnull Map<String, TacacsServer> _servers;
   private @Nullable String _sourceInterface;
   private @Nullable Integer _timeout;
-  private boolean _keyEncrypted;
+  private @Nullable Boolean _keyEncrypted;
 
   public Tacacs() {
     _servers = new LinkedHashMap<>();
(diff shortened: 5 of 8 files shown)


dhalperi (commented):
> @dhalperi+AGNT reviewed 8 files and all commit messages, and made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10161)* status: 0 of 8 files reviewed, 9 unresolved discussions (waiting on Murt2005).

___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvzJPDJn00h9bHvlw:-P-yvzJPDJn00h9bHvlx:b-fmq3zn) (no related file):*
Reviewed read-only against the FASTPATH CLI guide (§4.3 Table 4.2, §6.13.5, §6.14.1-6.14.10) linked from batfish/batfish#10144. No blocking finding: nothing in this change puts wrong data into the VI model.

Things I checked and found correct:

- `TACACS_TIMEOUT_RANGE` = 1-30 matches §6.14.4 (global) and §6.14.10 (per-host).
- `tsh_block`'s sub-keywords (`key`, `keystring`, `port`, `priority`, `timeout`) are exactly the TACACS Config mode commands in §6.14.5-6.14.8/§6.14.10, and none of them starts a top-level `statement`, so the `(...)*` loop can't swallow a following global command.
- Adding `ENCRYPTED`/`EXIT`/`KEY`/`KEYSTRING`/`PRIORITY` to DEFAULT mode doesn't regress previously-parsing input: DEFAULT has no general `WORD` rule, so those words never lexed there before, and `null_rest_of_line` is `~NEWLINE*` so it still accepts them.
- `convertTacacsServers`/`convertTacacsSourceInterface` mirror the `convertLogging*` pair exactly.
- `computeIfAbsent` keyed on unquoted host text is device-faithful per §6.14.1.

Not raised as findings, for the record: the `no tacacs-server {host|key|timeout|source-interface}` forms (§6.14.1.1, §6.14.2.1, §6.14.4.1, §6.14.9.1) aren't handled, but they shouldn't appear in `show running-config`; and `TacacsServer.getHost()` has no reader outside the map key, matching `LoggingServer.getHost()`.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 399 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-mdCCTo17Z5uMcqp:-P-z-mdCCTo17Z5uMcqq:brvcelo) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L399)):*
> ```ANTLR
> ;
> 
> tsh_keystring_null
> ```

Per the CLI guide, `keystring` takes no arguments and prompts interactively — §6.14.6 gives the syntax as the bare keyword and shows:

```
(Routing) (Config)#tacacs-server host 1.1.1.1
(Routing) (Tacacs)#keystring
Enter tacacs key:********
Re-enter tacacs key:********
```

§6.14.3 is the same for the global form. Batfish parses `show running-config` output, so on that reading neither form is reachable input, `null_rest_of_line` here and in `ts_keystring_null` never consumes anything, and the two `keystring` lines in the `fastpath_tacacs` fixture document behavior no device emits.

If instead you saw these lines in a real 3.4.3.x config, then the real syntax carries an argument the guide omits — in which case `null_rest_of_line` is silently swallowing it and it's probably worth extracting. Either way, worth saying which in the PR body.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java` line 17 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvmHwDtIdZaHV7AAA:-P-yvmHwDtIdZaHV7AAB:b-r11och) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java#L17)):*
> ```Java
>   private @Nullable Integer _priority;
>   private @Nullable Integer _timeout;
>   private boolean _keyEncrypted;
> ```

`_keyEncrypted` is a primitive `boolean`, so it reads `false` both for a host with no `key` line and for a host with a plaintext key. `testTacacsExtraction` asserts `false` for `1.1.1.1` (plaintext `key %CENSORED%`) and for `tacacs.example.com` (no `key` line at all) — two different device states, one value.

Siblings in this package use `@Nullable Boolean` for exactly this tri-state: `Logging._consoleEnabled`, `Logging._cliCommand`, `Logging._syslogEnabled`, `Dns._lookupEnabled`.

Same applies to `Tacacs._keyEncrypted` (`Tacacs.java:15`).
___
*[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java` line 151 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvsEs6cVOvlpoC72E:-P-yvsEs6cVOvlpoC72F:b-w4zfv9) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java#L151)):*
> ```Java
>   @Override
>   public void exitTsh_port(Tsh_portContext ctx) {
>     _currentTacacsServer.setPort(Integer.parseInt(ctx.port.getText()));
> ```

`port` and `priority` bypass `toIntegerInSpace`, unlike `exitLh_server` and `exitSntpc_port` in this same file. `docs/extraction/README.md:819-822` calls for routing numbers through it.

No value is currently mis-accepted: the CLI guide gives both `port` (§6.14.7) and `priority` (§6.14.8) a range of 0-65535, which is exactly what `uint16` admits, so the range check would be a no-op today. Consistency point rather than a defect.
___
*[`projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvvKT0a7-dFFv1-SI:-P-yvvKT0a7-dFFv1-SJ:b-r0bcue) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs#L7)):*
> ```
> hostname "fastpath_tacacs"
> !
> tacacs-server source-interface serviceport
> ```

`serviceport` is the only source-interface form the fixture covers, and it's the one form §6.14.9 does not document: `tacacs-server source-interface {slot/port | loopback loopback-id | vlan vlan-id}` (no `serviceport`, no `tunnel`).

Reusing the shared `interface_name` rule is right — it matches `sntp_source_interface` and `ls_source_interface` — but the documented forms (`0/1`, `loopback 0`, `vlan 10`) go untested. `testLoggingSourceInterfaceForms` is the pattern.
___
*[`projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java` line 424 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-rLT-0gLhkd-9nnH:-P-z-rLT-0gLhkd-9nnI:b1g3d2c) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java#L424)):*
> ```Java
>     // Syslog
>     assertThat(c.getLoggingServers(), equalTo(ImmutableSet.of("3.3.3.1", "3.3.3.2")));
>     assertThat(c.getLoggingSourceInterface(), equalTo("serviceport"));
> ```

`fastpath_complete_config` already ends with

```
tacacs-server host "9.9.9.9"
exit
tacacs-server timeout 10
```

Those were unrecognized lines before this PR and now extract, so this test's coverage changed with nothing pinning it. Worth a `// TACACS` block here alongside DNS/NTP/Syslog asserting `getTacacsServers()` contains `9.9.9.9`.

It covers an ordering `fastpath_tacacs` doesn't: a host block with no sub-commands followed by a global command. That fixture is also the only in-tree evidence that real configs terminate the host block with `exit`, which `tsh_block`'s mandatory `EXIT NEWLINE` depends on — so it's the assertion worth having.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvpuu2Bup1g5Rp_ue:-P-yvpuu2Bup1g5Rp_uf:b-huph3l) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L50)):*
> ```ANTLR
>   'host'
>   {
>     if (lastTokenType() == LOGGING || lastTokenType() == TACACS_SERVER) {
> ```

The two comment lines explaining this predicate were dropped rather than extended. The `ip host ...` note recorded why the predicate is needed at all, which is the part a future reader can't reconstruct from the condition.

Relatedly, the `mode M_HostValue` header comment (line 382) still says the mode is for `sntp server` / `logging host`.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 69 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvp9e-R1gtJbC0GeB:-P-yvp9e-R1gtJbC0GeC:bhce6ar) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L69)):*
> ```ANTLR
> IPV6: 'ipv6';
> 
> KEY
> ```

`KEY` pushes `M_Key` unconditionally, unlike `HOST` (line 46), `NAME`, and `SERVER`, which gate the push on `lastTokenType()`.

`key` is not only a `tacacs-server` sub-keyword. CLI guide §6.13.5 documents `radius server key {auth | acct} {ipaddr|dnsname} encrypted password` in Global Config. Once RADIUS is modeled, `radius server key acct 10.240.4.10 encrypted <pw>` will push `M_Key` after `key` and consume `acct` as the key `WORD`.

Nothing breaks today — no other modeled command reaches `key` in DEFAULT mode — but the hazard is the same one the `HOST`/`NAME`/`SERVER` predicates exist for.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 430 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvo2z4eOyc4Z05ueE:-P-yvo2z4eOyc4Z05ueF:bwkk34l) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L430)):*
> ```ANTLR
> ;
> 
> mode M_Key;
> ```

`M_Key` has no `DOUBLE_QUOTE` rule, unlike `M_Word` and `M_HostValue`, and `ts_key`/`tsh_key` require the raw `WORD` token rather than the shared `word` rule. A quoted key value therefore does not parse.

`tacacs-server key encrypted "abc def"` lexes as `TACACS_SERVER KEY ENCRYPTED` → lexer error on `"` → `WORD(abc)` (pops back to DEFAULT) → lexer error on `def` (DEFAULT has no `WORD` rule) → `DOUBLE_QUOTE NEWLINE`; `ts_key: KEY ENCRYPTED? WORD NEWLINE` (`FastpathParser.g4:424`) then fails at `DOUBLE_QUOTE`. `key "abc"` fails the same way.

CLI guide §4.3, Table 4.2: "Character strings — Use double quotation marks to identify character strings, for example, `"System Name with Spaces"`." §6.14.2 gives key-string a range of 0-128 characters, so a key containing a space or any character outside `F_WordChar` (`,` `[` `]` `?` `'` `~` `|` `\`) has to be quoted.

Separately, §6.14.2/§6.14.5 syntax is `key [key-string | encrypted key-string]` — bare `tacacs-server key` / `key` is valid, and `KEY ENCRYPTED? WORD NEWLINE` rejects it.

Mirroring `M_HostValue` (add `'"' -> type(DOUBLE_QUOTE), mode(M_DoubleQuote)`) and using `word` instead of `WORD` covers both. The value is discarded either way, so a null tail would work too.

Murt2005:
> _[a discussion](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvzJPDJn00h9bHvlw:-P-yvzJPDJn00h9bHvlx:b-fmq3zn) (no related file):_ Reviewed read-only against the FASTPATH CLI guide (§4.3 Table 4.2, §6.13.5, §6.14.1-6.14.10) linked from #10144. No blocking finding: nothing in this change puts wrong data into the VI model.
> 
> Things I checked and found correct:
> 
> * `TACACS_TIMEOUT_RANGE` = 1-30 matches §6.14.4 (global) and §6.14.10 (per-host).
> * `tsh_block`'s sub-keywords (`key`, `keystring`, `port`, `priority`, `timeout`) are exactly the TACACS Config mode commands in §6.14.5-6.14.8/§6.14.10, and none of them starts a top-level `statement`, so the `(...)*` loop can't swallow a following global command.
> * Adding `ENCRYPTED`/`EXIT`/`KEY`/`KEYSTRING`/`PRIORITY` to DEFAULT mode doesn't regress previously-parsing input: DEFAULT has no general `WORD` rule, so those words never lexed there before, and `null_rest_of_line` is `~NEWLINE*` so it still accepts them.
> * `convertTacacsServers`/`convertTacacsSourceInterface` mirror the `convertLogging*` pair exactly.
> * `computeIfAbsent` keyed on unquoted host text is device-faithful per §6.14.1.
> 
> Not raised as findings, for the record: the `no tacacs-server {host|key|timeout|source-interface}` forms (§6.14.1.1, §6.14.2.1, §6.14.4.1, §6.14.9.1) aren't handled, but they shouldn't appear in `show running-config`; and `TacacsServer.getHost()` has no reader outside the map key, matching `LoggingServer.getHost()`.
> 
> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 399 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-mdCCTo17Z5uMcqp:-P-z-mdCCTo17Z5uMcqq:brvcelo) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L399)):_
> 
> > ```antlr
> > ;
> > 
> > tsh_keystring_null
> > ```
> 
> Per the CLI guide, `keystring` takes no arguments and prompts interactively — §6.14.6 gives the syntax as the bare keyword and shows:
> 
> ```
> (Routing) (Config)#tacacs-server host 1.1.1.1
> (Routing) (Tacacs)#keystring
> Enter tacacs key:********
> Re-enter tacacs key:********
> ```
> 
> §6.14.3 is the same for the global form. Batfish parses `show running-config` output, so on that reading neither form is reachable input, `null_rest_of_line` here and in `ts_keystring_null` never consumes anything, and the two `keystring` lines in the `fastpath_tacacs` fixture document behavior no device emits.
> 
> If instead you saw these lines in a real 3.4.3.x config, then the real syntax carries an argument the guide omits — in which case `null_rest_of_line` is silently swallowing it and it's probably worth extracting. Either way, worth saying which in the PR body.
> 

keystring is an argument-less interactive command and nothing here is being swallowed by null_rest_of_line. The keystring attribute is not something i'm looking to model so it is null_rest_of_line. 

> _[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java` line 17 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvmHwDtIdZaHV7AAA:-P-yvmHwDtIdZaHV7AAB:b-r11och) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java#L17)):_
> 
> > ```java
> >   private @Nullable Integer _priority;
> >   private @Nullable Integer _timeout;
> >   private boolean _keyEncrypted;
> > ```
> 
> `_keyEncrypted` is a primitive `boolean`, so it reads `false` both for a host with no `key` line and for a host with a plaintext key. `testTacacsExtraction` asserts `false` for `1.1.1.1` (plaintext `key %CENSORED%`) and for `tacacs.example.com` (no `key` line at all) — two different device states, one value.
> 
> Siblings in this package use `@Nullable Boolean` for exactly this tri-state: `Logging._consoleEnabled`, `Logging._cliCommand`, `Logging._syslogEnabled`, `Dns._lookupEnabled`.
> 
> Same applies to `Tacacs._keyEncrypted` (`Tacacs.java:15`).
> 

Good catch, added.

> _[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java` line 151 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvsEs6cVOvlpoC72E:-P-yvsEs6cVOvlpoC72F:b-w4zfv9) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java#L151)):_
> 
> > ```java
> >   @Override
> >   public void exitTsh_port(Tsh_portContext ctx) {
> >     _currentTacacsServer.setPort(Integer.parseInt(ctx.port.getText()));
> > ```
> 
> `port` and `priority` bypass `toIntegerInSpace`, unlike `exitLh_server` and `exitSntpc_port` in this same file. `docs/extraction/README.md:819-822` calls for routing numbers through it.
> 
> No value is currently mis-accepted: the CLI guide gives both `port` (§6.14.7) and `priority` (§6.14.8) a range of 0-65535, which is exactly what `uint16` admits, so the range check would be a no-op today. Consistency point rather than a defect.
> 

I don't think adding a range check here is worth it since the grammar is already enforcing the range, added a one line comment to the exit methods for port and priority explaining this.

> _[`projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvvKT0a7-dFFv1-SI:-P-yvvKT0a7-dFFv1-SJ:b-r0bcue) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs#L7)):_
> 
> > ```
> > hostname "fastpath_tacacs"
> > !
> > tacacs-server source-interface serviceport
> > ```
> 
> `serviceport` is the only source-interface form the fixture covers, and it's the one form §6.14.9 does not document: `tacacs-server source-interface {slot/port | loopback loopback-id | vlan vlan-id}` (no `serviceport`, no `tunnel`).
> 
> Reusing the shared `interface_name` rule is right — it matches `sntp_source_interface` and `ls_source_interface` — but the documented forms (`0/1`, `loopback 0`, `vlan 10`) go untested. `testLoggingSourceInterfaceForms` is the pattern.
> 

Added a test for this.

> _[`projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java` line 424 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-rLT-0gLhkd-9nnH:-P-z-rLT-0gLhkd-9nnI:b1g3d2c) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java#L424)):_
> 
> > ```java
> >     // Syslog
> >     assertThat(c.getLoggingServers(), equalTo(ImmutableSet.of("3.3.3.1", "3.3.3.2")));
> >     assertThat(c.getLoggingSourceInterface(), equalTo("serviceport"));
> > ```
> 
> `fastpath_complete_config` already ends with
> 
> ```
> tacacs-server host "9.9.9.9"
> exit
> tacacs-server timeout 10
> ```
> 
> Those were unrecognized lines before this PR and now extract, so this test's coverage changed with nothing pinning it. Worth a `// TACACS` block here alongside DNS/NTP/Syslog asserting `getTacacsServers()` contains `9.9.9.9`.
> 
> It covers an ordering `fastpath_tacacs` doesn't: a host block with no sub-commands followed by a global command. That fixture is also the only in-tree evidence that real configs terminate the host block with `exit`, which `tsh_block`'s mandatory `EXIT NEWLINE` depends on — so it's the assertion worth having.
> 

Extended testFastPathConfig() to cover this.

> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvpuu2Bup1g5Rp_ue:-P-yvpuu2Bup1g5Rp_uf:b-huph3l) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L50)):_
> 
> > ```antlr
> >   'host'
> >   {
> >     if (lastTokenType() == LOGGING || lastTokenType() == TACACS_SERVER) {
> > ```
> 
> The two comment lines explaining this predicate were dropped rather than extended. The `ip host ...` note recorded why the predicate is needed at all, which is the part a future reader can't reconstruct from the condition.
> 
> Relatedly, the `mode M_HostValue` header comment (line 382) still says the mode is for `sntp server` / `logging host`.
> 

Added back the comments, and updated M_HostValue comment.

> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 69 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvp9e-R1gtJbC0GeB:-P-yvp9e-R1gtJbC0GeC:bhce6ar) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L69)):_
> 
> > ```antlr
> > IPV6: 'ipv6';
> > 
> > KEY
> > ```
> 
> `KEY` pushes `M_Key` unconditionally, unlike `HOST` (line 46), `NAME`, and `SERVER`, which gate the push on `lastTokenType()`.
> 
> `key` is not only a `tacacs-server` sub-keyword. CLI guide §6.13.5 documents `radius server key {auth | acct} {ipaddr|dnsname} encrypted password` in Global Config. Once RADIUS is modeled, `radius server key acct 10.240.4.10 encrypted <pw>` will push `M_Key` after `key` and consume `acct` as the key `WORD`.
> 
> Nothing breaks today — no other modeled command reaches `key` in DEFAULT mode — but the hazard is the same one the `HOST`/`NAME`/`SERVER` predicates exist for.
> 

Not really understanding the feedback here. KEY does push M_Key unconditionally here and there are other uses of key like in `crypto key`, `radius server key`, 'lacp admin key', and 'lacp actor admin key' but these aren't modeled yet.  If or when these commands are modeled then the key token should be changed for when it pushes M_Key.

> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 430 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvo2z4eOyc4Z05ueE:-P-yvo2z4eOyc4Z05ueF:bwkk34l) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L430)):_
> 
> > ```antlr
> > ;
> > 
> > mode M_Key;
> > ```
> 
> `M_Key` has no `DOUBLE_QUOTE` rule, unlike `M_Word` and `M_HostValue`, and `ts_key`/`tsh_key` require the raw `WORD` token rather than the shared `word` rule. A quoted key value therefore does not parse.
> 
> `tacacs-server key encrypted "abc def"` lexes as `TACACS_SERVER KEY ENCRYPTED` → lexer error on `"` → `WORD(abc)` (pops back to DEFAULT) → lexer error on `def` (DEFAULT has no `WORD` rule) → `DOUBLE_QUOTE NEWLINE`; `ts_key: KEY ENCRYPTED? WORD NEWLINE` (`FastpathParser.g4:424`) then fails at `DOUBLE_QUOTE`. `key "abc"` fails the same way.
> 
> CLI guide §4.3, Table 4.2: "Character strings — Use double quotation marks to identify character strings, for example, `"System Name with Spaces"`." §6.14.2 gives key-string a range of 0-128 characters, so a key containing a space or any character outside `F_WordChar` (`,` `[` `]` `?` `'` `~` `|` `\`) has to be quoted.
> 
> Separately, §6.14.2/§6.14.5 syntax is `key [key-string | encrypted key-string]` — bare `tacacs-server key` / `key` is valid, and `KEY ENCRYPTED? WORD NEWLINE` rejects it.
> 
> Mirroring `M_HostValue` (add `'"' -> type(DOUBLE_QUOTE), mode(M_DoubleQuote)`) and using `word` instead of `WORD` covers both. The value is discarded either way, so a null tail would work too.

Added M_Key_DOUBLE_QUOTE and used word instead of WORD in the ts_key/tsh_key rules. Also added a quoted key in the testconfig and updated tests to cover this. Kept the value discarded. Also changed it so that a bare key is accepted as well.

dhalperi (commented):
> @dhalperi+AGNT reviewed 7 files and all commit messages, made 8 comments, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10161)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on Murt2005).

___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 50 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvpuu2Bup1g5Rp_ue:-P-zs-fm71_aVyT4scH8:bam0shn) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L50)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Added back the comments, and updated M_HostValue comment.
</blockquote></details>

Verified at r2 — the `ip host` note is back and the `M_HostValue` header now lists all three commands.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 69 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvp9e-R1gtJbC0GeB:-P-zsL7N2PqFSgceB1YP:b-60s3ju) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L69)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Not really understanding the feedback here. KEY does push M_Key unconditionally here and there are other uses of key like in `crypto key`, `radius server key`, 'lacp admin key', and 'lacp actor admin key' but these aren't modeled yet.  If or when these commands are modeled then the key token should be changed for when it pushes M_Key.
</blockquote></details>

Fair enough — deferring is reasonable, and I'll close this. Restating the mechanism just so it's on the record for whoever adds those commands.

The concern isn't that `key` appears in unmodeled commands, it's that an unconditional `pushMode` makes the eventual breakage silent and non-local: someone adds `radius server key`, writes a correct `rs_key` parser rule, and it fails on real input because `M_Key` already consumed `auth`/`acct` as the key `WORD` several hundred lines away in the lexer. The `HOST`/`NAME`/`SERVER` predicates exist so the constraint sits next to the token that creates it.

For whenever you get there: `key` needs `M_Key` in exactly two positions today — after `tacacs-server`, and line-initial in the TACACS submode — so

```antlr
if (lastTokenType() == TACACS_SERVER || lastTokenType() == NEWLINE) {
  pushMode(M_Key);
}
```

separates all four of your examples, since `crypto key`, `radius server key`, `lacp admin key`, and `lacp actor admin key` each have a preceding token on the line (`CRYPTO`, `SERVER`, `ADMIN`). Gating on `NEWLINE` is weaker than the `HOST`/`SERVER` predicates though — a line-initial `key` in some other submode would still push — so it narrows the hazard rather than removing it.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4` line 430 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvo2z4eOyc4Z05ueE:-P-zs-AC21sKWC37AByw:b-dxf3tz) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4#L430)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Added M_Key_DOUBLE_QUOTE and used word instead of WORD in the ts_key/tsh_key rules. Also added a quoted key in the testconfig and updated tests to cover this. Kept the value discarded. Also changed it so that a bare key is accepted as well.
</blockquote></details>

Verified at r2. `M_Key_DOUBLE_QUOTE` uses `mode(M_DoubleQuote)` rather than `pushMode`, so the closing quote's `popMode` lands back in DEFAULT — same stack discipline as `M_Word_DOUBLE_QUOTE` and `M_HostValue_DOUBLE_QUOTE`. `key "%CENSORED% with spaces"` and the bare `key` on `4.4.4.4` cover both new paths, and `key encrypted` with no value is still rejected, which matches the `[key-string | encrypted key-string]` alternation.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 399 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-mdCCTo17Z5uMcqp:-P-zs4JB94IsB17pRsid:b-nxi68w) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L399)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

keystring is an argument-less interactive command and nothing here is being swallowed by null_rest_of_line. The keystring attribute is not something i'm looking to model so it is null_rest_of_line. 
</blockquote></details>

Understood — not modeling it is the right call, and a `_null` rule for a command that can't reach a saved config costs nothing. Leaving as-is.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java` line 151 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvsEs6cVOvlpoC72E:-P-zs3lL8BOKapzDBDun:b-xbdlfj) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/grammar/FastpathConfigurationBuilder.java#L151)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

I don't think adding a range check here is worth it since the grammar is already enforcing the range, added a one line comment to the exit methods for port and priority explaining this.
</blockquote></details>

Agreed — `uint16` and the documented 0-65535 range coincide exactly, so the check would be dead code. The comments record why, which is what was missing. Thanks.
___
*[`projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java` line 17 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvmHwDtIdZaHV7AAA:-P-zs0BKDXbyT-zZIGni:bhstht9) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/main/java/org/batfish/vendor/fastpath/representation/TacacsServer.java#L17)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Good catch, added.
</blockquote></details>

Verified at r2. The five fixture hosts now cover all three states distinctly: `tacacs.example.com` null (no `key` line), `1.1.1.1`/`2.2.2.2`/`4.4.4.4` false, `3.3.3.3` true.
___
*[`projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java` line 424 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-z-rLT-0gLhkd-9nnH:-P-zs2mE3OomsM1wY5Vb:buqo938) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/java/org/batfish/vendor/fastpath/grammar/FastpathGrammarTest.java#L424)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Extended testFastPathConfig() to cover this.
</blockquote></details>

Verified at r2 — the `// TACACS` block asserts `getTacacsServers()` is `{9.9.9.9}`.
___
*[`projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs` line 7 at r1](https://reviewable.io/reviews/batfish/batfish/10161#-P-yvvKT0a7-dFFv1-SI:-P-zs0pVA5i0BoRiJ97l:b3yo556) ([raw file](https://github.com/batfish/batfish/blob/dcbe3d469e92a84a6a10566128940550d8a7aa1a/projects/batfish/src/test/resources/org/batfish/vendor/fastpath/grammar/testconfigs/fastpath_tacacs#L7)):*
<details><summary><i>Previously, Murt2005 wrote…</i></summary><blockquote>

Added a test for this.
</blockquote></details>

Verified at r2 — `testTacacsSourceInterfaceForms` covers `loopback 0`, `0/1`, `1/0/1`, `tunnel 5`, and `vlan 10`, matching `testLoggingSourceInterfaceForms` including the undocumented-for-this-command `tunnel` form.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10161)* status: all files reviewed, 1 unresolved discussion (waiting on Murt2005).

dhalperi (commented):
> @dhalperi resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10161)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10162, opened 2026-08-26
PaloAlto: report an absent ipsec dh-group, and only warn on a genuinely ambiguous IKE version

dhalperi:
An ipsec-crypto-profile with no dh-group left the phase 2 policy's pfs
key groups empty, which negotiateIpsecP2 treats the same as no-pfs: the
tunnel cannot negotiate. Only the explicit no-pfs case warned, so the
absent case was silent. Both now report, and PAN-OS documents no default
dh group to fall back on. The IPSEC-AH profile in the fixture is in this
state, so the tunnel it backs never formed an edge.

hasAmbiguousIkeVersion fired whenever a gateway named a profile for
both IKE versions with no protocol version, including when both named
the same profile. Naming one profile for both versions is common and
leaves nothing to assume, so it no longer warns.

CryptoProfile.equals and hashCode omitted the protocol and no-pfs
fields. That is load-bearing: testIpsecCryptoProfiles compares parsed
profiles against expected ones and was passing only because the
protocol was ignored.

Also covers the IKE_CRYPTO_PROFILE_OR_NONE and
IPSEC_CRYPTO_PROFILE_OR_NONE paths, which existed to keep references to
the profiles PAN-OS ships predefined from being reported undefined but
had no test.

Left for follow-up: the post-quantum dh-group values and the
ike-crypto-profile ake subtree are parse errors on valid 11.1 config.
Both need keyword sets confirmed against a device rather than guessed.

For batfish/batfish#10156.

----

Prompt:
```
Decide which of the open follow-ups from the batfish/batfish#10156 review
to fix and send immediately, now that the PR has been squash-merged, and
implement them.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/palo_alto/CryptoProfile.java
@@ -79,7 +79,9 @@ public final class CryptoProfile implements Serializable {
         && Objects.equals(_dhGroups, rhs._dhGroups)
         && Objects.equals(_encryptionAlgorithms, rhs._encryptionAlgorithms)
         && Objects.equals(_hashAlgorithm, rhs._hashAlgorithm)
-        && Objects.equals(_lifetimeSeconds, rhs._lifetimeSeconds);
+        && Objects.equals(_lifetimeSeconds, rhs._lifetimeSeconds)
+        && Objects.equals(_protocol, rhs._protocol)
+        && _noPfs == rhs._noPfs;
   }
 
   public boolean getNoPfs() {
@@ -143,7 +145,9 @@ public final class CryptoProfile implements Serializable {
         _dhGroups,
         _encryptionAlgorithms,
         _hashAlgorithm,
-        _lifetimeSeconds);
+        _lifetimeSeconds,
+        _protocol,
+        _noPfs);
   }
 
   public void setAuthAlgorithm(@Nullable IpsecAuthenticationAlgorithm authAlgorithm) {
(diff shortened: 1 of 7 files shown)


dhalperi:
@gh-ajinks would you review this? It is the follow-up to the open items from the batfish/batfish#10156 review, now that that has merged.

Two are behaviour fixes to warnings that land on real snapshots as of this morning:

- An ipsec-crypto-profile with no `dh-group` leaves the phase 2 policy's pfs key groups empty, which `negotiateIpsecP2` treats exactly as `no-pfs` does — the tunnel cannot negotiate. Only the explicit `no-pfs` case warned. `IPSEC-AH` in your fixture is in that state, so `TUN-AH` never formed an edge.
- `hasAmbiguousIkeVersion()` fired even when both IKE versions named the same profile, where the preference order decides nothing. `GW-SAMEPROF` was already there to catch it.

Plus `CryptoProfile.equals`/`hashCode` picking up `_protocol` and `_noPfs`, and a test for the `*_OR_NONE` paths you added in r3 — those had no coverage, and the false positive they prevent was unguarded.

Worth knowing: including `_protocol` in `equals` broke `testIpsecCryptoProfiles`, which compares parsed profiles against expected ones and was passing only because the field was ignored. I updated the expectations rather than backing the change out.

I deliberately left `snicp_ike_ake_null` and the post-quantum `dh-group` values alone. I could not confirm those keyword sets against a device or the CLI reference, and you had the 11.1.13-h3 device in hand for the r3 enumeration — if you can check them, that is the last piece.

gh-ajinks:
Checked this against my configs — no ipsec-crypto-profile of mine is missing a `dh-group` or uses `no-pfs`, and no gateway of mine names a profile on both IKE versions. So neither case applies to me and nothing here breaks my use case.

Changes read correct to me either way. LGTM.

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10162)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10163, opened 2026-08-26
Junos: don't flag routing-instance default as an undefined reference

dhalperi:
Junos reserves the word default to mean the master routing instance
where a routing instance is referenced, e.g. firewall filter ... then
routing-instance default, and forbids it as the name of a configured
instance. Skip reference tracking for that name at all eight routing
instance reference sites.

The conversion path already produced the right VRF, since the VI default
VRF is itself named "default"; a test now pins that.

----

Prompt:
```
We're debugging an issue where we're getting undefined references for
Junos routing-instance default. Another model claims master and default
are aliases for the same default routing instance: master is the
canonical name reported by `show route instance`, default is used in
configuration statements such as `routing-instance default` in a
firewall filter, and Junos prevents creating a user routing instance
literally named default. Check the cited Juniper documentation and make
up your own mind.

Follow-ups: confirmed against a real Junos config whose only undefined
routing-instance reference is `then routing-instance default` on an
interface owned by a virtual-router instance. Decided the right fix is
to not add a reference at all when the value is specifically default,
rather than remapping it to master.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10163)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10165, opened 2026-08-26
feat(fastpath): add AAA support

Murt2005:
## Summary

Add FastPath extraction for the `aaa` command to capture the device's AAA method lists: `aaa authentication {login|enable}`, `aaa authorization {commands|exec}`, and `aaa accounting {exec|commands|dot1x}`. Each is recorded with its list name and the ordered methods the device will try. This is captured in a new vendor-specific `representation` `Aaa` model.

## Changes

- **Lexer:** Add lexer tokens `AAA`, `ACCOUNTING`, `AUTHENTICATION`, `AUTHORIZATION`, `COMMANDS`, `DENY`, `DOT1X`, `ENABLE`, `EXEC`, `IAS_USER`, `LINE`, `LOCAL`, `LOGIN`, `NONE`, `PASSWORD`, `RADIUS`, `SESSION_ID`, `START_STOP`, `STOP_ONLY`, `TACACS`, and `USERNAME`, plus a virtual `DEFAULT` token alongside the existing `WORD`/`QUOTED_TEXT`/`REMAINDER`.

- **Lexer:** Add an `M_AaaListName` mode so a list name accepts the `default` keyword, a quoted string, or a bare word. It is entered only from the default mode via `lastTokenType()` predicates on `COMMANDS`, `DOT1X`, `ENABLE`, `EXEC`, and `LOGIN`, so those generic keywords only push a mode in AAA position.

- **Lexer:** Add `PASSWORD` and `SESSION_ID` mode-pushes to `M_Remainder` so password and session-id operands are matched and dropped rather than retained, and `USERNAME` to `M_Word`.

- **Grammar:** Add the `s_aaa` grammar rule with `aaa_authentication`, `aaa_authorization`, and `aaa_accounting` families, each terminating in a leaf that consumes its own `NEWLINE`. Method and record-type sets are per family: authentication accepts `deny`/`enable`/`line`/`local`/`none`/`radius`/`tacacs`, authorization accepts `local`/`none`/`radius`/`tacacs` (`local` is shared by both authorization types even though the device rejects it for `commands` specifically), and accounting accepts `radius`/`tacacs`. `aaa_dot1x_accounting_record` is split from `aaa_accounting_record` because dot1x supports only `start-stop` and `none`, and `aaa_accounting_dot1x` hardcodes the `default` list and an optional `radius`, the only list and method it supports.

- **Grammar:** Add `aaa_ias_user` for the `aaa ias-user username <user>` … `exit` submode: an `aaaiu_block` of optional `aaaiu_password_null` lines terminated by `exit`, so the `password … encrypted` body is consumed by the grammar rather than surfacing as an unrecognized line whose warning text would repeat the password. Add a parsed-but-ignored `aaa_session_id_null`.

- **VS:** Add VS model classes in `representation`: `AaaMethod`, `Authentication`/`AuthenticationType`, `Authorization`/`AuthorizationType`, `Accounting`/`AccountingType` (carrying a non-null `RecordType`), and `Aaa`, which holds each family in an insertion-ordered map keyed by type and then list name, since the same name may be reused across types.

- **VS:** Add AAA state to `FastpathConfiguration.java`: an `Aaa` field with a `getAaa()` accessor.

- **Extraction:** Populate handlers in `FastpathConfigurationBuilder.java`. The list classes are immutable and `Aaa.define*` replaces any existing list of the same type and name, matching the device, where re-issuing a command for an existing list overwrites its record type and methods rather than appending. Method-keyword conversion asserts on the final alternative so adding a grammar alternative without extraction fails loudly. The `default` keyword is recorded as the literal list name `"default"` via a shared constant.

## Testing

- Add a `fastpath_aaa` testconfig covering all three families, both authentication types, both authorization types, all three accounting types, every list-name form, and both ias-user block shapes.

- Add `testAaaParsesWithoutWarnings`, `testAaaExtraction`, `testAaaListNameFormsExtraction`, `testAaaListRedefinitionReplaces`, `testAaaIasUserBlockIsSilent`, and `testCompleteConfigAaaExtraction` to `FastpathGrammarTest`.

- All tests pass locally and no reference tests are affected.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathLexer.g4
@@ -5,12 +5,21 @@ options {
 }
 
 tokens {
+  DEFAULT,
   QUOTED_TEXT,
   REMAINDER,
   WORD
 }
 
+AAA: 'aaa';
+
+ACCOUNTING: 'accounting';
+
 ALERT: 'alert';
+
+AUTHENTICATION: 'authentication';
+
+AUTHORIZATION: 'authorization';
 
 BROADCAST: 'broadcast';
 
@@ -20,15 +29,37 @@ CLI_COMMAND: 'cli-command';
 
 CLIENT: 'client';
 
+COMMANDS
+:
+  'commands'
+  {
+    if (lastTokenType() == AUTHORIZATION || lastTokenType() == ACCOUNTING) {
+      pushMode(M_AaaListName);
+    }
+  }
+;
+
 CONSOLE: 'console';
 
 CRITICAL: 'critical';
 
 DEBUG: 'debug';
 
+DENY: 'deny';
+
 DNS: 'dns';
 
 DOMAIN: 'domain';
+
+DOT1X
+:
+  'dot1x'
+  {
+    if (lastTokenType() == ACCOUNTING) {
+      pushMode(M_AaaListName);
+    }
+  }
+;
 
 EMAIL
 :
@@ -37,9 +68,29 @@ EMAIL
 
 EMERGENCY: 'emergency';
 
+ENABLE
+:
+  'enable'
+  {
+    if (lastTokenType() == AUTHENTICATION) {
+      pushMode(M_AaaListName);
+    }
+  }
+;
+
 ENCRYPTED: 'encrypted';
 
 ERROR: 'error';
+
+EXEC
+:
+  'exec'
+  {
+    if (lastTokenType() == AUTHORIZATION || lastTokenType() == ACCOUNTING) {
+      pushMode(M_AaaListName);
+    }
+  }
+;
 
 EXIT: 'exit';
 
@@ -60,6 +111,8 @@ HOSTNAME
   'hostname' -> pushMode ( M_Word )
 ;
 
+IAS_USER: 'ias-user';
+
 INFO: 'info';
 
 IP: 'ip';
@@ -75,9 +128,23 @@ KEY
 
 KEYSTRING: 'keystring';
 
+LINE: 'line';
+
 LIST: 'list';
 
+LOCAL: 'local';
+
 LOGGING: 'logging';
+
+LOGIN
+:
+  'login'
+  {
+    if (lastTokenType() == AUTHENTICATION) {
+      pushMode(M_AaaListName);
+    }
+  }
+;
 
 LOOKUP: 'lookup';
 
@@ -99,7 +166,14 @@ NAME
 
 NO: 'no';
 
+NONE: 'none';
+
 NOTICE: 'notice';
+
+PASSWORD
+:
+  'password' -> pushMode ( M_Remainder )
+;
 
 PERSISTENT: 'persistent';
 
@@ -115,6 +189,8 @@ PROMPT
 :
   'prompt' -> pushMode ( M_Word )
 ;
+
+RADIUS: 'radius';
 
 RECONFIGURE: 'reconfigure';
 
@@ -136,18 +212,29 @@ SERVER
 
 SERVICEPORT: 'serviceport';
 
+SESSION_ID
+:
+  'session-id' -> pushMode ( M_Remainder )
+;
+
 SET: 'set';
 
 SNTP: 'sntp';
 
 SOURCE_INTERFACE: 'source-interface';
 
+START_STOP: 'start-stop';
+
 STATUS
 :
   'status' -> pushMode ( M_Remainder )
 ;
 
+STOP_ONLY: 'stop-only';
+
 SYSLOG: 'syslog';
+
+TACACS: 'tacacs';
 
 TACACS_SERVER: 'tacacs-server';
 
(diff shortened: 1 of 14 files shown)


dhalperi (commented):
> @dhalperi+AGNT made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10165)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on Murt2005).

___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 458 at r1](https://reviewable.io/reviews/batfish/batfish/10165#-P0-T2ZE7bH5YiTjpy-J:-P0-T2ZE7bH5YiTjpy-K:b-drqq4o) ([raw file](https://github.com/batfish/batfish/blob/7993f007cf6c9709a2f83accb4dc0bcf1bad3a89/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L458)):*
> ```ANTLR
> ;
> 
> aaa_authentication
> ```

`aaa authentication dot1x default {[ias]|method1 [method2 [method3]]}` is the third documented Global Config form and isn't handled here, so it stays an unrecognized line and `ias` is missing from `AaaMethod`. Add it, or note in the commit message that dot1x authentication is deferred.
___
*[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 477 at r1](https://reviewable.io/reviews/batfish/batfish/10165#-P0-T0lb3-Law-H3Xt29:-P0-T0lb3-Law-H3Xt2A:b-fukxyy) ([raw file](https://github.com/batfish/batfish/blob/7993f007cf6c9709a2f83accb4dc0bcf1bad3a89/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L477)):*
> ```ANTLR
> ;
> 
> aaa_authentication_method
> ```

The FASTPATH CLI reference lists these per type rather than as one set: `login` takes `enable|local|none|radius|tacacs`, `enable` takes `deny|enable|line|none|radius|tacacs`. Split them, or keep one set with the same kind of note `aaa_authorization_method` carries; either way `fastpath_aaa:9` (`aaa authentication login default line none`) should use a method documented for `login`.

Murt2005:
> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 458 at r1](https://reviewable.io/reviews/batfish/batfish/10165#-P0-T2ZE7bH5YiTjpy-J:-P0-T2ZE7bH5YiTjpy-K:b-drqq4o) ([raw file](https://github.com/batfish/batfish/blob/7993f007cf6c9709a2f83accb4dc0bcf1bad3a89/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L458)):_
> 
> > ```antlr
> > ;
> > 
> > aaa_authentication
> > ```
> 
> `aaa authentication dot1x default {[ias]|method1 [method2 [method3]]}` is the third documented Global Config form and isn't handled here, so it stays an unrecognized line and `ias` is missing from `AaaMethod`. Add it, or note in the commit message that dot1x authentication is deferred.
> 

Added new ```aaa_authentication_dot1x``` rule with its own ```aaa_authentication_dot1x_method``` set (ias|local|none|radius) as well as IAS on AaaMethod and DOT1X on AuthenticationType. Only the default list is accepted which is the same as aaa_accounting_dot1x.

I modeled the operands as a flat method+ rather than following the documented ```{[ias]|[method1 [method2 [method3]]]}``` because the reference's own example ```aaa authentication dot1x default ias local radius none``` is ias plus three methods which that notation disallows. The flat list accepts both documented examples and both are covered by tests.

> _[`projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4` line 477 at r1](https://reviewable.io/reviews/batfish/batfish/10165#-P0-T0lb3-Law-H3Xt29:-P0-T0lb3-Law-H3Xt2A:b-fukxyy) ([raw file](https://github.com/batfish/batfish/blob/7993f007cf6c9709a2f83accb4dc0bcf1bad3a89/projects/batfish/src/main/antlr4/org/batfish/vendor/fastpath/grammar/FastpathParser.g4#L477)):_
> 
> > ```antlr
> > ;
> > 
> > aaa_authentication_method
> > ```
> 
> The FASTPATH CLI reference lists these per type rather than as one set: `login` takes `enable|local|none|radius|tacacs`, `enable` takes `deny|enable|line|none|radius|tacacs`. Split them, or keep one set with the same kind of note `aaa_authorization_method` carries; either way `fastpath_aaa:9` (`aaa authentication login default line none`) should use a method documented for `login`.

Split into ```aaa_authentication_login_method``` and ```aaa_authentication_enable_method```. On the specific sets however, line is documented for both. The aaa authentication login parameter table lists "At least one from the following: enable, line, local, none, radius, tacacs", so the actual difference is just that login takes local but not deny, and enable takes deny but not local. That makes fastpath_aaa:9 (aaa authentication login default line none) valid as written, so I left it alone.

dhalperi (commented):
> @dhalperi reviewed 14 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10165)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Murt2005).

PR #10166, opened 2026-08-27
PaloAlto: drop log-forwarding payloads before they break config nesting

gh-ajinks:
An HTTP log-forwarding profile stores its webhook body as a quoted string, and PAN-OS emits that string with its inner double quotes **unescaped**:

```
payload "{\n  "attachments": [\n    {\n      "fallback": "$device_name reports $severity",\n ...
```

`F_QuotedString` is `'"' ~'"'* '"'`, so the token ends at the first inner quote — after `"{\n  "`. Everything past that lexes as ordinary tokens, including the body's `{` and `}`, and those are structural. Config nesting desynchronises for the remainder of the device.

The result is a device that converts with **no interfaces at all**, and nothing in the output identifies the cause. On a snapshot containing four PAN-OS firewalls alongside AWS and Arista devices, established BGP sessions drop from 200 to 182 and 16 sessions disappear from the model entirely.

### The fix

Consume the payload statement, terminator included, and skip it. Nothing is lost: `s_log_settings` offers only `config`, `profiles`, `syslog`, `system` and `userid`, and no `HTTP` token exists anywhere in the PAN-OS grammar, so this subtree is already unmodeled.

It has to be done in the lexer. A `_null` rule in the main grammar, or skipping the subtree in `PaloAltoNestedFlattener`, would both run after `PaloAltoNestedParser` has already built a mis-nested tree.

I considered the more general shape — treating a quoted value as running to the `"` that precedes its `;` — and decided against it. A bracketed list (`[ "a" "b" ]`) has no terminator, so that rule would consume across list elements. The targeted form has an unambiguous terminator, and the real payloads contain no `";` before their own.

### Tests

Two testconfigs that differ only in the payload block. `log-profile-control` establishes that the fixture is sound; `log-profile-payload` fails during parsing without this change and passes with it.

Green locally: the PAN-OS suite, both parsing corpora, checkstyle and `//tools/format:format.check`.

### Provenance

Observed on PA-1420s running PAN-OS 11.1.13-h3 and 11.1.13-h5. In the first 2000 characters of one payload: 47 bare `"`, zero `\"`.

### DIFF
--- a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto_nested/PaloAltoNestedLexer.g4
+++ b/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto_nested/PaloAltoNestedLexer.g4
@@ -54,6 +54,23 @@ SHOW_CONFIG_LINE
   F_NonNewlineChar* 'show' F_WhitespaceChar+ 'config' F_NonNewlineChar* F_NewlineChar+ -> channel(HIDDEN)
 ;
 
+// An HTTP log-forwarding profile stores its webhook body as a quoted string, and PAN-OS
+// emits that string with its inner double quotes unescaped. F_QuotedString stops at the
+// first of those, after which the body's own braces lex as OPEN_BRACE/CLOSE_BRACE and
+// desynchronise config nesting for the remainder of the device -- typically leaving it with
+// no interfaces at all, and no warning saying why.
+//
+// The body is not modeled: s_log_settings offers no http alternative, and no HTTP token
+// exists in the PAN-OS grammar. So consume the whole statement, terminator included, and
+// drop it. This must happen here rather than in PaloAltoNestedFlattener or the main
+// grammar, both of which run after the tree has already been mis-nested.
+LOG_PROFILE_PAYLOAD
+:
+   'payload'
+   {lastTokenType() == -1 || lastTokenType() == NEWLINE || lastTokenType() == SHOW_CONFIG_LINE}?
+   F_WhitespaceChar+ '"' .*? '";' -> skip
+;
+
 WORD
 :
    F_QuotedString
--- a/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java
+++ b/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java
@@ -436,6 +436,20 @@ public final class PaloAltoGrammarTest {
   }
 
   @Test
+  public void testLogProfilePayloadDoesNotBreakNesting() {
+    // PAN-OS emits an HTTP log-forwarding payload as a quoted string whose inner double
+    // quotes are unescaped. If the lexer stops at the first of those, the body's braces
+    // become structural and every later statement is mis-nested. The payload sits before
+    // the interface in this config, so the address proves nesting survived it.
+    Configuration c = parseConfig("log-profile-payload");
+
+    assertThat(c.getAllInterfaces(), hasKey("ethernet1/1"));
+    assertThat(
+        c.getAllInterfaces().get("ethernet1/1").getConcreteAddress(),
+        equalTo(ConcreteInterfaceAddress.parse("10.0.1.1/24")));
+  }
+
+  @Test
   public void testInterfaceSdwanAddress() {
     // On an SD-WAN interface the gateway is nested under the address, so flattening emits
     // one line carrying both and the address never appears on its own. ethernet1/1 is the
--- a/projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/log-profile-payload
+++ b/projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/log-profile-payload
@@ -1,0 +1,50 @@
+config {
+  shared {
+    log-settings {
+      http {
+        webhook-profile {
+          format {
+            system {
+              payload "{
+  "attachments": [
+    {
+      "fallback": "$device_name reports $severity",
+      "color": "danger",
+      "text": "*Time:* $time_generated"
+    }
+  ]
+}";
+            }
+          }
+        }
+      }
+    }
+  }
+  devices {
+    localhost.localdomain {
+      deviceconfig {
+        system {
+          hostname log-profile-payload;
+        }
+      }
+      network {
+        interface {
+          ethernet {
+            ethernet1/1 {
+              layer3 {
+                ip {
+                  10.0.1.1/24;
+                }
+              }
+            }
+          }
+        }
+        virtual-router {
+          default {
+            interface [ ethernet1/1 ];
+          }
+        }
+      }
+    }
+  }
+}


dhalperi:
Freaking broken PAN CLI.

Did you try the xml -> conversion path that is recommended? I imagine it might still have the same problem, but ... https://batfish.readthedocs.io/en/latest/formats.html#from-panorama-preferred

gh-ajinks:
Not Panorama-managed here, so I can't use that exact path — these are standalone firewalls. But you're right about the underlying idea, and testing it turned up something you may want to know.

**The payload is safe in set format.** Whichever way I render it, the value comes out **single**-quoted:

```
set shared log-settings http <profile> format system payload '{
  "attachments": [ ... "fallback": "..." ... ]
}'
```

22 bare `"` inside, zero escaped — but the enclosing quotes are `'`, so nothing terminates early and the braces never become structural. Same story via the XML API, where it's just a `<payload>` element and quoting is moot. So the bug genuinely cannot arise on the conversion path.

**But the documented standalone recipe doesn't produce set format on 11.1.13-h5.** Running exactly what the docs prescribe:

```
set cli config-output-format set
set cli pager off
show config merged
```

returns brace format — 365 KB, zero `set ` lines. Same with `config-output-format xml`. The setting is being applied; it just only affects `show` inside **configure** mode, where I do get set lines. If that holds on other releases, the "From Individual Devices" instructions produce precisely the format they tell you to avoid.

**The xml -> set path I have not managed to test end to end.** I pulled the config over the API and ran it through `panconf.py --set`, which gave 2276 clean set lines with correctly shaped interface statements. But both my attempts lost the vsys qualification — the import comes out as `set import network interface [ ... ]` with no `vsys vsys1` anywhere — and Batfish then parses the node with only 78 warnings but produces zero interfaces and zero defined structures. That looks like me driving `panconf` from the wrong root rather than anything wrong with the path; the documented flow uses `panxapi.py -sxr`, which presumably hands it the shape it expects, and I was hand-rolling the fetch. So treat the conversion path as unverified by me, not as ruled out.

Where that leaves the PR: brace format is still an accepted input that today mis-parses a whole device into no interfaces at all, silently and with no warning naming the cause. That seems worth fixing regardless of which collection method is recommended — but if you would rather close it and treat brace as unsupported, I have no objection, and I will move my own collection over to the conversion path either way.

dhalperi (commented):
> @dhalperi reviewed 2 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10166)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on gh-ajinks).

___
*[`projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto_nested/PaloAltoNestedLexer.g4` line 69 at r1](https://reviewable.io/reviews/batfish/batfish/10166#-P03-uotBM41nS4mpIkE:-P03-uotBM41nS4mpIkF:b-9ps129) ([raw file](https://github.com/batfish/batfish/blob/8ae898c87e9ca669a66236cf6fecafc06791229a/projects/batfish/src/main/antlr4/org/batfish/grammar/palo_alto_nested/PaloAltoNestedLexer.g4#L69)):*
> ```ANTLR
> LOG_PROFILE_PAYLOAD
> :
>    'payload' F_WhitespaceChar+ '"' .*? '";' -> skip
> ```

can we add the part that anchors the `payload` to the start of the line, similar to the COMMENT_LINE above?

`payload {lastTokenType() == NEWLINE || lastTokenType() == SHOW_CONFIG_LINE || lastTokenType() ==  COMMENT_LINE}? ...`
___
*[`projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java` line 445 at r1](https://reviewable.io/reviews/batfish/batfish/10166#-P030N3d4eLQFnC2Mvtd:-P030N3d4eLQFnC2Mvte:bwgfrgr) ([raw file](https://github.com/batfish/batfish/blob/8ae898c87e9ca669a66236cf6fecafc06791229a/projects/batfish/src/test/java/org/batfish/grammar/palo_alto/PaloAltoGrammarTest.java#L445)):*
> ```Java
>     // no interfaces. The interface below sits before the payload; the addresses prove the
>     // config still nests correctly around it.
>     Configuration ctrl = parseConfig("log-profile-control");
> ```

I don't think we need this at all, do we? May have been useful for development, but we just need the payload file for the test.
___
*[`projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/log-profile-payload` line 29 at r1](https://reviewable.io/reviews/batfish/batfish/10166#-P030Ef88N5szauIL8-I:-P030Ef88N5szauIL8-J:bq1bnrw) ([raw file](https://github.com/batfish/batfish/blob/8ae898c87e9ca669a66236cf6fecafc06791229a/projects/batfish/src/test/resources/org/batfish/grammar/palo_alto/testconfigs/log-profile-payload#L29)):*
> ```
>     }
>   }
>   shared {
> ```

can you put this earlier in the file so that we know it doesn't break things after it?

dhalperi (commented):
> @dhalperi made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10166)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on gh-ajinks).

___
*[a discussion](https://reviewable.io/reviews/batfish/batfish/10166#-P030_lJ5cwcBLjlMpJv:-P030_lKEO0NhRVn_h3G:bicdtus) (no related file):*
This PR needs a rebase after your other PR merged.

gh-ajinks:
All four addressed, and rebased onto master.

**Anchoring.** Added the predicate. One deviation from your suggestion: I left `COMMENT_LINE` out of the set, because it is `-> skip` and the base lexer only records a type on `emit()`, so it can never be `lastTokenType()` — that clause would be dead. I used the same set `COMMENT_LINE` itself uses:

```antlr
'payload'
{lastTokenType() == -1 || lastTokenType() == NEWLINE || lastTokenType() == SHOW_CONFIG_LINE}?
F_WhitespaceChar+ '"' .*? '";' -> skip
```

Say the word if you would rather have it in for symmetry.

**Control config.** Gone. You are right that it was development scaffolding — I added it after my first fixture attempt failed for an unrelated reason, and it had served its purpose by the time I opened the PR.

**Payload position.** Moved to the top of the file, ahead of the interface. That is a better test than what I had: with the payload last, nothing downstream of it was being exercised, so the fixture never actually tested the thing the bug breaks. Now the interface parsing after it is the assertion.

**Rebase.** Done, onto `8cde9750f2`.

Still green: the PAN-OS suite, both parsing corpora, checkstyle and `//tools/format:format.check`. The test fails with the rule removed and passes with it.

dhalperi (commented):
> @dhalperi reviewed 4 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10166)* status: all files reviewed, 2 unresolved discussions (waiting on gh-ajinks).

dhalperi (commented):
> @dhalperi resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10166)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #10168, opened 2026-08-27
AWS: build IKE phase 1 proposals from phase 1, and drop integrity for AEAD

gh-ajinks:
Two defects in `VpnConnection`'s IKE/IPsec conversion, both of which stop a correctly configured VPN from negotiating.

### Phase 1 proposals are built from the phase 2 algorithms

```java
for (Value ikePfs : ipsecTunnel.getIkePerfectForwardSecrecy()) {
  for (Value authAlgorithm : ipsecTunnel.getIpsecAuthProtocol()) {
    for (Value encryptionAlgorithm : ipsecTunnel.getIpsecEncryptionProtocol()) {
```

Those are the same two accessors the phase 2 builder uses. `IpsecTunnel` parses both phases correctly from `TunnelOptions`, but `getIkeAuthProtocol()` and `getIkeEncryptionProtocol()` are then read nowhere in `main/` — the phase 1 values are dead. Only the DH group loop uses a phase 1 field, which is why the group numbers look right while the algorithms do not.

A VPN configured with

| | Encryption | Integrity | DH |
|---|---|---|---|
| Phase 1 | `AES256` | `SHA2-256` | 19, 21, 20 |
| Phase 2 | `AES256-GCM-16` | *(omitted)* | 19, 21, 20 |

produced twelve phase 1 proposals, every one `AES256-GCM-16`. `toEncryptionAlgorithm` maps `AES256` to `AES_256_CBC` and `AES256-GCM-16` to `AES_256_GCM`, so a peer offering AES-256-CBC — which is what that tunnel's phase 1 actually asks for — cannot negotiate phase 1.

### An omitted phase 2 integrity list becomes HMACs paired with an AEAD cipher

AWS omits `Phase2IntegrityAlgorithms` when the phase 2 cipher is AEAD, since the cipher authenticates internally. `TunnelOptions.create` defaults each absent list to every supported value, so the omission became four HMAC algorithms, each paired with `AES256-GCM-16`. A peer configured for GCM offers no separate integrity algorithm, so none of those four could ever match.

Emit a single proposal with no authentication algorithm for an AEAD cipher instead, and name it without the null.

### Compatibility

This changes what the AWS side advertises, so it can change session outcomes for any vendor terminating a Site-to-Site VPN, not only the one I hit it with.

- Phase 1 and phase 2 configured the same: no change. This includes every VPN left at AWS defaults, where both phases fall back to the same permissive list — which is why this has gone unnoticed.
- They differ and the peer matches phase 1: previously failed, now negotiates.
- They differ and the peer happened to match phase 2: previously negotiated, now correctly fails.

The AEAD change only removes proposals that could not have matched a conforming peer, so it should turn non-negotiating tunnels into negotiating ones and not the reverse.

### Tests

`testIkePhase1ProposalsUsePhase1Algorithms` builds a tunnel whose phase 1 and phase 2 differ in encryption, integrity and DH group, and asserts the phase 1 proposal carries the phase 1 values. It fails on master with `Expected: <AES_256_CBC> but: was <AES_256_GCM>`.

`testIpsecPhase2ProposalsAeadHasNoAuthentication` gives a tunnel a GCM phase 2 cipher alongside the defaulted HMAC list, and asserts a single proposal with no authentication algorithm.

Measured on a snapshot of four PAN-OS firewalls plus AWS and Arista: established BGP sessions go from 200 to 208, and established IPsec sessions from 16 to 28.

Green locally: the AWS, PAN-OS, Cisco and common-util suites, the parsing corpus, checkstyle and `//tools/format:format.check`.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/aws/VpnConnection.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/aws/VpnConnection.java
@@ -400,8 +400,8 @@ final class VpnConnection implements AwsVpcEntity, Serializable {
   private static @Nonnull List<IkePhase1Proposal> toIkePhase1Proposals(IpsecTunnel ipsecTunnel) {
     List<IkePhase1Proposal> proposals = new ArrayList<>();
     for (Value ikePfs : ipsecTunnel.getIkePerfectForwardSecrecy()) {
-      for (Value authAlgorithm : ipsecTunnel.getIpsecAuthProtocol()) {
-        for (Value encryptionAlgorithm : ipsecTunnel.getIpsecEncryptionProtocol()) {
+      for (Value authAlgorithm : ipsecTunnel.getIkeAuthProtocol()) {
+        for (Value encryptionAlgorithm : ipsecTunnel.getIkeEncryptionProtocol()) {
           IkePhase1Proposal ikePhase1Proposal =
               new IkePhase1Proposal(
                   "ike_proposal_"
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10168)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #10169, opened 2026-08-27
AWS: tolerate a tunnel whose pre-shared key lives in Secrets Manager

gh-ajinks:
AWS Site-to-Site VPN offers two pre-shared key storage modes. From the [documentation](https://docs.aws.amazon.com/vpn/latest/s2svpn/enhanced-security-storage.html):

> **Standard** — The pre-shared key is stored directly in the Site-to-Site VPN service.
> **Secrets Manager** — The pre-shared key is stored using AWS Secrets Manager.

and, on switching a tunnel to the latter:

> The pre-shared key is removed from the Site-to-Site VPN service. A new Secrets Manager secret is created, if one doesn't already exist. The new pre-shared key is stored in Secrets Manager.

So for such a tunnel `DescribeVpnConnections` returns a top-level `PreSharedKeyArn` and the tunnel options carry no `PreSharedKey`. `TunnelOptions.create` rejects that:

```java
checkArgument(presharedKey != null, "PreSharedKey cannot be null");
```

Because the exception comes from a `@JsonCreator`, it fails deserialization of the enclosing object, and the whole file is discarded with `Exception while parsing 'VpnConnections' ... Cannot construct instance of VpnConnection$TunnelOptions`. Every VPN connection in that region is lost — the transit gateway ends up with no VPN attachment and no VPN interfaces, and the tunnels cannot be modelled at all.

### Why the field can go

`getPresharedKey()` has no caller anywhere in `main/`. The key used to build the phase 1 key comes from the customer gateway configuration:

```java
builder.setIkePreSharedKeyHash(
    CommonUtil.sha256Digest(
        Utils.textOfFirstXmlElementWithTag(ikeElement, AwsVpcEntity.XML_KEY_PRE_SHARED_KEY)
            + CommonUtil.salt()));
```

That element is still present under Secrets Manager storage, with its value masked. It hashes to something non-null and negotiates as an unreadable pre-shared key — which is exactly what it is, and what `negotiateIkePhase1Key` already handles for the mixed encrypted/plaintext case.

### Impact

On a snapshot where one VPN uses Secrets Manager storage, the parse error disappears and the transit gateway gains its four VPN interfaces (two external, two tunnel-inside). The tunnels reach `IPSEC_SESSION_ESTABLISHED` and their BGP sessions come up.

This affects anyone following AWS's own guidance — the Secrets Manager mode is presented as the more secure option, for "compliance with security best practices".

### Tests

`testDeserializationSecretsManagerPreSharedKey` uses a fixture shaped like real output: a `PreSharedKeyArn` on the connection, and tunnel options with no `PreSharedKey`. It fails on master and passes with this change.

Green locally: the AWS suite, the parsing corpus, checkstyle and `//tools/format:format.check`.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/representation/aws/VpnConnection.java
+++ b/projects/batfish/src/main/java/org/batfish/representation/aws/VpnConnection.java
@@ -805,7 +805,11 @@ final class VpnConnection implements AwsVpcEntity, Serializable {
         @JsonProperty(JSON_KEY_PRESHARED_KEY) @Nullable String presharedKey,
         @JsonProperty(JSON_KEY_PHASE2_LIFETIME_SECONDS) @Nullable Integer phase2LifetimeSeconds) {
       checkArgument(outsideIpAddress != null, "OutsideIpAddress cannot be null");
-      checkArgument(presharedKey != null, "PreSharedKey cannot be null");
+      // PreSharedKey is absent whenever the tunnel stores its key in Secrets Manager rather
+      // than in the Site-to-Site VPN service, in which case DescribeVpnConnections returns a
+      // PreSharedKeyArn instead. Nothing reads this field -- the key used for phase 1 comes
+      // from the customer gateway configuration -- so rejecting the tunnel would discard
+      // every VPN in the file over a value that is legitimately absent and never consumed.
       return new TunnelOptions(
           ikeVersions,
           firstNonNull(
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10169)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gh-ajinks).

PR #10171, opened 2026-08-27
claude: add a skill for drafting release notes

dhalperi:
The skill drafts the notes for a date-tagged release: read the previous
two releases for format, seed New Contributors from GitHub's
generate-notes API, survey the range by subject prefix to find the
themes, then curate down to what an operator upgrading can observe.
Internal work -- refactors, dependency churn, engine and symbolic
internals, grammar hygiene -- collapses to one line rather than being
enumerated, and breaking changes get their own list.

Credit does not come from `Fixes` trailers alone, since a maintainer
usually authors the fix and the reporter's name appears nowhere in the
commit; the skill sweeps issues closed over the release window and reads
their authors instead. Docker and pybatfish claims are grounded in those
repos, because image platforms and the MCP server are what users
experience but are not in this log.

The result is left as an unpublished draft. Notes that a draft's
`untagged-<hash>` id changes on every edit, so the URL has to be taken
from each command's output.

----

Prompt:

```
Can we create an in-repo skill for release notes?
```

Follows drafting the v2026.08.27 notes in the same session; the include
and exclude rules are the ones that came out of that, after being told
to drop the full commit list, the internal improvements, and the
symbolic engine work.

### DIFF
--- a/.claude/skills/release-notes/SKILL.md
+++ b/.claude/skills/release-notes/SKILL.md
@@ -1,0 +1,159 @@
+---
+name: release-notes
+description: Draft the GitHub release notes for a new Batfish release. Use when asked to prep or cut a release, draft release notes, or summarize what changed since the last tag. Produces a short curated summary written for someone deciding whether to upgrade, credits the people who filed the issues behind each change, and leaves it as an unpublished draft for review.
+---
+
+# Drafting a Batfish release
+
+A release covers 500+ commits, and almost none of them belong in the notes. The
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10171)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10179, opened 2026-08-31
StaticRoute: intern instances, including on deserialization

dhalperi:
Deployments that program the same static route onto every device in a
fabric produce snapshots holding enormous numbers of identical
StaticRoute objects. One such snapshot held 25.6M static routes with
only 16.1k distinct values.

Intern them through a Caffeine cache, as FibForward and BgpRoute's
attributes already do. Interning at construction is not sufficient by
itself: the vendor-independent configurations are serialized and
deserialized before the dataplane runs, and Java deserialization
rebuilds each object field by field without calling the builder. Nearly
all of the duplication was in those deserialized copies, so readResolve
is what collapses it. The JSON creator interns too, leaving all three
construction paths in agreement.

Mark the class final. Its only constructor was already private, so it
could not be subclassed outside the file, but interning is only sound if
no subclass can be conflated with an equal instance of the base class,
and final states that rather than leaving it to be re-derived.

The existing serialization tests asserted equality, so they passed
without readResolve; they now assert identity.

----

Prompt:
```
Let's land StaticRoute caching/readresolve - you have to do it in open
source batfish tho, ~/batfish/batfis
```

Context: found while investigating a dataplane memory regression on a
snapshot whose devices each carry ~14k identical static routes.

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/StaticRoute.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/StaticRoute.java
@@ -8,8 +8,12 @@ import static java.util.Objects.requireNonNull;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10179)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10180, opened 2026-08-31
rib: store backup routes without a per-prefix multimap

dhalperi:
AbstractRib keeps a backup for every route it is ever offered, and
LinkedHashMultimap spends a ValueSet with its own hash table per prefix
plus a linked ValueEntry per route on that. Nearly every prefix has one
or two backups, so hold the route directly when there is one and a small
list otherwise. On a snapshot with ~28M backups this cut peak heap by
8.5 GiB.

get() now returns an immutable copy rather than a live view, so
removeBackupRoute removes through the map instead of through the
returned set. values() groups by prefix rather than using global
insertion order; only getBackupRoutes reads it, as a Set.

----

Prompt:
```
Land the backup-route compaction in open source batfish, first part only
(the prefix -> route-or-list change, not the open-addressed index).
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/AbstractRib.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/AbstractRib.java
@@ -1,8 +1,6 @@
 package org.batfish.dataplane.rib;
 
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.LinkedHashMultimap;
-import com.google.common.collect.Multimap;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10180)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10180)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10181, opened 2026-08-31
ibdp: hold conditional static routes in a list, not a StaticRib

dhalperi:
Nothing looks a prefix up in _staticConditionalRib: it is populated once
and thereafter only iterated, and StaticRib#comparePreference is
constant, so the RIB contributes a prefix trie and a per-prefix element
set for no benefit. On a snapshot whose devices each carry ~14k
conditional statics that was ~51M trie nodes, 2.8 GiB of peak heap.

Build the routes through a StaticRib during init so their iteration
order is unchanged, then keep only the flattened list.

----

Prompt:
```
What's next? (landing the static-RIB trie removal from the dataplane
memory investigation)
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10181)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10182, opened 2026-08-31
ibdp: retain annotated main RIBs only when a debug flag asks

dhalperi:
IncrementalDataPlane always kept every VRF's pre-finalization main RIB,
whose only readers are tests. Because the dataplane outlives the
computation, that pinned each RIB's prefix trie, route annotations and
backup routes for as long as the dataplane was held, alongside the
FinalMainRibs that supersede them. On a snapshot with ~25M main RIB
routes it was 6.4 GiB of retained heap; peak is unaffected, since the
RIBs are live during convergence either way.

Populate the field only when the retainAnnotatedRibs debug flag is set,
which BatfishTestUtils now does for tests, and have getRibsForTesting
fail with that flag's name rather than returning nothing. Tests that
assert on route annotations keep working; those that only wanted routes
read getRibs() instead, as most already did.

----

Prompt:
```
Land the _annotatedRibs change, using a flag that tests set rather than
retaining the RIBs in production.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
@@ -432,6 +432,16 @@ final class IncrementalBdpEngine {
       TopologyContext initialTopologyContext,
       Set<BgpAdvertisement> externalAdverts,
       IpOwners initialIpOwners) {
+    return computeDataPlane(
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10182)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10183, opened 2026-09-01
ForwardingAnalysis: gather unowned ARP IPs in parallel, free the BDDs

dhalperi:
computeUnownedArpIps walked every FIB entry in the snapshot through a
sequential stream to build a set that stays in the thousands. On a large
snapshot that was one of the few single-threaded stretches of dataplane
computation, and it runs once per topology iteration: 38.1s over 4
calls, down to 2.2s.

Collect the ARP IPs in parallel, then test them against the BDD
sequentially, since BDDFactory is not thread safe. Also free each BDD
from IpSpaceToBDD#toBDD, which hands back a copy the caller owns, rather
than leaking one per ARP IP per iteration.

----

Prompt:
```
Draft the ARP-IP change (parallelizing
ForwardingAnalysisImpl.computeUnownedArpIps).
```

Follow-up: also free the BDD returned by toBDD.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10183)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10184, opened 2026-09-01
Vrf: hold deserialized static routes in an array, not a TreeSet

dhalperi:
A TreeSet costs a red-black-tree node per element, and nothing mutates
this collection after the vendor-independent configurations are built.
On a snapshot whose devices each carry ~14k static routes that was 25.6M
nodes, and replacing them took the minimum heap the dataplane needs from
40g to 36g.

Convert in readResolve rather than at construction, since vendor
extractors do mutate the set in place, but only before serialization.
ImmutableSortedSet rather than ImmutableSet because nothing looks routes
up by membership, so the hash table would be dead weight; copyOfSorted
rather than copyOf so a non-natural comparator is preserved.

----

Prompt:
```
Keep going on the memory fixes, especially if they don't hurt
performance.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
+++ b/projects/batfish/src/main/java/org/batfish/job/ConvertConfigurationJob.java
@@ -648,11 +648,16 @@ public class ConvertConfigurationJob extends BatfishJob<ConvertConfigurationResu
             });
   }
 
-  /** Remove and warn on undefined track references. */
+  /**
+   * Remove and warn on undefined track references.
+   *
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10184)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10184)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10185, opened 2026-09-01
Rib: do not memoize a constant resolution restriction

dhalperi:
ResolvabilityEnforcer wrapped its restriction in a map keyed by route,
so it spent an entry per route in the RIB to remember a value that
cannot vary when the restriction is alwaysTrue, which is every VRF
with no resolution policy. It also rebuilt the restriction on each
evaluation, inside the memoizing lambda.

On a snapshot with ~25M main RIB routes this halved live HashMap$Node,
119,995,569 down to 62,701,450 instances, 3.1 GiB counting the tables.
Wall time is unchanged: three runs each of 185-188s before and 186-191s
after.

----

Prompt:
```
Keep going on the memory fixes, especially if they don't hurt
performance.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10185)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10186, opened 2026-09-01
Fib: answer matching IPs per prefix instead of materializing a map

dhalperi:
Replace Fib.getMatchingIps() with Fib.matchingIps(Prefix), and store
only the forwarding prefixes that have a more specific forwarding prefix
beneath them. Every other prefix is the longest match for exactly the
IPs it contains, so its matching IPs come from the key. In one region
that was 27,695,281 of 27,891,892 forwarding prefixes.

----

Prompt:
```
Every stack I sampled dies in the same place — ForwardingAnalysisImpl.computeMatchingIps → FibImpl.getMatchingIps, i.e. forwarding-analysis during dataplane computation.

Is there something we can do ther?
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10186)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10188, opened 2026-09-01
ibdp: merge protocol-independent routes into the main RIB directly

dhalperi:
The independent RIB existed so that reinitForNewIteration() could
re-seed a main RIB that was rebuilt from scratch every iteration; the
main RIB has been final and persistent since. Merge connected, local,
unconditional static, kernel and RIP-internal routes straight into it
and drop the RIB. On a 2,135-device snapshot the re-import it fed ran
123,708 times across 53 iterations without once changing the main RIB.

Fixes an autostate bug that this exposes: withdrawing an interface's
connected and local routes from the main RIB left them in the
independent RIB, so the next iteration's re-import put them back.

----

Prompt:
```
Over and over again, we keep coming to RAM used in PrefixTrieMultiMap. Have we investigated how to improve its memory use?
```

Followed by a request to move everything importing into the independent
RIB to import into the main RIB instead, and to fix the autostate bug
with red/green TDD.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
@@ -979,7 +979,6 @@ final class IncrementalBdpEngine {
     vrs.parallelStream()
         .forEach(
             vr -> {
-              importRib(vr.getMainRib(), vr._independentRib);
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10188)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10191, opened 2026-09-02
RibTree: skip straight to the next shorter match when a longest match is unusable

dhalperi:
getLongestPrefixMatch looked up the longest match for every prefix
length from the maximum down to zero whenever the longest match held no
usable forwarding route. Each iteration searched the trie again and
re-tested the same set until the length dropped below that set's prefix.
Static-route activation under a resolution restriction hits this on
every check, and in one large snapshot it was 46% of dataplane CPU.

Jump to one less than the matched prefix's length instead, stop as soon
as a lookup finds nothing, and test the route sets with loops rather
than streams.

----

Prompt:
```
make a PR(s) for any other changes that aren't related to the
PrefixTrieMultiMap refactor. I see some more stream rewrites I think
```

Split out of a session optimizing PrefixTrieMultiMap, where profiling
showed this loop rather than the trie itself as the top cost.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibTree.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibTree.java
@@ -94,13 +94,21 @@ final class RibTree<R extends AbstractRouteDecorator> implements Serializable {
   }
 
   private boolean hasAllowedForwardingRoute(Set<R> routes, ResolutionRestriction<R> restriction) {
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10191)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10192, opened 2026-09-02
profiling: add a headless dataplane runner

dhalperi:
dataplaneRunner parses a snapshot directory and computes its dataplane
in one JVM with no coordinator or client, so a profiler attached to it
sees only that work. It prints parse and dataplane times, route and FIB
counts with order-independent hashes so two builds can be checked for
identical results, and on request a class histogram of the live heap
taken in-process, what a server would retain, a flight recording with
GC-root chains, or the largest main RIB's prefixes as benchmark input.

It exits through Runtime.halt: Batfish leaves non-daemon threads behind
that keep the JVM alive after main returns, and attaching jcmd for a
histogram hung against a large heap.

----

Prompt:
```
make a PR(s) for any other changes that aren't related to the
PrefixTrieMultiMap refactor. I see some more stream rewrites I think
```

Split out of a session optimizing PrefixTrieMultiMap, where this was
the harness for every before/after measurement.

### DIFF
--- a/tools/profiling/BUILD.bazel
+++ b/tools/profiling/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@rules_java//java:java_binary.bzl", "java_binary")
 load("@rules_python//python:defs.bzl", "py_binary")
 
 package(default_visibility = ["//visibility:public"])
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10192)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10193, opened 2026-09-02
PrefixTrieMultiMap: add Handle to read and then update one prefix in a single search

dhalperi:
RibTree.mergeRoute searched the trie for a prefix's routes and then
searched again to add or replace them; removeRouteGetDelta searched to
remove and again to see what was left. handle(prefix) finds or creates
the node once and returns a Handle whose get, add, replaceAll and remove
operate on that node. existingHandle(prefix) returns null rather than
creating anything, so removing a route that was never present adds no
node.

On 250 devices of one region the trie work under mergeRoute falls from
about 10.5% to 8.6% of dataplane CPU samples, around 2% overall and
within run-to-run noise of wall time. The API is groundwork for a
change of node representation, where a second search costs more.

----

Prompt:
```
Would it be useful to send Handle as its own PR?
Do it. including measuring perf
```

Split out of a session optimizing PrefixTrieMultiMap.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibTree.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibTree.java
@@ -50,14 +50,14 @@ final class RibTree<R extends AbstractRouteDecorator> implements Serializable {
     assert reason != Reason.ADD : "cannot remove a route with reason ADD";
 
     Prefix network = route.getNetwork();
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10193)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10194, opened 2026-09-02
PrefixTrieMultiMap: pack nodes into int records

dhalperi:
Store trie nodes as records in an int[] instead of objects: three ints
(prefix bits, left child and length, right child and flags) for the
branching nodes that path compression creates, four for nodes that hold
elements, whose fourth int indexes parallel element and key arrays. A
node with one element stores the element itself rather than a singleton
set. Lookups read one array location per step and never follow a
reference until they have found their answer. equals and hashCode are
over the entries rather than the node structure.

Add traverseEntriesAround for RibResolutionTrie, which visited a
prefix's ancestors and subtree through the general predicate traversal
and would otherwise pay to materialize a Prefix at every branching node
it passed. FinalMainRib and FibImpl trim their tries once built.

On a 250-device subset of a large snapshot, with identical RIB and FIB
contents before and after: the trie's share of the live heap falls from
684 MB to 326 MB, the live heap from 1.34 GB to 1.00 GB, and dataplane
CPU samples by about a fifth, of which the trie's own fall by about a
quarter. On the whole snapshot (2267 devices, 30M main RIB routes) the
dataplane completes with 9.6 GB live; before, a 20 GB heap ran out of
memory during forwarding analysis. On one of those RIBs the JMH
benchmark in tools/benchmarks shows LPM 1.8x, build 2.5x and
getAllElements 2.2x faster than the object trie.

----

Prompt:
```
PrefixTrieMultiMap is our biggest memory use and biggest CPU use. How
can we improve it by 2x-3x on both? If you want a snapshot to test with,
use a recent snapshot of a large network.

Goal set: design, implement, and test improvements to the way we
represent and lookup/LPM prefixes. Don't run out of disk space. Continue
until you've made a 2x improvement or are truly out of ideas.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibResolutionTrie.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/RibResolutionTrie.java
@@ -3,7 +3,6 @@ package org.batfish.dataplane.rib;
 import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 import java.util.function.BiConsumer;
-import java.util.function.BiPredicate;
 import javax.annotation.Nonnull;
 import org.batfish.datamodel.Ip;
 import org.batfish.datamodel.Prefix;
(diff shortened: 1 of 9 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10194)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10196, opened 2026-09-03
ibdp: re-evaluate a conditional static route only when its next hop's resolution can have changed

dhalperi:
activateStaticRoutes ran a longest-prefix match and a merge or remove
for every conditional static route in every iteration, and the removal
allocated a fresh AnnotatedRoute each time, even in iterations where the
main RIB had not changed at all. On a large snapshot whose devices carry
many recursive statics this was ~13% of dataplane wall time, most of it
in iterations that produced no static route change.

Whether a next-hop-IP static activates depends only on the main RIB
routes whose network contains the next hop, so within a topology round
it can only change if the previous round's main RIB delta added or
removed such a route. Group conditional statics by next hop IP and
re-evaluate only the groups whose IP lies inside a delta prefix. Statics
whose next hop is an interface, discard, VRF, or VTEP depend on
interface and track state, which only change between rounds. Every route
is still evaluated at init and at the start of each topology round, when
those inputs and the RIB may have changed without going through the
delta.

The resulting RIBs and FIBs are identical on the snapshot above; the
static pass drops from ~13% of wall time to ~3%, and total wall time
by ~8%.

----

Prompt:
```
Implement the top-ranked item from the dataplane profiling
investigation: stop re-evaluating every conditional static route each
iteration, and only re-evaluate the ones whose next hop resolution can
have changed. Measure correctness and benefit internally, then port to
open source, make sure master is up to date, and send the PR without
revealing internal details.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/VirtualRouter.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/VirtualRouter.java
@@ -38,6 +38,7 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10196)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10197, opened 2026-09-03
Bgpv4Rib: track next hop IPs in an ordered set instead of mirroring every main RIB prefix

dhalperi:
The resolvability enforcer kept a RibResolutionTrie holding a copy of
every main RIB prefix, maintained from each round's main RIB delta,
purely so that the next hop IPs affected by a changed prefix could be
found while pruning those shadowed by a more specific route. That is a
second prefix trie the size of the main RIB in every BGP VRF. On a large
snapshot it was 13% of the live heap during BGP convergence.

Keep only the next hop IPs, in a TreeSet, and answer "which next hops
lie inside this prefix" with a sub-set lookup. Shadowed next hops are
now visited, so record each next hop's resolvability as of the last
time all of its routes were (de)activated together and skip the ones
whose resolvability did not change; re-(de)activating them was a no-op
before. A next hop whose routes arrived under different main RIB states
within one round is marked mixed and always revisited. That also fixes a
latent gap: RibDelta.Builder cancels a withdrawal and re-add of the same
route within one delta, so a BGP route that arrived while its next hop's
covering route was transiently absent used to stay withheld.

RIBs and FIBs are identical on the snapshot above; live heap at a
steady-state iteration drops by 1.3 GiB, all of it PrefixTrieMultiMap
storage.

----

Prompt:
```
Next item from the dataplane profiling investigation: drop the main RIB
prefix mirror in the BGP resolvability enforcer and keep a next-hop-only
structure. Measure correctness and benefit internally, then port to open
source and send the PR without revealing internal details.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/Bgpv4Rib.java
@@ -55,19 +55,40 @@ public final class Bgpv4Rib extends BgpRib<Bgpv4Route> {
      */
     private final @Nonnull SetMultimap<Prefix, Bgpv4Route> _bgpRoutesByPrefix;
 
-    private final @Nonnull RibResolutionTrie _mainRibPrefixesAndBgpNhips;
+    /** The keys of {@link #_bgpRoutesByNhip}, ordered so the NHIPs inside a prefix can be found. */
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10198, opened 2026-09-03
profiling: delete the dataplane runner's temp storage on exit, and accept debug flags

dhalperi:
The runner parses into a temp directory under java.io.tmpdir and left
it behind, several GB of snapshot storage per run. On a host where tmp
is a tmpfs that is RAM, and enough leftover runs got later JVMs killed
by the OOM killer.

Also add `--debug-flag FLAG` so the runner can enable the same debug
flags a production client would pass to generate-dataplane.

----

Prompt:
```
Side fix while profiling the dataplane: the headless runner leaves its
temp snapshot storage in /tmp, which is a tmpfs here and was causing
OOM kills. Make it clean up, and let it take debug flags like the
production client does. Send as an open-source PR.
```

### DIFF
--- a/tools/profiling/BUILD.bazel
+++ b/tools/profiling/BUILD.bazel
@@ -81,6 +81,7 @@ java_binary(
         "//projects/batfish:batfish_testlib",
         "//projects/common",
         "@maven//:com_google_code_findbugs_jsr305",
+        "@maven//:com_google_guava_guava",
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10198)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10199, opened 2026-09-03
ForwardingAnalysisImpl: group each VRF's FIB entries by interface and ARP IP, not per route

dhalperi:
Per VRF, forwarding analysis built a route -> interface -> ARP IPs map
and then, for every interface and again for every neighbor of it, ran a
stream over every route out that interface to look up its ARP IPs and
test them against the neighbor's ARP replies. Every one of those
decisions depends only on the (interface, ARP IP) pair, and a VRF has
orders of magnitude fewer distinct next hops than routes, so the work
was dominated by stream setup, map lookups, and repeated Fib#matchingIps
calls for the same prefix. On a large snapshot this was 23% of all
dataplane CPU, more than BGP message passing.

Replace the per-route maps with a VrfForwardingIndex built in one pass
over the FIB: interface -> ARP IP -> routes, plus route -> ARP IPs for
the owned/unowned classification. ARP reply checks run once per group.
Fib#matchingIps is memoized per VRF across the unions that share a
prefix, the dest-IP match condition is computed once per interface
rather than once per neighbor edge, and the network-wide route ->
interface map that was held across both phases is gone. The IPs routed
out an interface are read straight from the FIB entries.

Results are unchanged: on two large snapshots every IpSpace of the new
analysis was compared as a BDD against the old implementation's and
found equal, and the analysis takes half the wall time it did. Line
order inside some AclIpSpace unions changes, so structural equality with
the old output does not hold.

----

Prompt:
```
A CPU profile of a full dataplane computation showed forwarding
analysis as the largest bucket, 23% of CPU, with the self time in Java
stream machinery and hash lookups inside the per-route ARP helpers.
Rewrite those to group routes by next hop and verify the output is
semantically identical.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/ForwardingAnalysisImpl.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/ForwardingAnalysisImpl.java
@@ -10,7 +10,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableTable;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> @dhalperi partially reviewed 2 files.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10199)* status: 1 of 2 files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10199)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10200, opened 2026-09-04
rib: keep only the routes that lost in BackupRoutes, not every route ever offered

dhalperi:
BackupRoutes recorded every route a RIB was offered, keyed by prefix,
so that the best remaining route could be restored when the held one
was withdrawn. That made it a hash map with an entry for every prefix
in the RIB, on top of the prefix trie that already indexes them. On a
large snapshot the map nodes and tables were the largest class in the
heap, 36% of the live set during convergence, while only a small
fraction of prefixes had ever been offered a losing route.

Track only the routes the RIB does not hold: a merged route that lost
on preference, and the routes displaced when a better one arrives, both
read off the merge delta. A restored route leaves the set when the
withdrawal delta promotes it, and a losing route withdrawn while in
backup is removed directly. Callers that want every offered route, the
dataplane's BGP backup routes and the same-prefix eviction in BgpRib,
now read the trie's routes together with the losers, using the trie
rather than getRoutes() because a multipath BgpRib hides same-next-hop
duplicates there. Rib's resolvability enforcer only needed the "in
backup" test, which is now a direct lookup.

DataplaneUtil's TODO on this, that backups should not store best
routes, still holds for the public getBackupRoutes(); this change keeps
that view identical and shrinks the storage behind it.

----

Prompt:
```
Heap histograms of large dataplane runs show HashMap nodes and tables
as the largest class, and they trace to BackupRoutes holding an entry
for every prefix in every main RIB. Store only the losing routes
instead, keeping the RIB results identical.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/rib/AbstractRib.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/rib/AbstractRib.java
@@ -1,8 +1,8 @@
 package org.batfish.dataplane.rib;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10200)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10201, opened 2026-09-04
parsing: rule out format heuristics with literal prefilters, and skip charset detection for ASCII files

dhalperi:
Format detection ran about 40 multiline regexes over the whole text of
every file, and each one that does not match, which is nearly all of
them, scans every position. Each heuristic is now a Tell: the regex plus
literals of which every match must contain at least one, checked with
String.contains first. The regex runs only when a literal is present.
On a corpus of thousands of large configurations the detected formats
are unchanged and detection CPU dropped by more than an order of
magnitude.

Reading a file ran ICU charset detection on every one. When all bytes
are printable ASCII, tab, or line-ending characters, every charset the
detector could report decodes them identically (no BOM or byte pattern
for UTF-16/32, no escape sequences for ISO-2022), so such files are
decoded as ASCII directly. The check is done a word at a time.

----

Prompt:
```
Upstream the parse-phase changes first: format detection with literal
prefilters per regex, and the printable-ASCII fast path that skips ICU
charset detection in StreamDecoder.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/grammar/VendorConfigurationFormatDetector.java
+++ b/projects/batfish/src/main/java/org/batfish/grammar/VendorConfigurationFormatDetector.java
@@ -1,7 +1,11 @@
 package org.batfish.grammar;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10201)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10202, opened 2026-09-04
init: serialize the convert answer element per file in parallel

dhalperi:
ConvertConfigurationAnswerElement's defined and referenced structures
are by far the largest part of it, a line number per statement that
references a structure, and default serialization wrote them through one
ObjectOutputStream on one thread after the parallel per-configuration
writes had finished. Each file's entry is now serialized to its own byte
array in parallel and the arrays are written in order; reading does the
reverse. This is factored into ParallelMapSerializer for other
per-device maps. On a large snapshot the write went from about ten
seconds to three and the read from about two seconds to a fraction of
one, for a file roughly a tenth larger.

----

Prompt:
```
Upstream the second batch: chunked parallel serialization of the
convert configuration answer element's defined and referenced
structures.
```

An earlier revision also kept the converted configurations in the cache
instead of reloading them from disk during initialization; that was
dropped because the reload is what checks that a cold cache sees the
same data as a hot one.

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/util/ParallelMapSerializer.java
+++ b/projects/common/src/main/java/org/batfish/common/util/ParallelMapSerializer.java
@@ -1,0 +1,85 @@
+package org.batfish.common.util;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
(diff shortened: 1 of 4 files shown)


dhalperi (commented):
> @dhalperi partially reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10202)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10204, opened 2026-09-04
datamodel: compact serialized forms for AclIpSpace, IpWildcardSetIpSpace, and PrefixTrieMultiMap

dhalperi:
A stored dataplane is mostly these three: a forwarding analysis holds an
ACL line per route out of each interface and a wildcard per route under
each prefix, and every RIB and FIB is a prefix trie. Default
serialization wrote three objects per ACL line (line, space, address)
and an object and an Ip per wildcard; the trie went through a
SerializedForm of two lists that wrote a Prefix, an Ip, and a set with
its backing array per key and rebuilt the trie on reading. Each class
now writes the numbers instead: an action and address bytes per line,
with spaces other than a single prefix or address still written as
objects; address and mask per wildcard; and for the trie its packed
records as ints, each slot's prefix as address and length, and each
slot's value as its elements, read straight back into the arrays with
nothing rebuilt. The SerializedForm is gone. On a large snapshot storing
the dataplane went from about nine seconds to three, and loading gains
similarly.

----

Prompt:
```
Upstream the second half of the storage batch: compact writeObject and
readObject for AclIpSpace lines, IpWildcardSetIpSpace wildcards, and the
PrefixTrieMultiMap serialized form.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/datamodel/AclIpSpace.java
+++ b/projects/common/src/main/java/org/batfish/datamodel/AclIpSpace.java
@@ -10,6 +10,10 @@ import com.google.common.collect.Comparators;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.Streams;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serial;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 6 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10204)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10204)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10205, opened 2026-09-04
coordinator: read an uploaded snapshot once, and extract and copy its files in parallel

dhalperi:
Uploading a snapshot wrote the zip to storage, read it back as a stream,
inflated it entry by entry on one thread, and then copied the extracted
files into the snapshot's input store one at a time. The upload is now
read once into a local file; storage gets a stream from that file, and
UnzipUtility extracts the file's entries in parallel through ZipFile,
directories first. The copy into storage materializes the directory walk
and runs in parallel too. UnzipUtility keeps its stream entry point for
callers that have only a stream, spooling to a file first. On a snapshot
of a few thousand large files the time from upload to the start of
parsing went from about eight seconds to under two.

----

Prompt:
```
Upstream the snapshot upload changes: parallel unzip and parallel copy
of input files. Reworked after discussion to read the upload into one
local file that is both stored and extracted, rather than storing it,
reading it back, and spooling it again, keeping StorageProvider stream
based so that it could be backed by S3.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/util/UnzipUtility.java
+++ b/projects/common/src/main/java/org/batfish/common/util/UnzipUtility.java
@@ -3,15 +3,16 @@ package org.batfish.common.util;
 import static com.google.common.io.MoreFiles.createParentDirectories;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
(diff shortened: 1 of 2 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10205)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10206, opened 2026-09-04
BGP: import from all neighbors in parallel, then merge into the RIBs in order

dhalperi:
A node pulled routes from its neighbors one edge at a time: for each
edge, run the neighbor's export (its export policy on every changed
route), then our import transform and import policy, then merge into our
RIBs. A node with many neighbors held up its whole schedule step. The
first two parts are now done for all edges in parallel and only the
merges stay sequential, edge by edge in the same order as before, with
each edge's routes in the same order (the sort is stable). Everything
the parallel part reads is fixed for the iteration: the neighbor's
export state, which several importers in one color already read at once,
our configuration and policies, the session properties, and the watched
track states; PrefixTracer is backed by concurrent collections. Route
transformation does not consult our RIBs, so deferring the merges cannot
change what is merged. Only the routes that pass import are held until
they are merged, as the transformed route and a withdrawn flag; rejected
routes are reported to the prefix tracer as they are seen, as are
installed ones, since its collections are concurrent and order does not
matter to it. Peak heap measured on a large snapshot is unchanged. On a
large snapshot the BGP portion of the dataplane went from about 32 to 25
seconds, with identical routes at every step.

----

Prompt:
```
Upstream the dataplane parallelism batch, starting with the two-phase per-neighbor BGP import.
```

Split from a PR that also built NetworkConfigurations once per
iteration; that is its own PR now.

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/BgpRoutingProcess.java
@@ -945,9 +945,19 @@ final class BgpRoutingProcess implements RoutingProcess<BgpTopology, BgpRoute<?,
       _externalAdvertisements = null;
     }
 
-    // Process updates from each neighbor
-    for (EdgeId edgeId : _bgpv4Edges) {
-      pullV4UnicastMessages(bgpTopology, nc, nodes, edgeId, _unicastEdgesWentUp.contains(edgeId));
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10206)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10207, opened 2026-09-04
BGP: build NetworkConfigurations once per iteration instead of once per process per step

dhalperi:
executeIteration also rebuilt the NetworkConfigurations map of every
node on every call, once per BGP process per schedule step. The engine
already has it and passes it down; the old signature remains and builds
the map itself.

----

Prompt:
```
Upstream building NetworkConfigurations once per dataplane iteration
instead of inside every BGP process's executeIteration, as its own PR.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10207)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10208, opened 2026-09-04
ibdp: pass NetworkConfigurations to every RoutingProcess.executeIteration

dhalperi:
The interface took only the node map, so BgpRoutingProcess kept a
one-argument executeIteration that built the NetworkConfigurations map
itself and existed only to satisfy the interface and for tests, and
EigrpRoutingProcess still rebuilt that map on every call. The interface
method now takes the NetworkConfigurations the engine already has, all
three processes receive it, and the IGP entry points and the OSPF
initialization loop pass it through.

----

Prompt:
```
Change RoutingProcess.executeIteration to take NetworkConfigurations so
that BgpRoutingProcess has a single executeIteration, as its own PR
after the per-iteration NetworkConfigurations change merged.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10208)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10209, opened 2026-09-04
statics: resolve a conditional route group's next hop once, and skip merging routes already present

dhalperi:
Conditional static routes are grouped by next hop IP for re-evaluation,
and the routes of a group differ only in prefix and distance, yet each
was resolved with its own longest prefix match. StaticRouteHelper now
offers resolving a next hop IP, which depends on the IP and on whether
the route is recursive, separately from the check that a route's own
prefix does not swallow it; the combined form is gone, and VirtualRouter
resolves a group at most once per recursiveness and hands the result to
the next hop check. A route that should be active and is already in the
main RIB is not merged again, which was a no-op; deactivation still
runs, since the route may be a backup. On a large snapshot the
per-iteration recomputation of conditional statics went from about five
seconds to under three, with identical routes at every step.

----

Prompt:
```
Upstream 5b of the dataplane batch: static-route activation by next-hop
group.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10209)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10210, opened 2026-09-05
FibImpl: resolve each next hop IP once per VRF and share the result across routes

dhalperi:
Building a FIB resolved every route's next hop IP independently, walking
the RIB down to interface routes each time. A VRF has orders of
magnitude fewer distinct next hop IPs than routes, and the resolution of
an IP depends only on the IP, on whether the top-level route is a static
route and if so whether it is recursive, and on the RIB itself.

Keep the outcome of each resolution keyed by that triple: the FIB
actions it produced and the networks it traversed. Reuse the actions for
a later route with the same key, unless its own network is among the
traversed ones (that route would have been declared a loop instead) or
the first resolution itself declared a loop (an outcome specific to the
route that hit it). Resolve such routes the old way.

Move FibImpl from org.batfish.datamodel to org.batfish.dataplane, next
to its only caller and to its test. Nothing in common used it; the Fib
interface stays.

On a large snapshot FIB construction went from about seven seconds to
under two per dataplane, with identical FIBs.

----

Prompt:
```
Measure the FibImpl next-hop resolution cache in isolation and see
whether it is useful. If so, upstream it; drop the rest of the small
CPU cleanup batch.
```


dhalperi (commented):
> @dhalperi partially reviewed 4 files.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10210)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10210)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10211, opened 2026-09-05
topology: filter synthesized L3 edges as they are produced, and synthesize prefix buckets in parallel

dhalperi:
computeRawLayer3Topology built the full set of same-subnet interface
pairs, sorted it, wrapped it in a Topology with its per-node and
per-interface indexes, and only then dropped the pairs that are not in
the same L1 broadcast domain. On networks with large shared subnets that
set is quadratic in the subnet size and the filter discards most of it.

Test the keep predicate on each candidate edge before collecting it, and
synthesize the buckets in parallel: they are independent, and the
collector sorts the result, so bucket order does not matter.
synthesizeL3Topology keeps every edge as before.

On a large snapshot the initial and per-round L3 topology computations
went from about seventeen seconds to two, with identical topologies.

----

Prompt:
```
Upstream the next performance change: the L3 topology synthesis in
TopologyUtil that filters edges during synthesis instead of after
building a Topology.
```

### DIFF
--- a/projects/common/src/main/java/org/batfish/common/topology/TopologyUtil.java
+++ b/projects/common/src/main/java/org/batfish/common/topology/TopologyUtil.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10211)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10212, opened 2026-09-05
ibdp: hash and count RIB routes without materializing the route set each iteration

dhalperi:
Every iteration hashes each VRF's main RIB and OSPF RIBs for oscillation
detection and counts the main RIB routes for statistics. Both went
through getRoutes(), which builds an ImmutableSet of every route in the
RIB whenever the RIB changed since the last call, so a changed RIB paid
a full copy per iteration and the copies of all main RIBs stayed live
between iterations.

Maintain Set.hashCode() of the routes in AbstractRib as merges and
removals report their deltas, and expose it with getRoutesHashCode();
expose the tree's route count with getNumRoutes(). Hash the RIB state
and count main RIB routes with those. The hash value is unchanged, so
oscillation detection behaves as before, and getRoutesHashCode() asserts
against the materialized value.

On a large snapshot the per-iteration statistics went from about seven
seconds to two over a dataplane computation, with identical results.

----

Prompt:
```
Upstream the next performance change: maintain the main RIB's
Set.hashCode incrementally so iteration hashing and statistics need not
materialize the route set.
```

### DIFF
--- a/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
+++ b/projects/batfish/src/main/java/org/batfish/dataplane/ibdp/IncrementalBdpEngine.java
@@ -1143,7 +1143,7 @@ final class IncrementalBdpEngine {
     ae.getBgpMultipathRibRoutesByIteration()
         .put(dependentRoutesIterations, numBgpMultipathRibRoutes);
     int numMainRibRoutes =
-        vrs.parallelStream().mapToInt(vr -> vr.getMainRib().getRoutes().size()).sum();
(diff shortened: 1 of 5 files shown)


dhalperi (commented):
> @dhalperi partially reviewed 5 files.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10212)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on dhalperi).

dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10212)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10213, opened 2026-09-05
topology: compute per-device ARP replies, host subnet IPs, owned IP spaces, and location info in parallel

dhalperi:
Four per-round computations ran sequentially over all devices, each a
pure function of one device's configuration and read-only shared inputs:
ARP replies and interface host subnet IPs in the forwarding analysis,
the owned-IP inversion in IpOwnersBaseImpl, and computeLocationInfo.

Add CollectionUtil.toImmutableMapParallel, the parallel form of
toImmutableMap with the same iteration order, and use it for the first
three. Run computeLocationInfo as a parallel stream; its collector keeps
the encounter order. In the owned-IP inversion, collect each interface's
IPs into a list first and build the IpWildcardSetIpSpaces, the expensive
part, per device in parallel.

Worth about two seconds per dataplane on a large snapshot, with
identical results.

----

Prompt:
```
Send the two small remaining OSS PRs: the parallel per-device ARP
replies and location info, and the schedule computed once per topology
round.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10213)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10214, opened 2026-09-05
ibdp: compute the node schedule once per topology round instead of once per iteration

dhalperi:
Every iteration rebuilt the node schedule from the topology, though
within a round the nodes and topology are fixed and only the schedule
type can change, when oscillation detection switches to NODE_SERIALIZED.

Compute the schedule's steps once and reuse them for every iteration
until the type changes. Worth about two seconds per dataplane on a large
snapshot, with identical results.

----

Prompt:
```
Send the two small remaining OSS PRs: the parallel per-device ARP
replies and location info, and the schedule computed once per topology
round.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10214)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10215, opened 2026-09-09
ParallelMapSerializer: pair each key with its value in one traversal

dhalperi:
writeMap serialized the values from byKey.values().parallelStream() and
then read the keys from a separate byKey.keySet() loop, pairing the two
by index. That relies on the two views being traversed in corresponding
order, and on a parallel collect over an unordered spliterator, such as
a HashMap's, reporting its elements in the source's iteration order.
Neither is specified, though both hold for the JDK maps today and the
only caller passes TreeMaps.

Take the keys and the values in one forEach into two lists, and
serialize the value list, whose spliterator is ordered.

Add a round trip test, including a HashMap whose values name their own
keys. It passes against the old code too: nothing mispairs on the
current JDK.

----

Prompt:
```
Got this review comment on the parallel map serializer we just added:

  This pairs keys with chunks by index across two separate traversals:
  chunks is built from byKey.values().parallelStream(), while keys come
  from a later byKey.keySet() loop. This is fragile because it relies on
  values() and keySet() being traversed in corresponding order, and for
  a Map whose spliterator is unordered (e.g. HashMap), a parallel
  collect(Collectors.toList()) is not guaranteed to preserve the
  values() iteration order -- chunks can be silently paired with the
  wrong key. Since the method accepts an arbitrary Map<String, ?>, build
  key+value together from a single entrySet() traversal. Alternatively,
  document and enforce that only maps with a defined iteration order
  (e.g. LinkedHashMap/TreeMap) may be passed.

Is it right? If so, what should we do about it?
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10215)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10216, opened 2026-09-09
Junos: report undefined references for apply-groups and apply-groups-except

dhalperi:
Applying a group that does not exist was only a red-flag warning, so it
never reached the undefined references answer. Track the group as a
structure instead: new APPLY_GROUP type, with APPLY_GROUPS and
APPLY_GROUPS_EXCEPT usages.

Definitions and references come from opposite ends of the pipeline.
Preprocessing deletes every `set groups` line before
ConfigurationBuilder walks the tree, so Hierarchy retains the
S_groups_named context of each one and FlatJuniperControlPlaneExtractor
defines the structures before the walk. The apply-groups and
apply-groups-except lines do survive preprocessing, so
ConfigurationBuilder records the references and gets flattened line
numbers the usual way.

apply-groups ${node} names a cluster variable, not a group, and Batfish
applies both node groups, so it credits a reference to whichever of
node0 and node1 exists. The existence check used to run before ${node}
was expanded, making every apply-groups ${node} a false warning.

Drops the warning: the undefined reference covers the same statements
and reports a line number rather than a hierarchy path. Two configs in
the parsing-tests snapshot do apply undefined groups, so the updated
refs show those.

----

Prompt:
```
Use red/green TDD to fix a user report that Batfish doesn't report
undefined reference when apply-group is not defined. This may be
annoying as apply-groups are mostly dealt with in preprocessor, let me
know if you want to discuss.
```

Follow-ups: named the structure type apply-group; removed the red-flag
warning now that the undefined reference supersedes it; extended
reference tracking to apply-groups-except.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10216)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10217, opened 2026-09-11
junit_tests: drop every attribute the receiving java rule does not accept

dhalperi:
junit_tests passes its keyword arguments to both a java_library holding
the test sources and the java_test that runs them, but only removed size
from the library's copy. Every other test-only attribute reached
java_library and failed to load, and the four java_library-only
attributes reached java_test the same way.

Replace the single pop with the full lists, 18 test-only attributes and
4 library-only ones. rules_java defines the split: java_test takes
java_binary's attributes plus the ones Bazel adds to every test rule,
while java_library takes only the compilation attributes.

Neither rule is introspectable from a macro, since
@rules_java//java:defs exports wrapper functions and the rule objects
live in a repository not visible to the main one. So
//skylark:junit_attrs_test declares a probe java_library and java_test,
reads their attribute names back with native.existing_rule, and diffs
each rule's exclusive attributes against the lists in junit.bzl. A
rules_java upgrade that moves an attribute fails there instead of in the
next BUILD file to use it.

One gap: native.existing_rule omits attributes left at a None default,
so the probes set testonly and timeout explicitly, and a future
attribute that also defaults to None would escape the check.

----

Prompt:
```
Seems like our skylark/junit rule doesn't properly remove all args that
java_test takes and java_library doesn't. It removes size, but not env,
for example. Can you figure the source of truth and flesh out the list?
Bonus points if it's programtically driven instead of enumerated.
```


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10217)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10222, opened 2026-09-14
coordinator: delete a snapshot's temp trees on every exit, and verify zip entry CRCs

dhalperi:
Forking copies the base snapshot's inputs into a temporary directory and
deleted only the input subdirectory, in a finally around initSnapshot
alone. Anything that threw before that, such as a bad user zip, left a
full copy of the snapshot's configurations in java.io.tmpdir, and every
successful fork left the enclosing directory behind. The fork's own
extraction directory was deleted outside a finally. All of it now sits
under one directory that a single finally removes, and the extraction
directory has a finally of its own. On the upload path the extraction
directory is now deleted for an unchecked failure as well, not only for
an IOException. deleteOnExit was the only backstop, and it does not fire
in a process that runs for months, nor can it remove a directory that is
not empty.

Reading entries through ZipFile instead of ZipInputStream in
batfish/batfish#10205 stopped verifying them: a snapshot zip with a
corrupt entry uploaded successfully and stored the wrong bytes.
Extraction now checks each entry's CRC as it copies, and reports a write
failure as an IOException rather than a BatfishException, so a failed
extraction is reported as one.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10222)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10223, opened 2026-09-15
Junos: parse bgp metric-out, and keep groups whose statements are all inactive

dhalperi:
Two problems kept the config in the prompt from parsing cleanly.

metric-out had a lexer token and no parser rule. Add b_metric_out with
a subrule per documented form: a constant MED, igp with an optional
offset and delay-med-update, or minimum-igp with an optional offset.
The MED itself is still unmodeled, so the leaves are _null. An offset
may be negative, so metric-out pushes a lexer mode where its sign lexes
as DASH instead of being swallowed by UNRECOGNIZED_WORD.

Group definitions came from the set lines surviving preprocessing, so a
group whose every statement is inactive looked undefined and applying
it was an undefined reference. Record definitions in a new pass that
runs before deactivated lines are pruned; a group's inherited contents
still come from the surviving lines only.

----

Prompt:
```
Make this config parse without undefined references:

groups {
    replace: G1 {
        /* A comment. */
        protocols {
            bgp  {
                replace: inactive: metric-out 2000000000;
            }
        }
    }
}
protocols {
    bgp {
        replace: apply-groups G1;
    }
}
```

Follow-ups: took the metric-out syntax from the Junos CLI reference
rather than a null_filler catch-all, gave each form its own subrule,
and lexed the offset sign as DASH rather than adding a signed-integer
token.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10223)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10224, opened 2026-09-16
Junos: parse empty containers, domain-path-id, filter group, and quoted group wildcards

dhalperi:
A group is defined by the set lines under `set groups <name>`, so a line
Batfish cannot parse takes the group's definition with it, and every
apply-groups of that group is then reported as an undefined reference.
Four unrelated gaps surfaced that way.

`routing-options srlg` required both a name and one of srlg-cost or
srlg-value, though the device accepts the container and the named entry
empty; flattening `srlg { }` inside a group produces exactly the bare
form. Split out ro_srlg_named and allow an empty tail at both levels.
`class-of-service interfaces`, and an interface within it, get the same
treatment.

`domain-path-id` had no rule at all. Add it with receive and send
subrules at the bgp, group, and neighbor levels per the CLI reference.
The BGP domain path attribute itself is unmodeled.

`family inet filter group <number>` had no rule either. `group` now
lexes as a keyword in M_Filter, so filter_name joins it to the input and
output it already accepts as filter names, keeping a firewall filter
actually named `group` working.

A group wildcard whose pattern is quoted, `<"ae[0-9][0-9]">`, cost the
whole file rather than one group: flattening splits the word at the
quote, and the resulting `< "ae[0-9][0-9]" >` lexes as a single
wildcard token that GroupWildcard rejects, failing the parse. Hierarchy
now trims and unquotes inside the angle brackets. Set-format config read
off a device puts the quotes outside the brackets, where unquote already
handled them, so only the hierarchical form was affected.

----

Prompt:
```
Even with our last fix, this gives an undefined reference:

groups {
    G_SRLG {
        routing-options {
            srlg {
            }
        }
    }
}
routing-options {
    /* Apply the SRLG group. */
    apply-groups G_SRLG;
}

In configs collected off the box, it just comes out as

set groups G_SRLG
set routing-options apply-groups G_SRLG

how can we best fix this?
```

Follow-ups: four more hierarchical group configs from the same source
(bgp domain-path-id under a routing-instance wildcard, lacp under
aggregated-ether-options, family inet filter group under interface
wildcards including a quoted pattern, and an empty class-of-service
interfaces container), and a decision not to make group definitions
survive a line Batfish cannot parse -- that remains a separate gap.


dhalperi (commented):
> @dhalperi reviewed 14 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10224)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10225, opened 2026-09-16
VRRP: honor `no vrrp <group> preempt` on IOS and Arista

### LINKED ISSUE #10220: no vrrp <group> preempt parses but has no effect on CISCO_IOS and ARISTA; Preempt stays true
**Version:** `2026.08.27.3685`

## Summary

`no vrrp <group> preempt` parses cleanly on `CISCO_IOS` and `ARISTA` — no parse warning, file status `PASSED` — and `vrrpProperties.Preempt` still answers `True` for the interface. Both platforms default preempt-on and print only the negation, so this is the only form a real running-config of either contains.

A VRRP member configured not to preempt is therefore indistinguishable from one configured to preempt, with nothing anywhere reporting a problem.

## Reproducer

Four single-device configs. Two exhibit the bug; two are controls that rule out the alternative explanations.

**`ios-vrrp.cfg`** — detected `CISCO_IOS`

```
!
hostname ios-vrrp
!
interface GigabitEthernet0/0
 ip address 10.10.1.2 255.255.255.0
 vrrp 2 ip 10.10.1.1
 vrrp 2 priority 90
 no vrrp 2 preempt
!
end
```

**`eos-vrrp.cfg`** — detected `ARISTA`

```
! device: eos-vrrp (DCS-7050QX-32, EOS-4.20.1F)
!
hostname eos-vrrp
!
interface Ethernet1
   no switchport
   ip address 10.10.3.2/24
   vrrp 1 ip 10.10.3.1
   vrrp 1 priority 90
   no vrrp 1 preempt
!
end
```

**`ios-hsrp.cfg`** (control) — same platform, HSRP instead of VRRP

```
!
hostname ios-hsrp
!
interface GigabitEthernet0/0
 ip address 10.10.2.2 255.255.255.0
 standby 1 ip 10.10.2.1
 standby 1 priority 90
 no standby 1 preempt
!
end
```

**`junos-vrrp.cfg`** (control) — detected `FLAT_JUNIPER`

```
set system host-name junos-vrrp
set interfaces ge-0/0/0 unit 0 family inet address 10.10.4.2/24 vrrp-group 1 virtual-address 10.10.4.1
set interfaces ge-0/0/0 unit 0 family inet address 10.10.4.2/24 vrrp-group 1 priority 90
set interfaces ge-0/0/0 unit 0 family inet address 10.10.4.2/24 vrrp-group 1 no-preempt
```

## Observed

| file | format | parse | `Preempt` | expected |
|---|---|---|---|---|
| `ios-vrrp.cfg` | `CISCO_IOS` | `PASSED`, 0 warnings | **`True`** | `False` |
| `eos-vrrp.cfg` | `ARISTA` | `PASSED`, 0 warnings | **`True`** | `False` |
| `ios-hsrp.cfg` | `CISCO_IOS` | `PASSED`, 0 warnings | `False` | `False` |
| `junos-vrrp.cfg` | `FLAT_JUNIPER` | `PASSED`, 0 warnings | `False` | `False` |

The two controls are the point. `hsrpProperties.Preempt` on the **same platform** honours its negation, and `vrrpProperties.Preempt` reports `False` for Junos — so the column is capable of `False`, and the parser is capable of honouring a preempt negation. The failure is specific to VRRP on `CISCO_IOS` and `ARISTA`.

`Priority = 90` is read correctly on both failing rows, so the VRRP block is not being skipped wholesale.

## Where it looks like it goes wrong

The negation is a recognised production, which is consistent with there being no parse warning. From `Cisco_interface.g4` on `master`:

```
if_vrrpno
:
   NO VRRP groupnum = dec
   (
      ifvrrpno_preempt
   )
;

ifvrrpno_preempt
:
   PREEMPT NEWLINE
;
```

So the line parses and the extracted model is unaffected — the effect appears to be dropped after parsing rather than the syntax being unrecognised. I have read the grammar on `master` and tested the behaviour on `2026.08.27.3685`; I have not read the grammar at the tested build, so treat the file quote as indicative of where to look rather than as the state of that build.

## Secondary, smaller

Bare `vrrp <group> preempt` — the affirmative form, valid on IOS — is **not** recognised:

```
interface GigabitEthernet0/0
 vrrp 3 preempt
```

yields `PARTIALLY_UNRECOGNIZED` with a warning on that line, because the affirmative rule requires a delay argument:

```
ifvrrp_preempt
:
   PREEMPT DELAY
   (
      MINIMUM
      | RELOAD
   ) dec NEWLINE
;
```

This is far less harmful than the above, since it is loud rather than silent, and since a config relying on the default rarely states it.

## Why it is worth fixing rather than documenting

The silent direction is the one real configurations take. Because preempt defaults on, a device that preempts says nothing and a device that does not says `no vrrp N preempt` — so every genuinely misconfigured member is exactly the case the model gets wrong, and it gets it wrong in the reassuring direction. Any analysis built on `vrrpProperties.Preempt` will report a first-hop pair as consistent when it is not, with no warning to contradict it.

dhalperi:
Both grammars had a rule for the negation and no listener for it, so
`no vrrp 1 preempt` parsed without a warning and left Preempt at its
default of true. Because preempt defaults on and only the negation is
printed, every member configured not to preempt was indistinguishable
from one that does. Implement the listener on both, and collapse the
repeated per-interface group lookup that surrounds it into a
currentVrrpGroups helper.

The affirmative form required a delay argument, so bare
`vrrp 1 preempt` drew a parse warning. Cisco's FHRP command reference
gives `vrrp group preempt [delay minimum seconds]`, and Arista's own
pyeapi builds `vrrp <vrid> preempt` and captures it in a `show
running-config all` fixture, so make the delay clause optional in both
grammars.

The ASA grammar carries the same unimplemented rules, but the ASA does
not run VRRP: its configuration guide mentions VRRP only as traffic
that can pass through. Left alone.

Fixes batfish/batfish#10220.

----

Prompt:
```
Fix https://github.com/batfish/batfish/issues/10220
```

Follow-up: ground the syntax changes in the vendor manuals rather than
inferring them.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10225)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10226, opened 2026-09-16
storage: reclaim deleted data by moving it aside, not by scanning for it

dhalperi:
Deleting a network or snapshot removed its name-to-ID mapping and queued
a pass that scanned every network directory, every snapshot directory
and every cache blob, deleting whatever no mapping pointed at and whose
file timestamps were all older than a ten minute skew allowance. The
skew was needed because data is written before the mapping that makes it
reachable, so data mid-creation is indistinguishable from garbage. It
also meant a snapshot uploaded and deleted within ten minutes was
skipped and never revisited, since nothing re-runs the pass but another
delete, another upload, or a restart.

Storage now offers deleteNetwork and deleteSnapshot, which must return
in time independent of how much data is stored, plus a maintenance
lifecycle that the coordinator starts once at startup. FileBasedStorage
deletes by renaming into <base>/trash, where nothing is reachable by
name or by ID, so a background thread empties it without regard for what
else is running and retries whatever it could not delete. Where the
rename fails, because the base directory spans filesystems or the
platform refuses to rename a directory holding an open file, a tombstone
naming the data is written instead and the drainer deletes it in place.

Data that no mapping points at is now only left behind by a crash
between deleting a mapping and deleting the data, so that scan runs at
startup, where nothing is in flight and no skew is needed. It also
reclaims the directories of deleted questions, which the old pass never
looked at, and a snapshot whose initialization fails now deletes its own
data rather than waiting for the scan.

The skew is gone everywhere, including from parse cache eviction, which
now keeps entries written before the oldest snapshot that still exists.
A snapshot's own entries are written while it is parsed, after its
creation time, and an entry evicted in error costs one reparse, since a
miss parses the input again. Eviction runs on a maintenance pass that a
deletion asked for, rather than on every delete and every upload: a
deletion is the only thing that makes more of the cache evictable.

Deleting a snapshot with work in flight defers deleting its data until
that work leaves the queue, which the coordinator can answer exactly and
the timestamps only approximated.

Deleting a network also deletes the mappings of everything inside it,
under <base>/ids: its snapshots, questions and node roles. Those were
unreachable and never collected, so they accumulated one small directory
per deleted network.

The tests cover the crash orderings by leaving the storage in the state
a crash would leave it in and restarting on the same data, and race
creation against deletion and the maintenance thread, asserting that
what a name still resolves to is intact and that nothing unreachable
survives. Each new test was checked against a deliberately broken
implementation.

----

Prompt:
```
Our garbage collection story (delete the metadata on network creation,
delete the files lazily 10m later after upload) is crap. Can you suggest
a better way to do it?

If I recall,
1/ we went with garbage collection because we didn't want delete_network
to be a long blocking call.
2/ we went with delayed GC because we were worried GC could run in the
middle of **creating** a network and break things.
```

Followups: confine the mechanism to the storage engine, since a
hypothetical S3 backend might leave reclamation to a lifecycle policy;
keep FileBasedStorage reclaiming promptly, because leaking data on a
local disk is not acceptable; and drop the ten minute pad on cache
eviction, which only held bytes longer.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10226)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

PR #10227, opened 2026-09-16
Junos: parse vlans filter and mac-move-limit, vrrp-inet6-group, and fix xSTP on tagged interfaces

dhalperi:
Three constructs had no parser rule, so a group whose entire body was
one of them recorded no definition and every apply-groups of it was an
undefined reference: a VLAN filter under forwarding-options, a VLAN
mac-move-limit, and vrrp-inet6-group under family inet6. None of the
three is modeled, so the leaves are _null.

The xSTP check resolved an interface named without a unit to unit 0,
then reported that unit as not enabled for Ethernet switching even when
the configuration has no unit 0 at all. Skip the check when the
physical interface is vlan-tagged, since such an interface carries its
logical interfaces on the tagged units.

----

Prompt:
```
Fix Junos false positives found by an audit of a large Junos fleet.
Four groups whose entire body is one construct are reported undefined
wherever they are applied, consistent with the construct not parsing
and so no group definition being recorded:

    vlans <*|NAME> forwarding-options filter input <filter>
    vlans <*> switch-options mac-move-limit <n> packet-action log
        interface <iface> action-priority <n>
    interfaces irb unit <*> family inet6 address <*>
        vrrp-inet6-group <*>
    routing-options bmp station <*> hold-down <n> flaps <n> period <n>

A fifth finding is in a check rather than the grammar: an interface
with flexible-vlan-tagging and only a tagged unit, referenced from
protocols vstp, is reported as "XSTP : Interface <name>.0 is not
enabled for Ethernet Switching" -- a unit that does not exist. Work out
what the check should require for a tagged L2 interface; the device is
running this config, so Junos accepted it.

Make each construct parse rather than recording group definitions from
lines that failed to parse; nulled leaves are fine where the semantics
are not modeled. Prefer red/green, with one focused test config per
construct.
```

Follow-ups: the bmp station construct parses already, and that group is
undefined for a different reason -- the config defines it in one module
and then wipes it with a later `replace: groups`, which is also what
Junos does with the same file -- so nothing is changed for it here.


dhalperi (commented):
> *[Reviewable](https://reviewable.io/reviews/batfish/batfish/10227)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dhalperi).

